PR #61, opened 2023-09-13
Nithin/eng 3813 store subgraphs in router config so the router is subgraph

thisisnithin:
## Motivation and Context


cli/src/commands/router/commands/compose.ts
```diff
@@ -86,6 +86,7 @@ export default (opts: BaseCommandOptions) => {
       argumentConfigurations: result.federationResult.argumentConfigurations,
       federatedSDL: printSchema(result.federationResult.federatedGraphSchema),
       subgraphs: config.subgraphs.map((s, index) => ({
+        id: `${index}`,
```
StarpTech, line 89: ID can be removed. I think we can match only on the routingUrl. The name can be used for debugging. Ensure the routingUrl is stored normalized without parameters etc...

StarpTech (changes requested):

cli/src/commands/router/commands/compose.ts
thisisnithin: Let's keep ID for the sake of having a unique identifier. I've added url normalization before storing on control plane.

thisisnithin: Also done

shared/src/utils/util.ts
```diff
@@ -15,3 +15,8 @@ export function splitLabel(label: string) {
 export function joinLabel({ key, value }: { key: string; value: string }) {
   return key + labelSeparator + value;
 }
+
+export function normalizeURL(url: string): string {
```
StarpTech, line 19: Please add a few tests.

StarpTech (changes requested):

shared/src/utils/util.ts
thisisnithin: Done

cli/src/commands/router/commands/compose.ts
```diff
@@ -28,7 +28,9 @@ type Config = {
 
 export default (opts: BaseCommandOptions) => {
   const command = new Command('compose');
-  command.description('Generates the router config locally. The output can be piped to a file.');
+  command.description(
+    'Generates the router config locally. The output can be piped to a file. To generate config for production use please use router fetch instead.',
```
StarpTech, line 32: > .. To generate config for production use please use router fetch instead.

"Generates a router config from a local composition file. This makes it easy to test your router without a control-plane connection. For production, please use the "router fetch" command"

StarpTech (changes requested):

router/core/transport.go
```diff
@@ -122,6 +127,19 @@ func (t TransportFactory) RoundTripper(transport http.RoundTripper, enableStream
 					}
 					span.SetAttributes(otel.WgOperationType.String(operation.opType))
 				}
+
+				var subgraph *nodev1.Subgraph
```
StarpTech, line 131: We should create a separate struct for subgraphs and embed them in the RequestContext e.g. `subgraphs()` (read-only).
Additionally, for convenience, we should expose a method like `activeSubgraph()` to get the current subgraph based on the subgraph request. This could return `nil` when evaluated at an early stage like [`middleware`](https://github.com/wundergraph/cosmo/blob/main/router/core/modules.go#L56). The data can be checked [here](https://github.com/wundergraph/cosmo/blob/main/router/core/transport.go#L41).

That also means you could replace the code with:

```go
rc := getRequestContext(r.Context())
if s:= rc.activeSubgraph(); s != nil {
	// .....
}
```

StarpTech (changes requested):

router/core/context.go
```diff
@@ -15,6 +15,12 @@ const requestContextKey = key("request")
 
 var _ RequestContext = (*requestContext)(nil)
 
+type Subgraph struct {
+	Id   string
+	Name string
+	Url  string
```
StarpTech, line 21: Make it a .net.URL for easier access.

router/core/transport.go
thisisnithin: Done

router/core/context.go
```diff
@@ -266,6 +277,15 @@ func (c *requestContext) GetStringMapStringSlice(key string) (smss map[string][]
 	return
 }
 
+func (c *requestContext) ActiveSubgraph(subgraphRequest *http.Request) *Subgraph {
+	for _, sg := range c.subgraphs {
```
StarpTech, line 281: We can do this once before we run the module handlers.

router/core/context.go
```diff
@@ -266,6 +277,15 @@ func (c *requestContext) GetStringMapStringSlice(key string) (smss map[string][]
...
+		if sg.Url == subgraphRequest.URL.String() {
```
StarpTech, line 282: Should we normalize the request URL before comparison? Host:port match should be enough.

router/core/router.go
```diff
@@ -583,6 +582,16 @@ func (r *Router) newServer(ctx context.Context, routerConfig *nodev1.RouterConfi
 			subChiRouter.Use(metricHandler.Handler)
 		}
 
+		var subgraphs []Subgraph
```
StarpTech, line 585: Pre-allocate the size based on the fixed numbers of subgraphs.

StarpTech (changes requested):

router/core/context.go
StarpTech: Let's stick with that approach because we don't need to deal with concurrent access now. Every subgraph request has a different active subgraph.

shared/src/router-config/builder.ts
```diff
@@ -24,6 +24,8 @@ export interface Input {
 }
 
 export interface Subgraph {
+  id: string;
+  name: string;
   sdl: string;
```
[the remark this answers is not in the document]
thisisnithin: It isnt available

router/core/router.go
thisisnithin: Done

router/core/context.go
thisisnithin: We do it on the control plane already. But updated comparison here anyways

router/core/context.go
thisisnithin: Done

router/core/router.go
```diff
@@ -594,6 +595,21 @@ func (r *Router) newServer(ctx context.Context, routerConfig *nodev1.RouterConfi
 			subChiRouter.Use(metricHandler.Handler)
 		}
 
+		subgraphs := make([]Subgraph, len(routerConfig.Subgraphs))
+		for _, s := range routerConfig.Subgraphs {
+			parsedURL, err := url.Parse(s.RoutingUrl)
+			if err != nil {
+				r.logger.Error("Failed to parse subgraph url", zap.Error(err))
```
StarpTech, line 602: Please specify the subgraph url that could not be parsed otherwise it is hard to debug.

router/core/transport.go
```diff
@@ -122,6 +123,15 @@ func (t TransportFactory) RoundTripper(transport http.RoundTripper, enableStream
 					}
 					span.SetAttributes(otel.WgOperationType.String(operation.opType))
 				}
+
+				reqContext := getRequestContext(r.Context())
```
StarpTech, line 127: The operation context is inside the request context so the code above is not needed.

StarpTech (changes requested):

router/core/router.go
thisisnithin: Added

router/core/transport.go
thisisnithin: Done

StarpTech (approved):
LGTM

PR #79, opened 2023-09-16
feat: implement backoff jitter retry mechanism

StarpTech:
## Motivation and Context

Implements backoff jitter retry for query operations
https://cosmo-docs.wundergraph.com/~/changes/VKKWOoLA1Kg0m70D5CDW/router/traffic-shaping

Introduce a breaking change where delays, and timeouts are parsed as Go duration (10s, 100ms) instead of suffixing all environment variables and config properties with _seconds)

### DIFF
--- a/router/core/factoryresolver.go
+++ b/router/core/factoryresolver.go
@@ -1,15 +1,9 @@
 package core
 
 import (
-	"context"
-	"crypto/tls"
-	"crypto/x509"
-	"errors"
 	"fmt"
-	"net"
 	"net/http"
 	"net/url"
-	"strings"
 	"time"
 
 	"github.com/wundergraph/graphql-go-tools/v2/pkg/engine/datasource/graphql_datasource"
@@ -30,7 +24,7 @@ type FactoryResolver interface {
 }
 
 type ApiTransportFactory interface {
-	RoundTripper(transport *http.Transport, enableStreamingMode bool) http.RoundTripper
+	RoundTripper(transport http.RoundTripper, enableStreamingMode bool) http.RoundTripper
 	DefaultTransportTimeout() time.Duration
 	DefaultHTTPProxyURL() *url.URL
 }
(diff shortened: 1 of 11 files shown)


router/core/factoryresolver.go
```diff
@@ -66,138 +60,13 @@ func NewDefaultFactoryResolver(transportFactory ApiTransportFactory, baseTranspo
...
-		// if the data source uses a custom proxy, create a dedicated client
-		if dataSourceUsesHTTPProxy(ds) {
-			_, found := config.LookupStringVariable(cfg.HttpProxyUrl)
-			return found
-		}
-	}
-	return false
-}
-
-// customTLSTransport returns a TLS *http.Transport with the given key and certificates loaded
```
StarpTech, line 89: Legacy code from WunderGraph. No code was actually used.

router/internal/retrytransport/retry_transport.go
```diff
@@ -0,0 +1,118 @@
...
+		resp, err = rt.RoundTripper.RoundTrip(req)
+
+	}
+
+	return resp, err
+}
+
+func IsRetryableError(err error, resp *http.Response) bool {
+
+	if resp != nil {
```
jensneuse, line 96: I'd put err check above res check.

router/internal/retrytransport/retry_transport.go
```diff
@@ -0,0 +1,118 @@
...
+			zap.Int("retry", retries),
+			zap.String("url", req.URL.String()),
+			zap.Duration("sleep", sleepDuration),
+		)
+
+		// Wait for the specified backoff period
+		time.Sleep(sleepDuration)
+
+		// Retry the request
+		resp, err = rt.RoundTripper.RoundTrip(req)
```
jensneuse, line 87: I'd add below pseudocode:
If err is nil and res status code is 200, return it.
No need to evaluate further and it makes the flow clear that once we have a valid response, we return it immediately.

router/internal/retrytransport/retry_transport.go
```diff
@@ -0,0 +1,118 @@
...
+	return &RetryHTTPTransport{
+		RoundTripper: roundTripper,
+		RetryOptions: retryOptions,
+		Logger:       logger,
+	}
+}
+
+func (rt *RetryHTTPTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+
+	resp, err := rt.RoundTripper.RoundTrip(req)
```
jensneuse, line 60: I'd add pseudocode after this:
If err is nil and resp status code is 200, return immediately.
No need to create a backoff or do further evaluations if we should retry.

jensneuse (changes requested):

router/internal/retrytransport/retry_transport.go
StarpTech: :+1:

jensneuse (approved):

PR #82, opened 2023-09-18
feat(config): support configuration of graphql path

paulpdaniels:
## Motivation and Context

Users should be able to modify the path that the gateway uses when serving graphql requests. Currently this path is hardcoded to `/graphql` but similar to the `ListenAddr` it should be possible for the user to change it via configuration. This PR adds support for a new config `GraphQLPath` defined in the yaml and adds it to the configurator code at startup.


StarpTech:
Hi @paulpdaniels thank you for the PR. Looks great. I'll take a deeper look in the afternoon.

StarpTech:
I made this PR a feat

StarpTech (approved):
LGTM

PR #99, opened 2023-09-20
feat: advanced analytics

Pagebakers:
## Motivation and Context

Screenshot 2023-09-20 at 17 16 00


thisisnithin (approved):

controlplane/src/core/bufservices/PlatformService.ts
```diff
@@ -1254,6 +1256,104 @@ export default function (opts: RouterOptions): Partial<ServiceImpl<typeof Platfo
...
+        const params = {
+          range: req.range,
+          params: {
+            organizationId: authContext.organizationId,
+            graphId: graph.id,
+            graphName: req.federatedGraphName,
+          },
+        };
+
+        const requests = await repo.getRequestRateMetrics(params);
```
StarpTech, line 1296: Run them in parallel with `Promise.all`.

controlplane/src/core/repositories/analytics/MetricsDashboardRepository.ts
```diff
@@ -0,0 +1,364 @@
...
+        endDate: endDate - range * 60 * 60 * 1000,
+      }),
+    });
+
+    return {
+      value: latencyP95.data.result[0]?.value[1],
+      previousValue: prevLatencyP95.data.result[0]?.value[1],
+      top: top5.data.result.map((v: any) => ({
+        name: v.metric.wg_operation_name,
+        value: v.value[1],
```
StarpTech, line 207: A potential out-of-range exception?

controlplane/src/core/repositories/analytics/MetricsDashboardRepository.ts
```diff
@@ -0,0 +1,364 @@
...
+  }
+
+  /**
+   * Get latency metrics
+   */
+  public async getLatencyMetrics({ range = 24, endDate = getUtcUnixTime(), params }: GetMetricsProps) {
+    const prange = '15m';
+
+    // get p95 of requests in last [range] hours
+    const latencyP95 = await this.client.query({
```
StarpTech, line 162: I think you missed handling error cases. I'd ensure that the client `query` always returns a valid result when it doesn't throw an error. You need to check for errors props in the response.

See https://prometheus.io/docs/prometheus/latest/querying/api/

controlplane/src/core/repositories/analytics/MetricsDashboardRepository.ts
```diff
@@ -0,0 +1,364 @@
...
+
+    // get top 5 operations in last [range] hours
+    const top5 = await this.client.query({
+      query: `topk(5, sum by (wg_operation_name) (rate(cosmo_router_http_request_duration_milliseconds_bucket{${this.getQueryLabels(
+        params,
+      )}}[${range}h])))`,
+    });
+
+    // get p95 latency [range] hours in series of [step]
+    const result = await this.client.queryRange({
```
StarpTech, line 182: Should we run all those queries in parallel? How fast is it?

controlplane/src/core/repositories/analytics/MetricsDashboardRepository.ts
```diff
@@ -0,0 +1,364 @@
...
+    });
+
+    return {
+      series: this.mapSeries(endDate, range, result.data.result[0].values),
+      errorSeries: this.mapSeries(endDate, range, errorRate.data.result[0].values),
+    };
+  }
+
+  protected getQueryLabels = (params: MetricsParams) => {
+    return `wg_organization_id="${params.organizationId}", wg_federated_graph_id="${params.graphId}", wg_router_graph_name="${params.graphName}"`;
```
StarpTech, line 312: The graph id already unique. Filtering for the graph name is not needed.

controlplane/src/core/repositories/analytics/MetricsDashboardRepository.ts
```diff
@@ -0,0 +1,364 @@
...
+      ...this.getRangeQueryProps({
+        range,
+        endDate,
+      }),
+    });
+
+    const errorRate = await this.client.queryRange({
+      query: `sum(rate(cosmo_router_http_requests_total{${this.getQueryLabels(
+        params,
+      )}, http_status_code!~"2.."}[${prange}]) * 60)`,
```
StarpTech, line 298: We should only count 5xx as errors. Client errors are expected.

controlplane/src/core/repositories/analytics/MetricsDashboardRepository.ts
```diff
@@ -0,0 +1,364 @@
...
+    const prevResult = await this.client.queryRange({
+      query: `sum (rate(cosmo_router_http_requests_total{${this.getQueryLabels(params)}}[${prange}]))`,
+      ...this.getRangeQueryProps({
+        range,
+        endDate: endDate - range * 60 * 60 * 1000,
+      }),
+    });
+
+    return {
+      value: percentage.data.result[0]?.value[1],
```
StarpTech, line 262: That can throw. data might not be set when the request errors.

controlplane/src/core/repositories/analytics/MetricsDashboardRepository.ts
```diff
@@ -0,0 +1,364 @@
...
+   * Get error metrics
+   */
+  public async getErrorMetrics({ range = 24, endDate = getUtcUnixTime(), params }: GetMetricsProps) {
+    const prange = '15m';
+
+    // get error percentage of requests in last [range] hours
+    const percentage = await this.client.query({
+      query: `sum(rate(cosmo_router_http_requests_total{${this.getQueryLabels(
+        params,
+      )}, http_status_code!~"2.."}[${range}h])) / sum(rate(cosmo_router_http_requests_total{${this.getQueryLabels(
```
StarpTech, line 223: Same here.

controlplane/test/api-keys.test.ts
```diff
@@ -57,6 +58,10 @@ describe('API Keys', (ctx) => {
 
     const platformWebhooks = new MockPlatformWebhookService();
 
+    const prometheus = new PrometheusClient({
```
StarpTech, line 61: How will it behave in tests when Prometheus is not running? We should either provide a mock implementation or conditionally disable it when we don't require prometheus in tests.

StarpTech (changes requested):

controlplane/src/core/repositories/analytics/MetricsDashboardRepository.ts
Pagebakers: Locally it's blazing fast, but with larger amounts of load and traffic it might be slower, i will use promise.all

controlplane/src/core/repositories/analytics/MetricsDashboardRepository.ts
Pagebakers: Hmm then the types of the client are incorrect

controlplane/src/core/repositories/analytics/MetricsDashboardRepository.ts
Pagebakers: I've updated the error handling, I'm returning all errors now from the repository methods, should we log them, pass them back to the client?

controlplane/src/core/repositories/analytics/MetricsDashboardRepository.ts
Pagebakers: What do you mean?

controlplane/test/api-keys.test.ts
Pagebakers: Add mock client

controlplane/src/core/bufservices/PlatformService.ts
Pagebakers, line 1296: I'll add promise.all here as well

StarpTech (approved):
LGTM

PR #106, opened 2023-09-21
feat: add logout command

JivusAyrus:

### DIFF
--- a/cli/src/commands/auth/commands/logout.ts
+++ b/cli/src/commands/auth/commands/logout.ts
@@ -3,19 +3,16 @@ import { Command } from 'commander';
 import pc from 'picocolors';
 import { configDir } from '../../../core/config.js';
 import { BaseCommandOptions } from '../../../core/types/types.js';
-import program from '../../index.js';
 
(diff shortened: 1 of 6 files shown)


cli/src/commands/auth/commands/logout.ts
```diff
@@ -0,0 +1,21 @@
...
+import { configDir, configFile } from '../../../core/config.js';
+import program from '../../index.js';
+import { BaseCommandOptions } from '../../../core/types/types.js';
+
+export default (opts: BaseCommandOptions) => {
+  const logoutCommand = new Command('logout');
+  logoutCommand.description('Logout a user.');
+
+  logoutCommand.action(() => {
+    if (!existsSync(configFile)) {
```
StarpTech, line 13: I think we can skip this. Logout can be idempotent.

StarpTech (changes requested):

cli/src/commands/auth/commands/logout.ts
JivusAyrus: Done

StarpTech (approved):
LGTM

PR #108, opened 2023-09-24
feat: expand environmental variables in config.yaml

cs-clarence:
## Motivation and Context
To allow expansion of referenced environmental variables in `config.yaml`:

```yaml
# Header manipulation
headers:
  all: # Header rules for all origin requests.
    request:
      - op: "propagate"
        named: Subgraph-Secret
       # The value will be replaced with the value of SUBGRAPH_SECRET (or empty string if undefined)
        default: "${SUBGRAPH_SECRET}"

```


StarpTech:
Hi @cs-clarence thank you for the PR. That's a great addition.

StarpTech (approved):
LGTM

PR #109, opened 2023-09-24
feat: add support for overriding subgraph routing urls

cs-clarence:
## Motivation and Context
Allows overriding routing urls for subgraphs via `config.yaml`:
```yaml
override_routing_url:
  subgraphs:
    some-subgraph: http://new-subgraph-routing-url.com/graphql
```


StarpTech:
Hi @ cs-clarence, can you give more context on why this feature is valuable to add to the core?

The routing URL is stored in the route config which is fetched from the controlplane. The integrity of the URL is important to enrich requests with proper trace attributes. If we allow arbitrary subgraph URLs we lose the connection.
In order to change the subgraph URL you have to update the subgraph with the wgc cli.

cs-clarence:
Hey @StarpTech, I've seen this feature in the apollo router [here](https://www.apollographql.com/docs/router/configuration/overview#subgraph-routing-urls). 

It would be useful for routers inside a k8s cluster so they can use the cluster-local DNS name while keeping the public DNS name config in the control plane. I understand that this is a niche use case, so feel free to close this if you deem this a not-useful feature.

StarpTech:
> It would be useful for routers inside a k8s cluster so they can use the cluster-local DNS name while keeping the public DNS name config in the control plane. I understand that this is a niche use case, so feel free to close this if you deem this a not-useful feature.

This is a good use case! I'll come back later.

jgautheron:
Got a similar use case here, currently replacing it before deployment with `sed` but would be cleaner to allow an override either by config and/or env var.

StarpTech:
@jgautheron thanks for the feedback. Will look into it.

StarpTech (changes requested):

router/internal/config/config.go
```diff
@@ -83,21 +83,21 @@ type TrafficShapingRules struct {
...
 
 type BackoffJitterRetry struct {
-	Enabled     bool          `yaml:"enabled" default:"true" envconfig:"RETRY_ENABLED"`
-	Algorithm   string        `yaml:"algorithm" default:"backoff_jitter" validate:"oneof=backoff_jitter"`
-	MaxAttempts int           `yaml:"max_attempts" default:"5" validate:"required,min=1,required_if=Algorithm backoff_jitter"`
-	MaxDuration time.Duration `yaml:"max_duration" default:"10s" validate:"required,min=1s,required_if=Algorithm backoff_jitter"`
-	Interval    time.Duration `yaml:"interval" default:"3s" validate:"required,min=100ms,required_if=Algorithm backoff_jitter"`
+	Enabled     bool          `yaml:"enabled"      default:"true"           envconfig:"RETRY_ENABLED"`
+	Algorithm   string        `yaml:"algorithm"    default:"backoff_jitter"                           validate:"oneof=backoff_jitter"`
+	MaxAttempts int           `yaml:"max_attempts" default:"5"                                        validate:"required,min=1,required_if=Algorithm backoff_jitter"`
```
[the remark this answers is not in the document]
cs-clarence: Oh I didn't realize I had it on. Will revert it

router/core/router.go
```diff
@@ -224,9 +227,32 @@ func NewRouter(opts ...Option) (*Router, error) {
 // updateServer starts a new Server. It swaps the active Server with a new Server instance when the config has changed.
 // This method is safe for concurrent use. When the router can't be swapped due to an error the old server kept running.
 func (r *Router) updateServer(ctx context.Context, cfg *nodev1.RouterConfig) error {
+	overrides := make(map[string]string)
+
+	for _, sg := range cfg.Subgraphs {
```
StarpTech, line 232: Put this logic in a method for better readability e.g. `configureSubgraphOverwrites`.

StarpTech (changes requested):

router/core/router.go
```diff
@@ -588,9 +619,17 @@ func (r *Router) newServer(ctx context.Context, routerConfig *nodev1.RouterConfi
 
 		subgraphs := make([]Subgraph, len(routerConfig.Subgraphs))
 		for _, s := range routerConfig.Subgraphs {
-			parsedURL, err := url.Parse(s.RoutingUrl)
+			rawURL := s.RoutingUrl
```
StarpTech, line 622: Move the validation logic into the proposed `configureSubgraphOverwrites` and return the error when a subgraph has an invalid url. Although the URLs are already validated through the validate config package. I'm also fine with both.

StarpTech (changes requested):

router/core/router.go
```diff
@@ -473,6 +528,11 @@ func (r *Router) newServer(ctx context.Context, routerConfig *nodev1.RouterConfi
 	)
 
 	httpRouter := chi.NewRouter()
+	httpRouter.Use(func(h http.Handler) http.Handler {
```
StarpTech, line 531: Mistake? This is an empty middleware.

router/core/router.go
```diff
@@ -221,12 +223,63 @@ func NewRouter(opts ...Option) (*Router, error) {
 	return r, nil
 }
 
+func (r *Router) configureSubgraphOverwrites(cfg *nodev1.RouterConfig) ([]Subgraph, error) {
+	overrides := make(map[string]string)
+
+	// Validate all the URLs first before we start overriding
```
StarpTech, line 229: Much better :+1:

StarpTech (changes requested):

router/core/router.go
cs-clarence: Oh yeah, I forgot to remove it

StarpTech:
@cs-clarence ready for review again?

cs-clarence:
@StarpTech yes

StarpTech:
I continue your work here #152 I found a couple of issues and fixed them there. I hope it's okay. I could not push the branch here.

cs-clarence:
@StarpTech yeah that's alright, it's better for you to take over because although I tested the code, I haven't verified that it works without issues.

StarpTech:
@cs-clarence thank you again for pushing this initiative.

PR #110, opened 2023-09-24
feat: support per subgraph header rules

cs-clarence:
## Motivation and Context
Supports the following in the `config.yaml`:
```yaml
# Header manipulation
headers:
  subgraphs:
    specific-subgraph: # Will only affect this subgraph
      request:
        - op: "propagate"
          named: Subgraph-Secret
         # The value will be replaced with the value of SUBGRAPH_SECRET (or empty string if undefined)
          default: "some-secret"
```

### DIFF
--- a/router/core/header_rule_engine.go
+++ b/router/core/header_rule_engine.go
@@ -2,9 +2,10 @@ package core
 
 import (
 	"fmt"
-	"github.com/wundergraph/cosmo/router/internal/config"
 	"net/http"
 	"regexp"
+
+	"github.com/wundergraph/cosmo/router/internal/config"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


StarpTech:
Hi @cs-clarence thank you for the PR. Would you mind adding a test to header_rule_engine_test?

cs-clarence:
@StarpTech added it

router/core/header_rule_engine_test.go
```diff
@@ -167,3 +169,69 @@ func TestInvalidRegex(t *testing.T) {
...
+						Default:   "Test-Value-1",
+					},
+				},
+			},
+			"subgraph-2": {
+				Request: []config.RequestHeaderRule{
+					{
+						Operation: "propagate",
+						Named:     "X-Test-Subgraph-2",
+						Default:   "Test-Value-2",
```
StarpTech, line 190: Please ensure we have test cases that cover the default and propagation case. Currently, we have two equal test cases.

StarpTech (changes requested):

router/core/header_rule_engine_test.go
StarpTech: In order to resolve this, you would need to remove the default value of one of these subgraphs.

StarpTech (changes requested):

cs-clarence (commented):
Accept changes

StarpTech (approved):
LGTM

PR #114, opened 2023-09-25
feat: add composition library for Go

fiam:
This PR adds a composition library as a Go package, by running our already existing TypeScript composition inside
a JavaScript runtime. The process is entirely opaque for callers of this package, which accepts and returns Go
types.

Benchmarks indicate that when running with the v8 backend, performance is as good as running the composition
algorithm in Node.

## Motivation and Context

The composition library in TypeScript is not convention to run for clients that have all their tooling in Go. This wrapper
provides a turnkey solution that enables them to compose federated graphs from Go.

- [x] Documentation is changed or added on [https://app.gitbook.com/](https://app.gitbook.com/)


devsergiy (approved):

Aenimus (approved):

composition-go/shim/package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "@wundergraph/composition-shim",
+  "version": "1.0.0",
+  "description": "",
+  "main": "./dist/index.js",
+  "types": "./dist/index.d.ts",
+  "files": [
+    "dist"
+  ],
+  "scripts": {
```
StarpTech, line 10: Do we want to release and publish this package?

StarpTech (changes requested):

composition-go/shim/package.json
fiam: No, strictly for bundling it with the Go library, but we reference it in `generate.sh` that's why I gave it a `@wundergraph/*` name.

StarpTech: ok, then you would need to make it a private package to prevent it from publishing.

fiam: I didn't even know that was a thing! I've added it, PTAL.

StarpTech:
@Aenimus This PR implies that every feature we develop in composition has to be tested with the go package and we can't use Node.js specific things anymore. Are you aware of this?

composition-go/composition_test.go
```diff
@@ -0,0 +1,141 @@
...
+			nest: Nested4
+		  }
+		  
+		  type Nested4 {
+			age: Int
+		  }`,
+	},
+}
+
+func normalizeWhiteSpace(s string) string {
```
StarpTech, line 61: Do these tests cover both javascript engines?

composition-go/README.md
```diff
@@ -0,0 +1,80 @@
...
+}
+
+```
+
+## Performance considerations
+
+By default, this package uses the [goja] runtime to run the composition algorithm. This has no
+dependencies and supports all platforms, but depending on your inputs it can be too slow.
+
+For better performance, a [V8] backend is also available guarded under the `wg_composition_v8` build
```
StarpTech, line 75: Could you add an example to the readme?

StarpTech (changes requested):

composition-go/README.md
fiam: Added

composition-go/composition_test.go
StarpTech: Great

PR #115, opened 2023-09-25
fix: improved analytics queries and output

Pagebakers:
…ectly

## Motivation and Context


StarpTech (approved):
LGTM

controlplane/src/core/repositories/analytics/MetricsDashboardRepository.ts
```diff
@@ -321,7 +321,7 @@ export class MetricsDashboardRepository {
         top:
           top5Response.data?.result.map((v: any) => ({
             name: v.metric.wg_operation_name || 'unknown',
-            value: v.value[1] || '0',
+            value: !v.value[1] || v.value[1] === 'NaN' ? '0' : v.value[1],
```
thisisnithin, line 324: Maybe other places would face the same issue? I saw it with P95 latency too. (the big number in the card) We should have a resolveValue fucntion and apply it to every field here where we access value

thisisnithin (changes requested):

controlplane/src/core/repositories/analytics/MetricsDashboardRepository.ts
Pagebakers: Good point

StarpTech (approved):
LGTM

thisisnithin (approved):

PR #139, opened 2023-09-28
fix: collect request metrics as early as possible

StarpTech:
## Motivation and Context


router/core/graphql_handler.go
```diff
@@ -242,11 +243,6 @@ func (h *GraphQLHandler) preparePlan(requestOperationName []byte, shared *pool.S
 	}, nil
 }
 
-func respondWithInternalServerError(w http.ResponseWriter, requestLogger *zap.Logger) {
```
Aenimus, line 245: don't understand this change.

jensneuse (approved):

router/core/graphql_handler.go
StarpTech: We had multiple ways to set a status code. I removed this specific way and went with `w.WriteHeader(statusCode)`.

Aenimus (approved):

PR #174, opened 2023-10-09
feat: schema field level usage analytics

StarpTech:
## TODO

- [x] Store subgraph sourceIDs in clickhouse
- [x] Add service to Helm chart
- [ ] Add documentation
- [x] Add materialized view
- [x] Set SourceID's correctly
- [X] Add service to docker-compose, demo
- [X] Integrate service with release pipeline

----
## Motivation and Context

### Linear

Fixes ENG-4370


graphqlmetrics/server.go
```diff
@@ -0,0 +1,97 @@
...
+
+	return s
+}
+
+func (s *Server) bootstrap() {
+	mux := http.NewServeMux()
+	path, handler := graphqlmetricsv1connect.NewGraphQLMetricsServiceHandler(
+		s.metricsService,
+		// Compress responses with Brotli.
+		brotli.WithCompression(),
```
jensneuse, line 47: Does it make a difference? We're not sending back much, right?

graphqlmetrics/server.go
```diff
@@ -0,0 +1,97 @@
...
+	mux.Handle(path, handler)
+
+	s.server = &http.Server{
+		Addr: s.listenAddr,
+		// https://ieftimov.com/posts/make-resilient-golang-net-http-servers-using-timeouts-deadlines-context-cancellation/
+		ReadTimeout:       1 * time.Minute,
+		WriteTimeout:      2 * time.Minute,
+		ReadHeaderTimeout: 20 * time.Second,
+		// Use h2c so we can serve HTTP/2 without TLS.
+		Handler:  h2c.NewHandler(mux, &http2.Server{}),
```
jensneuse, line 58: Is this a good setup for prod as well?

proto/wg/cosmo/graphqlmetrics/v1/graphqlmetrics.proto
```diff
@@ -0,0 +1,48 @@
+// https://protobuf.dev/programming-guides/style/
+
+syntax = "proto3";
+
+package wg.cosmo.graphqlmetrics.v1;
+
+import "wg/cosmo/common/common.proto";
+
+message SchemaUsageInfo {
+  string OperationDocument = 1;
```
jensneuse, line 10: I would name this RequestDocument.
(A GraphQL Document can contain Operations as per the GraphQL spec)

proto/wg/cosmo/graphqlmetrics/v1/graphqlmetrics.proto
```diff
@@ -0,0 +1,48 @@
...
+  repeated TypeFieldUsageInfo TypeFieldMetrics = 2;
+  OperationInfo OperationInfo = 3;
+  RequestInfo RequestInfo = 4;
+  map<string,string> Attributes = 6;
+}
+
+message OperationInfo {
+  string OperationHash = 1;
```
jensneuse, line 18: Remove all prefixes (Operation)? e.g. just Hash

proto/wg/cosmo/graphqlmetrics/v1/graphqlmetrics.proto
```diff
@@ -0,0 +1,48 @@
...
+  string OperationName = 2;
+  string OperationType = 3;
```
jensneuse, line 20: Make Type an Enum?

proto/wg/cosmo/graphqlmetrics/v1/graphqlmetrics.proto
```diff
@@ -0,0 +1,48 @@
...
+}
+
+message RequestInfo {
+  // FederatedGraphID and OrganizationID are transport over JWT
+  string RouterConfigVersion = 3;
+}
+
+message TypeFieldUsageInfo {
```
jensneuse, line 28: Please add comments to all fields to properly explain them.

proto/wg/cosmo/graphqlmetrics/v1/graphqlmetrics.proto
```diff
@@ -0,0 +1,48 @@
...
+
+message PublishGraphQLRequestMetricsRequest {
+  repeated SchemaUsageInfo SchemaUsage = 1;
+}
+
+message PublishOperationCoverageReportResponse {
+}
+
+service GraphQLMetricsService {
+  rpc PublishGraphQLMetrics(PublishGraphQLRequestMetricsRequest) returns (PublishOperationCoverageReportResponse) {}
```
jensneuse, line 47: I think we should distinguish between "registering" a Document with TypeFieldUsageInfo and publishing the actual usage with attributes, etc.
Sending the Document as a string for each request will quickly become a problem, even with compression.
You've added a "Count" value, but this will only count up if the attributes are the same, right?
In addition, do you also want to report the request IDs? This could make batching worse because of cardinality.

jensneuse: One way you could solve this is by adding one more rpc endpoint that only accepts the Operation Hash and the Document. When publishing the Schema Usage, you simply omit the Document and only publish the Hash. WDYT?

router/core/graphql_handler.go
```diff
@@ -159,12 +169,43 @@ func (h *GraphQLHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		for i := range preparedPlan.schemaUsageInfo.TypeFields {
+			fieldUsageInfos[i] = &graphqlmetricsv1.TypeFieldUsageInfo{
+				Count:     1,
+				Path:      preparedPlan.schemaUsageInfo.TypeFields[i].Path,
+				TypeNames: preparedPlan.schemaUsageInfo.TypeFields[i].TypeNames,
+				SourceIDs: preparedPlan.schemaUsageInfo.TypeFields[i].Source.IDs,
+			}
+		}
+
+		h.gqlMetricsExporter.Record(&graphqlmetricsv1.SchemaUsageInfo{
```
jensneuse, line 193: At this point, I would only register the Hash and the Document.

router/core/graphql_handler.go
```diff
@@ -159,12 +169,43 @@ func (h *GraphQLHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		if !ok {
+			requestLogger.Error("unexpected prepared plan type")
 			w.WriteHeader(http.StatusInternalServerError)
+			writeRequestErrors(graphql.RequestErrorsFromError(internalServerErrorErr), w, requestLogger)
 			return
 		}
-		preparedPlan = sharedPreparedPlan.(planWithExtractedVariables)
+	}
+
+	if h.gqlMetricsExporter != nil {
```
jensneuse, line 181: How about moving this logic away from here and just calling one func?

router/internal/graphqlmetrics/batch_queue.go
```diff
@@ -0,0 +1,163 @@
...
+		stopped:         &atomic.Bool{},
+		dispatchedCount: 0,
+	}
+
+	return bq
+}
+
+// Enqueue adds an item to the queue. Returns false if the queue is stopped or not ready to accept items
+func (b *BatchQueue) Enqueue(item any) bool {
+	if b.stopped.Load() {
```
jensneuse, line 79: You're currently not evaluating the result of Enqueue in the http handler.
Is it really necessary to have this atomic and load it?
If the server behaviour will not adopt to it, could you not just omit it?

router/internal/graphqlmetrics/batch_queue.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+
+	if o.MaxQueueSize == 0 {
+		o.MaxQueueSize = defaultMaxQueueSize
+	}
+}
+
+// BatchQueue coordinates dispatching of queue items by time intervals
+// or immediately after the batching limit is met.
+type BatchQueue struct {
```
jensneuse, line 41: How confident are you that the implementation is "good"?
I could try and analyze it more, but would like to first ask if you have evaluated existing solutions?
Is there no batch library that is tested and does what we need?

jensneuse: I'll postpone evaluating the batch impl until answered.

jensneuse (changes requested):

router/internal/graphqlmetrics/batch_queue.go
StarpTech: Most of the code and tests were copied from https://github.com/wind-c/bqueue. I researched alternatives, but this is the only implementation I found good enough for our use case. It has to support

- Non-blocking
- Dispatching by interval and batch limit

router/internal/graphqlmetrics/batch_queue.go
StarpTech: We can omit it.

StarpTech: Done

graphqlmetrics/server.go
StarpTech: This is required to decompress brotli. The comment is misleading.

StarpTech: Fixed

graphqlmetrics/server.go
StarpTech: Good question. I have no experience in regard to buf connect. It is supported in our infrastructure. We can also use HTTP1.1

proto/wg/cosmo/graphqlmetrics/v1/graphqlmetrics.proto
StarpTech: All good points. Need to think about it a bit longer.

proto/wg/cosmo/graphqlmetrics/v1/graphqlmetrics.proto
StarpTech: Done

proto/wg/cosmo/graphqlmetrics/v1/graphqlmetrics.proto
StarpTech: Done

proto/wg/cosmo/graphqlmetrics/v1/graphqlmetrics.proto
StarpTech: Requires additional mapping but could be worth it from a protocol perspective.

StarpTech: Done

proto/wg/cosmo/graphqlmetrics/v1/graphqlmetrics.proto
StarpTech: Done

graphqlmetrics/server.go
StarpTech: Using HTTP1.1 for now.

router/core/graphql_handler.go
StarpTech: Done

router/internal/graphqlmetrics/batch_queue.go
StarpTech: Please test it :smiley_cat:

router/core/graphql_handler.go
StarpTech: As discussed. Let's keep super optimizations out of the initial version. At the collector and database level, we already ensure that operations aren't stored for every request. The exporter won't send the requested document for the same usage twice but for every batch which is done by default every 10s or every 512 requests in batches. This makes the current implementation very simple because we don't need to coordinate it with the collector.

proto/wg/cosmo/graphqlmetrics/v1/graphqlmetrics.proto
StarpTech: I refactored the aggregation logic. Now, only exact same schema usage information is aggregated. Please check.
For the second comment please see https://github.com/wundergraph/cosmo/pull/174#discussion_r1359881346

fiam (approved):
LGTM

PR #184, opened 2023-10-13
feat: auto ignore schema errors for check command if github is integrated

thisisnithin:
## Motivation and Context

We do not want to fail the command in the case we have a github integration. Only the commit check should be marked unsuccessful instead.


controlplane/src/core/bufservices/PlatformService.ts
```diff
@@ -2988,5 +2989,52 @@ export default function (opts: RouterOptions): Partial<ServiceImpl<typeof Platfo
...
+              code: EnumStatusCode.ERR_NOT_FOUND,
+              details: 'Organization not found',
+            },
+            isInstalled: false,
+          };
+        }
+
+        const githubRepository = new GitHubRepository(opts.db, opts.githubApp);
+        const isInstalled = await githubRepository.isAppInstalledOnRepo({
+          accountId: req.gitInfo?.accountId ?? '',
```
StarpTech, line 3026: How can it be empty?

cli/src/commands/subgraph/commands/check.ts
```diff
@@ -41,6 +41,19 @@ export default (opts: BaseCommandOptions) => {
...
+    if (gitInfo) {
+      const integrationCheckResponse = await opts.client.platform.isGitHubAppInstalled(
+        {
+          gitInfo,
+        },
+        {
+          headers: baseHeaders,
+        },
+      );
+      ignoreErrors = integrationCheckResponse.isInstalled;
```
StarpTech, line 54: Could we also log a message that the result depends on the PR check? This helps in debugging scenarios.

StarpTech (changes requested):

controlplane/src/core/bufservices/PlatformService.ts
thisisnithin: It can never be, atleast in this context. protobuf generates optional fields sadly. Added a check just in case.

cli/src/commands/subgraph/commands/check.ts
thisisnithin: Done

controlplane/src/core/bufservices/PlatformService.ts
StarpTech: ok

StarpTech (approved):
LGTM

PR #185, opened 2023-10-13
feat: add support for subscriptions

fiam:
- fix: paths to schemas when creating demo during initial setup
- fix: ammend command
- chore: add dummy CI check
- feat: add subscription support to demos
- feat: update graphiql
- feat: add support for serving susbcriptions over websockets

## Motivation and Context


Aenimus (approved):

proto/wg/cosmo/node/v1/node.proto
```diff
@@ -195,10 +195,18 @@ message MTLSConfiguration {
   bool insecureSkipVerify = 3;
 }
 
+enum GraphQLSubscriptionProtocol {
```
StarpTech, line 198: Configuring the router through this is legacy because the router should provide the config (ENV, config.yaml) but on the other hand, I see great potential in it too e.g. we can control certain values from control-plane. Overwriting them in the router is always possible. I'm open for both but we should have a plan. Let's discuss.

router/core/operation_metrics.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/wundergraph/cosmo/router/internal/otel"
+	"go.opentelemetry.io/otel/attribute"
+	semconv "go.opentelemetry.io/otel/semconv/v1.21.0"
+	"go.opentelemetry.io/otel/trace"
+)
+
+type OperationProtocol int
+
+const (
+	OperationProtocolHTTP OperationProtocol = iota + 1
```
StarpTech, line 19: Why not use just string constants if we never benefit from this structure?

StarpTech (changes requested):
Rudimentary review. I'll test it more in-depth later.

router/core/operation_metrics.go
fiam: They have a String() method which is what is written to the traces, but I agree with you. Let's cut the middleman and make them strings directly.

proto/wg/cosmo/node/v1/node.proto
fiam: Another option I explored was allowing configuration options to be passed in the router URL. e.g. http://my-subgraph.example.com/graphql#subscribe_with=sse - It has the advantage that wherever we accept a subgraph URL we automatically accept any options too, but it's less discoverable than a separate field in the configuration. Keep in mind that in any case we can always override all this by generating a static router configuration locally on demand. What do you think about the URL approach?

router/core/operation_metrics.go
fiam: Changed

router/internal/graphiql/README.hacking.md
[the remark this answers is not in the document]
fiam: Renamed

proto/wg/cosmo/node/v1/node.proto
StarpTech: I like it but I'd not ship it in the initial version. My question here is: Should we stop storing router configurations in the config and only use it for data-source information?

StarpTech: As clarified. We go with the current proposal that the control plane has the power to change it.

demo/pkg/subgraphs/demo.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/99designs/gqlgen/graphql/handler"
+	"github.com/99designs/gqlgen/graphql/handler/extension"
+	"github.com/99designs/gqlgen/graphql/handler/transport"
+	"github.com/gorilla/websocket"
+)
+
+func NewDemoServer(schema graphql.ExecutableSchema) *handler.Server {
+	srv := handler.New(schema)
+	// SSE needs to be the first transport
+	srv.AddTransport(transport.SSE{})
```
jensneuse, line 17: Why? Can you explain in the comment?

router/core/graphql_prehandler.go
```diff
@@ -113,168 +85,36 @@ func (h *PreHandler) Handler(next http.Handler) http.Handler {
...
-		err = printer.Print(doc, h.executor.Definition, hash)
-		if err != nil {
-			requestLogger.Error("unable to print document", zap.Error(err))
-			statusCode = http.StatusInternalServerError
-			w.WriteHeader(statusCode)
-			writeRequestErrors(graphql.RequestErrorsFromError(internalServerErrorErr), w, requestLogger)
+			var reportErr ReportError
+			var inputErr *inputError
+			switch {
+			case errors.As(err, &inputErr):
```
jensneuse, line 93: Should it be errors.Is?

router/core/graphql_prehandler.go
```diff
@@ -113,168 +85,36 @@ func (h *PreHandler) Handler(next http.Handler) http.Handler {
...
+				statusCode = http.StatusBadRequest
+				requestLogger.Error(inputErr.Error())
+				w.WriteHeader(statusCode)
+				w.Write([]byte(inputErr.Error()))
+			case errors.As(err, &reportErr):
+				report := reportErr.Report()
```
jensneuse, line 99: The report can contain internal and external errors.
According to the graphql-over-http spec, external errors should be propagated as 200 OK, whereas internal errors should be propagated as 500.

router/core/wsproto/proto.go
```diff
@@ -0,0 +1,57 @@
+package wsproto
```
jensneuse, line 1: Why did you re-implement this in the router? Is it not already implemented in graphql-go-tools?

jensneuse (approved):

router/internal/graphiql/playground.go
```diff
@@ -7,6 +7,6 @@ import (
 //go:embed graphiql.html
 var s string
 
-func GetGraphiqlPlaygroundHTML() string {
+func PlaygroundHTML() string {
```
StarpTech, line 10: Subscriptions in the Studio don't work

image

StarpTech (changes requested):

router/core/router.go
```diff
@@ -660,28 +680,28 @@ func (r *Router) newServer(ctx context.Context, routerConfig *nodev1.RouterConfi
 			subChiRouter.Use(traceHandler.Handler)
 		}
 
+		subChiRouter.Use(NewWebsocketMiddleware(rootContext, WebsocketMiddlewareOptions{
```
StarpTech, line 683: In the last meeting, we decided to not call the module hooks unless they are working as intended. Currently, the OnOriginRequest is called for upgrade requests. OnOriginRequest should only be called for subgraph requests.

StarpTech (changes requested):

router/internal/graphiql/playground.go
jensneuse: Does your router already support them?

router/core/wsproto/graphql_ws.go
StarpTech, line 1: Please move this to `internal`. The current design is to put all packages to internal that contribute to the core package.

StarpTech (changes requested):

router/core/operation_parser.go
```diff
@@ -0,0 +1,159 @@
...
+		executor: executor,
+		documentPool: &sync.Pool{
+			New: func() interface{} {
+				return ast.NewSmallDocument()
+			},
+		},
+	}
+}
+
+func (p *OperationParser) ParseReader(r io.Reader) (*ParsedOperation, error) {
```
StarpTech, line 55: It's not used anywhere

router/core/operation_parser.go
```diff
@@ -0,0 +1,159 @@
...
+	defer pool.PutBytesBuffer(buf)
+	if _, err := io.Copy(buf, r); err != nil {
+		return nil, fmt.Errorf("failed to read request body: %w", err)
+	}
+	body := buf.Bytes()
+	return p.Parse(body)
+}
+
+func (p *OperationParser) Parse(body []byte) (*ParsedOperation, error) {
+	requestQuery, _ := jsonparser.GetString(body, "query")
```
StarpTech, line 66: During the merging of ["limit request body size"](https://github.com/wundergraph/cosmo/pull/185/commits/e2f3c685fc1240a6a5f37e95d7fa20afb2f54745#diff-0c6c16945a1e9a216631c6bd16bcfbe9c07a6c96b1cf9b3aa93385436fab41e8R65) you forgot to move the logic to this place to respect the limit for WS as well.

StarpTech (changes requested):

router/internal/graphiql/playground.go
StarpTech: Yes, @fiam just forgot to update the playground to support subscriptions.

proto/wg/cosmo/node/v1/node.proto
```diff
@@ -195,10 +195,18 @@ message MTLSConfiguration {
...
+  GRAPHQL_SUBSCRIPTION_PROTOCOL_SSE_POST = 2;
+}
+
 message GraphQLSubscriptionConfiguration {
   bool enabled = 1;
   ConfigurationVariable url = 2;
-  bool useSSE = 3;
+  /* @deprecated - Kept for backwards compatibility when decoding. Use protocol instead. */
+  optional bool useSSE = 3;
+  GraphQLSubscriptionProtocol protocol = 4;
```
StarpTech, line 209: How will an old config behave on the router? Is it backward compatible? Please test.

StarpTech (changes requested):

StarpTech:
How did you handle graceful shutdown of WS connections when the schema is updated in between?

demo/pkg/subgraphs/demo.go
fiam: gqlgen requires it, I'd guess because other transports might incorrectly handle the request otherwise or modify the request, but I haven't dug into their code. I found it here https://gqlgen.com/recipes/subscriptions/. I've clarified the comment.

router/core/graphql_prehandler.go
fiam: No, because we do use `inputErr` inside the case. However, we should try to convert the to interface instead of the concrete type. I've amended it.

fiam:
> How did you handle graceful shutdown of WS connections when the schema is updated in between?

All websocket connections check a context while awaiting for messages. If that context is cancelled, then the read is aborted and the connection closed from the server side https://github.com/wundergraph/cosmo/blob/85555d55365e4d916ad9ac01bc1c639345e5bc44/router/core/websocket.go#L97 - The context is initialized here and cancelled during shutdown https://github.com/wundergraph/cosmo/blob/85555d55365e4d916ad9ac01bc1c639345e5bc44/router/core/router.go#L529

router/core/wsproto/proto.go
fiam: graphql-go-tools implements a client, while in the router we have to implement the server side. I could have taken the struct types and the message strings but I felt that graphql-go-tools should probably not export those at all. Additionally, the types in graphql-go-tools are named incorrectly in my opinion. Turns out the protocol named `graphql-ws` uses a websocket subprotocol called `graphql-transport-ws` while the protocol named `subscriptions-transport-ws` uses a websocket subprotocol named `graphql-ws`. The implementation in graphql-go-tools chose to use name the types after the websocket subprotocol, and I think using the official names is slightly less confusion.

router/core/wsproto/graphql_ws.go
fiam: I've moved it to internal

proto/wg/cosmo/node/v1/node.proto
fiam: I've tested it with an static config generated with wgc from npx. It works and uses ws for the subscriptions (which is the default)

router/core/router.go
fiam: Good catch. I've amended it to skip upgrade requests.

router/core/operation_parser.go
fiam: It stopped being used after merging the limited reader PR, I've removed this function

router/core/operation_parser.go
fiam: I'd rather have the upper layer do the request limiting, to provide better context for the errors. I'm now passing the limit to the websocket handler and producing an error there.

StarpTech (approved):
LGTM

PR #187, opened 2023-10-13
fix: allow absolute paths for the wgc compose output file

fiam:
The join is not necessary for relative paths and breaks absolute paths

## Motivation and Context


cli/src/commands/router/commands/compose.ts
```diff
@@ -99,7 +99,7 @@ export default (opts: BaseCommandOptions) => {
     });
 
     if (options.out) {
-      await writeFile(join(process.cwd(), options.out), routerConfig.toJsonString());
```
StarpTech, line 102: Is it backwards compatible? Wouldn't `resolve()` solve it too?

StarpTech (changes requested):

cli/src/commands/router/commands/compose.ts
fiam: Arguably, `pathe.join()` should be smart enough to yield the absolute path when `process.out` is absolute but that's not the case.

On the other hand, there's no need to prepend the current directory to `options.out`, a relative path is already resolved relative to the current working directory. Joining `cwd()` with it manually is just incorrect.

fiam: Forgot to answer your actual question 🙈. The change is backwards compatible, it only affects absolute paths which are not working correctly right now (they're interpreted as relative to the current working directory)

StarpTech (approved):
LGTM

PR #190, opened 2023-10-16
feat: secure api from simple overload attacks

StarpTech:
## Motivation and Context

While this approach is obviously not bulletproof it provides some protection from misuse. Query-cost analysis is the next logical step but even with there must exist some upper limit. I think 5MB is reasonable and big enough to run into any shortcomings from regular usage.

### DIFF
--- a/router/core/graphql_prehandler.go
+++ b/router/core/graphql_prehandler.go
@@ -38,17 +38,19 @@ type PreHandlerOptions struct {
 }
 
 type PreHandler struct {
-	log            *zap.Logger
-	executor       *Executor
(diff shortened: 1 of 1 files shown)


fiam (approved):
LGTM

jensneuse:
Sounds good. I'd make it configurable though. If someone runs into a limitation, we don't want to have to ship an update.

StarpTech:
@jensneuse addressed. I added support for human-friendly byte size syntax.

jensneuse (approved):
Don't forget to update docs.

PR #200, opened 2023-10-24
feat: add support for authentication via JWKS

fiam:
This adds support for authenticating users via JWKS. By default, requests without authentication are still allowed, but
they can be optionally disabled by requiring authentication in the authorization configuration.

When authentication succeeds, its result - including claims and the provider that authenticated the request - are propagated to the context.Context associated with the current http.Request.

## Motivation and Context


router/authentication/jwks.go
```diff
@@ -0,0 +1,98 @@
+package authentication
```
StarpTech, line 1: Can we put this package to `internals`?

router/core/router.go
```diff
@@ -493,6 +497,37 @@ func (r *Router) bootstrap(ctx context.Context) error {
...
+			if err != nil {
+				w.WriteHeader(http.StatusForbidden)
+				_, _ = io.WriteString(w, http.StatusText(http.StatusForbidden))
+				return
+			}
+			if auth != nil {
+				r = r.WithContext(authentication.NewContext(r.Context(), auth))
+				w.Header().Set("X-Authenticated-By", auth.Authenticator())
+			} else if authenticationRequired {
+				w.WriteHeader(http.StatusUnauthorized)
```
StarpTech, line 522: This needs to be aligned with https://github.com/wundergraph/cosmo/blob/main/router/core/graphql_handler.go#L301 to track errros and respond in proper GraphQL response. Any particular reason, why you don't implement it as a custom module inside `Middleware(ctx RequestContext, next http.Handler)` ? In that way, you can add this module at server startup.

StarpTech (changes requested):

router/authentication/jwks.go
fiam: No, otherwise we can't call it from the integration tests

StarpTech: Why do we have to? An integration test must not rely on internals.

router/cmd/custom-jwt/module/module.go
```diff
@@ -0,0 +1,79 @@
...
+	// Register your module here
+	core.RegisterModule(&JWTModule{})
+}
+
+const ModuleID = "com.example.custom-jwt"
+
+// JWTModule is a module that signs outgoing requests with a JWT token
+// based on the authentication information of the received request
+type JWTModule struct {
+	SecretKey string `mapstructure:"secret_key"`
```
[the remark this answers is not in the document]
fiam: Done

router/cmd/custom-jwt/module/module.go
```diff
@@ -0,0 +1,79 @@
...
+		claims := jwt.MapClaims(auth.Claims())
+		if claims == nil {
+			claims = make(jwt.MapClaims)
+		}
+		claims["iss"] = "cosmo-router"
+		t := jwt.NewWithClaims(jwt.SigningMethodHS256, claims)
+		signed, err := t.SignedString([]byte(m.SecretKey))
+		if err != nil {
+			return nil, &http.Response{
+				StatusCode: http.StatusInternalServerError,
```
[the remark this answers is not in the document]
fiam: Done

router/cmd/custom-jwt/module/module.go
```diff
@@ -0,0 +1,79 @@
...
+				Body:       io.NopCloser(strings.NewReader(fmt.Sprintf("signing token: %s", err))),
```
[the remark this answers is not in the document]
fiam: Done

router/cmd/custom-jwt/module/module.go
```diff
@@ -0,0 +1,79 @@
...
+			}
+		}
+		request.Header.Add("Authorization", "Bearer "+signed)
```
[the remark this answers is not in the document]
fiam: Done

router/authentication/jwks.go
StarpTech: Can we split it out so we don't need to reference the router code?

router/core/router.go
StarpTech: Smiliar as in https://github.com/wundergraph/cosmo/blob/main/router/core/header_rule_engine.go.

router/authentication/jwks.go
fiam: We have no means to set the router configuration from a config file other than `cmd/router/main.go`, but this starts the whole router. For integration tests we call directly into `core.NewRouter()` which requires the arguments to be already parsed. In order to pass an array of `authentication.Authenticator()`, the package needs to be importable by `router-tests`. If we move it to `internal`, then `router-tests` can't pass authenticators.

StarpTech: ok

router-tests/jwks/jwks.go
```diff
@@ -0,0 +1,91 @@
...
+		Type:      signingMethodType,
+		Algorithm: signingMethod.Name,
+		Use:       "sig",
+		ID:        jwtKeyID,
+		Exponent:  base64.URLEncoding.EncodeToString(big.NewInt(int64(s.privateKey.E)).Bytes()),
+		Modulus:   base64.URLEncoding.EncodeToString(s.privateKey.N.Bytes()),
+	}
+	data, err := json.Marshal(jsonWebKeySet{Keys: []jsonWebKey{k}})
+	if err != nil {
+		panic(err)
```
Aenimus, line 69: Just checking we definitely want to panic here.

router/authentication/doc.go
```diff
@@ -0,0 +1,3 @@
+// Package authentication implements the types and functions
+// used to load and validate user authentication
+package authentication
```
Aenimus, line 3: Empty package?

Aenimus: If you already saw a comment by me on this, please ignore it. It looked to me like my first attempt didn't send.

router/core/router.go
fiam: As discussed, this has been refactored to play nicely with metrics and tracing

router-tests/jwks/jwks.go
fiam: We do, this is just for testing

router/authentication/doc.go
fiam: This is a convention for Go packages. The package documentation goes into a file called `doc.go` which has nothing else. See, for example, net/http https://cs.opensource.google/go/go/+/master:src/net/http/doc.go

StarpTech (approved):
LGTM

Issue #202, opened 2023-10-25
Support linux/arm64/v8


harrytang:
### Component(s)

router, controlplane, studio

### Is your feature request related to a problem? Please describe.

Hi Team,

Could you add multi-arch (arm64) support to the Docker images soon? This will help Cosmo reaches more users on ARM64 systems. Thoughts?

Thanks,
Harry

### Describe the solution you'd like

multi-arch support

Aenimus:
Hi @harrytang,

As mentioned in [your PR](https://github.com/wundergraph/cosmo/pull/204), we will work on multi-architecture images soon.

Thank you!

PR #204, opened 2023-10-25
feat: adds multi platforms image building

harrytang:
this fixes the issue https://github.com/wundergraph/cosmo/issues/202


StarpTech:
Hi @harrytang, did you test the change? It will not work because the images must be written for multi-arch in mind. Therefore, the CI/CD would be super slow. We will work on multi arch images very soon. Thank you for your contribution!

PR #206, opened 2023-10-25
chore(ci): add staticcheck and go vet linter

StarpTech:
## Motivation and Context


.github/workflows/router-ci.yaml
```diff
@@ -49,6 +49,17 @@ jobs:
...
+      - name: Run go vet
+        run: "go vet ./..."
+        working-directory: ./router
+
+      - name: Run staticcheck linter
+        uses: dominikh/staticcheck-action@v1.3.0
+        with:
+          version: "2023.1.6"
+          install-go: false
+          working-directory: ./router
```
fiam, line 61: Can we run these in router-tests too? Maybe it is more convenient to put both checks into a reusable action, then call it from `graphqlmetrics-ci` once and thrice from `router-ci` (router, router-tests and demo)

StarpTech: It is already an action. So you want to put an action into an action? :smile:

router/internal/unsafebytes/unsafebytes.go
```diff
@@ -1,52 +1,18 @@
...
-	out, _ := strconv.ParseInt(*(*string)(unsafe.Pointer(&byteSlice)), 10, 64)
-	return out
-}
-
-func BytesToFloat32(byteSlice []byte) float32 {
-	out, _ := strconv.ParseFloat(*(*string)(unsafe.Pointer(&byteSlice)), 64)
-	return float32(out)
-}
-
+// BytesToString converts a byte slice to a string without a memory allocation.
```
StarpTech, line 7: Available since go 1.20 https://blog.devops.dev/fast-string-to-byte-and-byte-to-string-conversion-in-go-1-20-85bdb859ee67

StarpTech: To fix vet issues

.github/workflows/router-ci.yaml
StarpTech: I have added it to the router tests too.

fiam: I mean an action for vet + staticcheck. The idea is that we apply the same set of linters across the codebase. If we copy and paste the block across workflows, it might get out of sync.

fiam (approved):
LGTM

.github/workflows/router-ci.yaml
fiam: Discussed via messages

PR #212, opened 2023-10-27
feat: add support for header forwarding with websocket requests

fiam:
- chore: add new test1 subgraph to demo
- feat: add support for header forwarding with websocket requests

## Motivation and Context


fiam:
Needs a graphql-go-tools bump after https://github.com/wundergraph/graphql-go-tools/pull/647 is merged

demo/pkg/injector/context.go
```diff
@@ -0,0 +1,33 @@
+package injector
```
StarpTech, line 1: The name is very generic. Can it be more specific? What does it inject?

router/core/factoryresolver.go
```diff
@@ -200,6 +209,40 @@ func (l *Loader) Load(engineConfig *nodev1.EngineConfiguration) (*plan.Configura
 					subscriptionUseSSE = *in.CustomGraphql.Subscription.UseSSE
 				}
 			}
+			// Header rewrite rules are specified by subgraph name, but data source
+			// only has routing and subscription URLs, so we must manually match them
+			var dataSourceRules []config.RequestHeaderRule
```
StarpTech, line 214: Please put this logic into a separate function. `header_rule_engine.go` could be a good place.

demo/pkg/injector/ws.go
```diff
@@ -0,0 +1,14 @@
+package injector
+
+import (
+	"context"
+
+	"github.com/99designs/gqlgen/graphql/handler/transport"
+)
+
+func InitPayloadFunc(ctx context.Context, initPayload transport.InitPayload) (context.Context, *transport.InitPayload, error) {
```
StarpTech, line 9: Is this only used for testing? If so, please put it in a test package.

demo/pkg/injector/context.go
```diff
@@ -0,0 +1,33 @@
...
+
+func NewContextWithHeader(ctx context.Context, hdr http.Header) context.Context {
+	return context.WithValue(ctx, headerKey, hdr)
+}
+
+func NewContextWithInitPayload(ctx context.Context, initPayload transport.InitPayload) context.Context {
+	return context.WithValue(ctx, payloadKey, initPayload)
+}
+
+func Header(ctx context.Context) http.Header {
```
StarpTech, line 25: Used at all?

demo/pkg/injector/context.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"context"
+	"net/http"
+
+	"github.com/99designs/gqlgen/graphql/handler/transport"
+)
+
+type contextKey string
+
+const (
```
StarpTech, line 12: Could you explain a bit why we need to store the payload and header?

StarpTech (changes requested):

demo/pkg/injector/ws.go
fiam: No, it needs to be able to run in the demo subgraphs as a normal program, otherwise you can't run them standalone.

demo/pkg/injector/context.go
fiam: It's used here https://github.com/wundergraph/cosmo/blob/dfa1687064a3404547d4cab9f422a2bb0c795984/demo/pkg/subgraphs/test1/subgraph/schema.resolvers.go#L20

demo/pkg/injector/context.go
fiam: We need it because gqlgen doesn't expose the request header nor the WS initial payload to resolver functions. I'll add a doc.go file explaining that.

demo/pkg/injector/context.go
fiam: It injects HTTP headers and initial WS payload into the context.Context so the schema resolvers can read that information (gqlgen doesn't pass it). What name do you have in mind?

demo/pkg/injector/context.go
StarpTech: Oh it is part of the demo. I didn't see it.

router/core/factoryresolver.go
fiam: I've updated it, PTAL

demo/pkg/injector/context.go
StarpTech: Fine for me. I missed the context. I comment why we created this package would be great.

fiam: I've just added a doc.go file to the package with a bit of context https://github.com/wundergraph/cosmo/pull/212/files#diff-b671d1b1d70e1a60cf24e556532a391eeafa95b8fd18c59cea390153f4670250

StarpTech (approved):
LGTM

PR #218, opened 2023-10-31
feat: add support for router config generation to composition-go

fiam:
Also, use it to generate the router configuration when running
integration tests.

## Motivation and Context


composition-go/composition.go
```diff
@@ -41,9 +50,60 @@ var (
...
+		req.Header.Add("Content-Type", "application/json")
+		resp, err := http.DefaultClient.Do(req)
+		if err != nil {
+			return nil, fmt.Errorf("could not retrieve schema for subgraph %s: %w", subgraph.Name, err)
+		}
+		defer resp.Body.Close()
+		if resp.StatusCode != http.StatusOK {
+			return nil, fmt.Errorf("could not retrieve schema for subgraph %s: unexpected status code %d", subgraph.Name, resp.StatusCode)
+		}
+		var sdlResult struct {
```
jensneuse, line 73: How about extracting this into a struct?

composition-go/composition.go
```diff
@@ -41,9 +50,60 @@ var (
...
 )
 
+func updateSchemas(subgraphs []*Subgraph) ([]*Subgraph, error) {
+	updatedSubgraphs := make([]*Subgraph, 0, len(subgraphs))
+	for _, subgraph := range subgraphs {
+		if subgraph.Schema != "" {
+			updatedSubgraphs = append(updatedSubgraphs, subgraph)
+			continue
+		}
+		req, err := http.NewRequest("POST", subgraph.URL, strings.NewReader(sdlQuery))
```
jensneuse, line 60: How about moving the introspection of a subgraph into a separate function?

jensneuse (approved):

composition-go/composition.go
fiam: Sounds reasonable, I've moved it to a type

composition-go/composition.go
fiam: Good idea, that way we can also simplify the error messages a bit. I've refactor it, PTAL.

fiam:
> Looks great except I miss comprehensive documentation of keeping in sync with composition library and using the go library to generate the config. Currently, it is spread around different folders. A README with some instructions would be great.

I've added a new section to the README that documents how to update the bundled code, PTAL.

StarpTech (approved):
LGTM

PR #224, opened 2023-11-03
feat: automatically translate failed requests to localhost to docker.host.internal

fiam:
When running inside docker, if a request to localhost fails with ECONNREFUSED retry
the request to docker.host.internal, returning the response if it succeeds but returning
the original error if it fails.

This is enabled by default, but only loaded when actually running inside docker, and
can be disabled with a new setting that defaults to true.

## Motivation and Context


StarpTech (changes requested):
I'm strongly against this change. A user can connect to the host by overwriting their subgraph URLs in the config. We only have to document it. 

```
override_routing_url:
	subgraphs:
	  products: host.docker.internal:4001
```

We should not start to make our core logic aware of a specific environment. I think managing this with a manual overwrite is a good solution for an edge problem. 

https://docs.docker.com/desktop/networking/#i-cannot-ping-my-containers

StarpTech:
I changed my mind after @jensneuse @fiam  explained the onboarding issues when running the router along subgraphs that run on localhost. I have some concerns about the implementation. I'll create a different review.

router/core/transport.go
```diff
@@ -84,68 +96,159 @@ func (ct *CustomTransport) RoundTrip(req *http.Request) (*http.Response, error)
 	return resp, err
 }
 
+func isRunningInsideDocker() bool {
```
StarpTech, line 99: Please let us put this logic in a separate package so it doesn't bother us here too much. A simple test would be also great.

router/core/transport.go
```diff
@@ -84,68 +96,159 @@ func (ct *CustomTransport) RoundTrip(req *http.Request) (*http.Response, error)
...
+	// Check if we are running inside docker by
+	// testing by checking if /.dockerenv exists
+	st, err := os.Stat("/.dockerenv")
+	return err == nil && !st.IsDir()
+}
+
+// localhostFallbackRoundTripper is an http.RoundTripper that will retry failed
+// requests to localhost by rewriting the request to use is targetHost. Only
+// requests that fail with ECONNREFUSED will be retried.
+type localhostFallbackRoundTripper struct {
```
StarpTech, line 109: Please make name it like `DockerHostTransport` to make it clear it is related to docker only.

router/core/transport.go
```diff
@@ -15,6 +21,12 @@ import (
 	"go.uber.org/zap"
 )
 
+const (
+	// dockerInternalHost is the hostnamed used by docker to access the host machine.
+	// We use it for automatic fallbacks when requests to localhost fail
+	dockerInternalHost = "host.docker.internal"
```
StarpTech, line 27: This won't work on linux. I tested it yesterday after I had to connect docker prometheus with my local router. The solution is https://github.com/wundergraph/cosmo/blob/main/docker-compose.yml#L75 Please add this flag to the router run command in studio.

router/core/transport.go
```diff
@@ -84,68 +96,159 @@ func (ct *CustomTransport) RoundTrip(req *http.Request) (*http.Response, error)
 	return resp, err
 }
 
+func isRunningInsideDocker() bool {
+	// Check if we are running inside docker by
+	// testing by checking if /.dockerenv exists
+	st, err := os.Stat("/.dockerenv")
```
StarpTech, line 102: Could you reference and comment the source that says that's the way to identify docker?

router/core/transport.go
```diff
@@ -84,68 +96,159 @@ func (ct *CustomTransport) RoundTrip(req *http.Request) (*http.Response, error)
...
+	return newReq, nil
+}
+
+func (t *localhostFallbackRoundTripper) RoundTrip(r *http.Request) (*http.Response, error) {
+	// If the request has a body, we need to buffer it, otherwise it will
+	// get consumed
+	var bodyData []byte
+	if r.Body != nil {
+		var err error
+		bodyData, err = io.ReadAll(r.Body)
```
StarpTech, line 148: What happens when the body was already read by a module?

router/core/transport.go
```diff
@@ -84,68 +96,159 @@ func (ct *CustomTransport) RoundTrip(req *http.Request) (*http.Response, error)
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to read request body: %w", err)
+		}
+		r.Body.Close()
+		r.Body = io.NopCloser(bytes.NewReader(bodyData))
+	}
+	resp, err := t.transport.RoundTrip(r)
+	if err != nil && t.pointsToLocalhost(r) && errors.Is(err, syscall.ECONNREFUSED) {
```
StarpTech, line 156: Wouldn't it be safe to enable it by default? In docker you can't connect to localhost expect you want to talk to another process in the same container which is out of the scope.

router/core/transport.go
StarpTech: AFAIK `host.docker.internal` only works when the user uses the "bridge" network.

router/core/transport.go
```diff
@@ -84,68 +96,159 @@ func (ct *CustomTransport) RoundTrip(req *http.Request) (*http.Response, error)
...
+	// If the request has a body, we need to buffer it, otherwise it will
+	// get consumed
+	var bodyData []byte
+	if r.Body != nil {
+		var err error
+		bodyData, err = io.ReadAll(r.Body)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read request body: %w", err)
+		}
+		r.Body.Close()
```
StarpTech, line 152: For reference: Very unlikely but there are solutions like docker swarm or similar that use docker for production workloads. We should ensure to not create a solution that might conflict with it.

StarpTech (changes requested):

StarpTech:
Please document this in the router docs as well.

router/core/transport.go
fiam: If the body has already been read by a module without restoring it, the module has botched up the `http.Request`, hasn't it?

router/core/transport.go
fiam: It's not Docker specific, it takes a field to indicate the target host.

StarpTech (approved):
LGTM

router/core/transport.go
```diff
@@ -85,67 +86,80 @@ func (ct *CustomTransport) RoundTrip(req *http.Request) (*http.Response, error)
...
+	PreHandlers                   []TransportPreHandler
+	PostHandlers                  []TransportPostHandler
+	RetryOptions                  retrytransport.RetryOptions
+	RequestTimeout                time.Duration
+	LocalhostFallbackInsideDocker bool
+	Logger                        *zap.Logger
 }
 
 func NewTransport(opts *TransportOptions) *TransportFactory {
+	if opts.LocalhostFallbackInsideDocker && docker.Inside() {
```
StarpTech, line 109: I'd expect this this called on config load or inside the custom transport to keep this clean here.

router/core/transport.go
```diff
@@ -85,67 +86,80 @@ func (ct *CustomTransport) RoundTrip(req *http.Request) (*http.Response, error)
...
-						span.SetAttributes(otel.WgOperationName.String(operation.name))
-					}
-					if operation.opType != "" {
-						span.SetAttributes(otel.WgOperationType.String(operation.opType))
-					}
-					if operation.hash != 0 {
-						span.SetAttributes(otel.WgOperationHash.String(strconv.FormatUint(operation.hash, 10)))
-					}
+	if t.localhostFallbackInsideDocker && docker.Inside() {
+		transport = docker.NewLocalhostFallbackRoundTripper(transport)
```
StarpTech, line 127: Please mention here or in the package description that this module / transport exist only to simplify local development. It should not be used for production. Please also add some docs on gitbook about this smooth integration.

StarpTech (changes requested):

router/core/transport.go
fiam: I've moved it to `newServer()`, PTAL

router/core/transport.go
fiam: Added

StarpTech (approved):
LGTM

PR #230, opened 2023-11-03
feat: provide debug mode in gql collector

StarpTech:
## Motivation and Context

Unfortunately, async inserts are way to slow for big batches. The request timeouts. The new implementation will batch them in less than 10ms locally compared to > 10s with async. We need to keep an eye of the memory consumption. Once we hit a certain load, we need to make the ingestion pipeline smarter.


fiam (approved):
LGTM

PR #233, opened 2023-11-04
fix: ensure proper graphql error response when passing nil error

StarpTech:
## Motivation and Context

Passing nil to `graphql.RequestErrorsFromError(err)` panics. We have to ensure a `graphql.RequestErrors` instance is allocated always. If err is nil `WriteResponseError(ctx, nil)`, we return

```json
{
  "errors": [{ "message": "Internal Error" }]
}
```


router/core/modules.go
```diff
@@ -100,7 +100,16 @@ type ModuleContext struct {
 }
 
 // WriteResponseError writes the given error as a GraphQL error response to the http.ResponseWriter
-// associated with the given RequestContext
+// associated with the given RequestContext. It also takes care of logging and tracking the error in the
+// underlying telemetry system. Please never write errors directly to the http.ResponseWriter.
 func WriteResponseError(ctx RequestContext, err error) {
```
fiam, line 105: Isn't the real problem that somewhere else we're calling WriteResponseError with a nil error? This will silence that without providing any information to the user, the erros array will be empty.

StarpTech: Absolutely, but it's better than panicking and the result is still a valid error response which was intended by calling `WriteResponseError`

fiam: I see your point, but I think that be very difficult to debug if a user sees an error response with an empty array. Can we log an error with the call stack? RequestContext has access to the logger.

StarpTech: I've added a warn log.

fiam (approved):
LGTM

PR #246, opened 2023-11-08
feat: implement @override

Aenimus:
## Motivation and Context


JivusAyrus (approved):

cli/src/commands/router/commands/compose.ts
```diff
@@ -94,14 +94,21 @@ export default (opts: BaseCommandOptions) => {
...
-        name: s.name,
-        url: normalizeURL(s.routing_url),
-        sdl: sdls[index],
-        subscriptionUrl: s.subscription?.url ?? s.routing_url,
-        subscriptionProtocol: s.subscription?.protocol ?? 'ws',
-      })),
+      subgraphs: config.subgraphs.map((s, index) => {
+        const subgraphConfig = result.federationResult!.subgraphConfigBySubgraphName.get(s.name);
+        const schema = subgraphConfig?.schema;
+        const configurationDataMap = subgraphConfig?.configurationDataMap;
```
StarpTech, line 100: Could you describe briefly what this data structure is used for?

StarpTech (changes requested):

cli/src/commands/router/commands/compose.ts
Aenimus: `schema` is the normalized GraphQLSchema (graphql.js) that originally comes from `normalizeSubgraph`.
`ConfigurationDataMap` is the intermediate representation of the engine config that is turned into proto.

Before, both of these structures were returned by calling `normalizeSubgraph` again in `buildRouterConfig`, but now we normalize once and pass the schema and config up to the federator, and finally to `buildRouterConfig`.

Aenimus: `subgraphConfig` is just a container for the schema and the ConfigurationDataMap. It's mapped by subgraph name so we can keep track of which schema and config is for which graph (because this returned from batch normalize).

StarpTech: Could you add these as comments to the types in composition lib?

shared/test/router.config.test.ts
```diff
@@ -18,6 +22,14 @@ describe('Router Config Builder', () => {
       url: 'https://wg-federation-demo-accounts.fly.dev/graphql',
       subscriptionUrl: '',
       subscriptionProtocol: 'ws',
+      schema: buildSchema(
+        fs.readFileSync(path.join(__dirname, 'testdata', 'subgraphConfigs', 'normalized-accounts.graphql')).toString(),
+      ),
+      configurationDataMap: new Map<string, ConfigurationData>(JSON.parse(
+        fs.readFileSync(path.join(__dirname, 'testdata', 'subgraphConfigs', 'accounts-configuration-data-map.json'))
```
StarpTech, line 29: Can we not just generate those artifacts right here with the library? Maintaining them is error prone and manual.

StarpTech (changes requested):

shared/test/testdata/subgraphConfigGenerator.ts
```diff
@@ -0,0 +1,80 @@
+import * as fs from 'node:fs';
+import * as process from 'node:process';
+import { federateSubgraphs, Subgraph } from '@wundergraph/composition';
+import { parse } from 'graphql';
+import { printSchemaWithDirectives } from '@graphql-tools/utils';
+
+function replacer<K, V>(key: K, value: V) {
```
StarpTech, line 7: You could put this in a util file and call it from all tests. I think we can remove the pre build step.

StarpTech (changes requested):

shared/test/testdata/utils.ts
```diff
@@ -0,0 +1,31 @@
+import * as fs from 'node:fs';
+import { federateSubgraphs, FederationResultContainer, Subgraph } from '@wundergraph/composition';
+import { parse } from 'graphql';
+
+export function federateTestSubgraphs(): FederationResultContainer {
```
StarpTech, line 5: So much better!

StarpTech (approved):
LGTM

PR #273, opened 2023-11-18
fix: [connect] Move devDependencies to dependencies

clayne11:
The package is unable to function at runtime without these dependencies, so they're actually runtime dependencies, not devDepedencies.

### DIFF
--- a/connect/package.json
+++ b/connect/package.json
@@ -21,7 +21,9 @@
   },
   "dependencies": {
     "@bufbuild/protobuf": "^1.4.1",
-    "@connectrpc/connect-query": "^0.6.0",
+    "@connectrpc/connect-query": "^0.6.0"
+  },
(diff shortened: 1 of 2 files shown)


connect/package.json
```diff
@@ -19,7 +19,7 @@
       "default": "./dist/*.js"
     }
   },
-  "devDependencies": {
+  "dependencies": {
```
StarpTech, line 22: `typescript` and `de-cli` are dev dependencies for sure.

StarpTech (changes requested):
Good finding. I made a comment because not all of them are prod deps.

connect/package.json
clayne11: Fixed.

StarpTech (approved):
LGTM

PR #286, opened 2023-11-23
fix: variables normalization

devsergiy:
## Motivation and Context


router/core/operation_parser.go
```diff
@@ -323,6 +322,16 @@ func (p *OperationParser) parse(ctx context.Context, clientInfo *ClientInfo, bod
...
 
+	// set variables to empty object if they are null or not present
+	if requestVariableBytes == nil || bytes.Equal(requestVariableBytes, []byte("null")) {
+		requestVariableBytes = []byte("{}")
+	}
+
+	// set variables on doc input before normalization
+	// IMPORTANT: this is required for the normalization to work correctly!
+	// Normalization reads/rewrites/adds variables
+	kit.doc.Input.Variables = requestVariableBytes
```
jensneuse, line 333: In line 341/350 we write the Variables into the keyGen.
These must only be the normalized variables, not the request variables.
Otherwise, we would hash the request variables and would get a cache miss when the user variables are different.
Essentially, we have to omit user supplied variables from the keyGen which, with your change, is not correct anymore.

jensneuse (changes requested):

PR #298, opened 2023-11-24
feat: add support for enabling pprof handlers

fiam:
Handlers are disabled by default and must enabled at compile time.
When compiled in, the handlers are run by default on port 6060 but
can be disabled via command line.

## Motivation and Context


router/cmd/pprof.go
StarpTech, line 1: The design of "cmd" is to host only entry points. Can we move this to internals?

StarpTech (changes requested):

router/cmd/pprof.go
fiam: Sounds good. Should I move the rest of the code related to profiling into an internal package too?

StarpTech: Yes, please.

fiam: I've moved all code for profiling to `internal`, PTAL

StarpTech: Nice, ty.

StarpTech (approved):
LGTM

PR #318, opened 2023-11-29
feat: register router on the controlplane

StarpTech:
## Motivation and Context

This PR adds a new self-registration feature to the router. Self-registration is used to fetch account limits and tokens from the controlplane. Self-registration is not required for the router to function.

1. Register Router and exchange a secret that is used to verify request signatures from the Cosmo Studio. This makes ART safe to use in production. We plan to expand this to make secure introspections in public.
2. Fetch for account limits e.g. trace-sampling-rate to ensure limits are enforced.
3. Introduce a new flag `DEV_MODE` which enables the development mode thus ART without request signing and pretty log output.

Router registration is only required when telemetry data is sent to cloud or Advanced-Request-Tracing (ART) enabled in production. The feature can be disabled by `ROUTER_REGISTRATION=false`.


router/config/config.go
```diff
@@ -267,6 +271,10 @@ func LoadConfig(envOverride string) (*Config, error) {
 		c.ConfigPath = defaultConfigPath
 	}
 
+	if c.DevelopmentMode {
+		c.JSONLog = false
```
fiam, line 275: 👏

fiam (approved):
LGTM

router/config/config.go
StarpTech: :1st_place_medal:

jensneuse (approved):

controlplane/src/core/repositories/FederatedGraphRepository.ts
```diff
@@ -757,4 +767,102 @@ export class FederatedGraphRepository {
...
+    });
+    const publicKey = (keys.publicKey as KeyObject).export({
+      format: 'pem',
+      type: 'spki',
+    });
+
+    const items = await this.db
+      .insert(graphRequestKeys)
+      .values({
+        privateKey: privateKey.toString(),
```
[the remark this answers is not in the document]
StarpTech: That wouldn't be much safer. In cloud we ensure that all data is encrypted at rest. You would need system access to read the data. If the attacker would get access to the process he can also read the environment variable and because it is OSS he also knows how to decrypt it. We can't salt or hash the private key because we need the orginal content. Does it makes sense?

PR #323, opened 2023-11-30
feat: persist ops from playground and view all client ops

thisisnithin:
## Motivation and Context

https://github.com/wundergraph/cosmo/assets/44922285/37efbbed-67bf-49ff-bd4a-10a81ee1d08b


jensneuse:
improvements I'd make:
- move the "Persist Operation" button into a button with a save icon in the middle bar of the playground
- annotate analytics so that we know an operation was persisted
- when an operation document is being persisted, and the document contains exactly one named operation, extract the name and store it. This will make it easier for the user to find an operation because it's more meaningful than a hash
- when clicking on a client in the clients overview, consider turning the UI into a landing page with the following structure: basic client info, like name, created, etc. a weekly overview of the traffic by the client, similar to the federated graph overview, a list of operations where you can click on each to get more details in a slider, I think that's better than the current choice when considering that a developer who created the client would like to get an overview of his "application" and then drill into the details

PR #325, opened 2023-11-30
feat: add compositions page

JivusAyrus:

https://github.com/wundergraph/cosmo/assets/52773149/d94111c0-84e9-4d8c-a586-a2b72e212ed3


controlplane/src/core/bufservices/PlatformService.ts
```diff
@@ -4464,5 +4468,177 @@ export default function (opts: RouterOptions): Partial<ServiceImpl<typeof Platfo
...
+        });
+
+        const changelogs = await fedRepo.fetchChangelogByVersion({
+          schemaVersionId: composition.schemaVersionId,
+        });
+
+        let addCount = 0;
+        let minusCount = 0;
+        for (const log of changelogs) {
+          if (log.changeType.includes('REMOVED')) {
```
StarpTech, line 4562: Can we make it stricter and check for all cases explicitly? E.g similar as [here](https://github.com/wundergraph/cosmo/blob/main/controlplane/src/core/services/SchemaUsageTrafficInspector.ts#L190) but I would put it in a separate function.

controlplane/src/core/repositories/GraphCompositionRepository.ts
```diff
@@ -46,21 +48,125 @@ export class GraphCompositionRepository {
...
+      const isCurrentDeployed = await fedRepo.isLatestValidSchemaVersion(fedGraphTargetId, r.schemaVersionId);
+
+      compositions.push({
+        id: r.id,
+        schemaVersionId: r.schemaVersionId,
+        createdAt: r.createdAt.toISOString(),
+        isComposable: r.isComposable || false,
+        compositionErrors: r.compositionErrors || undefined,
+        createdBy: r.createdBy || undefined,
+        isCurrentDeployed,
```
StarpTech, line 166: We don't know if it was deployed. We only know it is latest.

controlplane/src/core/repositories/analytics/TraceRepository.ts
```diff
@@ -78,4 +78,45 @@ export class TraceRepository {
       },
     }));
   }
+
+  public async getRouterVersion(organizationID: string, startDate: string): Promise<string> {
+    const parsedDateRange = isoDateRangeToTimestamps({ start: startDate, end: startDate });
```
StarpTech, line 83: Why start equals end date?

controlplane/src/core/repositories/analytics/TraceRepository.ts
```diff
@@ -78,4 +78,45 @@ export class TraceRepository {
...
+    const [start] = getDateRange(parsedDateRange);
+
+    const queryWithTimeConstraint = `
+    WITH 
+      toDateTime('${start}') AS startDate
```
StarpTech, line 88: We have no end date?

controlplane/src/core/repositories/analytics/TraceRepository.ts
```diff
@@ -78,4 +78,45 @@ export class TraceRepository {
...
+    WHERE Timestamp >= startDate AND SpanAttributes['wg.router.version'] != '' AND SpanAttributes['wg.organization.id'] = '${organizationID}' 
+    ORDER BY Timestamp ASC
+    LIMIT 1 OFFSET 0
+    `;
+
+    const res = await this.client.queryPromise(queryWithTimeConstraint);
+
+    if (Array.isArray(res)) {
+      if (res.length === 0) {
+        const query = `
```
StarpTech, line 101: What's the reason behind this? If a router version can't be found in the range of the UI we will look for any version?

StarpTech (changes requested):

controlplane/src/core/repositories/analytics/TraceRepository.ts
JivusAyrus: as we are checking the otel traces table, if the users have not run an operation after the start date, I assume that whatever version they were using before the start date as the router version

controlplane/src/core/repositories/analytics/TraceRepository.ts
JivusAyrus: no its not required, as i only care about what the router version was when the composition happened

controlplane/src/core/repositories/analytics/TraceRepository.ts
JivusAyrus: wanted to repurpose the same function, now that I think about it, I will make it an empty string

controlplane/src/core/repositories/GraphCompositionRepository.ts
JivusAyrus: do you mean to rename it to latestValid?

controlplane/src/core/bufservices/PlatformService.ts
StarpTech: ok, my thought here is if the underlying library introduce a new type without that prefix we would loose it.

controlplane/src/core/repositories/analytics/TraceRepository.ts
StarpTech: ok, sounds better

controlplane/src/core/repositories/analytics/TraceRepository.ts
StarpTech: Can we limit it somehow to an end date because it has drastic impact on the performance of the query. A composition usually has an end date as long as it is not the latest.

controlplane/src/core/repositories/analytics/TraceRepository.ts
```diff
@@ -78,4 +78,45 @@ export class TraceRepository {
...
+  public async getRouterVersion(organizationID: string, startDate: string): Promise<string> {
+    const parsedDateRange = isoDateRangeToTimestamps({ start: startDate, end: startDate });
+    const [start] = getDateRange(parsedDateRange);
+
+    const queryWithTimeConstraint = `
+    WITH 
+      toDateTime('${start}') AS startDate
+    SELECT  
+        SpanAttributes['wg.router.version'] as routerVersion
+    FROM ${this.client.database}.otel_traces
```
StarpTech, line 91: Using `otel_traces` is not correct because the retention of this table is much smaller than the materialized view.

StarpTech: I'd also prefer to use the request metrics MV because it is not subject to sampling and has less data to scan.

controlplane/src/core/repositories/analytics/TraceRepository.ts
StarpTech: The problem is that this will result in a fullscan. If no router version can be found after the composition, I'd communicate it "No activity"

StarpTech (changes requested):

controlplane/src/core/repositories/GraphCompositionRepository.ts
JivusAyrus: Done

controlplane/src/core/repositories/analytics/TraceRepository.ts
JivusAyrus: we dont store it in request metrics mv

PR #327, opened 2023-12-01
feat: display available operation names after pushing an operation

fiam:
operation names are available in both the JSON and text output. If a persisted
operation has no named operations, the array is empty in JSON and not shown
in the text output.

## Motivation and Context

### DIFF
--- a/cli/src/commands/operations/commands/push.ts
+++ b/cli/src/commands/operations/commands/push.ts
@@ -2,6 +2,7 @@ import crypto from 'node:crypto';
 import { readFile } from 'node:fs/promises';
 
 import { Command } from 'commander';
+import { parse, visit } from 'graphql';
 import pc from 'picocolors';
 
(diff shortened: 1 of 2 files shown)


jensneuse:
Is it possible to do this in the backend? then it could be used not just from the cli but also from the Playground in the studio. If not feasible, we can copy the logic into the Playground which nithin is currently working on. I suggest you coordinate this with nithin. Otherwise looks good.

fiam:
> Is it possible to do this in the backend? then it could be used not just from the cli but also from the Playground in the studio. If not feasible, we can copy the logic into the Playground which nithin is currently working on. I suggest you coordinate this with nithin. Otherwise looks good.

If we need this in the studio, I'd rather move it to the backend and use the response from the RPC, to avoid duplication. I'll update this PR.

thisisnithin (approved):

PR #333, opened 2023-12-04
feat: add NATS to the router

fiam:
## Motivation and Context


jensneuse (approved):
LGTM but please check with @StarpTech as well

Makefile
```diff
@@ -122,3 +122,6 @@ docker-build-minikube: docker-build-local
 	minikube image load ghcr.io/wundergraph/cosmo/keycloak:latest & \
 	minikube image load ghcr.io/wundergraph/cosmo/cdn:latest
 	minikube cache reload
+
+run-local-demo:
```
StarpTech, line 126: Can we be more specifc? e.g. `run-subgraphs-local` ?

router/config/config.go
```diff
@@ -226,6 +230,7 @@ type Config struct {
 	LocalhostFallbackInsideDocker bool                        `yaml:"localhost_fallback_inside_docker" default:"true" envconfig:"LOCALHOST_FALLBACK_INSIDE_DOCKER"`
 	CDN                           CDNConfiguration            `yaml:"cdn"`
 	DevelopmentMode               bool                        `yaml:"dev_mode" default:"false" envconfig:"DEV_MODE"`
+	NATS                          NATSConfiguration           `yaml:"nats"`
```
StarpTech, line 233: I thought we want to keep it generic e.g. adding kafka support later. I would also describe it from a feature perspective e.g.

```yaml
eventing:
	provider: "NATS"
	url: "..."
```

router/internal/pubsub/nats.go
```diff
@@ -0,0 +1,89 @@
...
+		return err
+	}
+	return p.conn.Publish(topic, data)
+}
+
+func (p *natsPubSub) Request(ctx context.Context, topic string, data []byte, w io.Writer) error {
+	if err := p.ensureConn(); err != nil {
+		return err
+	}
+	msg, err := p.conn.RequestWithContext(ctx, topic, []byte(`{"id": 3}`))
```
StarpTech, line 82: Why is the id harcoded?

router/internal/pubsub/nats.go
```diff
@@ -0,0 +1,89 @@
...
+	if err != nil {
+		return fmt.Errorf("error requesting NATS topic %s: %w", topic, err)
+	}
+	_, err = w.Write(msg.Data)
```
StarpTech, line 86: Don't swallow the errors here.

StarpTech (changes requested):

router/internal/pubsub/nats.go
fiam: Do you mean `msg.Ack()`? Needs to be swallowed AFAIK because if there's no jetstream configured Ack returns an error.

router/internal/pubsub/nats.go
fiam: Leftover test data, removed

Makefile
fiam: Renamed

router/config/config.go
fiam: I've refactored it to use a more generic config format, which should allow us to support multiple data source types moving forward. PTAL

router/internal/pubsub/nats.go
StarpTech: If we don't use jetstream API why do we ACK at all?

fiam: I was under the impression that the user can enable JetStream in the NATS server and then we would have to deal with it. I've removed the Ack, PTAL.

StarpTech (approved):
LGTM

PR #339, opened 2023-12-04
feat: implement wgc router download-binary

Aenimus:
## Motivation and Context


cli/package.json
```diff
@@ -33,24 +33,29 @@
     "@bufbuild/protobuf": "^1.4.1",
     "@connectrpc/connect": "^1.1.3",
     "@connectrpc/connect-node": "^1.1.3",
+    "@types/decompress": "^4.2.7",
```
StarpTech, line 36: dev dependency

cli/src/commands/router/commands/download.ts
```diff
@@ -0,0 +1,157 @@
...
+    let routerRelease;
+    for (const release of releases.data) {
+      if (!release.tag_name || !release.tag_name.startsWith('router@')) {
+        continue;
+      }
+      routerRelease = release;
+      break;
+    }
+    let routerTarget = '';
+    switch (process.platform) {
```
StarpTech, line 49: Please put this in a `getBinaryName` function

cli/src/commands/router/commands/download.ts
```diff
@@ -0,0 +1,157 @@
...
+    if (!response.body) {
+      console.log(pc.red(`Response had no body`));
+      process.exit(1);
+    }
+    console.log(`Beginning download for ${routerTarget} from ${url}`);
+    let loaded = 0;
+    const total = Number.parseInt(contentLength, 10);
+    const bar = new cliProgress.SingleBar({});
+    bar.start(total, 0);
+    const res = new Response(new ReadableStream({
```
StarpTech, line 137: You don't have to wrap it into a Response. I found this great example

```ts
const response = await fetch('http://localhost:3000/measurements.json')
// response.body is a ReadableStream
const reader = response.body.getReader();
for await (const chunk of readChunks(reader)) {
    console.log(`received chunk of size ${chunk.length}`);
}

// readChunks() reads from the provided reader and yields the results into an async iterable
function readChunks(reader) {
    return {
        async* [Symbol.asyncIterator]() {
            let readResult = await reader.read();
            while (!readResult.done) {
                yield readResult.value;
                readResult = await reader.read();
            }
        },
    };
}
```

fiam (approved):
LGTM 👏

cli/src/commands/router/commands/download.ts
```diff
@@ -0,0 +1,132 @@
...
+import { Octokit } from 'octokit';
+import { Command, program } from 'commander';
+import pc from 'picocolors';
+import cliProgress from 'cli-progress';
+import decompress from 'decompress';
+import { join } from 'pathe';
+import { BaseCommandOptions } from '../../../core/types/types.js';
+
+export default function (_: BaseCommandOptions) {
+  const command = new Command('download');
```
StarpTech, line 11: Let's rename the command. It might create confusion between `fetch` and `download`. Both are very similar but do completely different things. Let's call it `wgc router binary download` or `wgc router-download`

cli/src/commands/router/commands/download.ts
```diff
@@ -0,0 +1,132 @@
...
+    }
+    const response = await fetch(url);
+    const contentLength = response.headers.get('content-length');
+    if (!contentLength) {
+      program.error(pc.red(`Could not get content-length of file`));
+    }
+    if (!response.body) {
+      program.error(pc.red(`Response had no body`));
+    }
+    console.log(`Beginning download for ${routerTarget} from ${url}`);
```
StarpTech, line 49: Can we inform the user to what path the router files have been extracted?

cli/src/commands/router/commands/download.ts
```diff
@@ -0,0 +1,132 @@
...
+
+export default function (_: BaseCommandOptions) {
+  const command = new Command('download');
+  command.description('Downloads the latest router binary for the detected machine platform and architecture.');
+  command.option('-o, --out [string]', 'Destination directory for the downloaded router binary.');
+  command.action(async (options) => {
+    const path = join(process.cwd(), options.out ?? './router');
+    const fullPath = path + '/router';
+    if (fs.existsSync(fullPath)) {
+      console.log(pc.red(`${path}/router already exists`));
```
StarpTech, line 18: Path normalization is wrong (see double slash)

```
> tsx src/index.ts "router" "download" "-o" "../"

/home/starptech/p/wundergraph/cosmo//router already exists

```

StarpTech (changes requested):

cli/src/commands/router/commands/download.ts
StarpTech, line 18: I'd assume it should exit 1 when the download operation could not be run successfully

StarpTech (changes requested):

cli/src/commands/router/commands/download.ts
Aenimus: Would you like me to change this to an error?

PR #345, opened 2023-12-05
chore: add go.work file

clayne11:
This is needed for go mod to work properly.

- go work init
- go work use router


go.work
```diff
@@ -0,0 +1,3 @@
+go 1.21.4
```
StarpTech, line 1: I removed the patch so not everyone has to use the exact same version. I also added the other go modules to the workspace so we can remove require statements.

StarpTech:
Continued in https://github.com/wundergraph/cosmo/pull/353

PR #347, opened 2023-12-07
refactor: make the main HTTP router not dependent on the active GraphQL server

clayne11:
## Motivation and Context

By decoupling the top level router's HTTP router and the currently active GraphQL server, we enable the router to be decoupled from running the HTTP server.

This separates concerns between configuring and executing the HTTP handler from running the server, and enables consumers to run their own HTTP handler while the internal logic can hot-swap when a new graph configuration is available.

Some consumers may already have an HTTP server running and simply want the router to be attached on a given path. This is the first PR along the path of enabling this capability.

This PR is purely an internal refactor. There are zero logical or API changes.


router/core/router.go
```diff
@@ -119,7 +119,7 @@ type (
 		configPoller configpoller.ConfigPoller
 		selfRegister selfregister.SelfRegister
 
-		registrationInfo *nodev1.RegistrationInfo
+		publicKey *ecdsa.PublicKey
```
clayne11, line 122: `registrationInfo` was only ever used to create the `publicKey`. We eagerly create the key instead on the first `router.Start`.

router/core/router.go
```diff
@@ -135,7 +135,11 @@ type (
 		rootContext       context.Context
 		rootContextCancel func()
 		routerConfig      *nodev1.RouterConfig
-		healthChecks      health.Checker
```
clayne11, line 138: There's no reason for the healthChecks to be on the currently active server vs the main router. I pulled it up to the main Router instance.

router/core/router.go
```diff
@@ -871,6 +856,42 @@ func (r *Router) newServer(ctx context.Context, routerConfig *nodev1.RouterConfi
 	return ro, nil
 }
 
+func (r *Router) buildHTTPRouter() http.Handler {
```
clayne11, line 859: This HTTP router now has 0 dependencies on the active server or its config. In the next PR I will modify the API to only call this once and move the HTTP server into `cmd.Main`.

router/core/router.go
```diff
@@ -326,9 +336,67 @@ func NewRouter(opts ...Option) (*Router, error) {
 		r.logger.Warn("Development mode enabled. This should only be used for testing purposes")
 	}
 
+	r.httpRouter = r.buildHTTPRouter()
+
 	return r, nil
 }
 
+func (r *Router) buildHTTPRouter() http.Handler {
```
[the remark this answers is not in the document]
StarpTech: What is your motivation? Do you want to run the router along another app or just have flexibility? Are you aware that the router shutdown itself on schema updates?

StarpTech (approved):
I like the changes. I re-tested the relevant places. LGTM

StarpTech:
Unfortunately, we found a [breaking change](https://github.com/wundergraph/cosmo/pull/347/files#diff-aa58824b56f499a6a7d39dd7d905130a1aed0683be3aa6826a63dfd5d5047edcR362) in this PR. The playground was exposed on all methods which caused conflicts when using `/` as graphql path. During working on the fix, I recognized that the code is harder to reason about compared to the previous code. We need to find another way to decouple it. We have plans to make the router working with AWS lambda. I think this is the right moment to overhaul it.

PR #350, opened 2023-12-08
fix: race condition due to premature buffer reuse

fiam:
When parsing the operation we were sending the buffer back to the reuse
pool while the operation had not yet been parsed, corrupting some of
the operations. Add also an integration test to prevent regressions.

## Motivation and Context


jensneuse:
Good work!

PR #382, opened 2023-12-18
feat: improve subgraph error handling

jensneuse:
Doesn't swallow subgraph errors anymore. Collects them and wraps them as a multi-error which is then appended to the trace span to surface subgraph issues, e.g. in Prometheus.


router/core/graphql_handler.go
```diff
@@ -186,6 +187,22 @@ func logInternalErrorsFromReport(report *operationreport.Report, requestLogger *
 	}
 }
 
+func propagateSubgraphErrors(ctx *resolve.Context) {
```
StarpTech, line 190: Can we call this method inside `writeRequestErrors` to not duplicate the context and span logic?

StarpTech (changes requested):

StarpTech (approved):
LGTM

PR #387, opened 2023-12-20
fix: handle shutdown more gracefully

SvenW:
## Motivation and Context
Graceful shutdown was not graceful. As soon as the signal was received the inner http server was terminated. Requests still in flight would then be terminated straight away. This makes sure we stop responding to the readiness check but still respond to the liveness check which means we are able to finish all requests coming in


router/cmd/main.go
```diff
@@ -82,5 +82,5 @@ func Main() {
 
 	profile.Finish()
 	logger.Debug("Server exiting")
-	os.Exit(0)
+	stop()
```
brodin, line 85: 💅 Maybe stop should be kept as a defer? Or is there some reason it can't? Avoid missing to call it for different paths where we return early?

StarpTech:
Great finding. I'm closing this in favor of https://github.com/wundergraph/cosmo/pull/393

PR #388, opened 2023-12-20
refactor(prometheus): prometheus integration

StarpTech:
## Motivation and Context

This PR creates a different OTLP MeterProvider and Meter to track metrics that should be exported through Prometheus. This allow us to apply filter rules correctly and only for Prometheus.


JivusAyrus (approved):

jensneuse (approved):

router/internal/metric/prometheus.go
```diff
@@ -0,0 +1,57 @@
...
+		Registry:          registry,
+		Timeout:           0,
+	}))
+
+	svr := &http.Server{
+		Addr:              listenAddr,
+		ReadTimeout:       1 * time.Minute,
+		WriteTimeout:      2 * time.Minute,
+		ReadHeaderTimeout: 20 * time.Second,
+		ErrorLog:          zap.NewStdLog(logger),
```
jensneuse, line 39: Are you sure about these timeouts?
I have no experience with Prom but these seem to be quite high.

StarpTech: I just set some because by default there are none. I made them smaller

PR #413, opened 2024-01-10
fix: update organization details

JivusAyrus:


controlplane/src/core/util.ts
```diff
@@ -187,6 +187,8 @@ export const getHighestPriorityRole = ({ userRoles }: { userRoles: string[] }) =
 };
 
 export const isValidOrganizationSlug = (slug: string): boolean => {
+  const reservedSlugs = ['login', 'signup', 'create', 'account'];
```
StarpTech, line 190: Good catch. How can we prevent more in the future? I think we should introduce a different path prefix.

StarpTech, line 190: Good catch.

StarpTech (approved):
LGTM

PR #424, opened 2024-01-15
feat: audit logs

StarpTech:
## Motivation and Context

image


controlplane/src/db/models.ts
```diff
@@ -19,3 +19,60 @@ export type NewOrganizationFeature = typeof organizationFeatures.$inferInsert;
 export type NewBillingSubscription = typeof billingSubscriptions.$inferInsert;
 export type NewBillingPlan = typeof billingPlans.$inferInsert;
 export type MemberRole = (typeof memberRoleEnum.enumValues)[number];
+
+export type AuditableType =
+  | 'organization'
```
Aenimus, line 24: Should the string we reuse be consts? Not a big deal; just a suggestion.

For instance, 'api_key' is used in a few places.

StarpTech: As long as we use the type it is ensured. If you mean something different, please make an example.

Aenimus: I just mean reusing a string literal as a constant

```
// constants.ts
const example = 'example';

// file.ts
type X = {
 aString: example
 };
 ```
 
 It's not important. I just see some repeated literals.

Comment (StarpTech) [reply controlplane/src/db/models.ts:24]: For exact that reason the type exists. There are two ways to in TS to work with "enums". The traditional way to use  enums or string constants. If you go with string constants you need to write in anytime.

**Review by JivusAyrus — COMMENTED** [review]
We should have an audit log when the subscription is upgraded right?

One more doubt, what is the diff between subscription.created and subscription.activated

**Comment by StarpTech** [comment]
@JivusAyrus good points. I missed the event.

> one more doubt, what is the diff between subscription.created and subscription.activated

`created` when the user creates a subscription for the first time. `Activated` once stripe reports that the subscription is active.

File: studio/src/components/audit-log-table.tsx [hunk 2]
```diff
@@ -0,0 +1,171 @@
...
+            actorDisplayName,
+            actorType,
+            auditAction,
+            createdAt,
+            action,
+            auditableDisplayName,
+            targetDisplayName,
+            targetType,
+          }) => {
+            let a = null;
```
thisisnithin, line 72: Tiny nitpick, better variable names here?

thisisnithin (approved):
LGTM

studio/src/components/audit-log-table.tsx
StarpTech: Done

PR #425, opened 2024-01-15
feat: improve subscriptions

jensneuse:


router/core/router.go
```diff
@@ -799,6 +799,12 @@ func (r *Router) newServer(ctx context.Context, routerConfig *nodev1.RouterConfi
 		MaxOperationSizeInBytes: int64(r.routerTrafficConfig.MaxRequestBodyBytes),
 		PersistentOpClient:      r.cdnPersistentOpClient,
 	})
+	// Pre-hash all data source IDs to avoid races
```
StarpTech, line 802: Could you add a few more comments why we need to add this here? For me it's also a TODO to do it in a different place at the engine side.

router/core/websocket_stats.go
```diff
@@ -75,12 +79,14 @@ func (s *WebSocketStats) run(ctx context.Context) {
 			s.reportConnections()
 		case <-s.update:
 			s.mu.Lock()
-			for _, subscriber := range s.subscribers {
+			for ctx, subscriber := range s.subscribers {
 				// non-blocking send
```
StarpTech, line 83: This comment is no longer valid.

router/internal/epoller/epoll_linux.go
```diff
@@ -99,7 +103,7 @@ func (e *Epoll) Wait(count int) ([]net.Conn, error) {
 	events := make([]unix.EpollEvent, count)
 
 retry:
-	n, err := unix.EpollWait(e.fd, events, -1)
+	n, err := unix.EpollWait(e.fd, events, e.timeoutMsec)
```
StarpTech, line 106: Did you find out why -1 is used on linux? Without any good argument I would not change the default.

StarpTech (changes requested):

router/internal/epoller/epoll_linux.go
jensneuse: -1 = no timeout

router/core/router.go
jensneuse: done

StarpTech (approved):
LGTM

PR #431, opened 2024-01-16
fix: skip healthchecks requests for tracing

StarpTech:
## Motivation and Context

### DIFF
--- a/router/internal/trace/utils.go
+++ b/router/internal/trace/utils.go
@@ -3,6 +3,7 @@ package trace
 import (
 	"context"
 	"net/http"
+	"strings"
 
 	"go.opentelemetry.io/otel"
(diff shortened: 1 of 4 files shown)


router/internal/trace/utils.go
```diff
@@ -20,7 +21,12 @@ func TracerFromContext(ctx context.Context) (tracer trace.Tracer) {
 }
 
 func RequestFilter(r *http.Request) bool {
-	if r.URL.Path == "/health" || r.URL.Path == "/favicon.ico" || r.Method == "OPTIONS" {
+
+	// /health and /health/live, /health/ready are ignored
+	if strings.HasPrefix(r.URL.Path, "/health") {
```
jensneuse: can the user configure the health endpoint via config? if so, should we pick this up?

PR #434, opened 2024-01-16
fix: navigation from invitations page

JivusAyrus:

### DIFF
--- a/studio/src/components/layout/sidenav.tsx
+++ b/studio/src/components/layout/sidenav.tsx
@@ -100,7 +100,7 @@ const Organizations = () => {
             );
             if (currentOrg) {
               router.replace(
-                currentPage === "graph"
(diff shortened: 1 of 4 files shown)


studio/src/components/layout/sidenav.tsx
```diff
@@ -100,7 +100,7 @@ const Organizations = () => {
             );
             if (currentOrg) {
               router.replace(
-                currentPage === "graph"
+                currentPage === "graph" || currentPage === "invitations"
```
StarpTech, line 103: Can we generalize more to avoid specifying pages manually?

StarpTech (changes requested):

studio/src/components/layout/sidenav.tsx
JivusAyrus: Done

PR #444, opened 2024-01-18
feat: implement pagination and date filter for audit logs

JivusAyrus:

### DIFF
--- a/controlplane/src/core/bufservices/PlatformService.ts
+++ b/controlplane/src/core/bufservices/PlatformService.ts
@@ -5790,6 +5790,26 @@ export default function (opts: RouterOptions): Partial<ServiceImpl<typeof Platfo
               details: `The user doesnt have the permissions to perform this operation`,
             },
             logs: [],
+            count: '',
+          };
+        }
+
+        const { dateRange } = validateDateRanges({
+          limit: 7,
+          dateRange: {
+            start: req.startDate,
+            end: req.endDate,
+          },
+        });
+
+        if (!dateRange) {
+          return {
+            response: {
+              code: EnumStatusCode.ERR,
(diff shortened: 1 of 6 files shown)


controlplane/src/core/bufservices/PlatformService.ts
```diff
@@ -5775,13 +5776,51 @@ export default function (opts: RouterOptions): Partial<ServiceImpl<typeof Platfo
...
+              details: 'Invalid date range',
+            },
+            logs: [],
+            count: '',
+          };
+        }
+
+        let limit = req.limit;
+        // check that the limit is less than the max option provided in the ui
+        if (limit > 50) {
```
StarpTech, line 5809: I'd check for 100 and throw an error if someone goes beyond otherwise it will hard to understand.

controlplane/src/core/bufservices/PlatformService.ts
```diff
@@ -5775,13 +5776,51 @@ export default function (opts: RouterOptions): Partial<ServiceImpl<typeof Platfo
               details: `The user doesnt have the permissions to perform this operation`,
             },
             logs: [],
+            count: '',
```
StarpTech, line 5779: What's the reason that count is a string?

StarpTech (changes requested):

controlplane/src/core/bufservices/PlatformService.ts
JivusAyrus: Done, but kept the limit to 50 as that's the highest on the frontend

StarpTech: ok

StarpTech (approved):
LGTM

PR #452, opened 2024-01-24
feat: allow custom playground path

StarpTech:
## Motivation and Context

### DIFF
--- a/router/core/router.go
+++ b/router/core/router.go
@@ -89,6 +89,7 @@ type (
 		listenAddr               string
 		baseURL                  string
 		graphqlWebURL            string
+		playgroundPath           string
 		graphqlPath              string
 		playground               bool
 		introspection            bool
@@ -171,6 +172,10 @@ func NewRouter(opts ...Option) (*Router, error) {
 
 	if r.graphqlWebURL == "" {
 		r.graphqlWebURL = r.graphqlPath
+	}
+
+	if r.playgroundPath == "" {
+		r.playgroundPath = "/"
 	}
 
 	// Default values for trace and metric config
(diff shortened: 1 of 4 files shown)


router/core/router.go
```diff
@@ -870,7 +875,7 @@ func (r *Router) newServer(ctx context.Context, routerConfig *nodev1.RouterConfi
 	var graphqlPlaygroundHandler func(http.Handler) http.Handler
 
 	if r.playground {
-		r.logger.Info("Serving GraphQL playground", zap.String("url", r.baseURL))
+		r.logger.Info("Serving GraphQL playground", zap.String("url", r.baseURL+r.playgroundPath))
```
jensneuse, line 878: consider using url join instead of concatenation
what happens if the baseURL doesn't have a trailing slash and the playgroundpath starts without a slash?

router/pkg/config/config.go
```diff
@@ -246,6 +246,7 @@ type Config struct {
 	ReadinessCheckPath            string                      `yaml:"readiness_check_path" default:"/health/ready" envconfig:"READINESS_CHECK_PATH" validate:"uri"`
 	LivenessCheckPath             string                      `yaml:"liveness_check_path" default:"/health/live" envconfig:"LIVENESS_CHECK_PATH" validate:"uri"`
 	GraphQLPath                   string                      `yaml:"graphql_path" default:"/graphql" envconfig:"GRAPHQL_PATH"`
+	PlaygroundPath                string                      `yaml:"playground_path" default:"/" envconfig:"PLAYGROUND_PATH"`
```
jensneuse, line 249: Can we add validation that paths are valid paths?

jensneuse (changes requested):

router/core/router.go
Aenimus: playgroundPath defaults to "/", but isn't this different to the current behaviour? Before, it's just the baseURL. Now it's the baseURL + "/"

Perhaps I am misunderstanding something.

StarpTech: Done

router/pkg/config/config.go
StarpTech: Added URI validator

router/core/router.go
Aenimus: Reading further ahead makes it clear :D

StarpTech: Default was not changed.

jensneuse (approved):

jensneuse (approved):

PR #456, opened 2024-01-25
feat: produce spans for handler and engine work

StarpTech:
## Motivation and Context

image


router/core/request_tracing.go
```diff
@@ -23,11 +23,11 @@ func ParseRequestTraceOptions(r *http.Request) (options resolve.RequestTraceOpti
 	var (
 		values []string
 	)
-	if r.Header.Get(RequestTraceHeader) != "" {
+	if r.Header.Get(RequestTraceHeader) == "true" {
 		options.Enable = true
 		values = r.Header.Values(RequestTraceHeader)
 	}
-	if r.URL.Query().Get(RequestTraceQueryParameter) != "" {
+	if r.URL.Query().Get(RequestTraceQueryParameter) == "true" {
```
jensneuse, line 30: I think this is not correct, any non-empty string is ok, except "false".

jensneuse (changes requested):

router/core/request_tracing.go
StarpTech: I reverted it. I thought supporting false and true as values would be more reasonable because we never communicated that `wg_trace` is the same as `wg_trace=true`

jensneuse: Ok

jensneuse (approved):

router/core/graphql_prehandler.go
```diff
@@ -142,42 +150,81 @@ func (h *PreHandler) Handler(next http.Handler) http.Handler {
...
-		validatedReq, err := h.accessController.Access(w, r)
-		if err != nil {
-			hasRequestError = true
-			requestLogger.Error(err.Error())
-			writeRequestErrors(r, http.StatusUnauthorized, graphql.RequestErrorsFromError(err), w, requestLogger)
-			return
+		// If we have authenticators, we try to authenticate the request
+		if len(h.accessController.authenticators) > 0 {
+			authenticateSpanCtx, authenticateSpan := h.tracer.Start(r.Context(), "Authenticate",
+				trace.WithSpanKind(trace.SpanKindServer),
```
Blacksmoke16, line 156: Does it make sense for this to be a `server` span? Based on their [docs](https://opentelemetry.io/docs/concepts/signals/traces/#server):

> A server span represents a synchronous incoming remote call such as an incoming HTTP request or remote procedure call.

It doesn't seem appropriate given, if anything, it would be making HTTP requests so `client` might make sense since it can't be `internal` due to crossing process boundaries?

PR #466, opened 2024-01-29
feat: cosmo ai, generate docs on publish

StarpTech:
## Motivation and Context

image

This PR introduces "Cosmo AI" the first step to bring AI capabilities into Cosmo. The user has to opt-in to use it. The first feature is the ability to auto generate subgraph documentation based on the provided schema on `wgc subgraph publish`. 

In order to not block the publish operation for several seconds, we integrated [BullMQ](https://docs.bullmq.io/) a message queue for Node.js and offload the task on a worker. The readme is updated asynchronously after publish.
This also sets the foundation for alerting and recurrent jobs.


controlplane/src/core/bufservices/PlatformService.ts
```diff
@@ -1519,6 +1532,27 @@ export default function (opts: RouterOptions): Partial<ServiceImpl<typeof Platfo
           targetNamespaceDisplayName: subgraph.namespace,
         });
 
+        if (
+          opts.openaiApiKey &&
+          // Avoid calling OpenAI API if the schema is too big.
+          // Best effort approach. This way of counting tokens is not accurate.
+          subgraph.schemaSDL.length <= 10_000
```
jensneuse, line 1539: You can count tokens e.g. using a library like this:
https://github.com/dqbd/tiktoken

controlplane/src/core/bufservices/PlatformService.ts
```diff
@@ -1519,6 +1532,27 @@ export default function (opts: RouterOptions): Partial<ServiceImpl<typeof Platfo
...
+        ) {
+          const orgRepo = new OrganizationRepository(opts.db, opts.billingDefaultPlanId);
+          const feature = await orgRepo.getFeature({
```
jensneuse, line 1542: what happens if one of these awaits fails?

controlplane/src/core/openai-graphql/index.ts
```diff
@@ -0,0 +1,107 @@
...
+        frequency_penalty: 0,
+        presence_penalty: 0,
+        stream: false,
+      },
+      {
+        timeout: 60_000,
+      },
+    );
+
+    return { readme: response.choices[0].message.content ?? '' };
```
jensneuse, line 47: What if choices is not an array?

controlplane/src/core/workers/AIGraphReadmeWorker.ts
```diff
@@ -0,0 +1,112 @@
...
+    },
+  ) {
+    this.oaig = new OpenAIGraphql({
+      openAiApiKey: input.openAiApiKey,
+    });
+    this.input.log = input.log.child({ worker: this.constructor.name });
+  }
+
+  public async handler(job: Job<CreateReadmeInputEvent>) {
+    if (job.data.type === 'subgraph') {
```
jensneuse, line 58: use switch + move each job to another func?

jensneuse (changes requested):

controlplane/src/core/bufservices/PlatformService.ts
StarpTech: Thanks, I saw it but I'm a bit reluctant to run WASM in our hot paths. It requires more testing.

controlplane/src/core/bufservices/PlatformService.ts
StarpTech: What do you expect? An error is returned to the client.

controlplane/src/core/workers/AIGraphReadmeWorker.ts
StarpTech: I've moved both into separate functions. I don't see a value here to use switch though.

jensneuse (approved):

PR #484, opened 2024-02-02
feat: show span error and allow filter by span status code and trace id

StarpTech:
## Motivation and Context

Display Span Status and allow to filter for Span Status and Trace Id.

image

## TODO


controlplane/src/core/repositories/analytics/util.ts
```diff
@@ -189,10 +191,22 @@ export function buildCoercedFilterSqlStatement(
...
-      } else if (column.type === 'string') {
-        sql = `${dbField} ${operatorSql} {${id}:String}`;
+      switch (column.type) {
+        case 'number': {
+          sql = `${dbField} ${operatorSql} {${id}:Float64}`;
+          break;
+        }
+        case 'string': {
+          sql = `${dbField} ${operatorSql} {${id}:String}`;
+          break;
```
StarpTech, line 201: This is not necessary but I also don't want to remove it because it provides support for boolean values.

thisisnithin (approved):

controlplane/src/core/repositories/analytics/util.ts
```diff
@@ -276,10 +292,22 @@ export function coerceFilterValues(
...
+        case 'number': {
+          result[id] = Number.parseFloat(filter.value);
+          break;
+        }
+        case 'string': {
+          result[id] = filter.value;
+          break;
+        }
+        case 'boolean': {
+          result[id] = filter.value === 'true';
```
JivusAyrus, line 305: Will the value be a string?

StarpTech: Yes

PR #485, opened 2024-02-03
fix: refactor router trace instrumentation

StarpTech:
## Motivation and Context

This PR improves the tracing instrumentation of the router to ensure that we always attach as much information as possible in error cases. In order to achieve this, I had to split the `OperationParser` in multiple steps:

- Parse: Parse the request body to get information like operation name, variables.
- Normalize: Normalize operation content and variables. 
- Validate: Validates if the operation is valid according to the schema.
- Plan: Creates the engine plan.

For each step, we create new spans.

Additionally, I added support for graphql variables in span attributes. This features can be disabled via config or env flag.

```yaml
telemetry:
  tracing:
      export_graphql_variables: true # TRACING_EXPORT_GRAPHQL_VARIABLES
```

image

## TODO


router/core/graphql_prehandler.go
```diff
@@ -114,14 +131,115 @@ func (h *PreHandler) Handler(next http.Handler) http.Handler {
...
 
+		/**
+		* Parse the operation
+		 */
+
+		engineParseCtx, engineParseSpan := h.tracer.Start(r.Context(), "Operation - Parse",
+			trace.WithSpanKind(trace.SpanKindServer),
+		)
+
+		parser, err := h.parser.NewParser(body)
```
jensneuse, line 150: Does this create one parser per request without being able to re-use it?
If yes, is there a way to re-use parsers more?
If parsers are already being re-used, close this comment.

jensneuse (approved):

router/core/graphql_prehandler.go
StarpTech: I'll try to improve the naming. Only one parser is created but we create an OperationKit per request that yields the result of the parsed, normalized and validated result.

PR #489, opened 2024-02-05
fix: load schema from CP and sync router playground

### LINKED ISSUE #473: Playground could take schema from registry not from introspection
### Component(s)

studio

Due to security reasons we don't want to expose introspection in our production clusters, but that is a problem with Cosmo's playground as it relies on introspection, without that we are not able to use it.

It could automatically take latests compose schema from registry instead.

_No response_

### Additional context

_No response_

thisisnithin:
## Motivation and Context

Completes ENG-4833
Fixes #473


studio/src/pages/[organizationSlug]/[namespace]/graph/[slug]/playground.tsx
```diff
@@ -556,6 +565,7 @@ const PlaygroundPage: NextPageWithLayout = () => {
                 showAttribution: false,
               }),
             ]}
+            schema={isLoading ? null : parseSchema(data?.sdl) ?? undefined}
```
Comment (jensneuse) [inline studio/src/pages/[organizationSlug]/[namespace]/graph/[slug]/playground.tsx:568]: Will "parseSchema" need to run on every render?
Would it be better to use something like useMemo to parse it only once?

jensneuse (changes requested):

studio/src/pages/[organizationSlug]/[namespace]/graph/[slug]/playground.tsx
Comment (thisisnithin) [reply studio/src/pages/[organizationSlug]/[namespace]/graph/[slug]/playground.tsx]: Yep, moved it inside a useMemo

StarpTech (approved):
LGTM

PR #512, opened 2024-02-09
feat(helm): support existing secrets

### LINKED ISSUE #453: Enable helm chart to use existing secrets
### Component(s)

helm

The problem with the current helm chart is that you need to provide secret values in plain text in the `values.yaml`. For users who follow a gitops workflow, this is not the ideal solution because the secrets need to be pushed into a repo.

Normally helm chart provide a possibility to define an existing secret name for the chart. In case this value is provided the helm chart does not create its secret resource. 

_No response_

### Additional context

_No response_

alejandrobelleza:
## Motivation and Context
Support existing secrets to:
*  avoid having sensitive credentials in Cosmo values.yaml.
* Support gitops workflow. ( external-secrets )

This PR closes #453

## TODO

### DIFF
--- a/helm/cosmo/charts/router/CHART.md
+++ b/helm/cosmo/charts/router/CHART.md
@@ -25,6 +25,7 @@ WunderGraph Cosmo router.
 | configuration.routerConfigPath | string | `""` | A possible to solution could be to use an init container to download the file from a CDN. If set, polling of the config is disabled. |
 | deploymentStrategy | object | `{}` |  |
 | existingConfigmap | string | `""` | If this is set, the commonConfiguration section is ignored. |
+| existingSecret | string | `""` | Existing secret in the same namespace containing the graphApiToken. The secret key has to match with current secret. |
 | extraEnvVars | list | `[]` | Allows to set additional environment variables on the container |
 | extraEnvVarsCM | string | `""` | Name of existing ConfigMap containing extra env vars |
 | extraEnvVarsSecret | string | `""` | Name of existing Secret containing extra env vars |
--- a/helm/cosmo/charts/cdn/CHART.md
+++ b/helm/cosmo/charts/cdn/CHART.md
@@ -17,6 +17,7 @@ WunderGraph Cosmo CDN
 | autoscaling.targetCPUUtilizationPercentage | int | `80` |  |
 | configuration | string | `nil` |  |
 | deploymentStrategy | object | `{}` |  |
+| existingSecret | string | `""` | Existing secret in the same namespace containing the authJwtSecret and s3StorageUrl. The secret keys have to match with current secret. |
 | extraEnvVars | list | `[]` | Allows to set additional environment variables on the container. Useful for global application non-specific settings. |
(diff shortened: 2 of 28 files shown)


helm/cosmo/charts/router/CHART.md
```diff
@@ -22,6 +22,7 @@ WunderGraph Cosmo router.
 | configuration.otelCollectorUrl | string | `""` | The URL of the Cosmo GraphQL OTEL Collector. Should be internal to the cluster. Default to cloud if not set. |
 | deploymentStrategy | object | `{}` |  |
 | existingConfigmap | string | `""` | The name of the configmap to use for the router configuration. The key "config.yaml" is required. |
+| existingSecret | string | `""` | Existing secret containing the graphApiToken |
```
StarpTech, line 25: Please document that the secret has to be in the same namespace, named like the chart fullname and the keys has to match with current secret.

StarpTech (changes requested):
Thanks for the PR. We should only improve docs around the usage.

StarpTech (approved):
LGTM

helm/cosmo/charts/router/CHART.md
alejandrobelleza: Sure, I've added a _helper function to set the name. And improved the doc.

StarpTech:
@alejandrobelleza you need to run `make docs` to update docs.

PeteMac88:
@alejandrobelleza Thanks for the PR. I was also working on one :D

alejandrobelleza:
> @alejandrobelleza you need to run `make docs` to update docs.

Sure, done

PeteMac88:
Can we merge this PR?

PeteMac88:
@alejandrobelleza  @StarpTech Wouldn't it be better not to create the secrets in the helm chart in case you pass the existingSecret value? I think this will be problematic because when you create the secret through an external secret provider you will get conflicts since the secret in the chart has the same name.

StarpTech:
@PeteMac88 I agree, is that your use case?

@alejandrobelleza Do I read it correctly, that after setting `existingSecret` the old secret is still created? I'd expect that it is not. Similar to how it works in the majority of bnitnami charts https://github.com/bitnami/charts/blob/3a0dfccb3e0eb0ecd344eaae5f53b547135da938/bitnami/nats/templates/secrets.yaml#L6

alejandro-belleza-globant:
> @PeteMac88 I agree, is that your use case?
> 
> @alejandrobelleza Do I read it correctly, that after setting `existingSecret` the old secret is still created? I'd expect that it is not. Similar to how it works in the majority of bitnami charts https://github.com/bitnami/charts/blob/3a0dfccb3e0eb0ecd344eaae5f53b547135da938/bitnami/nats/templates/secrets.yaml#L6

Yes, there was an if there to support secret creation. Sorry during rebase that change got lost. Going to submit a quick fix.

alejandrobelleza:
https://github.com/wundergraph/cosmo/pull/518 - @StarpTech , the fix

StarpTech:
@PeteMac88 does it address your concern too?

PR #515, opened 2024-02-12
feat: router fleet management

StarpTech:
## Motivation and Context

This is the first version of router fleet management. It allows the user to keep track of all running routers. This includes resource utilization like CPU and memory as well as to validate the current deployed router configuration.

Router runtime metrics are pushed automatically to Cosmo Cloud. This can be configured through:

```yaml
telemetry:
  metrics:
    router_runtime: false
```
In order to build a logical cluster, you have to specify a cluster name like this:

```yaml
instance_id: my_replica_1
cluster:
  name: us-central1-cosmo-cloud 
telemetry:
  service_name: my-application
```

Optionally, we accept a custom instance ID which makes your instance stable for metrics. Otherwise, a restart of your router will generate a new instance ID by default. You could always specify a service name for telemetry data. We use this information to render a human friendly name.

image

## TODO


router/pkg/metric/otlp_metric_store.go
```diff
@@ -0,0 +1,125 @@
...
+	baseKeys = append(baseKeys, attr...)
+
+	baseAttributes := otelmetric.WithAttributes(baseKeys...)
+
+	if c, ok := h.requestMeasures.counters[ResponseContentLengthCounter]; ok {
+		c.Add(ctx, size, baseAttributes)
+	}
+}
+
+func (h *OtlpMetricStore) MeasureLatency(ctx context.Context, requestStartTime time.Time, attr ...attribute.KeyValue) {
```
jensneuse, line 107: Will this work for Subscriptions?

jensneuse (approved):

jensneuse:
LGTM
take a look at latency vs Subscriptions time
don't forget to document the configuration and add a guide/docs in the Router section on how to setup clusters with instance ID, etc...

router/pkg/metric/otlp_metric_store.go
StarpTech: For subscriptions we will create different metrics. As today, we don't collect metrics for subscriptions.

Issue #534, opened 2024-02-15
Router - Forward generated Request-Id to subgraph


oxyno-zeta:
### Component(s)

router

### Is your feature request related to a problem? Please describe.

As an operation team member, I would like to forward the generated request-id (generated by chi middleware if request doesn't have it) to subgraph.

### Describe the solution you'd like

In configuration and using the "Header manipulation" feature, I would like to be able to forward X-Request-Id (even if it is generated by router backend) to subgraph backends.

Something like this:

```yaml
headers:
  all: # Header rules for all origin requests.
    request:
      - op: "propagate"
        named: X-Request-Id
```

oxyno-zeta:
I have something working on my computer, I'm just lacking time to make tests for the moment.

Moreover, can you tell me where I can add tests for this ?

I've started modifying the router.go file in core package with this:

(start at line 776)
```golang
	httpRouter.Use(middleware.RequestID)
	// Rewrite request id in request header to allow "simple" forward to subgraph
	httpRouter.Use(func(h http.Handler) http.Handler {
		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
			// Get request id
			reqID := middleware.GetReqID(r.Context())

			// Save it in request header
			// Note: if it wasn't generated by middleware, it will override with the same value
			// Note: so nothing important is made in this case
			r.Header.Set(middleware.RequestIDHeader, reqID)

			// Next
			h.ServeHTTP(w, r)
		})
	})
```

flymedllva:
I think being able to write my own
```go
func newHTTPTransport(opts *SubgraphTransportOptions) http.RoundTripper {
```
with its own `http.RoundTripper` chain would solve this problem.

oxyno-zeta:
@flymedllva : Yes custom modules can help for this but I proposed this because I can imagine multiple people wanted the same thing. If this is rejected by maintainers, that won't a problem for me, I will try to do it only via custom modules.

PR #535, opened 2024-02-15
feat(router): Forward generated request id to subgraph

oxyno-zeta:
Add generated request id into request to allow forward to subgraph

Related to issue #534

## Motivation and Context

That will allow to forward generated request id to subgraphs

## TODO


oxyno-zeta:
Feel free to tell me where to add tests or anything !

router/core/router.go
```diff
@@ -795,6 +795,21 @@ func (r *Router) newServer(ctx context.Context, routerConfig *nodev1.RouterConfi
 	})
 	httpRouter.Use(recoveryHandler)
 	httpRouter.Use(middleware.RequestID)
+	// Rewrite request id in request header to allow "simple" forward to subgraph
```
StarpTech, line 798: Hi @oxyno-zeta, this will work but we put such logic into [`preOriginHandlers`](https://github.com/wundergraph/cosmo/blob/a1005919a28ee8c9e9c3b87c730e47db336bceec/router/core/router.go#L258). This is the same way as custom module are developed and they are executed before the request to the subgraph. There is also a unit test example.

StarpTech (changes requested):

router/core/router.go
oxyno-zeta: Hello @StarpTech , are you saying that you want me to create a preOriginHandlers at this line ?

Like this ?

```golang
	r.preOriginHandlers = append(r.preOriginHandlers, func(req *http.Request, ctx RequestContext) (*http.Request, *http.Response) {
		// Get request id
		reqID := middleware.GetReqID(req.Context())

		// Save it in request header
		// Note: if it wasn't generated by middleware, it will override with the same value
		// Note: so nothing important is made in this case
		req.Header.Set(middleware.RequestIDHeader, reqID)

		// Next
		return req, nil
	})
	r.preOriginHandlers = append(r.preOriginHandlers, r.headerRuleEngine.OnOriginRequest)
```

StarpTech: Take a look at this example. It implements the `EnginePreOriginHandler` interface. https://github.com/wundergraph/cosmo/blob/a1005919a28ee8c9e9c3b87c730e47db336bceec/router/core/header_rule_engine.go
Test: https://github.com/wundergraph/cosmo/blob/a1005919a28ee8c9e9c3b87c730e47db336bceec/router/core/header_rule_engine_test.go

oxyno-zeta: I'm very sorry @StarpTech . Maybe because it is late here but I don't understand what you looking for. Do you want to create a full header rule engine for this ?

StarpTech: We must be in a similar timezone :smile: This

```go
package core

import (
	"net/http"
)

var (
	_ EnginePreOriginHandler = (*ForwardRequestHeader)(nil)
)

type ForwardRequestHeader struct {
}

func NewHeaderTransformer(rules config.HeaderRules) *ForwardRequestHeader {
	return &ForwardRequestHeader{}
}

func (h ForwardRequestHeader) OnOriginRequest(request *http.Request, ctx RequestContext) (*http.Request, *http.Response) {

	ctx.Request().Header.Set("X-Request-ID", request.Header.Get("X-Request-ID"))

	return request, nil
}

```

Then you can append  on `preOriginHandlers`

oxyno-zeta: Yes I think we are in the same one ;)

Oh I see what you are requesting. I will try that way !

oxyno-zeta: I've updated the code with your recommendation. This is working well.

StarpTech:
Hi @oxyno-zeta, I'm closing this PR in favor of https://github.com/wundergraph/cosmo/pull/573 I had to modify a few places.

oxyno-zeta:
Hello @StarpTech , sure no problem! Thanks for taking this time !

PR #538, opened 2024-02-15
chore: fix command in aws-lambda-router readme

vasily-polonsky:
## Motivation and Context

There is a typo in the readme file. wgc federated-graph fetch command fetches graphql sdl file, but we need to fetch router.json config file.

btw, contributors guide is not public, so cannot be accessed https://github.com/wundergraph/cloud/blob/main/CONTRIBUTING.md

## TODO


StarpTech:
HI @vasily-polonsky, thank you for the PR. Nice catch!

> btw, contributors guide is not public, so cannot be accessed 

Good point, will create a separate one here.

StarpTech (approved):
LGTM

PR #550, opened 2024-02-18
fix: don't expose token on wgc list command

StarpTech:
## Motivation and Context

For security reasons, graph tokens should never be available after creation. We already warn the user that we won't show the token again. During this, I added the information who created the token. I also simplified the list command.

```
┌─────────┬─────────────────────┬──────────────┬──────────────────────────┐
│ NAME    │ AUTHOR              │ LAST_USED_AT │ CREATED_AT               │
├─────────┼─────────────────────┼──────────────┼──────────────────────────┤
│ mytoken │ foo@wundergraph.com │ Never        │ 2024-02-18T18:32:17.568Z │
└─────────┴─────────────────────┴──────────────┴──────────────────────────┘
```

## TODO


JivusAyrus (approved):

controlplane/src/core/repositories/FederatedGraphRepository.ts
```diff
@@ -891,13 +895,24 @@ export class FederatedGraphRepository {
...
+    const userRepo = new UserRepository(this.db);
+    const user = await userRepo.byId(input.createdBy);
+
+    if (!user) {
+      throw new Error('User not found');
+    }
+
     const key = keys[0];
 
+    // TOOD
```
JivusAyrus, line 907: Required?

StarpTech: No, will remove, thanks.

PR #562, opened 2024-02-21
feat: version update check

thisisnithin:
## Motivation and Context

This introduces a update check for the latest version of CLI on running a command. The cooldown is 24 hours. It can be disabled by setting env variable `DISABLE_UPDATE_CHECK=true`.

image

## TODO

### DIFF
--- a/cli/src/utils.ts
+++ b/cli/src/utils.ts
@@ -5,6 +5,7 @@ import { program } from 'commander';
 import yaml from 'js-yaml';
 import pc from 'picocolors';
 import { config, configDir, configFile } from './core/config.js';
+import { KeycloakToken } from './commands/auth/utils.js';
 
 export interface Header {
   key: string;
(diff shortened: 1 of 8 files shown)


cli/src/utils.ts
```diff
@@ -73,3 +76,48 @@ export function checkAPIKey() {
...
+    );
+
+    const newData = yaml.dump({
+      ...configFileData,
+      lastUpdateCheck: currentTime,
+    });
+    mkdirSync(configDir, { recursive: true });
+    writeFileSync(configFile, newData);
+  } catch {
+    console.log(pc.red('Failed to check for updates'));
```
StarpTech, line 121: We should throw with an error so they can open a proper report on github.

cli/src/utils.ts
```diff
@@ -73,3 +76,48 @@ export function checkAPIKey() {
...
+        borderColor: 'yellow',
+        borderStyle: 'round',
+      }),
+    );
+
+    const newData = yaml.dump({
+      ...configFileData,
+      lastUpdateCheck: currentTime,
+    });
+    mkdirSync(configDir, { recursive: true });
```
StarpTech, line 118: Please put this into a "pre" config action.

cli/src/utils.ts
```diff
@@ -73,3 +76,48 @@ export function checkAPIKey() {
...
+    writeFileSync(configFile, newData);
```
StarpTech, line 119: Please also provide helper methods to write and read from the user config. That's a common use case.

cli/src/utils.ts
```diff
@@ -73,3 +76,48 @@ export function checkAPIKey() {
...
+export const checkForUpdates = async () => {
+  try {
+    if (config.disableUpdateCheck === 'true') {
+      return;
+    }
+
+    const currentTime = Date.now();
+
+    const data = existsSync(configFile) ? yaml.load(readFileSync(configFile, 'utf8')) ?? {} : {};
+    const configFileData = JSON.parse(JSON.stringify(data));
```
StarpTech, line 89: We should deal with the config in a type-safe way.

StarpTech (changes requested):

cli/src/utils.ts
thisisnithin: Done

cli/src/utils.ts
thisisnithin: Done

cli/src/utils.ts
thisisnithin: Done

cli/src/utils.ts
thisisnithin: Done

StarpTech (approved):
LGTM

PR #564, opened 2024-02-21
feat: add new subgraph fetch cli command

co-sic:
## Motivation and Context
Currently there is no way to fetch a subgraph SDL through the CLI. This is very helpful if you have a lot of subgraphs and want to do a local composition for development where you just changed one subgraph and you don't want to start all the other subgraphs just for the schema introspection.

## TODO


co-sic:
Not sure how to acutally change the docs. How do you use gitbook?

jensneuse:
I like it. Just one question, would it make sense to add another command to load all subgraph SDLs and the federated SDL, and store it all in a directory?

co-sic:
@jensneuse sounds like a good idea to me! Should i try to add this here, or maybe in another PR?

co-sic:
I'm also thinking if it might be a good idea to modify the federated-graph fetch command to have the option to also fetch the subgraphs and save it to a folder or if it is better to have a new command. I think having an option to include subgraph sounds fine, but might be a little confusing if you save to a single file in the standard way but then if you include you save to a dir.

jensneuse:
> I'm also thinking if it might be a good idea to modify the federated-graph fetch command to have the option to also fetch the subgraphs and save it to a folder or if it is better to have a new command. I think having an option to include subgraph sounds fine, but might be a little confusing if you save to a single file in the standard way but then if you include you save to a dir.

I think I'd keep the two separate.

co-sic:
Would help a lot if this could be merged soon, so i can continue the migration process :sweat_smile:

StarpTech:
I need to do a final test. We can merge it Monday.

co-sic:
Perfect, thanks :)

StarpTech (approved):
LGTM

StarpTech:
Hi @co-sic, the implementation is not correct. `getLatestValidSubgraphSDLByName` expects the name of the federated-graph but it was not passed. The graph name should not be required when fetching the schema of a subgraph only.

cli/src/commands/subgraph/commands/fetch.ts
```diff
@@ -0,0 +1,47 @@
...
+export default (opts: BaseCommandOptions) => {
+  const command = new Command('fetch');
+  command.description('Fetches the latest valid SDL of a subgraph. The output can be piped to a file.');
+  command.argument('<name>', 'The name of the subgraph to fetch.');
+  command.option('-n, --namespace [string]', 'The namespace of the subgraph.');
+  command.option('-o, --out [string]', 'Destination file for the SDL.');
+  command.action(async (name, options) => {
+    const resp = await opts.client.platform.getLatestValidSubgraphSDLByName(
+      {
+        name,
```
StarpTech, line 19: The current implementation expects the name of the federated graph. Please see https://github.com/wundergraph/cosmo/pull/564#issuecomment-1964537496

StarpTech (changes requested):

PR #579, opened 2024-02-27
feat: enrich logger for platform calls

thisisnithin:
## Motivation and Context

Add user and org id to log info .This helps in debugging user issues better.

## TODO

### DIFF
--- a/controlplane/src/core/services/Authentication.ts
+++ b/controlplane/src/core/services/Authentication.ts
@@ -15,6 +15,7 @@ const maxAuthCacheTtl = 30 * 1000; // 30 seconds
 export interface Authenticator {
   authenticate(headers: Headers): Promise<AuthContext>;
   authenticateRouter(headers: Headers): Promise<GraphKeyAuthContext>;
+  extractUserAndOrgId(headers: Headers): Promise<{ userId?: string; organizationId?: string }>;
 }
 
 export class Authentication implements Authenticator {
@@ -27,6 +28,19 @@ export class Authentication implements Authenticator {
     private graphKeyAuth: GraphApiTokenAuthenticator,
     private orgRepo: OrganizationRepository,
   ) {}
+
+  public async extractUserAndOrgId(headers: Headers): Promise<{ userId?: string; organizationId?: string }> {
+    try {
+      const user = await this.webAuth.authenticate(headers);
+      const organization = await this.orgRepo.bySlug(user.organizationSlug);
+      return {
+        userId: user.userId,
+        organizationId: organization?.id,
+      };
+    } catch {
+      return {};
+    }
+  }
 
   /**
    * Authenticate a user for an organization.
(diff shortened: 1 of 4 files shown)


controlplane/src/core/services/Authentication.ts
```diff
@@ -28,6 +29,19 @@ export class Authentication implements Authenticator {
     private orgRepo: OrganizationRepository,
   ) {}
 
+  public async extractUserAndOrgId(headers: Headers): Promise<{ userId?: string; organizationId?: string }> {
+    try {
+      const user = await this.webAuth.authenticate(headers);
```
StarpTech, line 34: Why do we need do this again? Everything should be available in the authcontext. Also some requests are issued by tokens, let's make this visible in the logs too. I recommend to pass the authcontext to `enrichLogger`.

controlplane/src/core/util.ts
```diff
@@ -54,6 +56,21 @@ export async function handleError<T extends ResponseMessage>(
...
+export const enrichLogger = async (req: object, ctx: HandlerContext, logger: Logger, authenticator: Authenticator) => {
+  const authContext = await authenticator.extractUserAndOrgId(ctx.requestHeader);
+  return logger.child({
+    service: ctx.service.typeName,
+    method: ctx.method.name,
+    actor: {
+      userId: authContext.userId,
+      organizationId: authContext.organizationId,
+    },
+    req: {
```
StarpTech, line 68: The request could contain sensitive information and also files. Let's start with the org and user context.

StarpTech (changes requested):

controlplane/src/core/util.ts
thisisnithin: The request body should never contain sensitive information. Its public by default.

controlplane/src/core/services/Authentication.ts
thisisnithin: The logger enriching should happen outside the error handler for us to have information in the handler itself

controlplane/src/core/util.ts
StarpTech: Email is already sensitive. We use that information in several places.

controlplane/src/core/services/Authentication.ts
StarpTech: We need to find a solution without duplicating this. Based on your thoughts, we would also fail to get context when the `authenticate` fails for whatever reasons. I suggest a pragmatic approach e.g.

```ts
const logger = await enrichLogger(ctx, opts.logger, authContext);
```

If we ensure that this

```ts
const authContext = await opts.authenticator.authenticate(ctx.requestHeader);
```

is the first line of the handler, it has the same consistency as you proposed but without duplicating code.

StarpTech (approved):
LGTM

PR #593, opened 2024-03-01
fix: handle empty files in the cli itself

JivusAyrus:
## TODO


cli/src/commands/subgraph/commands/fix.ts
```diff
@@ -30,11 +30,18 @@ export default (opts: BaseCommandOptions) => {
       );
     }
 
+    const schema = await readFile(schemaFile);
+    if (schema.length === 0) {
```
jensneuse, line 34: What if a schema contains some whitespace, like a newline, space or tab?

jensneuse (changes requested):

cli/src/commands/subgraph/commands/fix.ts
StarpTech: Do a `trim()` before :smile:

StarpTech (approved):
LGTM

jensneuse (approved):

PR #601, opened 2024-03-05
fix: crash due to missing explicit type cast

thisisnithin:
## Motivation and Context

If the array is empty, I think some postgres version crashes if type isn't explicitly specified. 

## TODO


thisisnithin:
This would never reach. Closing

PR #603, opened 2024-03-06
fix: subscription deduplication

jensneuse:
This PR...
- properly handles subscription deduplication
- adds trigger counting & defining trigger count expectations in tests (e.g. wait until trigger count == 0)
- fixes a flaky WebSockets test
- fix memory leak in WebSockets implementation
- adds a test with 3 overlapping Subscriptions from 2 Subgraphs


StarpTech (approved):
LGTM

PR #610, opened 2024-03-08
chore: koala identify call

thisisnithin:
## Motivation and Context

Identifies users on Koala.

Completes ENG-4941

## TODO

### DIFF
--- a/studio/src/components/app-provider.tsx
+++ b/studio/src/components/app-provider.tsx
@@ -99,6 +99,10 @@ const fetchSession = async () => {
     }
     return null;
   } catch (e) {
+    // Reset koala if custom head scripts are found and user is not authenticated
+    if (process.env.CUSTOM_HEAD_SCRIPTS) {
+      window.ko.reset();
+    }
     throw e;
   }
 };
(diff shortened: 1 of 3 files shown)


studio/src/components/app-provider.tsx
```diff
@@ -164,6 +170,17 @@ export const AppProvider = ({ children }: { children: ReactNode }) => {
         invitations: data.invitations,
       });
 
+      window.ko?.identify(data.email, {
```
StarpTech, line 173: Can we make it more obvious that this only works when the script was provided? Maybe add a condition around with a few comments?

StarpTech (changes requested):

studio/src/components/app-provider.tsx
thisisnithin: Done. Added reset too

studio/src/components/app-provider.tsx
```diff
@@ -164,6 +174,20 @@ export const AppProvider = ({ children }: { children: ReactNode }) => {
...
+      // If custom head scripts are available, enable identify call for koala script
+      if (process.env.CUSTOM_HEAD_SCRIPTS) {
+        window.ko?.identify(data.email, {
+          id: data.id,
+          $account: {
+            organizationId: organization.id,
+            organizationName: organization.name,
+            organizationSlug: organization.slug,
+            plan: organization.plan,
+            roles: organization.roles,
```
StarpTech, line 186: We don't need to send the roles.

StarpTech (changes requested):

studio/src/components/app-provider.tsx
```diff
@@ -93,6 +99,10 @@ const fetchSession = async () => {
     }
     return null;
   } catch (e) {
+    // Reset koala if custom head scripts are found and user is not authenticated
+    if (process.env.CUSTOM_HEAD_SCRIPTS) {
+      window.ko?.reset();
```
StarpTech, line 104: We also need to call `reset` when the user clicks on logout.

StarpTech (changes requested):

studio/src/components/app-provider.tsx
thisisnithin: Removed.

studio/src/components/app-provider.tsx
thisisnithin: Done.

StarpTech (approved):
Lgtm

PR #614, opened 2024-03-10
fix(controlplane): fix in webDomain replace

d4rkstar:
Added a check on webDomain: the replace should happen only when webDomain is not an IPv4 address.

## Motivation and Context

I tried to host locally on IPv4 static address, without a domain name and the sessions are never written to database.


d4rkstar:
Sorry team, I have added and reverted another commit I would like to send in another PR

PR #628, opened 2024-03-13
feat: router config signature validation through custom admission webhooks

StarpTech:
## TODO


controlplane/src/core/webhooks/utils.ts
```diff
@@ -1,55 +1,33 @@
...
-        log = log.child({ message: error.message });
-      }
-
-      if (logLevel === 'error') {
-        log.error('Could not send webhook event');
+    .catch((error: AxiosError) => {
+      if (error instanceof AxiosError) {
+        logger.debug({ statusCode: error.response?.status, message: error.message }, 'Could not send webhook event');
       } else {
-        log.debug('Could not send webhook event');
```
thisisnithin, line 52: Please log as errors instead of debug for platform webhooks. It has helped me to see if something was wrong before.

admission-server/README.md
```diff
@@ -0,0 +1,38 @@
...
+```bash
+wgc subgraph publish employees --schema ../demo/pkg/subgraphs/employees/subgraph/schema.graphqls --labels team=A
+```
+
+3. Start the router with the same signing key as the admission server
+```yaml
+version: '1'
+
+graph: 
+  sign_key: 'sign_key'
```
jensneuse, line 35: how about renaming to "config_signature_key" and adding "config_signature_algorithm" enum with a default value of what you've implemented to keep the alg flexible.

admission-server/src/index.ts
```diff
@@ -0,0 +1,49 @@
...
+    },
+  });
+
+  if (!resp.ok) {
+    return c.json({ error: 'Failed to fetch private config' }, 500);
+  }
+
+  const hmacDigest = await makeSignature(await resp.text(), signatureKey);
+
+  return c.json({ signatureSHA256: hmacDigest }, 200);
```
jensneuse, line 40: How does the user know what alg the Router uses for validation?

cdn-server/cdn/src/index.ts
```diff
@@ -2,22 +2,32 @@ import { JWTVerifyResult, jwtVerify } from 'jose';
 import { Context, Env, Hono, Next, Schema } from 'hono';
 import { stream } from 'hono/streaming';
 
+export const signatureSha256Header = 'X-Signature-SHA256';
```
jensneuse, line 5: Should the header be independent from the alg so that the user can use a different alg if they want?

cdn-server/cdn/src/index.ts
```diff
@@ -45,81 +56,92 @@ declare module 'hono' {
...
     }
-    const [type, token] = authHeader.split(' ');
-    if (type !== 'Bearer' || !token) {
-      return c.text('Unauthorized', 401);
+
+    let jwt = '';
+    if (queryToken) {
+      jwt = queryToken;
+    } else if (authHeader) {
+      const [type, token] = authHeader.split(' ');
```
jensneuse, line 68: Is this always a Bearer token?

thisisnithin (changes requested):

cli/src/commands/router/commands/fetch.ts
```diff
@@ -70,31 +75,34 @@ export default (opts: BaseCommandOptions) => {
...
-      );
-
-      if (latestConfigResp.response?.code !== EnumStatusCode.OK) {
-        console.log(`${pc.red(`No router config could be fetched for federated graph ${pc.bold(name)}`)}`);
-        if (resp.response?.details) {
-          console.log(pc.red(pc.bold(resp.response?.details)));
-        }
+
+    const body = await response.text();
+    const signature = response.headers.get('X-Signature-SHA256');
```
jensneuse, line 80: Should the header name be a const and re-used across all methods?

controlplane/src/core/webhooks/utils.ts
StarpTech: Good point. Done.

thisisnithin (approved):
LGTM

router/internal/cdn/router_config.go
```diff
@@ -110,29 +119,71 @@ func (cdn *RouterConfigClient) RouterConfig(ctx context.Context, version string)
...
+
+	if cdn.signatureKey != "" {
+
+		configSignature := resp.Header.Get(sigResponseHeaderName)
+		if configSignature == "" {
+			return nil, errors.New("signature header not found in CDN response")
+		}
+
+		// create a signature of the received config body
+		hasher := hmac.New(sha256.New, []byte(cdn.signatureKey))
```
jensneuse, line 170: Should we allow for selecting an alg?

router/pkg/config/config.schema.json
```diff
@@ -19,6 +19,12 @@
         "token": {
           "type": "string",
           "description": "The token used to authenticate with other component from Cosmo. Can be ommitted if the router is started with a static execution config."
+        },
+        "sign_key": {
```
jensneuse, line 23: how about "config_signature_key" and adding "config_signature_algorithm" with an enum value, defaulting to "hmac-sha-256" to allow for changes in the future?

router/pkg/config/config.schema.json
```diff
@@ -19,6 +19,12 @@
...
+          "type": "string",
+          "minLength": 32,
+          "maxLength": 32,
+          "description": "The key used to verify the graph config signature when downloading from the CDN. The same key was used to create the signature in the admission webhook '/validate-config'. If the key is not set, the router will not verify the graph configuration. The key must be a 32 byte long string."
```
jensneuse, line 27: Should we add a note here on how to properly generate such a secret?

jensneuse (changes requested):

admission-server/README.md
StarpTech: That would require to match the right algorithm in the admission handler. I'd like to be opinionated here to make it simpler. HMAC256 is a reasonable default which is used across the industry. Let's wait for feedback also.

cdn-server/cdn/src/index.ts
StarpTech: I see no reason why we should make it flexible right now. See my comment above.

cdn-server/cdn/src/index.ts
StarpTech: Yes

cli/src/commands/router/commands/fetch.ts
StarpTech: It is used only here in the cli.

router/internal/cdn/router_config.go
StarpTech: I say no. See comment above.

router/pkg/config/config.schema.json
StarpTech: This should solve the documentation, I'll add later.

router/pkg/config/config.schema.json
StarpTech: Let's wait for feedback. We can always make it flexible while preserving the defauilt.

jensneuse (approved):

PR #631, opened 2024-03-14
fix: improve ui and logs

JivusAyrus:
## TODO

### DIFF
--- a/controlplane/src/core/bufservices/PlatformService.ts
+++ b/controlplane/src/core/bufservices/PlatformService.ts
@@ -1227,7 +1227,7 @@ export default function (opts: RouterOptions): Partial<ServiceImpl<typeof Platfo
 
         const schemaChanges = await getDiffBetweenGraphs(subgraph.schemaSDL, newSchemaSDL);
         if (schemaChanges.kind === 'failure') {
-          logger.debug(`Error finding diff between graphs: ${schemaChanges.error}`);
+          logger.error(`Error finding diff between graphs: ${schemaChanges.error}`);
           return {
             response: {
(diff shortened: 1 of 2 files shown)


controlplane/src/core/bufservices/PlatformService.ts
```diff
@@ -1227,7 +1227,7 @@ export default function (opts: RouterOptions): Partial<ServiceImpl<typeof Platfo
 
         const schemaChanges = await getDiffBetweenGraphs(subgraph.schemaSDL, newSchemaSDL);
         if (schemaChanges.kind === 'failure') {
-          logger.debug(`Error finding diff between graphs: ${schemaChanges.error}`);
+          logger.error(`Error finding diff between graphs: ${schemaChanges.error}`);
```
StarpTech, line 1230: This is a user error correct? We should never log users errors as errors because we have infrastructure alerts on it.

StarpTech (changes requested):

controlplane/src/core/bufservices/PlatformService.ts
JivusAyrus: will logger.warn also trigger a alert?

debug logs are not seen on the log explorer as I think its configured with info

StarpTech: Yes, let's do warn.

StarpTech (approved):
LGTM

controlplane/src/core/bufservices/PlatformService.ts
JivusAyrus: Done, changed to warning

PR #635, opened 2024-03-15
feat: support authentication and multiple sources for nats edfs

Aenimus:
Relies on https://github.com/wundergraph/graphql-go-tools/pull/757

Fixes ENG-4945
Fixes ENG-4692
Fixes ENG-4959

## TODO


demo/pkg/subgraphs/subgraphs.go
```diff
@@ -125,75 +125,89 @@ func subgraphHandler(schema graphql.ExecutableSchema) http.Handler {
...
-		url = u
+	if defaultSourceNameURL := os.Getenv("NATS_URL"); defaultSourceNameURL != "" {
+		url = defaultSourceNameURL
+	}
+	defaultConnection, err := nats.Connect(url)
+	if err != nil {
+		log.Printf("failed to connect to nats source \"default\": %v", err)
 	}
-	nc, err := nats.Connect(url)
+	secondUrl := "nats://127.0.0.1:4223"
```
StarpTech, line 172: Will this be handled gracefully on fly?

demo/pkg/subgraphs/subgraphs.go
```diff
@@ -125,75 +125,89 @@ func subgraphHandler(schema graphql.ExecutableSchema) http.Handler {
...
-	if srv := newServer("employees", config.EnableDebug, config.Ports.Employees, employees.NewSchema(nc)); srv != nil {
+	if srv := newServer("employees", config.EnableDebug, config.Ports.Employees, employees.NewSchema(natsConnectionBySourceName)); srv != nil {
 		servers = append(servers, srv)
 	}
-	if srv := newServer("family", config.EnableDebug, config.Ports.Family, family.NewSchema(nc)); srv != nil {
+	if srv := newServer("family", config.EnableDebug, config.Ports.Family, family.NewSchema(natsConnectionBySourceName)); srv != nil {
 		servers = append(servers, srv)
 	}
-	if srv := newServer("hobbies", config.EnableDebug, config.Ports.Hobbies, hobbies.NewSchema(nc)); srv != nil {
+	if srv := newServer("hobbies", config.EnableDebug, config.Ports.Hobbies, hobbies.NewSchema(natsConnectionBySourceName)); srv != nil {
```
StarpTech, line 195: Why does every subgraph needs a nats connection?

router/.env.example
```diff
@@ -8,5 +8,6 @@ DEV_MODE=true
 GRAPH_API_TOKEN=<generated_with_wgc>
 CDN_URL=http://127.0.0.1:11000
 NATS_URL=nats://localhost:4222
+NATS_URL_TWO=nats://localhost:4223
```
StarpTech, line 11: Please remove

router/pkg/config/config.go
```diff
@@ -249,9 +249,20 @@ type CDNConfiguration struct {
 	CacheSize BytesString `yaml:"cache_size,omitempty" envconfig:"CDN_CACHE_SIZE" default:"100MB"`
 }
 
+type NatsAuthentication struct {
```
StarpTech, line 252: I'd split it in `TokenBasedAuthenthication` and `UsernamePasswordAuthenthication` and inline both in `Authentication`

StarpTech (changes requested):

router/.env.example
Aenimus: Based on our discussion, should I also remove `NATS_URL`?

demo/pkg/subgraphs/subgraphs.go
Aenimus: It was originally configured this way; no other reason.

demo/pkg/subgraphs/subgraphs.go
Aenimus: This is not used by Fly.

router/pkg/config/config.go
Aenimus: How do you feel about `CredentialsBasedAuthentication` rather than `UsernamePasswordAuthentication`?

demo/pkg/subgraphs/countries/subgraph/resolver.go
```diff
@@ -11,6 +11,6 @@ import (
 // It serves as dependency injection for your app, add any dependencies you require here.
 
 type Resolver struct {
-	NC  *nats.Conn
-	mux sync.Mutex
+	NatsConnectionBySourceName map[string]*nats.Conn
```
jensneuse, line 14: How about passing a map[string]PubSub?
That'll be more flexible in the future to add other PubSubs, instead of having to add Kafka, MQTT, etc...

demo/pkg/subgraphs/subgraphs.go
jensneuse: It allows us to publish messages from all subgraphs, just in case.
We only do this currently from one Subgraph I think, but it's flexible for testing purposes.

go.work
```diff
@@ -9,4 +9,4 @@ use (
 	./router-tests
 )
 
-//replace github.com/wundergraph/graphql-go-tools/v2 => ../../graphql-go-tools/v2
+//replace github.com/wundergraph/graphql-go-tools/v2 => ../graphql-go-tools/v2
```
jensneuse, line 12: please don't change this file
ideally, we could add it to the gitignore, but have a default somehow

router-tests/websocket_test.go
```diff
@@ -197,9 +199,9 @@ func TestWebSockets(t *testing.T) {
 				// Trigger the subscription via NATS
 				subject := "employeeUpdated.3"
 				message := []byte(`{"id":3,"__typename": "Employee"}`)
-				err := xEnv.NC.Publish(subject, message)
+				err := xEnv.NatsConnectionBySourceName[defaultSourceName].Publish(subject, message)
```
jensneuse, line 202: I think we will only ever have 2 nats instances in the testenv.
Could we make this static and just add the two instances as 2 fields instead of having to pick from a map, which might lead to errors by itself?

jensneuse (changes requested):

go.work
Aenimus: I wouldn't normally commit this, but based on the default replace in router go.mod, the extra `../` seems wrong.

I.e., router is one dir deeper and the `go.mod` is `../../`, but here we're one dir up and it's the same relative path.

router-tests/websocket_test.go
Aenimus: I thought about this, but I added some foundations for supporting however many sources we want, in case we want to extend with, for example, Kafka, or whatever.

I can make it explicit fields if you prefer, though.

demo/pkg/subgraphs/subgraphs.go
Aenimus: I thnk we do it from two: Mood and Availability.

router-tests/websocket_test.go
jensneuse: I understand, this makes sense, although each additional client will require the client setup anyways.
So I'm not sure what we win through having a map.

Aenimus: I'll change it to explicit fields 👍

router/core/operation_planner.go
```diff
@@ -106,7 +109,7 @@ func (p *OperationPlanner) Plan(operation *ParsedOperation, clientInfo *ClientIn
 		if err != nil {
 			return nil, err
 		}
-		opContext.preparedPlan = &prepared
```
StarpTech, line 109: Why was the type changed?

router/core/router.go
```diff
@@ -1039,18 +1042,6 @@ func (r *Router) newServer(ctx context.Context, routerConfig *nodev1.RouterConfi
 		MaxOperationSizeInBytes: int64(r.routerTrafficConfig.MaxRequestBodyBytes),
 		PersistentOpClient:      r.cdnPersistentOpClient,
 	})
-	// Pre-hash all data source IDs to avoid races
```
StarpTech, line 1042: Did we clean this up?

StarpTech (changes requested):

router/core/router.go
devsergiy: yes, it is not needed anymore

demo/cmd/all/main.go
```diff
@@ -36,9 +36,10 @@ func main() {
 		},
 		EnableDebug: *debug,
 	}
-	subgraphs, err := subgraphs.New(&config)
+	ctx := context.Background()
+	subgraphs, err := subgraphs.New(ctx, &config)
```
jensneuse, line 40: Wouldn't it make more sense to create a ctx in the testenv, which we cancel using textenv cleanup:
https://github.com/wundergraph/cosmo/blob/5643cc9268f9f7b3fc3de86eaf7f3da11178e660/router-tests/testenv/testenv.go#L68

router/core/operation_planner.go
devsergiy: preparePlan now returns pointer right away or nil in case of error
Jens asked for that

jensneuse (approved):

demo/cmd/all/main.go
devsergiy: It is not for test env, it is cmd in demo to run all subgraphs

Test env already uses context which it propagates to subgraphs
https://github.com/wundergraph/cosmo/blob/7616e360047dd4ab183c1ddf835a079747ff1e50/router-tests/testenv/testenv.go#L162

PR #638, opened 2024-03-16
feat: allow to update admission url

StarpTech:
## Motivation and Context

- Allow to update admission url.
- Allow to unset several other flags with an empty value
- Cleanup flag handling.
- Remove unused header option from commands and proto in a backwards compatible way
- Add vitest.workspace.yaml for vscode debugging support (Upgrade the vscode plugin to the latest pre-release version)

## TODO

### DIFF
--- a/shared/src/utils/util.ts
+++ b/shared/src/utils/util.ts
@@ -1,3 +1,5 @@
+import { SubscriptionProtocol } from '../router-config/builder.js';
+
 export function delay(t: number) {
   return new Promise((resolve) => setTimeout(resolve, t));
 }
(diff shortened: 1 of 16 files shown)


shared/src/utils/util.ts
```diff
@@ -48,3 +48,16 @@ export function isValidUrl(url: string) {
     return false;
   }
 }
+
+export function isValidSubscriptionProtocol(protocol: string) {
```
thisisnithin, line 52: Please use the `SubscriptionProtocol` type instead of string

thisisnithin (changes requested):

shared/src/utils/util.ts
StarpTech: Good eye.

StarpTech: Done

thisisnithin (approved):

PR #639, opened 2024-03-16
feat: configure CSP for studio

StarpTech:
## Motivation and Context

Test [CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy) configuration without blocking content.

## TODO

### DIFF
--- a/studio/next.config.mjs
+++ b/studio/next.config.mjs
@@ -1,10 +1,50 @@
 import withMarkdoc from "@markdoc/next.js";
 import pkg from "./package.json" assert { type: "json" };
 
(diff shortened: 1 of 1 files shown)


studio/next.config.mjs
```diff
@@ -1,10 +1,50 @@
 import withMarkdoc from "@markdoc/next.js";
 import pkg from "./package.json" assert { type: "json" };
 
+const isProd = process.env.NODE_ENV === "production";
+const debugCSP = true
```
StarpTech, line 5: On by default to test if everything works in prod

jensneuse (approved):

PR #643, opened 2024-03-18
feat: improve planning time of small queries

devsergiy:
ENG-4945

## TODO


router/core/operation_planner.go
```diff
@@ -66,7 +66,10 @@ func (p *OperationPlanner) preparePlan(requestOperationName []byte, requestOpera
 		return planWithMetaData{}, &reportError{report: &report}
 	}
 
-	planner := plan.NewPlanner(context.Background(), p.executor.PlanConfig)
+	planner, err := plan.NewPlanner(p.executor.PlanConfig)
+	if err != nil {
+		return planWithMetaData{}, err
```
jensneuse, line 71: Should we return pointer to planWithMetaData, so we can return nil?

jensneuse (changes requested):

router/core/operation_planner.go
devsergiy: updated

jensneuse (approved):

PR #644, opened 2024-03-18
feat: Add support for subset of config.yaml in AWS Lambda Router

Tehnix:
## Motivation and Context
It's been really great to see the release and native support for AWS Lambda via [aws-lambda-router](https://github.com/wundergraph/cosmo/releases?q=aws-lambda-router&expanded=true)!

Currently the Router only supports a few settings via environment variables. I needed support for `override_routing_url` and a few other things, so I tried to stitch in loading of the configuration in a backwards compatible way:
- If you don't specify it, you'll use the current configuration and defaults
- If you specify a path to a config.yaml via `CONFIG_PATH`, then it will try to load the values from there.

Currently I've only added a subset of configurations, but it is now very easy to add more :) 

Caveat: I'm entirely new to Go, so this may not at all be idiomatic code and there is probably a better way 😅 I did however update the tests to make sure everything still passes while loading the config :) 

## TODO


StarpTech:
Hi @Tehnix, thank you for the PR. The current configuration is optimized for fast cold starts. `config.LoadConfig` could have a significant impact because we do JSONSchema compilation and validation on every cold start. Also not all default values we will inherit by calling `config.LoadConfig`  are suitable for severless e.g. we register the router on every start on the controlplane. By default this is enabled. If you need support for `override_routing_url` I recommend to compile your own binary. We will iterate on this very soon to make the configuration more flexible.

Tehnix:
No inconvenience at all! I assumed there was probably a deeper reason for not loading everything in the Lambda version :) 

Would the same be true for modules? I previously added support for the same functionality via a custom module that'll overwrite the subgraph URLs on the fly, using the normal Router.

Am I right in that modules are not supported for the Lambda Router also?

StarpTech:
Custom module are absolutely fine but you would need to build your own binary.

StarpTech:
I'm closing this PR based on the reasonings of my last comment.

PR #652, opened 2024-03-19
feat: support file upload in router

pedraumcosta:
## Motivation and Context

This PR makes possible to support a file upload according to spec:

https://github.com/jaydenseric/graphql-multipart-request-spec

**This PR depends on this PR from `graphql-go-tools`:**

https://github.com/wundergraph/graphql-go-tools/pull/758

This PR parse multiform part http requests with files, saves in the file system and passed the file information along the graphql tools library to be forwarded to a subgraph that supports files uploads. In order to make tests possible, I added file uploads operations to the demo employee subgraph and some enhanced methods to test infrastructure. 

## TODO


router/core/graphql_prehandler.go
```diff
@@ -160,29 +163,69 @@ func (h *PreHandler) Handler(next http.Handler) http.Handler {
...
 		}()
 
+		var body, files []byte
+		var err error
 		// XXX: This buffer needs to be returned to the pool only
 		// AFTER we're done with body (retrieved from parser.ReadBody())
 		buf := pool.GetBytesBuffer()
 		defer pool.PutBytesBuffer(buf)
+		if r.Header.Get("Content-Type") == "application/json" {
+			body, err = h.operationProcessor.ReadBody(buf, r.Body)
```
jensneuse, line 173: Have you thought about creating temp files in the system temp dir, stream all multipart files into the file system, read them as streams, and clean up the temp files after the request lifecycle.
I'm concerned that big files will occupy a lot of memory, so ideally, we can stream as much as possible and use the filesystem for temporary storage. How doy ou think about this?

pedraumcosta: Hum ... this is the regular json request body reading. In fact, in the following lines I trigger the body read from the multipart form using the library you have suggested to me:

https://github.com/wundergraph/cosmo/pull/652/files#diff-0c6c16945a1e9a216631c6bd16bcfbe9c07a6c96b1cf9b3aa93385436fab41e8R188

If I understood right, this library will handle the file using a stream. Seems very straight forward. The library claims there's no need to use files and have a low memory footprint. Assuming this is true, I wonder how I would pass a stream to the library which actually forward the request.

By the way I have a working naive version for `graphql-go-tools`

https://github.com/wundergraph/graphql-go-tools/pull/758/files

PR #670, opened 2024-03-26
chore: add graph feature flag rfc

jensneuse:
RFC to add Graph Feature Flags, a new functionality to enable autonomous experimentation in staging environments and gradually rolling out features.


rfc/graph-feature-flags.md
```diff
@@ -0,0 +1,242 @@
...
+Let's say you're starting with a monolithic GraphQL API and want to migrate to a Federated GraphQL API.
+You've got existing clients that depend on the old schema and have certain expectations about the characteristics of the API,
+like the performance of certain GraphQL Operations.
+
+With feature flags, we can publish a new Subgraph Schema that `@overrides` a field in the old schema, but don't automatically enable it for all clients.
+Instead, we can send a small percentage of traffic to the new Subgraph in a "shadow mode" and monitor its performance.
+This means that the new Subgraph is used to compute the result of the GraphQL Operation, but the result is not returned to the client.
+Instead, we compare the result of the newly introduced Subgraph with the result of the existing Monolith in terms of correctness and performance.
+
+If everything goes well, we can enable the feature flag for 1% of your users, keep monitoring,
```
StarpTech, line 46: How do you want to implement this? A router fleet won't be able to serve an exact percentage of the traffic because of a split-brain scenario. In the simplest implementation, it would be a rough estimate. Based on the load-balancing algorithm of the LB (least-connections), it could diverge even further. Usually, this is a task of a ingress or LB.

Both approaches has benefits:

**Proxy**
- Almost every proxy like Nginx, Caddy, Traefik can gradually roll-out traffic based on the feature flag header, Session affinity, Cookie. This integrates with existing platform / monitoring tools.
- Requires to deploy a different router instance for every feature flag which is quite a burden but can be automated.

**Router**
- We can test / rollout feature flags without deploying additional routers.
- Cosmo would need to provide excellent integration with OTEL / Prometheus to make gradual roll-outs approachable, from an operational and observability standpoint. Cosmo Studio needs to be super well integrated with feature-flag across the board.

In summary, I think the complexity we would hold back from the customer with an integrated solution will be huge. When we are doing it right, it can enable small companies very fast to start experiments. Bigger companies could use Cosmo or build upon this foundation to provide GraphQL specific tools. The tooling and integration are the keys here, but this is feasible.

rfc/graph-feature-flags.md
```diff
@@ -0,0 +1,242 @@
...
+This feature is called Graph Feature Flags, or GFF for short.
+
+GFF allows you to publish Subgraph Schema Changes under a feature flag.
+The runtime can dynamically enable or disable feature flags for each request.
+
+## Motivation
+
+There are four main reasons, and the correlated problems to solve, why we want to add Graph Feature Flags to Open Federation:
+
+1. **Feature Rollout**: Feature flags allow you to roll out new features gradually.
```
StarpTech, line 16: Config synchronization is a topic here. Currently, every router operates in a standalone mode. As a User, I'd like to know if my roll-out was successfully. According to the consistency mode, we might need to reconsiders other strategies to keep it in sync and to report status updates in realtime.

rfc/graph-feature-flags.md
```diff
@@ -0,0 +1,242 @@
...
+This means that you'll very likely have one or more frontend applications, and many Subgraphs that are part of the Federated API.
+To be able to test a new feature, you need the entire stack to be running, so it's unlikely that you can run the whole stack for each feature in isolation.
+As a consequence, you need a shared staging environment where developers can test their features without affecting other developers,
+hence the need for feature flags so that individual developers can test their changes in as much isolation as possible.
+
+In a staging environment, you can enable a feature flag by setting a specific cookie or header in your client.
+
+## Summarizing the Problem
+
+If we condense the four main reasons, we can summarize the problem as two main points:
```
StarpTech, line 82: It would be very valuable to draw the whole process of feature designing, roll-outs, analysis, rollbacks, and persistent. Then compare how well this can be integrated in existing CI/CD workflows.

rfc/graph-feature-flags.md
```diff
@@ -0,0 +1,242 @@
...
+
+1. **Feature Rollout**: We need a way to gradually roll out new features to our users
+2. **Experimental Features**: We need a way to test new features without affecting the "main" Graph
```
StarpTech, line 85: A problem I see are version compatibility between compositions and the routers. This is something we have to keep in mind. Ideally, we can validate this before enabling the feature flag.

rfc/graph-feature-flags.md
```diff
@@ -0,0 +1,242 @@
...
+
+What this would allow teams to do is to set up a new feature flag,
+and then let the system automatically roll out the feature without requiring the platform team to deploy extra infrastructure or handle the rollout manually.
+
+The goal here is to avoid the platform team becoming a bottleneck for rolling out new features,
+but instead, to empower the teams to roll out new features safely and quickly,
+and to provide them but also the platform team with the necessary tools to monitor and observe rollouts.
+
+As you scale your API across more and more teams,
+we want to make sure that the platform team can focus on the platform itself,
```
StarpTech, line 199: The more feature flags a subgraph has, the more variants it gets. Also here tooling is the key. I also want to understand what services "infrastructure" needs to run to serve all or X variants.

rfc/graph-feature-flags.md
```diff
@@ -0,0 +1,242 @@
...
+Existing clients will continue to use the previous version of the Schema.
+
+To enable the feature flag for a specific request, you can set a cookie or header in your client.
+The Header Syntax is `X-WG-Feature-Flag: products.v2`.
+Multiple feature flags can be enabled by separating them with a comma: `X-WG-Feature-Flag: products.v2,employees.userAvatar`.
+
+By default, the feature flag will implicitly have a `--feature-flag-mode=replace` flag.
+This means that the Subgraph with the feature flag will replace the default Subgraph.
+
+It's also possible to use the `--feature-flag-mode=combine` flag.
```
StarpTech, line 120: It could think of a nice integration in Cosmo , that allows to issue individual tokens with different feature flags to the users. Further, this could enable users to monetize their API based on features.

rfc/graph-feature-flags.md
```diff
@@ -0,0 +1,242 @@
...
+```
+
+We've deployed a copy of the API to a different port, and we've published a new Subgraph Schema under the feature flag `products.experimental`.
+We've also disabled composition checks for this Subgraph, as we're testing a new feature that might not be compatible with changes that other developers are making.
+
+To be able to test this new feature end-to-end, we need to enable the feature flag in our client.
+We can achieve this by setting a cookie:
+
+```bash
+curl -X POST http://localhost:4002/graphql -H "Content-Type: application/json" -H "Cookie: X-WG-Feature-Flag=products.experimental" --data '{ "query": "{ products { id name } }" }'
```
StarpTech, line 228: This is a request to the router correct? Port `4002` was used by the subgraph above.

rfc/graph-feature-flags.md
```diff
@@ -0,0 +1,242 @@
...
+```bash
+npx wgc subgraph publish products --schema ../demo/subgraphs/products/products.graphql --routing-url http://localhost:4001/v2/graphql --label=team=A --feature-flag=v2
+```
+
+This will publish the new Subgraph Schema under the feature flag `products-v2`.
+Existing clients will continue to use the previous version of the Schema.
+
+To enable the feature flag for a specific request, you can set a cookie or header in your client.
+The Header Syntax is `X-WG-Feature-Flag: products.v2`.
+Multiple feature flags can be enabled by separating them with a comma: `X-WG-Feature-Flag: products.v2,employees.userAvatar`.
```
akoenig, line 115: Just a minor addition: `X-WG-Feature-Flag: bleeding` for enabling all flags 🩸

rfc/graph-feature-flags.md
```diff
@@ -0,0 +1,242 @@
...
+We can add a `--feature-flag` flag to the `subgraph publish` command:
+
+```bash
+npx wgc subgraph publish products --schema ../demo/subgraphs/products/products.graphql --routing-url http://localhost:4001/v2/graphql --label=team=A --feature-flag=v2
+```
+
+This will publish the new Subgraph Schema under the feature flag `products-v2`.
+Existing clients will continue to use the previous version of the Schema.
+
+To enable the feature flag for a specific request, you can set a cookie or header in your client.
```
akoenig, line 113: There could be a nice integration in the Cosmo Playground where you toggle the respective feature flags then.

rfc/graph-feature-flags.md
```diff
@@ -0,0 +1,242 @@
...
+The Header Syntax is `X-WG-Feature-Flag: products.v2`.
```
akoenig, line 114: Wondering if this could be solved via directives instead of passing a HTTP header. Could imagine that the DX would be superb when using something like [gql.tada](https://github.com/0no-co/gql.tada).

rfc/graph-feature-flags.md
jensneuse: I just had this very idea.
How about using an alternative way to represent feature flags?
We could send them in the client extension:

```json
{
  "query": "{ products { id name } }",
  "extensions": {
    "featureFlags": [
      "products.experimental"
    ]
  }
}
```

jensneuse: We could also make FF introspect-able, so that the Playground can fetch them to show you FF toggles:

```json
{
  "query": "{ __featureFlags { name description } }",
}
```

This will return all feature flags that are currently enabled in the staging environment.

```json
{
  "data": {
    "__featureFlags": [
      {
        "name": "products.experimental",
        "description": "Enables the experimental feature for the products Subgraph",
      }
    ]
  }
}
```

rfc/graph-feature-flags.md
jensneuse: To make sth like tada work, I think we need to first introspect the API and download the Schema with the correct flags enabled. So we need to send the header or cookie at introspection time, no need for directives at runtime.

akoenig: Yap, you're right. The introspection happens first and the LSP plugin consumes the schema (with the then enabled feature flags). Nothing required during runtime.

rfc/graph-feature-flags.md
```diff
@@ -0,0 +1,307 @@
...
+```bash
+npx wgc subgraph publish products --schema ../demo/subgraphs/products/products.graphql --routing-url http://localhost:4001/graphql --label=team=A --remove-feature-flags
+```
+
+When adding a feature flag to a Subgraph, we create a matrix of possible combinations of feature flags.
+This matrix is used to compute the Supergraph of all combinations of Subgraphs and feature flags.
+By default, we will run composition checks against all possible combinations of Subgraphs and feature flags to ensure that the Supergraph is valid.
+This means that future changes to any Subgraph will run composition checks against the matrix of Subgraphs and feature flags.
+
+It's possible that you want to publish a Subgraph Schema with a feature flag for an experimental feature,
```
ccortezaguilera, line 147: Kind of curious on use cases, why isn't it the default to not run composition checks for publishing feature flags? Seems like if I do want publishing to behave similarly as normal I would want to force the composition checks for feature flags rather than running composition checks for features to be the default.

PR #674, opened 2024-03-28
Feat : Allow to rename client headers on propagate to subgraphs

Rutik7066:
## WIP
I Added a new field called "rename" to the config and added logic to rename the header only if the rename field is not an empty string.

I need your feedback on the solution I have provided.

## TODO

### DIFF
--- a/router/core/header_rule_engine.go
+++ b/router/core/header_rule_engine.go
@@ -2,9 +2,10 @@ package core
 
 import (
 	"fmt"
-	"github.com/wundergraph/cosmo/router/pkg/config"
 	"net/http"
 	"regexp"
+
+	"github.com/wundergraph/cosmo/router/pkg/config"
 
 	nodev1 "github.com/wundergraph/cosmo/router/gen/proto/wg/cosmo/node/v1"
 )
@@ -76,6 +77,17 @@ func (h HeaderRuleEngine) OnOriginRequest(request *http.Request, ctx RequestCont
 	for _, rule := range requestRules {
 		// Forwards the matching client request header to the upstream
 		if rule.Operation == config.HeaderRuleOperationPropagate {
+
+			// Rename the header
+			if rule.Rename != "" {
+				value := ctx.Request().Header.Get(rule.Named)
+				if value != "" {
+					request.Header.Set(rule.Rename, ctx.Request().Header.Get(rule.Named))
+				} else if rule.Default != "" {
+					request.Header.Set(rule.Rename, rule.Default)
+				}
+				continue
+			}
 
 			// Exact match
 			if rule.Named != "" {
--- a/router/pkg/config/config.go
+++ b/router/pkg/config/config.go
@@ -159,6 +159,8 @@ type RequestHeaderRule struct {
 	Matching string `yaml:"matching"`
 	// Named is the exact header name to match
 	Named string `yaml:"named"`
+	// Rename renames the header's key to the provided value
+	Rename string `yaml:"rename,omitempty"`
 	// Default is the default value to set if the header is not present
 	Default string `yaml:"default"`
 }


router/core/header_rule_engine_test.go
```diff
@@ -48,6 +49,43 @@ func TestNamedPropagateHeaderRule(t *testing.T) {
 	assert.Empty(t, updatedClientReq.Header.Get("X-Test-2"))
 }
 
+func TestReNamedPropagateHeaderRule(t *testing.T) {
```
StarpTech, line 52: Please add another test case when `matching` is used instead of `named`.

router/core/header_rule_engine.go
```diff
@@ -77,6 +78,17 @@ func (h HeaderRuleEngine) OnOriginRequest(request *http.Request, ctx RequestCont
 		// Forwards the matching client request header to the upstream
 		if rule.Operation == config.HeaderRuleOperationPropagate {
 
+			// Rename the header
+			if rule.Rename != "" {
+				value := ctx.Request().Header.Get(rule.Named)
+				if value != "" {
+					request.Header.Set(rule.Rename, ctx.Request().Header.Get(rule.Named))
```
StarpTech, line 85: The implementation does not remove the old header.

router/core/header_rule_engine_test.go
```diff
@@ -241,3 +279,74 @@ func TestSubgraphNamedHeaderRule(t *testing.T) {
 	assert.Empty(t, updatedClientReq2.Header.Get("X-Test-Subgraph-1"))
 	assert.Equal(t, "Test-Value-2", updatedClientReq2.Header.Get("X-Test-Subgraph-2"))
 }
+
+func TestSubgraphReNamedHeaderRule(t *testing.T) {
+	ht, err := NewHeaderTransformer(config.HeaderRules{
+		Subgraphs: map[string]config.GlobalHeaderRule{
+			"subgraph-1": {
```
StarpTech, line 286: Please separate the tests in smaller case like

- Rename named header
- Rename matched header
- Rename with default
- ...

StarpTech (changes requested):

router/core/header_rule_engine_test.go
Rutik7066: when we use ```matching```  instead of  ```named``` . we will need to be set ```default```  because it will use ```default``` value. if not set then it will set  empty string.

Rutik7066:
Hi @StarpTech Thank you for the quick response. I resolved and updated the changes.

Rutik7066:
Hi @StarpTech could you please review the changes.

StarpTech:
Hi @Rutik7066, I'll take a look tomorrow. Thank you.

StarpTech:
@Rutik7066 CI is still failing

router/core/header_rule_engine_test.go
```diff
@@ -48,6 +49,87 @@ func TestNamedPropagateHeaderRule(t *testing.T) {
...
+		request:        clientReq,
+		operation:      &operationContext{},
+	})
+
+	assert.Equal(t, "test1", updatedClientReq.Header.Get("X-Test-Renamed"))
+	assert.Empty(t, updatedClientReq.Header.Get("X-Test-1"))
+	assert.Empty(t, updatedClientReq.Header.Get("X-Test-2"))
+}
+
+func TestReNamedMatichingPropagateHeaderRule(t *testing.T) {
```
StarpTech, line 89: Typo

Rutik7066:
Hi @StarpTech now it is working as expected.

StarpTech:
Hi @Rutik7066, we'll take a look soon.

StarpTech:
The following is missing:
- Update the https://github.com/wundergraph/cosmo/blob/main/router/pkg/config/config.schema.json to include the rename option.

According to the issue description an integration test in https://github.com/wundergraph/cosmo/tree/main/router-tests is missing.

router/core/header_rule_engine.go
```diff
@@ -76,6 +77,19 @@ func (h HeaderRuleEngine) OnOriginRequest(request *http.Request, ctx RequestCont
 	for _, rule := range requestRules {
 		// Forwards the matching client request header to the upstream
 		if rule.Operation == config.HeaderRuleOperationPropagate {
+			// Rename the header
+			if rule.Rename != "" {
```
StarpTech, line 81: This means rename would only work with an exact match but we also provide a regex match.

StarpTech (changes requested):

StarpTech:
Please keep in mind that the challenge will expire in 3 weeks after the initial attempt.

router/core/header_rule_engine.go
Rutik7066: done. Added the logic in matching block.

Rutik7066:
Hi @StarpTech I have added the unit test and schema update. Could you please take a look at it.

router-tests/headers_test.go
```diff
@@ -227,3 +228,230 @@ func TestForwardHeaders(t *testing.T) {
...
+			ModifyEngineExecutionConfiguration: func(cfg *config.EngineExecutionConfiguration) {
+				cfg.WebSocketReadTimeout = time.Millisecond * 10
+			},
+			RouterOptions: []core.Option{
+				core.WithHeaderRules(headerRules),
+			},
+		}, func(t *testing.T, xEnv *testenv.Environment) {
+			for _, c := range cases {
+				headerName := c.headerName
+				log.Println(c.testName, c.headerName, c.headerRename)
```
StarpTech, line 292: Please remove the logs from the tests.

router-tests/headers_test.go
```diff
@@ -227,3 +228,230 @@ func TestForwardHeaders(t *testing.T) {
...
+			})
+			require.NoError(t, err)
+			var msg testenv.WebSocketMessage
+			err = conn.ReadJSON(&msg)
+			require.NoError(t, err)
+			require.Equal(t, `{"data":{"headerValue":{"value":"","initialPayload":{"extensions":{"token":"123","upgradeHeaders":{"User-Agent":["Go-http-client/1.1"]}}}}}}`, string(msg.Payload))
+		})
+	})
+
+	t.Run("ws with multiple conns", func(t *testing.T) {
```
StarpTech, line 395: Why do you think multiple connections makes a difference?

router-tests/headers_test.go
```diff
@@ -227,3 +228,230 @@ func TestForwardHeaders(t *testing.T) {
...
+					})
+					log.Println(res.Body)
+					require.Equal(t, `{"data":{"headerValue":"`+headerValue+`"}}`, res.Body)
+
+				})
+			}
+		})
+	})
+
+	t.Run("HTTP with client extension", func(t *testing.T) {
```
StarpTech, line 309: What does it test regarding header renaming? You could query `headerValue` twice.

router-tests/headers_test.go
```diff
@@ -227,3 +228,230 @@ func TestForwardHeaders(t *testing.T) {
...
+						Operation: config.HeaderRuleOperationPropagate,
+						Matching:  "(?i)^bar.*",
+						Rename:    headerRenameInSubgraphRule,
+					},
+				},
+			},
+		},
+	}
+
+	t.Run("HTTP", func(t *testing.T) {
```
StarpTech, line 271: Please also add a test that validates if "default" works with rename. Thank you.

router-tests/headers_test.go
```diff
@@ -227,3 +228,230 @@ func TestForwardHeaders(t *testing.T) {
...
+			RouterOptions: []core.Option{
+				core.WithHeaderRules(headerRules),
+			},
+		}, func(t *testing.T, xEnv *testenv.Environment) {
+			for _, c := range cases {
+				headerName := c.headerName
+				log.Println(c.testName, c.headerName, c.headerRename)
+				log.Println(headerName, headerValue)
+				t.Run(c.testName, func(t *testing.T) {
+					res := xEnv.MakeGraphQLRequestOK(testenv.GraphQLRequest{
```
StarpTech, line 295: This test does not validate that the old header was removed.

router/core/header_rule_engine.go
```diff
@@ -99,7 +113,21 @@ func (h HeaderRuleEngine) OnOriginRequest(request *http.Request, ctx RequestCont
 					// Headers are case-insensitive, but Go canonicalize them
 					// Issue: https://github.com/golang/go/issues/37834
 					if regex.MatchString(name) {
-						request.Header.Set(name, ctx.Request().Header.Get(name))
+						// Rename the header when matiching is provided
```
StarpTech, line 116: typo

StarpTech (changes requested):

router-tests/headers_test.go
Rutik7066: I copied and modified the TestForwardHeaders code.

Rutik7066:
I'm sorry for making a lot of mistakes. now I have fixed all the mistakes. Could you please review it again.

StarpTech:
@Rutik7066 no problem at all as long as you keep up and provide the improvements 👍

Rutik7066:
Is everything perfect?

Rutik7066:
Hi @StarpTech could you please review it again?

StarpTech:
@Rutik7066 asking multiple times will not speed it up. We will take a look as soon as we find the time.

Rutik7066:
Okay, thought your were free.

router/core/header_rule_engine.go
```diff
@@ -99,7 +113,21 @@ func (h HeaderRuleEngine) OnOriginRequest(request *http.Request, ctx RequestCont
...
+							if value != "" {
+								request.Header.Set(rule.Rename, ctx.Request().Header.Get(name))
+								request.Header.Del(name)
+							} else if rule.Default != "" {
+								request.Header.Set(rule.Rename, rule.Default)
+								request.Header.Del(name)
+							}
+
+						} else {
+
```
StarpTech, line 128: Please remove the empty line here.

router/core/header_rule_engine.go
```diff
@@ -99,7 +113,21 @@ func (h HeaderRuleEngine) OnOriginRequest(request *http.Request, ctx RequestCont
...
 					// Issue: https://github.com/golang/go/issues/37834
 					if regex.MatchString(name) {
-						request.Header.Set(name, ctx.Request().Header.Get(name))
+						// Rename the header when matching is provided
+						if rule.Rename != "" && rule.Named == "" {
+							value := ctx.Request().Header.Get(name)
+							if value != "" {
+								request.Header.Set(rule.Rename, ctx.Request().Header.Get(name))
+								request.Header.Del(name)
+							} else if rule.Default != "" {
```
StarpTech, line 122: Good catch! This will fix `default` in the regex case.

StarpTech (changes requested):
Looks good only one small formatting issue.

router/core/header_rule_engine.go
Rutik7066: thanks

StarpTech (approved):
LGTM

PR #690, opened 2024-04-04
chore: add RFC for Subgraph error extensions Schema

jensneuse:


rfc/Errors-Extension-Specification.md
```diff
@@ -0,0 +1,123 @@
...
+npx wgc federated-graph update production --errors-extensions-validation-behavior skip
+```
+
+The default behavior is to "skip" the field and log a warning to the logs.
+
+In addition to the above,
+we also propose to make the extensions Schema visible in Studio so that Subgraph owners can see what fields are allowed in the `extensions` object.
+
+To integrate this solution into the overall architecture and workflow,
+we propose that the Supergraph owner(s) setup a CI/CD pipeline that publishes the Errors Extension Schema from a git repository to the Supergraph.
```
ccortezaguilera, line 107: I like this a lot. If we can define the schema, I can image wanting the source of truth in code and thus colocated with the supergraph repo.

rfc/Errors-Extension-Specification.md
```diff
@@ -0,0 +1,123 @@
...
+We thought about allowing Subgraph owners to define the schema for the `extensions` field in the Subgraph itself,
+which would give them more flexibility and autonomy.
+However, this would lead to a more fragmented ecosystem of error codes and allowed fields in the `extensions` object,
+so that codes are not standardized across different Subgraphs and therefore less useful for the client.
+
+We also believe that it helps organizations to have a centralized place to define the schema for the `extensions` field,
+encouraging collaboration and standardization across different teams and Subgraphs.
+
+Another benefit of this solution is that clients can use the JSON Schema to generate client-side code to handle the errors more gracefully,
+e.g. by generating TypeScript types for the error object.
```
ccortezaguilera, line 123: plus one to being able to create error types.

rfc/Errors-Extension-Specification.md
```diff
@@ -0,0 +1,123 @@
...
+  "additionalItems": false
+}
+```
+
+The `additionalItems` field is set to `false` to ensure that only the fields defined in the schema are allowed.
+
+In addition to just defining the schema,
+we also propose to add an additional option to define the behavior when the schema validation fails.
+
+When the JSON Schema validation fails for an `extensions` field,
```
ccortezaguilera, line 85: when is this schema validation happening?

rfc/Errors-Extension-Specification.md
```diff
@@ -0,0 +1,123 @@
...
+The behavior can be defined in the Supergraph using the following command:
+
+```bash
+npx wgc federated-graph create production -r http://router.example.com/graphql --label-matcher team=A department=backend --errors-extensions-schema ./errors-extensions-schema.json --errors-extensions-validation-behavior skip
+```
+
+Example of updating the behavior for an existing Supergraph:
+
+```bash
+npx wgc federated-graph update production --errors-extensions-validation-behavior skip
```
ccortezaguilera, line 98:
```suggestion
npx wgc federated-graph update production --errors-extensions-schema ./errors-extensions-schema.json --errors-extensions-validation-behavior skip
```

rfc/Errors-Extension-Specification.md
jensneuse: At runtime when an error is rendered by the Router

rfc/Errors-Extension-Specification.md
```diff
@@ -0,0 +1,123 @@
...
+we also propose to add an additional option to define the behavior when the schema validation fails.
+
+When the JSON Schema validation fails for an `extensions` field,
+the Engine can either "skip" this field in the response object or "fail" the error rendering process and return a generic error message to the client.
+In addition, the Engine can log the error to the logs for debugging purposes and notify the Subgraph owner that their Subgraph is not returning errors in the correct format.
+
+The behavior can be defined in the Supergraph using the following command:
+
+```bash
+npx wgc federated-graph create production -r http://router.example.com/graphql --label-matcher team=A department=backend --errors-extensions-schema ./errors-extensions-schema.json --errors-extensions-validation-behavior skip
```
StarpTech, line 92: Wouldn't it be more natural if the errors were pushed from the subgraphs? In that way, we could avoid coordination across teams. At composition time, we could merge the schemas. It also allows us to show what errors can be reported by which subgraph. Makes it easy to create connections with metrics.

StarpTech (changes requested):

rfc/Errors-Extension-Specification.md
clayne11: I'm a bit worried about how we enforce the correct emission of errors from the subgraphs. Ideally at compile time we want to ensure that subgraphs are emitting errors in the correct format.

I know this is a language and subgraph server specific issue though, so maybe the solution to this can't necessarily live here.

rfc/Errors-Extension-Specification.md
```diff
@@ -0,0 +1,123 @@
...
+Example of updating the schema for an existing Supergraph:
+
+```bash
+npx wgc federated-graph update production --errors-extensions-schema ./errors-extensions-schema.json
+```
+
+The schema file should look like this:
+
+```json
+{
```
frekw, line 61: How would conflicting definitions between subgraphs be handled in this case? I.e different values for `required`?

rfc/Errors-Extension-Specification.md
SvenW: Isn't the idea to get one error schema for all errors? Sure, subgraphs could probably send in their respective codes and update that list on a per subgraph basis.

rfc/Errors-Extension-Specification.md
frekw: Nevermind I misread `subgraph` for `subgraph` on L52.

rfc/Errors-Extension-Specification.md
frekw: I think merging schemas would be quite hard, for example the `required` field doesn't have any obvious semantics where it can be easily merged in a way that works for all subgraphs. 🤔

rfc/Errors-Extension-Specification.md
```diff
@@ -0,0 +1,123 @@
...
+Example of updating the behavior for an existing Supergraph:
+
+```bash
+npx wgc federated-graph update production --errors-extensions-validation-behavior skip
+```
+
+The default behavior is to "skip" the field and log a warning to the logs.
+
+In addition to the above,
+we also propose to make the extensions Schema visible in Studio so that Subgraph owners can see what fields are allowed in the `extensions` object.
```
frekw, line 104: Love this!

rfc/Errors-Extension-Specification.md
jensneuse: Here's how I think this would be possible.
The JSON Schema is stored in the Schema Registry.
At build time, the Subgraph downloads the Schema using a `wgc` command.
The Subgraph codebase can then generate types/models from the Subgraph Error Extensions Schema.
A mismatch will result in a compiler error in most frameworks, given that the Subgraph uses the errors.
This would still require the developers to opt-in into the tooling, but if they do, we can catch errors at build time.
WDYT?

rfc/Errors-Extension-Specification.md
jensneuse: I think the Schema needs to live on the Supergraph.
It's the only way to enforce coordination between teams and ensure that there's consensus on how errors should look like.

PR #695, opened 2024-04-04
chore: update gqlgen to v0.17.45 for demo

clayne11:
## Motivation and Context

I'm going to put up a follow up PR that adds integration tests for the @requires Federation directive. Proper support for this was only added in v0.17.45 of gqlgen.

## Changes
Ran the following commands:

- cd demo && go get -d github.com/99designs/gqlgen@latest && go mod tidy.
- cd pkg/subgraphs/employees && go generate
- cd ../family && go generate
- cd ../hobbies && go generate
- cd ../products && go generate
- cd ../test1 && go generate
- cd ../availability && go generate
- cd ../mood && go generate
- cd ../countries && go generate
- cd ../employeeupdated && go generate

PR #704, opened 2024-04-07
test: introduce tests for @requires

clayne11:
## Motivation and Context

The Cosmo router [ostensibly supports the `@requires` directive](https://cosmo-docs-source.wundergraph.com/federation/federation-compatibility-matrix), however it doesn't seem to work properly. Added tests to router-tests to show the expected behavior.

These tests currently fail since the router logic seems to be incorrect. This is both true for @requires directives with flat fields as well as ones with nested selection sets, although each of these lead to different error patterns. 

Depends on #695 to separate out dependency updates from logic changes.


Aenimus:
Hi @clayne11 

Would you mind rebasing? If things are still failing, we'll investigate.

Thank you for the contribution!

clayne11:
Rebased — tests are still failing.

clayne11:
This is not stale, please don't close it.

Aenimus:
@clayne11 Hi Curtis,

Apologies for my lack of updates. We know what the issue is, and we're working on some fixes. 

We're still interested in the PR and don't plan to close this.

Thanks for your patience!

PR #709, opened 2024-04-08
fix: let everybody create orgs

JivusAyrus:
## TODO


studio/src/pages/create.tsx
```diff
@@ -297,10 +296,7 @@ const OrganizationForm = () => {
           className="ml-auto"
           isLoading={isPending}
           type="submit"
-          disabled={
-            !form.formState.isValid ||
-            !user?.currentOrganization.roles.includes("admin")
-          }
+          disabled={!form.formState.isValid}
```
StarpTech, line 299: Do we enforce max organizations?

StarpTech (changes requested):

studio/src/pages/create.tsx
JivusAyrus: No, as every new org is a paid org

StarpTech (approved):
LGTM

PR #710, opened 2024-04-08
fix: inform users if there is nothing new to publish

JivusAyrus:
## TODO

image

### DIFF
--- a/cli/src/commands/subgraph/commands/publish.ts
+++ b/cli/src/commands/subgraph/commands/publish.ts
@@ -95,7 +95,7 @@ export default (opts: BaseCommandOptions) => {
 
     switch (resp.response?.code) {
       case EnumStatusCode.OK: {
-        spinner.succeed('Subgraph published successfully.');
+        spinner.succeed(resp.response?.details || 'Subgraph published successfully.');
 
         break;
       }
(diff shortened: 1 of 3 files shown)


cli/src/commands/subgraph/commands/publish.ts
```diff
@@ -95,7 +95,7 @@ export default (opts: BaseCommandOptions) => {
 
     switch (resp.response?.code) {
       case EnumStatusCode.OK: {
-        spinner.succeed('Subgraph published successfully.');
+        spinner.succeed(resp.response?.details || 'Subgraph published successfully.');
```
StarpTech, line 98: The purpose of details was to communicate error details. We should not use it to communicate a state. Just add another property.

StarpTech (changes requested):

cli/src/commands/subgraph/commands/publish.ts
JivusAyrus: Done

cli/src/commands/subgraph/commands/publish.ts
```diff
@@ -95,7 +95,9 @@ export default (opts: BaseCommandOptions) => {
 
     switch (resp.response?.code) {
       case EnumStatusCode.OK: {
-        spinner.succeed('Subgraph published successfully.');
+        spinner.succeed(
+          resp?.hasChanged === false ? 'Detected no new changes to publish.' : 'Subgraph published successfully.',
```
StarpTech, line 99: "Skip publish, no changes to publish"

JivusAyrus: Done, confirmed it doesnt cause any compositions

StarpTech (approved):
LGTM

PR #713, opened 2024-04-09
fix: authenticate organization id before updating webhooks

StarpTech:
## TODO


controlplane/test/webhooks.test.ts
```diff
@@ -0,0 +1,273 @@
...
+            },
+          },
+        },
+      ],
+    });
+
+    await server.close();
+  });
+
+  test('Should be possible to subscribe for a federated graph that dont belong to the same organization', async (testContext) => {
```
JivusAyrus, line 61: Should it be "Should be possible to subscribe for a federated graph that belong to the same organization"

StarpTech: Indeed, Good catch!

JivusAyrus (approved):

PR #735, opened 2024-04-15
Feature/edfs subject validation

jaysneg:
## Motivation and Context

EDFS Subject validation

1 Template {{ args.id }} validations
2 Static subject separate.with.dots.lowercase validations
3 Nats [wildcard](https://docs.nats.io/reference/faq#does-nats-support-subject-wildcards) entities.* and entities.> validations

## TODO

### DIFF
--- a/composition/src/normalization/normalization-factory.ts
+++ b/composition/src/normalization/normalization-factory.ts
@@ -269,6 +269,49 @@ export function normalizeSubgraph(
   return normalizationFactory.normalize(document);
 }
 
+export function validateEventSubscribetionSubject(value: ConstValueNode, errorMessages: string[]): boolean {
+  const val = value.value;
+
+  if (value.kind !== Kind.STRING || val.length < 1) {
+    errorMessages.push(invalidEventSubjectsItemErrorMessage);
+
+    return false;
+  }
+
+  const regex = /^[a-zA-Z0-9.*>\{\}\s]+$/;
+  if (!regex.test(value.value)) {
+      errorMessages.push(invalidEventSubjectCharactersErrorMessage);
+     
+      return false;
+  }
+
+  if (val.includes('{{')) {
+    return validateSubjectTemplate(val, errorMessages)
+  }
+
+  return validateSubjectStatic(val, errorMessages);
+}
+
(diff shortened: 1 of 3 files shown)


composition/src/normalization/normalization-factory.ts
```diff
@@ -265,6 +274,112 @@ export function normalizeSubgraph(
...
+
+    const val = value.value;
+    if (val.includes('{{')) {
+      if (!validateSubjectTemplate(fieldDefinition, val, argsNames, errorMessages)) {
+        return false;
+      }
+    }
+  }
+
+  if (fieldDefinition.arguments?.length !== undefined && argsNames.length > 0) {
```
Aenimus, line 301: What happens if length is 0? Is the logic sound?

composition/src/normalization/normalization-factory.ts
```diff
@@ -265,6 +274,112 @@ export function normalizeSubgraph(
...
+    for (const argument of fieldDefinition.arguments) {
+      if (argsNames.indexOf(argument.name.value) === -1) {
+        errorMessages.push(invalidFieldDefinitionNoArgumentTemplateErrorMessage);
+
+        return false;
+      };
+    }
+  }
+
+  if (argsNames.length === 0 && fieldDefinition.arguments?.length !== undefined) {
```
Aenimus, line 311: Same here regarding length of arguments

composition/src/normalization/normalization-factory.ts
```diff
@@ -265,6 +274,112 @@ export function normalizeSubgraph(
...
+    errorMessages.push(invalidFieldDefinitionNoArgumentTemplateErrorMessage);
+
+    return false;
+  }
+
+  return true
+}
+
+function validateSubjectTemplate(fieldDefinition: FieldDefinitionNode, value: string, argsSet: Array<string>, errorMessages: string[]): boolean {
```
Aenimus, line 320: In the natsserver package in go, there is a method to validate a nats subject. It's likely that the nats js client has some thing similar. But if not, you could look what the go client does, and lift that.

I think the best way to validate is to match each argument template and validate it against the arguments.
After that is all clear, change each argument template to a dummy value and check whether it's a valid nats subject.

PR #737, opened 2024-04-17
chore: add entity caching RFC draft

jensneuse:
This RFC introduces entity caching

### DIFF
--- a/rfc/entity-caching-v1.md
+++ b/rfc/entity-caching-v1.md
@@ -140,7 +140,32 @@ In this example, the `updateEntity` mutation will invalidate the following cache
 
 ## Cache Invalidation through EDFS Events
 
+## Finding Caching Opportunities
 
+One concern regarding caching is that you might believe that there's no caching opportunity,
+the data might not be a good fit for caching,
+or you might think that your API is so dynamic that caching wouldn't improve anything.
+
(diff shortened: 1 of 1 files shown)


rfc/entity-caching-v1.md
```diff
@@ -0,0 +1,172 @@
+# RFC Entity Caching
+
+This document describes the design of the entity caching system for Cosmo Router.
+The goal is to provide a high-performance, scalable, and reliable caching system for entities.
+The Router connects to a Cache Service like Redis to store and retrieve entities instead of fetching them from Subgraphs.
+
+This document describes how caching can be implemented by the user in terms of TTL, cache eviction, and cache invalidation.
```
jfroundjian, line 7: I don't think this is called out anywhere in the doc but I'm assuming that the Router will fetch the entire entity from the Subgraph even if the Request is only asking for a few fields ?

jfroundjian (commented):
Thanks for sharing such a well thought out RFC @jensneuse

I'm not sure if you've already come across how Apollo has implemented this feature but at the risk of being redundant I'll share their public documentation on it: https://www.apollographql.com/docs/router/configuration/entity-caching

What I particularly like about this RFC is that the Router relies entirely on the Subgraph Schema to determine caching rules whereas the Apollo Router relies on the Cache control Http response header (probably some different tradeoffs).

PR #743, opened 2024-04-19
chore: add error propagation mode

jensneuse:


StarpTech:
Don't forget to update the cosmo docs.

PR #752, opened 2024-04-24
feat: improve Subgraph error propagation

jensneuse:


router/pkg/config/config.schema.json
```diff
@@ -1135,15 +1135,30 @@
...
+          "description": "Rewrite the paths of the Subgraph errors. If the value is true (default), the paths of the Subgraph errors will be rewritten to match the Schema of the Federated Graph."
+        },
+        "omit_locations": {
+          "type": "boolean",
+          "default": true,
+          "description": "Omit the location field of Subgraph errors. If the value is true, the location field of Subgraph errors will be omitted. This is useful because the locations of a Subgraph error is internal to the Subgraph and not relevant to the client."
+        },
+        "omit_extensions": {
+          "type": "boolean",
+          "default": false,
```
StarpTech, line 1158: Correct, that this is on by default?

router/pkg/config/config.schema.json
```diff
@@ -1135,15 +1135,30 @@
...
+          "description": "Omit the extensions field of Subgraph errors. If the value is true, the extensions field of Subgraph errors will be omitted. This is useful because the extensions of a Subgraph error is internal to the Subgraph and not relevant to the client."
```
StarpTech, line 1159: Could you also mention what is the default in the desc like in `rewrite_paths`? I like it.

StarpTech (commented):
Looks great, only a few minor things.

router/pkg/config/config.schema.json
jensneuse: yes

StarpTech (approved):
LGTM

PR #763, opened 2024-05-02
fix: validate graphql operation in playground

thisisnithin:
## Motivation and Context

Validates operation before making the request.

## TODO


StarpTech (approved):
LGTM

playground/src/App.tsx
```diff
@@ -1,15 +1,51 @@
...
-const graphiQLFetch = async (onFetch: any, ...args: any) => {
+const graphiQLFetch = async (
+  schema: GraphQLSchema | null,
+  clientValidationEnabled: boolean,
+  onFetch: any,
+  ...args: any
+) => {
   try {
+    if (schema && clientValidationEnabled) {
+      const query = JSON.parse(args[1].body as string)?.query as string;
```
jensneuse, line 24: can accessing args[1] throw or is it 100% safe?

playground/src/App.tsx
```diff
@@ -1,15 +1,51 @@
...
+
+      const errors = validate(schema, parse(query));
+
+      if (errors.length > 0) {
+        const responseData = {
+          message: 'Client-side validation failed for operation. Request not made to the router.',
```
jensneuse, line 30: Rephrase:
```
Client-side validation failed. The request was not sent to the Router.
```

playground/src/App.tsx
```diff
@@ -142,46 +213,75 @@ export default function App() {
...
+      const toggleClientValidation = document.createElement('div');
+      toggleClientValidation.id = 'toggle-client-validation';
+      toolbar.append(toggleClientValidation);
+    }
+
     setIsMounted(true);
   }, [isMounted]);
 
+  useEffect(() => {
+    const getSchema = async () => {
```
jensneuse, line 228: Are we not already loading the Schema from the backend using an API?

jensneuse (changes requested):

playground/src/App.tsx
thisisnithin: This is in router playground. There is a race when accessing the one that graphiql provides in the callback. So doing it separately.

playground/src/App.tsx
thisisnithin: It will exist. Refactored a bit.

playground/src/App.tsx
thisisnithin: Done

jensneuse (approved):

PR #764, opened 2024-05-02
feat: support inaccessible and add foundation for contracts

Aenimus:
## TODO


StarpTech (changes requested):
Good work! Like the way how you write everything with tests and clear expectations! I recommend to merge this PR into nithin's PR so we can first test it end to end before releasing it.

proto/wg/cosmo/node/v1/node.proto
```diff
@@ -78,6 +78,7 @@ message EngineConfiguration {
   string graphqlSchema = 4;
   repeated TypeConfiguration type_configurations = 5;
   map<string, string> string_storage = 6;
+  optional string graphql_client_schema = 7;
```
StarpTech, line 81: Out of curiosity. If you remove `optional` what will be the type? It will be still optional correct?

devsergiy: without optional
```go
type Some struct {
  ClientSchema string
}
```

with optional
```go
type Some struct {
  ClientSchema *string
}
```

jensneuse (approved):

proto/wg/cosmo/node/v1/node.proto
jensneuse: I personally like optional better because we can then assume that a non nil string will contain a valid schema.

StarpTech: Thanks both.

StarpTech (approved):
LGTM

shared/test/__snapshots__/router.config.test.ts.snap
```diff
@@ -262,7 +262,8 @@ exports[`Router Config Builder > Build Subgraph schema > router.config.json 1`]
...
       "64a29f293bb6d35526ce51e1715b8032cdfb31bf": "directive @extends on INTERFACE | OBJECT\\n\\ndirective @external on FIELD_DEFINITION | OBJECT\\n\\ndirective @key(fields: openfed__FieldSet!, resolvable: Boolean = true) repeatable on INTERFACE | OBJECT\\n\\ndirective @provides(fields: openfed__FieldSet!) on FIELD_DEFINITION\\n\\ndirective @requires(fields: openfed__FieldSet!) on FIELD_DEFINITION\\n\\ndirective @tag(name: String!) repeatable on ARGUMENT_DEFINITION | ENUM | ENUM_VALUE | FIELD_DEFINITION | INPUT_FIELD_DEFINITION | INPUT_OBJECT | INTERFACE | OBJECT | SCALAR | UNION\\n\\ntype Product @key(fields: \\"upc\\") {\\n  inStock: Boolean\\n  price: Int @external\\n  shippingEstimate: Int @requires(fields: \\"price weight\\")\\n  upc: String! @external\\n  weight: Int @external\\n}\\n\\nscalar openfed__FieldSet"
-    }
+    },
+    "graphqlClientSchema": "type Query {}"
```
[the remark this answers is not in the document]
Aenimus: Yes. It's the same a few properties up for graphQLSchema. We're just propagating the dummy values.

PR #770, opened 2024-05-07
feat: refactor edfs and add kafka support

StarpTech:
## Motivation and Context

Can be tested by

1. `make create-demo`
2. `make edfs-demo-up`
3. `make dc-subgraphs-demo`

Run

```gql
subscription {
  employeeUpdatedMyKafka(employeeID:2) {
    id
    tag
    details {
      surname
    }
  }
}
```

New Router config

```yaml
events:
  providers:
    nats:
      - id: default
        url: "nats://localhost:4222"
      - id: my-nats
        url: "nats://localhost:4222"
        authentication:
          # token: "token"
          user_info:
            username: "admin"
            password: "admin"
    kafka:
      - id: my-kafka
        brokers:
          - "localhost:9092"
        tls:
          enabled: true
        authentication:
          sasl_plain:
            username: "admin"
            password: "admin"
```

4. Create topics and publish an event via tests https://github.com/wundergraph/cosmo/pull/770/files#diff-96865f74f62689c7dc186e47fec31633a8c5cca2eb3cf6218055e850415ff35eR21-R28

## TODO


router/config.yaml
```diff
@@ -3,4 +3,18 @@
 # See pkg/config/config.go for the full list of configuration options.
 # This file is used to debugging purposes only.
 
-version: "1"
\ No newline at end of file
+version: "1"
+
+instance_id: test
```
StarpTech, line 8: Needs to be removed

router/pkg/pubsub/kafka/kafka.go
```diff
@@ -0,0 +1,380 @@
...
+					var kErr *kerr.Error
+					if errors.As(fetchError.Err, &kErr) {
+						if !kErr.Retriable {
+							p.logger.Error("unrecoverable fetch error",
+								zap.Error(fetchError.Err),
+								zap.String("topic", fetchError.Topic),
+							)
+
+							// If the error is not recoverable, return it and abort the poller
+							return fetchError.Err
```
jensneuse, line 143: Are you sure that we want to kill all Subscriptions that depend on this poller when such an error occurs? What errors are possible? Should we handle this more gracefully?

router/pkg/pubsub/kafka/kafka.go
```diff
@@ -0,0 +1,380 @@
...
+// removeSubscription removes the subscription from the topics and closes the client if there are no more subscriptions for the topic.
+func (p *kafkaPubSub) removeSubscription(sub *subscription, event *pubsub_datasource.KafkaSubscriptionEventConfiguration) {
+
+	// Remove the subscription from the topics
+	for _, topic := range event.Topics {
+
+		p.mu.RLock()
+		// Zero alloc due to slice reuse
+		tmp := p.subscriptions[topic][:0]
+		for _, s := range p.subscriptions[topic] {
```
jensneuse, line 251: Have you considered using a map with sub as the key?

router/pkg/pubsub/kafka/kafka.go
```diff
@@ -0,0 +1,380 @@
...
+// Subscribe subscribes to the given topics and updates the subscription updater.
+// The engine already deduplicates subscriptions with the same topics, stream configuration, extensions, headers, etc.
+func (p *kafkaPubSub) Subscribe(ctx context.Context, event pubsub_datasource.KafkaSubscriptionEventConfiguration, updater resolve.SubscriptionUpdater) error {
+
+	p.logger.Debug("subscribe",
+		zap.Strings("topics", event.Topics),
+		zap.String("providerID", event.ProviderID),
+	)
+
+	sub := &subscription{
```
jensneuse, line 193: I think the concept of subscription here makes the whole thing more complex than it needs to be.

As an alternative, you could create a client in Subscribe, then start the poller, and update through the updater until the ctx is done.

You're optimizing for the case that one Trigger depends on multiple topics, which I think is very rare.

router/pkg/pubsub/kafka/kafka.go
```diff
@@ -0,0 +1,380 @@
...
+// It uses a single write client to produce messages and a client per topic to consume messages.
+// Each client polls the Kafka topic for new records and updates the subscriptions with the new data.
+type kafkaPubSub struct {
+	ctx           context.Context
+	opts          []kgo.Opt
+	logger        *zap.Logger
+	clients       map[string]*kgo.Client
+	subscriptions map[string][]*subscription
+	writeClient   *kgo.Client
+	mu            sync.RWMutex
```
jensneuse, line 86: If we would remove the concept of subscription here and ignore possible duplication, we could get rid of the mutex and save ourselves from potential bugs.

router/pkg/pubsub/kafka/kafka.go
```diff
@@ -0,0 +1,380 @@
...
+	fErr := p.flush(ctx)
+	if fErr != nil {
+		p.logger.Error("error flushing clients", zap.Error(err))
+		err = errors.Join(err, fErr)
+	}
+
+	p.writeClient.Close()
+
+	for _, client := range p.clients {
+		client.Close()
```
jensneuse, line 373: You already close the clients when p.ctx.Done() is called in the poller. What happens when client.Close gets called twice?
If you'd delete the client from clients after close, with the mutex, would this prevent the issue?

router/pkg/pubsub/kafka/kafka.go
jensneuse, line 86: It seems like you're using the mutext for both subscriptions and clients.
I guess at least for clients we should have it?
In general, I'd always use one mutex per object (map, slice) that's being protected by it.

router/pkg/pubsub/kafka/kafka.go
```diff
@@ -0,0 +1,380 @@
...
+
+type connector struct {
+	writeClient *kgo.Client
+	opts        []kgo.Opt
+	logger      *zap.Logger
+}
+
+func NewConnector(logger *zap.Logger, opts []kgo.Opt) (pubsub_datasource.KafkaConnector, error) {
+
+	writeClient, err := kgo.NewClient(append(opts,
```
jensneuse, line 44: Should we always create a write client, even though we're not publishing anything?
As an alternative, you could use `sync.Once` to lazily init it once necessary.

jensneuse (changes requested):

router/pkg/pubsub/kafka/kafka.go
StarpTech: Yes, the library retries in case of retriable errors. If an unretriable errors occurs e.g. misconfigurations or misbehavior of the cluster side can't be replied. If we would swallow it, it would produce a lot logs.

router/pkg/pubsub/kafka/kafka.go
StarpTech: How would it help me to manage subscriptions to topics?

router/pkg/pubsub/kafka/kafka.go
StarpTech: Nothing will happen. It can be called multiple times.

router/pkg/pubsub/kafka/kafka.go
StarpTech: Absolutely, the client should be ready as soon as possible. Instantiating it on demand could result in latency and I'd need to protect the creation through a mutex again.

shared/src/router-config/graphql-configuration.ts
```diff
@@ -99,25 +101,54 @@ export function configurationDataMapToDataSourceConfiguration(
...
+                    }),
+                  }
+                : {}),
+            }),
+          );
+          break;
+        }
+        default: {
+          // TODO propagate this properly
+          throw new Error(`Unknown event provider.`);
```
Aenimus, line 146: How do we want to handle this? Theoretically, it should never happen.

router/pkg/pubsub/kafka/kafka.go
jensneuse: sync.Once handles all of this, but the additional latency might be an argument for the pre-creation
One thing we could consider is to not create a writing client if we don't have any publish operations configured.

router-tests/testenv/testenv.go
```diff
@@ -53,14 +57,16 @@ import (
 )
 
 const (
-	defaultSourceName = "default"
+	defaultSourceName = "nats"
 	myNatsSourceName  = "my-nats"
+	myKafkaSourceName = "my-kafka"
```
Aenimus, line 62: Should probably be changed to ProviderId for clarity

router/pkg/config/testdata/config_full.json
```diff
@@ -217,17 +217,42 @@
...
-      },
-      "default": {
-        "Provider": "NATS",
-        "URL": "nats://localhost:4222",
-        "Authentication": null
-      }
+    "Providers": {
+      "Nats": [
+        {
+          "ID": "default",
```
Aenimus, line 223: The default nats provider id is now "nats" rather than "default".

Just making sure that "default" here is intended.

router/pkg/pubsub/kafka/kafka.go
StarpTech: Honestly, this is negligible. A client does not have much overhead and we only have a single write client.

shared/src/router-config/graphql-configuration.ts
StarpTech: This is a question for you. Ideally, we can exhaust the switch by using types. I'm fine with both implementation. An error would highlight that something significantly is broken.

router-tests/testenv/testenv.go
StarpTech: Done

router/pkg/config/fixtures/full.yaml
```diff
@@ -182,13 +182,27 @@ cdn:
...
-  sources:
-    default:
-      provider: NATS
-      url: "nats://localhost:4222"
-    another-nats:
-      provider: NATS
-      url: "nats://localhost:4223"
+  providers:
+    nats:
+      - id: default
```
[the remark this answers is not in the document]
StarpTech: We go with id as discussed later. In my last commit I changed the default provider ID to "default" again.

router/pkg/config/testdata/config_full.json
StarpTech: Done

router/pkg/pubsub/kafka/kafka.go
```diff
@@ -0,0 +1,257 @@
...
+						p.logger.Error("fetch error", zap.Error(fetchError.Err), zap.String("topic", fetchError.Topic))
+					}
+				}
+			}
+
+			iter := fetches.RecordIter()
+			for !iter.Done() {
+				r := iter.Next()
+
+				rec := &record{
```
jensneuse, line 133: Is this necessary?

router/pkg/pubsub/kafka/kafka.go
```diff
@@ -0,0 +1,257 @@
...
+				p.logger.Debug("poller canceled", zap.Error(err))
+			} else {
+				p.logger.Error("poller error", zap.Error(err))
+
+			}
+			return
+		}
+	}()
+
+	go func() {
```
jensneuse, line 187: This seems unnecessary?

jensneuse (changes requested):

jensneuse (approved):

Issue #784, opened 2024-05-13
Router UpdateServer function doesn't work as advertised


wainglaister:
### Component(s)

router

### Component version

router@0.85.2

### wgc version

0.53.2

### controlplane version

N/A

### router version

0.85.2

## Description
Core router.go provides an exported UpdateServer function whose description says it creates **and starts**  a new server and replaces the old one. This is inaccurate, the new server is not started.
There is a non-exported function (updateServerAndStart) which uses the exported function and does start the new server.

## Steps to Reproduce
(by some means) call `err := router.UpdateServer(ctx, newRouterConfig)`

## Expected Result
Calling the exported function will replace the existing server with a newly configured and started server.

## Actual Result
Calling the exported function results in an inoperable system who's server exists but is not listening for connections.

## Environment
Problem is unrelated to environment
Source is this PR: https://github.com/wundergraph/cosmo/pull/446/files#diff-aa58824b56f499a6a7d39dd7d905130a1aed0683be3aa6826a63dfd5d5047edcL373

### Log output

```shell
15:59:38 PM WARN Advanced Request Tracing (ART) is enabled in development mode but requires a graph token to work in production. For more information see https://cosmo-docs.wundergraph.com/router/advanced-request-tracing-art {"component": "@wundergraph/router", "service_version": "dev"}
15:59:38 PM INFO Serving GraphQL playground {"component": "@wundergraph/router", "service_version": "dev", "url": "http://0.0.0.0:3002/"}
15:59:38 PM INFO Rate limiting disabled {"component": "@wundergraph/router", "service_version": "dev"}
15:59:38 PM INFO GraphQL endpoint {"component": "@wundergraph/router", "service_version": "dev", "method": "POST", "url": "http://0.0.0.0:3002/graphql"}
15:59:38 PM INFO Gracefully shutting down the router ... {"component": "@wundergraph/router", "service_version": "dev", "config_version": "", "grace_period": "20s"}
15:59:38 PM INFO Server stopped {"component": "@wundergraph/router", "service_version": "dev", "config_version": ""}
15:59:42 PM INFO WebSocket Stats {"component": "@wundergraph/router", "service_version": "dev", "open_connections": 0, "active_subscriptions": 0}
15:59:42 PM INFO Memory usage {"component": "@wundergraph/router", "service_version": "dev", "alloc_mb": "6.5 MB", "total_alloc_mb": "20 MB", "sys_mb": "20 MB", "num_gc": 9}
```

### Additional context

[...]

PR #787, opened 2024-05-14
fix: issue 784 router update start new server listening

wainglaister:
Fix issue where router update creates a new server and replaces the old server but doesn't start the new server listening
Add config hot-reload to router when using a local config file (using the router update fix)

## Motivation and Context

Fix corrects a mismatch between the documentation and implementation for router updating

Addition for hot-reload improves developer experience when running with a local config file

## TODO


wainglaister:
https://github.com/wundergraph/cosmo/issues/784

StarpTech:
Hi @wainglaister,

Thanks so much for your contribution! I've updated the comments for the methods. Unfortunately, we can't merge the PR just yet. It needs tests, and the watcher functionality should be properly abstracted into a package. Given this, the priority of this feature is quite low at the moment, and we don't have the resources to guide you through these changes or to ensure that this feature is working as expected.

To prevent such situations in the future, we recommend creating an issue first. Thank you!

StarpTech:
As described earlier in comment. No action is left here.

PR #792, opened 2024-05-15
chore: add graph feature flag adr

jensneuse:


adr/graph-feature-flags.md
```diff
@@ -0,0 +1,212 @@
...
+### Feature Flags always replace the subgraph
+
+Feature Flags are always replacing the subgraph they are associated with.
+This means that a feature flag can only be associated with one subgraph at a time.
+If a feature flag is associated with a subgraph, it will replace the subgraph in the federated graph.
+
+### Conflicting Feature Flags that replace the same subgraph are allowed
+
+It's possible to have multiple feature flags that replace the same subgraph.
+This is useful and allowed for testing different variants of the same subgraph.
```
ccortezaguilera, line 110: Does this mean then we can reference multiple flags for the same subgraph in the same request?

adr/graph-feature-flags.md
```diff
@@ -0,0 +1,212 @@
...
+With 20 active feature flags and 5 groups, we would have 20+5=25 possible combinations.
+If all 20 feature flags were combined, we would have 20*20=400 possible combinations.
+
+Each time a subgraph gets published, the Control Plane is required to check if all feature flags are still valid,
+which means that the number of possible combinations defines the composition check runs required.
+
+### Composition Checks need to be Feature Flag aware
+
+When a subgraph gets published, the Control Plane needs to check if the subgraph is compatible with all active feature flags.
+If errors occur, they need to be reported in such a way that the user can easily identify which feature flag caused the error.
```
ccortezaguilera, line 144: Will this be reported in the CLI and studio?

adr/graph-feature-flags.md
```diff
@@ -0,0 +1,212 @@
...
+Within the same namespace, feature flags and group feature flags must have unique names.
+It's not allowed to have a group feature flag with the same name as a feature flag and vice versa.
+It's not allowed to have a group feature flag with the same name as another group feature flag.
+It's not allowed to have a feature flag with the same name as another feature flag.
+
+### Feature Flags are namespaced
+
+Feature Flags are namespaced.
+
+### The Router Engine Config is aware of all Feature Flags
```
ccortezaguilera, line 162: for users to get all the feature flags would it be
```
wgc feature-flags list
``` ?

File: adr/graph-feature-flags.md [hunk 4]
```diff
@@ -0,0 +1,212 @@
...
+
+- Subscriptions over WebSocket (header, cookie, extensions)
+- Queries and Mutations over HTTP (header, cookie, extensions)
+- Subscriptions over SSE (header, cookie, extensions)
+
+### Introspection needs to be aware of Feature Flags
+
+Depending on the selected feature flag, the introspection query should return the schema that is associated with the feature flag.
+
+### No security through obscurity
```
ccortezaguilera, line 202: will there be a setting to disable feature flag functionality i.e. only enable it on the dedicated federated graph and not on others? Meaning if enabled the flag is respected on the request but if disabled it returns a 400?

adr/graph-feature-flags.md
```diff
@@ -0,0 +1,212 @@
...
+```shell
+wgc federated-graph create production --label-matcher team=A,team=B --routing-url https://graph.domain.com/graphql
+```
+
+4. Create a feature flag for the users subgraph
+
+```shell
+wgc feature-flag create users-v2 --label team=B --routing-url https://users-v2.domain.com --subgraph users
+# alias: wgc ff create
+wgc feature-flag publish users-v2 --schema ./subgraph-users-v2/schema.graphql
```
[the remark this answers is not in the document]
jensneuse: yes

adr/graph-feature-flags.md
```diff
@@ -0,0 +1,212 @@
...
+wgc ff delete users-v2
+wgc ff delete posts-v2
+```
+
+## Context
+
+### Feature Flags always replace the subgraph
+
+Feature Flags are always replacing the subgraph they are associated with.
+This means that a feature flag can only be associated with one subgraph at a time.
```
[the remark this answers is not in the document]
jensneuse: You could only have one FF per subgraph at a time. So it can be either posts (original), posts-v2, or posts-v2.1.
Do you need other semantics?

adr/graph-feature-flags.md
jensneuse: No, you can only have one feature flag at a time per Subgraph.
Does this work for your use case?

adr/graph-feature-flags.md
jensneuse: yes, both

adr/graph-feature-flags.md
jensneuse: yes

adr/graph-feature-flags.md
jensneuse: Yes, this is a good idea. We can add it to the Router Config.

adr/graph-feature-flags.md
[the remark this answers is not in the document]
ccortezaguilera: ok, at a time, meaning when making a request not that there could only exist 1 feature flag for the subgraph at all times.

adr/graph-feature-flags.md
ccortezaguilera: One feature flag at a time per subgraph does meet the use case. My follow up question is what would happen if there are multiple flags for the same subgraph i.e someone forgot to remove a cookie for the other feature flag?

adr/graph-feature-flags.md
ccortezaguilera: Great!

ccortezaguilera (approved):

adr/graph-feature-flags.md
```diff
@@ -0,0 +1,291 @@
...
+If feature flags are disabled in the router.yaml or a feature flag is not in the allow list, the Router will return a 400 Bad Request error.
+If feature flags are enabled and the allow list is empty, all feature flags are allowed.
+By default, feature flags are enabled and all feature flags are allowed.
+
+```yaml
+featureFlags:
+  enabled: true
+```
+
+### Feature Flags can be enabled or disabled using the CLI
```
bbil, line 276: If a FF is disabled, and a client makes a request trying to use it, will the default sub-graph just get used?

And if the FF version of the sub-graph has some additional field, that doesn't exist in the default sub-graph, what happens then?

jensneuse: In this case, the router would return a 400 bad request. Do you think this would be the expected behavior?

bbil: I think that's fine. Just means that we need to be careful that our configuration on clients should be updated first (probably with our own Feature Flags), before the feature flag is disabled/removed on the WG side.

Aenimus:
The implementation was changed from this RFC, but the functionality remains.

Please see the [Feature Flag documentation](https://cosmo-docs.wundergraph.com/concepts/feature-flags) and the [Feature Flag tutorial](https://cosmo-docs.wundergraph.com/tutorial/gradual-and-experimental-feature-rollout-with-feature-flags) for more information.

PR #794, opened 2024-05-16
feat: standalone binary distributions for `wgc`

samjcombs:
## Motivation and Context

In enterprise settings, installing a cli via npm globally isnt always possible.  Potential blockers include: 
- complete lack of familiarity with npm ecosystem (android, ios app teams, service teams, etc)
- conflicts with existing node ecosystem (perhaps teams are forced to use node14 or otherwise have conflicting dependencies)

Generally it is industry standard to offer cli tools in standalone binaries.  

## Details

Implementation is using [vercel/pkg](https://github.com/vercel/pkg) to produce the binaries. 

While there are a few options for modern node binary distribution ([deno](https://docs.deno.com/runtime/manual/tools/compiler), [bun](https://bun.sh/docs/bundler/executables), [node-22](https://nodejs.org/api/single-executable-applications.html)), `pkg` allows for easy export of cross-platform node16 binaries without fundamentally altering the runtime. 

NOTICE: `pkg` is deprecated as of early 2024 as the maintainer couldnt keep up with the ecosystem.  Case in point, `pkg` offers no [support for es6 modules](https://github.com/vercel/pkg/issues/1291).   

Luckily, there is a [workaround](https://github.com/vercel/pkg/issues/1291#issuecomment-1733514629). 

The workaround involves using [rollup](https://github.com/rollup/rollup) to bundle the output into a single `.cjs` file.  

Rollup [doesn't seem to support top-level await](https://github.com/rollup/rollup/issues/3623) so I [made IIFEs out of of the two top-level awaits](https://github.com/wundergraph/cosmo/commit/553302028b71fe60a5b72508882fea4deb251027) in the in the package. 

## TODO

- [ ] **Update stub URLS in install scripts with real release locations**
- [ ] **Develop github workflow to build binaries in CI/CD pipeline**

### DIFF
--- a/cli/.eslintrc
+++ b/cli/.eslintrc
@@ -1,8 +1,6 @@
 {
   "extends": ["eslint-config-unjs", "plugin:require-extensions/recommended"],
-  "plugins": [
-        "require-extensions"
-  ],
+  "plugins": ["require-extensions"],
   "rules": {
     "space-before-function-paren": 0,
     "arrow-parens": 0,
@@ -23,6 +21,7 @@
     "@typescript-eslint/no-non-null-assertion": 0,
     "unicorn/expiring-todo-comments": 0,
     "no-useless-constructor": 0,
-    "unicorn/no-process-exit": 0
+    "unicorn/no-process-exit": 0,
+    "unicorn/prefer-top-level-await": 0
   }
 }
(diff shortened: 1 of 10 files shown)


samjcombs:
One of my outstanding questions with the native Node [SEA](https://nodejs.org/api/single-executable-applications.html#single-executable-applications) approach was cross platform executable support. 
Also, last time I checked it was still a preview feature, but since Node-22 is available as `current` as of [2 weeks ago](https://nodejs.org/en/blog/announcements/v22-release-announce) and it looks like it supports [cross platform executable generation](https://github.com/nodejs/single-executable/discussions/89), using native SEA approach would be the ideal path.  

The above `pkg` method could be used as a stopgap until `wgc` is tested/upgraded to support node-22.

PR #803, opened 2024-05-18
feat: add hashHeadersWhitelist configuration for selective  header hashing

glower:
Added support for a new configuration option `hashHeadersWhitelist` in the WebSocket configuration. This allows specifying a whitelist of headers that should be hashed and forwarded during WebSocket connections. 

The implementation includes updates to the schema and handling of the new configuration option in the `WebSocketConnectionHandler`

## Motivation and Context

Implement an allow-list to have fine-grained control over which headers are hashed in the initial subscription upgrade request.

## TODO

- [ x] Tests or benchmark included
- [ x] PR title must follow [conventional-commit-standard](https://github.com/wundergraph/wundergraph/blob/main/CONTRIBUTING.md#conventional-commit-standard)

### DIFF
--- a/router/core/websocket_test.go
+++ b/router/core/websocket_test.go
@@ -1,0 +1,186 @@
+package core
+
+import (
+	"bufio"
+	"bytes"
+	"context"
+	"encoding/json"
+	"net"
+	"net/http"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/wundergraph/cosmo/router/internal/wsproto"
+	"github.com/wundergraph/cosmo/router/pkg/config"
+	"go.uber.org/zap"
+)
+
+type MockConn struct {
+	net.Conn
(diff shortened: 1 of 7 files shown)


router/core/websocket_test.go
```diff
@@ -0,0 +1,186 @@
...
+	"net/http"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/wundergraph/cosmo/router/internal/wsproto"
+	"github.com/wundergraph/cosmo/router/pkg/config"
+	"go.uber.org/zap"
+)
+
+type MockConn struct {
```
glower, line 18: Typically, I would utilize tools like gomock, testify/mock, or mockery to automatically generate mocks for improved testing. However, since this project does not currently use these tools, I chose to manually create the mocks to avoid introducing additional external dependencies.

PR #804, opened 2024-05-19
feat: allow-list to have fine-grained control over headers

henvic:
Adding allow-list to have fine-grained control over headers using a new configuration specifically for WebSocket connections). There are some pre-existing integration tests that are flaky. Let's discuss some approaches to fix this.


StarpTech:
@henvic as discussed. Thank you for your contribution.

PR #806, opened 2024-05-20
fix: add tests and implement graphql over http cases

jensneuse:
This PR Fixes a few issues with malformed requests, ensuring that we properly implement GraphQL over HTTP.
Valid JSON but invalid variables is a 200 with errors.
Invalid request with invalid GraphQL Operation is 400 with errors.


router-tests/graphql_over_http_test.go
```diff
@@ -0,0 +1,109 @@
...
+
+	"github.com/stretchr/testify/require"
+	"github.com/wundergraph/cosmo/router-tests/testenv"
+)
+
+func TestGraphQLOverHTTPCompatibility(t *testing.T) {
+	t.Parallel()
+
+	testenv.Run(t, &testenv.Config{}, func(t *testing.T, xEnv *testenv.Environment) {
+		t.Run("correct baseline", func(t *testing.T) {
```
StarpTech, line 17: Could you be a bit more specific about what this test is validating?

router-tests/graphql_over_http_test.go
```diff
@@ -0,0 +1,109 @@
...
+			}
+			body := []byte(`{"query":"query Find($criteria: SearchInput!) {findEmployees(criteria: $criteria){id details {forename surname}}}","variables":{"criteria":{"nationality":GERMAN}}}`)
+			res, err := xEnv.MakeRequest("POST", "/graphql", header, bytes.NewReader(body))
+			require.NoError(t, err)
+			require.Equal(t, http.StatusBadRequest, res.StatusCode)
+			data, err := io.ReadAll(res.Body)
+			require.NoError(t, err)
+			require.Equal(t, `{"errors":[{"message":"error parsing request body"}],"data":null}`, string(data))
+		})
+		t.Run("missing variables should not be 400", func(t *testing.T) {
```
StarpTech, line 82: Maybe a bit more concrete? `missing variable returns 200 status code`

router/core/operation_processor.go
```diff
@@ -189,6 +190,13 @@ func (o *OperationKit) Parse(ctx context.Context, clientInfo *ClientInfo, log *z
 				parseErr = invalidExtensionsTypeError(valueType)
 				return
 			}
+			if !gjson.ValidBytes(value) {
+				parseErr = &inputError{
+					message:    "error parsing request body",
```
StarpTech, line 195: Two pointa

1. Can we do this check on the whole request body instead of only for extensions?
2. If not, I'd try to be more concrete with the error message otherwise it is hard to debug.

StarpTech (changes requested):

router/core/operation_processor.go
jensneuse: Adding a ticket to address it later.

StarpTech (approved):
LGTM

Issue #826, opened 2024-05-27
Add metrics container port to kubernetes Deployment and Service helm chart of Router


miguel-cardoso-mindera:
### Component(s)

router

### Is your feature request related to a problem? Please describe.

We currently have an issue while scraping our cosmo router instance, as by default prometheus listens to the 8088 port, however this port is not defined neither in the containerPort spec of the Deployment neither in the Service ports spec

### Describe the solution you'd like

Add both entries to their respective files, using a new variable called metricsPort or something like that here:

```
service:
  type: ClusterIP
  port: 3002
  metricsPort: 8088
```

and in the service for example:
```
apiVersion: v1
kind: Service
metadata:
  name: {{ include "router.fullname" . }}
  annotations:
  {{- with .Values.serviceAnnotations }}
    {{- toYaml . | nindent 4 }}
  {{- end }}
  labels:
    {{- include "router.labels" . | nindent 4 }}
spec:
  type: {{ .Values.service.type }}
  ports:
    - port: {{ .Values.service.port }}
      targetPort: http
      protocol: TCP
      name: http
    - port: {{ .Values.service.metricsPort }}
      targetPort: metrics
      protocol: TCP
      name: metrics
  selector:
    {{- include "router.selectorLabels" . | nindent 4 }}
```

Do the same for the deployment containerPorts. Also maybe add by default PROMETHEUS_LISTEN_ADDR env var with the port specified

### Additional context

I'll happily open a PR if this is accepted, otherwise I would like to explore alternatives because it is quite bad currently the way we have manually edit the deployment and service resources on our cluster

StarpTech:
Hi @miguel-cardoso-mindera, Thank you for opening this issue. We're aware of that limitation and will address this soon. We would be happy to review a PR.

miguel-cardoso-mindera:
hey @StarpTech I opened the initial PR here: https://github.com/wundergraph/cosmo/pull/828

I'm unsure how many safety checks we want to add to this, for example, if we want to gate this behind some sort of "prometheus.enabled" gate, if we want to auto configure the prometheus listen address port with this variable etc

Just wanted to get the PR out there, I'm open to feedback and work on it further if needed.

Thanks in advance

StarpTech:
Hi @miguel-cardoso-mindera I like your second approach. Instead of defining the port in the service section, we can create a `prometheus` section and pass those values as environment variables to the deployment.

```yaml
configuration:
  prometheus:
[...]

PR #828, opened 2024-05-28
feat: add metrics port to service and deployment spec of router helm chart

miguel-cardoso-mindera:
## Motivation and Context

Created by: https://github.com/wundergraph/cosmo/issues/826

For prometheus scraping to work correctly, both the deployment spec and the service spec should expose the metrics port


helm/cosmo/charts/router/templates/deployment.yaml
```diff
@@ -123,6 +128,14 @@ spec:
                   name: {{ include "router.secretName" . }}
                   key: graphApiToken
             {{- end }}
+            - name: PROMETHEUS_ENABLED
```
StarpTech, line 131: Please write all values into the configmap first and then reference them here.

helm/cosmo/charts/router/values.yaml
```diff
@@ -211,3 +211,8 @@ configuration:
   # -- The path to the router execution config file. Before, you have to mount the file as a volume and set the path here.
   # -- A possible to solution could be to use an init container to download the file from a CDN. If set, polling of the config is disabled.
   routerConfigPath: ""
+
+  prometheus:
```
StarpTech, line 215: Don't forget to document each property and run `make docs` in the helm repository.

StarpTech (changes requested):

helm/cosmo/charts/router/templates/deployment.yaml
miguel-cardoso-mindera: Please let me know if it is fixed now

helm/cosmo/charts/router/values.yaml
miguel-cardoso-mindera: Please let me know if it is fixed now

helm/cosmo/charts/router/templates/deployment.yaml
```diff
@@ -123,6 +128,23 @@ spec:
...
+              valueFrom:
+                configMapKeyRef:
+                  name: {{ include "router.fullname" . }}
+                  key: prometheusEnabled
+            {{ if .Values.configuration.prometheus.enabled }}
+            - name: PROMETHEUS_LISTEN_ADDR
+              valueFrom:
+                configMapKeyRef:
+                  name: {{ include "router.fullname" . }}
+                  key: prometheusPort
```
StarpTech, line 141: Hi, this won't work because the router expects a hostname-port combination. You can construct the `prometheusListenAddr`. similar as with `listenAddress`

miguel-cardoso-mindera: You're right, when changing to the ConfigMap reference I forgot to build the full address

StarpTech (approved):
LGTM

StarpTech:
Has been released as 0.3.4 `ghcr.io/wundergraph/cosmo/helm-charts/router:0.3.4`

PR #840, opened 2024-05-30
feat: use Helm tpl function on .Values.commonConfiguration

bbil:
## Motivation and Context

Allows users of the Helm chart to use dynamic content for `.Values.commonConfiguration` string -- used in the Config Map.

```
commonConfiguration: |-
  headers:
    subgraphs:
      someSubgraph:
        {{- include "subgraph.headers" . | nindent 6 }}
```

Without this change, `.Values.commonConfiguration` is just a string, and the dynamic part does not get expanded.        

## TODO

### DIFF
--- a/helm/cosmo/charts/router/templates/config-map.yaml
+++ b/helm/cosmo/charts/router/templates/config-map.yaml
@@ -12,7 +12,7 @@ data:
   config.yaml: |-
     # User-supplied common configuration:
     {{- if .Values.commonConfiguration }}
-    {{- .Values.commonConfiguration | nindent 4 }}
+    {{- tpl .Values.commonConfiguration . | nindent 4 }}
     {{- end }}
   logLevel: "{{ .Values.configuration.logLevel }}"
   devMode: "{{ .Values.configuration.devMode }}"
--- a/helm/cosmo/charts/router/Chart.yaml
+++ b/helm/cosmo/charts/router/Chart.yaml
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.3.4
+version: 0.4.0
 
 # This is the version number of the application being deployed. This version number should be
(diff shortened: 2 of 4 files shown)


bbil:
This will also require a bump on the Chart version, should I also be doing this bump in this PR?

Also, I'm wondering if there's a good way to test this out locally... I've tested this works with a dummy config map yaml file, so I think it should be fine. But messing with Helm is always kinda sketchy to me, with understanding if this is going to be fine in a full end-to-end setup.

jeffutter:
👍 I like this idea. As far as I understand, shouldn't change anything for existing users that don't have things that look like templates in their configs.

StarpTech:
Hi @bbil, thank you for the PR. We have established a local setup with minikube to test our Helm charts. I can guide you through at some point when you're doing more significant work in the Helm charts.

> This will also require a bump on the Chart version, should I also be doing this bump in this PR?

Yes, please bump minor. The release is still manual after merge to main.

helm/cosmo/charts/router/templates/config-map.yaml
```diff
@@ -12,7 +12,7 @@ data:
   config.yaml: |-
     # User-supplied common configuration:
     {{- if .Values.commonConfiguration }}
-    {{- .Values.commonConfiguration | nindent 4 }}
+    {{- tpl .Values.commonConfiguration . | nindent 4 }}
```
StarpTech, line 15: Could you document this behaviour in the chart comments of `commonConfiguration ` ? After running `make docs`, the docs are regenerated.

StarpTech (changes requested):

jeffutter:
@StarpTech Brandon is out this week, so I'll be helping out here. I bumped the minor version and updated the docs. In updating the docs I noticed that there were a few other comments which had multiple lines that were only including the last line. I fixed those as well.

Please let me know if there are any other changes you would like to see.

StarpTech:
@jeffutter thank you 👏

StarpTech (approved):
LGTM

StarpTech:
Published as 0.4.0

PR #852, opened 2024-06-04
feat: admission webhook signature

thisisnithin:
## Motivation and Context

Sign requests made to the admission webhook endpoint if applicable.

## TODO

### DIFF
--- a/controlplane/src/core/services/AdmissionWebhookController.ts
+++ b/controlplane/src/core/services/AdmissionWebhookController.ts
@@ -1,3 +1,4 @@
+import { createHmac } from 'node:crypto';
 import axios, { AxiosError, AxiosInstance } from 'axios';
 import { JWTPayload } from 'jose';
 import axiosRetry, { exponentialDelay } from 'axios-retry';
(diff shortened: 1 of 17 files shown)


controlplane/src/core/services/AdmissionWebhookController.ts
```diff
@@ -50,10 +52,19 @@ export class AdmissionWebhookController {
     this.logger.debug({ url, path: '/validate-config', ...req }, 'Sending admission validate-config webhook request');
 
     try {
+      const headers: Record<string, string> = {};
+      if (this.graphAdmissionWebhookSecret) {
+        const dataString = JSON.stringify(req);
```
jensneuse, line 57: I would expect that you hash the exact content you're sending as "data" on the request (line 66) but it seems like that's not the case? Although it might result in the same outcome in the current case, wouldn't it be better to hash the actual content of the request and not a JSON encoded representation of the payload which is passed to data?

jensneuse (changes requested):

controlplane/src/core/services/AdmissionWebhookController.ts
thisisnithin: createHmac expects a string

jensneuse: can you pass the exact same string to the request (data)

thisisnithin: You shouldnt need to parse a json response. The signature on the other hand you have to do it no matter what.

jensneuse (approved):

StarpTech (approved):
LGTM

PR #859, opened 2024-06-06
fix: ignore directive changes instead of throwing

thisisnithin:
## Motivation and Context

Throwing this makes it such that the entire check is not inspectable. A check may have both inspectable and non inspectable changes. We should still respect checking for ones that are inspectable.

## TODO


StarpTech (approved):
LGTM

PR #864, opened 2024-06-11
chore: RFC - Graph Pruning / Graph Garbage Collection

jensneuse:
This PR introduces a new workflow, Graph Pruning / Graph Garbage Collection.
The idea is to remove unused fields, keep the Schema more maintainable and reduce technical debt.

Please add your feedback directly on the md file in the form of PR comments.
If you're thinking of alternative solutions, please consider opening up another PR to propose another workflow.

### DIFF
--- a/rfc/graph-pruning-garbage-collection.md
+++ b/rfc/graph-pruning-garbage-collection.md
@@ -1,0 +1,147 @@
+---
+title: "Graph Pruning / Garbage Collection of unused fields"
+author: Jens Neuse
+---
+
+This RFC describes a new workflow to continuously clean up the GraphQL Schema.
+The goal is to remove unused fields from the schema to keep it clean and more maintainable.
+
+## Problem
+
+The GraphQL Schema is a living document that evolves over time.
+Fields are added, modified, and removed.
+Over time, fields that are no longer used are not removed from the schema.
+This leads to a bloated schema that is hard to understand and maintain.
+
+Cosmo already has metrics that show which fields are used or not,
+but there is no automated way or workflow to remove unused fields from the schema.
+
+Removing unused fields through an automated process has the following benefits:
+1. The schema is easier to read understand
+2. The schema is easier to maintain
+3. Unused Code in resolvers can be removed, reducing complexity and technical debt
+4. The schema is smaller, which can lead to better performance
+
+## Solution
+
+The simplest possible solution is to re-use the existing workflow for schema composition checks to also check for unused fields.
+Schema composition checks are already part of the CI/CD pipeline and are executed whenever a new version of a subgraph is published.
+We can extend this workflow to also check for unused fields.
+
+We have several tools to facilitate this.
+We can have strict rules (fail CI) for unused fields, or we can have warnings.
+Depending on how strict the organization wants to operate,
+they can choose the appropriate level of strictness.
+
+In addition, you should be able to immediately remove unused fields or deprecate them first.
+
+## Implementation
+
+First, we need to define the rules for unused fields.
+We might want to have different rules for different environments,
+so it would make sense to define these rules on the namespace level.
+
+Let's create a new namespace:
+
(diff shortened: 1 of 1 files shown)


rfc/graph-pruning-garbage-collection.md
```diff
@@ -0,0 +1,147 @@
...
+## Escape Hatch
+
+It's possible that some fields are not used in the Schema but there are other reasons to keep them,
+e.g. for business reasons or because there are plans to use them in the future.
+
+In this case, you can add a special comment to the field:
+
+```graphql
+type Query {
+  # wgc-check:ignore-unused
```
maxkomarychev, line 144: I believe this could be challenging in workflows where the schema is an artifact generated from the code.

I'd prefer if we could implement ignore rules based on tag associated with fields or entire types

rfc/graph-pruning-garbage-collection.md
```diff
@@ -0,0 +1,147 @@
...
+## Implementation
+
+First, we need to define the rules for unused fields.
+We might want to have different rules for different environments,
+so it would make sense to define these rules on the namespace level.
+
+Let's create a new namespace:
+
+```shell
+npx wgc namespace create production --unused-fields-warn=30 --unused-fields-fail=60
```
maxkomarychev, line 47: I wonder if this could be configured in the same place where linter rules are configured to keep things related similar.

rfc/graph-pruning-garbage-collection.md
```diff
@@ -0,0 +1,147 @@
...
+
+Removing unused fields through an automated process has the following benefits:
+1. The schema is easier to read understand
+2. The schema is easier to maintain
+3. Unused Code in resolvers can be removed, reducing complexity and technical debt
+4. The schema is smaller, which can lead to better performance
+
+## Solution
+
+The simplest possible solution is to re-use the existing workflow for schema composition checks to also check for unused fields.
```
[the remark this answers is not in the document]
jensneuse: Yes. How do you think about that?

rfc/graph-pruning-garbage-collection.md
jensneuse: Lint rules are per org. I'm assuming that unused field rules would differ across namespaces (environments).

rfc/graph-pruning-garbage-collection.md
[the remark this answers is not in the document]
SteWootten: Works for us 👍

rfc/graph-pruning-garbage-collection.md
jensneuse: How are you generating your Schema? Are you using a code-first approach? Do you think a directive could work better? E.g. like this:

```graphql
type Query {
  oldField: String @keep(reason: "This field will be used in the future")
}
```

Or we put the directive on the type and it affects all enclosing fields:

```graphql
type Query @keep(reason: "This field will be used in the future") {
  oldField: String
}
```

maxkomarychev: Currently we are using code-first approach with [nestjs](https://docs.nestjs.com/graphql/quick-start#code-first) and [graphql-kotlin](https://opensource.expediagroup.com/graphql-kotlin/docs/).

Great idea! A dedicated directive could be a nice option too.

2c: "keep" is a bit too ambiguous and also too specific to my taste 😅 . I wonder if there is a possibility to create more generic directive to override linter checks in general. Like `@SuppressWarning` in java.

jensneuse: Based on other feedback, I'm leaning towards adding this feature into the linter category. If we're doing that, we could do a more generic approach like this:

```graphql
type Query {
  oldField: String @ignoreLint(reason: "This field will be used in the future")
}
```

What are your thoughts on this?

PR #874, opened 2024-06-19
fix: client schema usage, update wgc router compose to use client schema

devsergiy:
## TODO

### DIFF
--- a/router/core/executor.go
+++ b/router/core/executor.go
@@ -8,22 +8,20 @@ import (
 	"net/http"
 	"time"
 
+	"github.com/nats-io/nats.go"
 	"github.com/twmb/franz-go/pkg/kgo"
 	"github.com/twmb/franz-go/pkg/sasl/plain"
 	"go.uber.org/zap"
 
+	nodev1 "github.com/wundergraph/cosmo/router/gen/proto/wg/cosmo/node/v1"
 	"github.com/wundergraph/cosmo/router/pkg/config"
-
-	"github.com/nats-io/nats.go"
-
 	"github.com/wundergraph/graphql-go-tools/v2/pkg/ast"
 	"github.com/wundergraph/graphql-go-tools/v2/pkg/astparser"
 	"github.com/wundergraph/graphql-go-tools/v2/pkg/asttransform"
 	"github.com/wundergraph/graphql-go-tools/v2/pkg/engine/datasource/introspection_datasource"
 	"github.com/wundergraph/graphql-go-tools/v2/pkg/engine/plan"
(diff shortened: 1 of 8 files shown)


router/core/executor.go
```diff
@@ -72,32 +72,43 @@ func (b *ExecutorConfigurationBuilder) Build(ctx context.Context, routerConfig *
...
+		if report.HasErrors() {
+			return nil, fmt.Errorf("failed to parse graphql client schema from engine config: %w", report)
+		}
+		err = asttransform.MergeDefinitionWithBaseSchema(&clientSchema)
+		if err != nil {
+			return nil, fmt.Errorf("failed to merge graphql client schema with base schema: %w", err)
+		}
+		clientSchemaDefinition = &clientSchema
+	} else {
+		clientSchemaDefinition = &routerSchemaDefinition
```
jensneuse, line 103: This is a backwards compatible fallback, right? Can we mark it as such and put a date for removal?

router/core/operation_planner.go
```diff
@@ -60,8 +61,8 @@ func (p *OperationPlanner) preparePlan(requestOperationName []byte, requestOpera
 
 	validation := astvalidation.DefaultOperationValidator()
 
-	// validate the document before planning
-	state := validation.Validate(&doc, p.executor.Definition, &report)
+	// validate the document against client schema before planning
+	state := validation.Validate(&doc, p.executor.ClientSchema, &report)
```
jensneuse, line 65: Are we normalizing, etc... against the client schema as well? It would sound reasonable to me.

jensneuse (changes requested):

StarpTech (approved):
LGTM

router/core/operation_planner.go
devsergiy: yeah, everything except planning should be done against client schema

router/core/executor.go
devsergiy: yes

What should be a timeframe for this?

Should we log warning? e.g. "you are using outdated wgc version"

StarpTech: No, the ClientSchema is optional and only used with schema contracts. If schema contracts aren't used the router schema is the client and router schema.

devsergiy: not fully true

In case schema has `inaccessible` directive there will be 2 schemas

Aenimus: The client schema is a product of `@inaccessible` AND/OR `@tag` directives.

jensneuse (approved):

router/core/executor.go
StarpTech: Yes, but still it is nothing we want to get rid of in the nearly future.

PR #880, opened 2024-06-21
fix: allow configuration of `AllowWildcard`

### LINKED ISSUE #879: Cors AllowWildcard never set - Wildcard origins not taken into account
router v0.92.3
 
## Description
 
Using only a `config.yaml` to setup the router, it is impossible to make use of wildcards in `cors.allow_origins` as `AllowWildcard` is never set and therefore `parseWildcardRules()` doesn't pickup any rules.
 
## Steps to Reproduce
 
Start a router with the following in the config:
```yaml
cors:
  allow_origins: [http://*]
  allow_headers: ["*"]
  allow_credentials: true
```
 
Use chrome to make a fetch request to your router and you will get 403 Forbidden in the OPTIONS request.
 
## Expected Result
 
OPTIONS request should go through
 
## Actual Result
 
403 Forbidden in the OPTIONS request
 
## Useful links
- `allow_wildcard` should be added here ? https://github.com/wundergraph/cosmo/blob/main/router/pkg/config/config.go#L112
- AllowWildcard is never set anywhere (Except in tests where its manually set to true) https://github.com/search?q=repo%3Awundergraph%2Fcosmo+AllowWildcard&type=code
saved_reply_id:
issue[textarea_helper_element]: ## Description
 
Using only a `config.yaml` to setup the router, it is impossible to make use of wildcards in `cors.allow_origins` as `AllowWildcard` is never set and therefore `parseWildcardRules()` doesn't pickup any rules.
 
## Steps to Reproduce
 
Start a router with the following in the config:
```yaml
cors:
  allow_origins: [http://*]
  allow_headers: ["*"]
  allow_credentials: true
```
 
Use chrome to make a fetch request to your router and you will get 403 Forbidden in the OPTIONS request.
 
## Expected Result
 
OPTIONS request should go through
 
## Actual Result
 
403 Forbidden in the OPTIONS request
 
## Useful links
- `allow_wildcard` should be added here ? https://github.com/wundergraph/cosmo/blob/main/router/pkg/config/config.go#L112
- AllowWildcard is never set anywhere (Except in tests where its manually set to true) https://github.com/search?q=repo%3Awundergraph%2Fcosmo+AllowWildcard&type=code
 

## Config

```yaml
version: '1'
 
dev_mode: true
router_config_path: config.json
graph:
  token: ''
 
listen_addr: 0.0.0.0:3002
 
graphql_path: '/'
 
headers:
  all:
    request:
      - op: 'propagate'
        named: authorization
      - op: 'propagate'
        named: Cookie
      - op: 'propagate'
        named: origin
 
cors:
  allow_origins: [http://*]
  allow_headers: ["*"]
  allow_credentials: true
 
telemetry:          
  prometheus:
    enabled: false # Side node: disabling prometheus doesn't work too

```

YassineElbouchaibi:
This MR fixes #879

## Motivation and Context

Wildcard added to `cors.allow_origins` won't work without this MR as `AllowWildcard` is always `false`.

Let me know if my PR is missing stuff!

### DIFF
--- a/router/pkg/config/config.schema.json
+++ b/router/pkg/config/config.schema.json
@@ -519,6 +519,11 @@
           "default": true,
           "description": "The allowed credentials. The default value is to allow credentials. This allows the browser to send cookies and authentication headers."
         },
+        "allow_wildcard": {
+          "type": "boolean",
+          "default": true,
+          "description": "Allow wildcard origins in the allowed origins list. e.g. http://some-domain/*, https://api.* or http://some.*.subdomain.com."
+        },
         "max_age": {
           "type": "string",
           "duration": {
--- a/router/cmd/instance.go
+++ b/router/cmd/instance.go
@@ -138,6 +138,7 @@ func NewRouter(params Params, additionalOptions ...core.Option) (*core.Router, e
 			AllowOrigins:     cfg.CORS.AllowOrigins,
 			AllowMethods:     cfg.CORS.AllowMethods,
 			AllowCredentials: cfg.CORS.AllowCredentials,
+			AllowWildcard: 	  cfg.CORS.AllowWildcard,
 			AllowHeaders:     cfg.CORS.AllowHeaders,
 			MaxAge:           cfg.CORS.MaxAge,
 		}),
--- a/router/pkg/config/config.go
+++ b/router/pkg/config/config.go
@@ -114,6 +114,7 @@ type CORS struct {
 	AllowMethods     []string      `yaml:"allow_methods" default:"HEAD,GET,POST" envconfig:"CORS_ALLOW_METHODS"`
 	AllowHeaders     []string      `yaml:"allow_headers" default:"Origin,Content-Length,Content-Type" envconfig:"CORS_ALLOW_HEADERS"`
 	AllowCredentials bool          `yaml:"allow_credentials" default:"true" envconfig:"CORS_ALLOW_CREDENTIALS"`
+	AllowWildcard    bool	       `yaml:"allow_wildcard" default:"true" envconfig:"CORS_ALLOW_WILDCARD"`
 	MaxAge           time.Duration `yaml:"max_age" default:"5m" envconfig:"CORS_MAX_AGE"`
 }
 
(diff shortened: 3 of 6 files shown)


router/pkg/config/config.schema.json
```diff
@@ -519,6 +519,11 @@
           "default": true,
           "description": "The allowed credentials. The default value is to allow credentials. This allows the browser to send cookies and authentication headers."
         },
+        "allow_wildcard": {
```
StarpTech, line 522: @YassineElbouchaibi You also need to pass it to `core.WithTLSConfig` in `instance.go` otherwise the setting is not applied. How did you test it?

StarpTech (changes requested):

router/pkg/config/config.schema.json
YassineElbouchaibi: Ah yes sorry missed that... I have added it to `core.WithCors` now. I did not test unfortunately as I have made these changes directly in the web and I don't have a working local setup. I was hoping the pipeline tests would pickup anything wrong. Let me know if you think something else is missing.

StarpTech: I have approved the run. After the build, you can test the image with

`docker pull ghcr.io/wundergraph/cosmo/router:pr-880`

Would that work for you?

YassineElbouchaibi: Yes I definitely can test that, thanks!

StarpTech: Ideally, you can add an integration test to https://github.com/wundergraph/cosmo/tree/main/router-tests but since we did not have any tests for CORS, I can help you with that (if needed). Let me know what you think.

StarpTech: My bad. Images from contributors aren't pushed due to insufficient permissions. You can build the image locally with `cd router && docker build -t router:latest .`

YassineElbouchaibi: Sorry for the delay, my local setup was not able to build the image locally because of the following:

image

so I had to go through various hoops to get it working..

But I can now confirm wild card origins work and stop working when the setting is changed to `false`.

Regarding integration tests, It will be hard for me to find time for it.. I'm not very familiar with golang let alone testing in golang.. but if you give me enough time I can look into it.

PR #884, opened 2024-06-24
fix: cleanup inspectable check

thisisnithin:
## Motivation and Context

The method no longer throws for any case. So right now we will always have the check to be inspectable.

## TODO


StarpTech (approved):
LGTM

PR #893, opened 2024-06-26
Issue #391: added composition tests for Query and Mutation fields that returns Query and Mutation

maxrad82:
Added a test for issue #391. The composition fails but should not.


Aenimus:
Thank you for this @maxrad82 

I've included this test as part of the internal graph rewrite:
https://github.com/wundergraph/cosmo/pull/964

PR #898, opened 2024-06-27
chore: refactor operation processor to allow caching of normalized op…

jensneuse:


router/core/operation_processor.go
```diff
@@ -363,30 +448,182 @@ func (o *OperationKit) Normalize() error {
...
+	o.cache.persistetOperationVariableNamesLock.RUnlock()
+	if !ok {
+		return 0, false
+	}
+	key = o.generatePersistedOperationCacheKey(variableNames)
+	return key, true
+}
+
+func (o *OperationKit) generatePersistedOperationCacheKey(skipIncludeVariableNames []string) uint64 {
+	o.kit.keyGen.Reset()
```
StarpTech, line 559: A minor thing but IMO we should always clear memory after a task otherwise the memory of the last task is preserved for no reason.

router/core/operation_processor.go
```diff
@@ -363,30 +448,182 @@ func (o *OperationKit) Normalize() error {
...
+
+var (
+	literalF    = []byte("f")
+	literalT    = []byte("t")
+	literalNull = []byte("null")
+)
+
+func (o *OperationKit) loadPersistedOperationCacheKey(persistedQuerySha256Hash string) (key uint64, ok bool) {
+	o.cache.persistetOperationVariableNamesLock.RLock()
+	variableNames, ok := o.cache.persistetOperationVariableNames[persistedQuerySha256Hash]
```
StarpTech, line 549: We create an operationKit per request. From an abstraction perspective, it would be great if the kit wouldn't know anything from the cache so that the kit and the cache are not coupled. Pseudocode:

```go
kit := NewOperationKit()
normalization := kit.Normalize()
parsedOp, ok := cache.Get(sha256)
if !ok {
  kit.Parse(...)
}

```

router/core/operation_processor.go
```diff
@@ -363,30 +448,182 @@ func (o *OperationKit) Normalize() error {
...
+		return false, &reportError{
+			report: report,
+		}
+	}
+
+	return true, nil
+}
+
+func (o *OperationKit) savePersistedOperationToCache(skipIncludeVariableNames []string, exportedVariables []byte) {
+	if o.cache == nil {
```
StarpTech, line 519: I'd write it in way that a function call is not made when the cache is not set because the function has no meaning without the cache.

router/core/operation_processor.go
```diff
@@ -299,51 +317,118 @@ func (o *OperationKit) Parse(ctx context.Context, clientInfo *ClientInfo, log *z
...
+	// Hash the normalized operation with the static operation name to avoid different IDs for the same operation
+	err = o.kit.printer.Print(o.kit.doc, o.operationParser.executor.ClientSchema, o.kit.keyGen)
+	if err != nil {
+		return errors.WithStack(fmt.Errorf("failed to print normalized operation: %w", err))
 	}
 
-	if extensions.PersistedQuery != nil {
-		o.parsedOperation.PersistedID = extensions.PersistedQuery.Sha256Hash
+	// Generate the operation ID
+	o.parsedOperation.ID = o.kit.keyGen.Sum64()
```
StarpTech, line 398: It would be more idiomatic and clear if the hash is reset after his usage here.

router/core/operation_processor.go
jensneuse: Yes but you cannot have a global cache here. The problem is that the overall architecture of the operation processor needs to be rewritten. You cannot really do dependency injection with the current way things are organized. Once all of this works and is well tested, I can refactor the whole pipeline. Keep in mind that the kit doesn't know anything about the cache, it's just a pointer for DI purposes. Any other suggestions to improve it without globals?

StarpTech: I understand, then I'd postpone the discussion to the general refactor.

StarpTech (approved):
LGTM

Issue #915, opened 2024-07-04
Handle websocket authentication via initial payload


alexandra-c:
### Component(s)

router

### Is your feature request related to a problem? Please describe.

The Cosmo router's built-in authentication currently looks for the JWT token only in the request header. For websocket requests made from a browser, headers cannot be altered, so the token can only be sent via URL Query, Cookie, or initial payload.

Currently, if the router configuration is set to authorization >> require_authentication: true, websocket requests will receive an Unauthorized response, causing Subscriptions to fail.

### Describe the solution you'd like

The Cosmo router's websocket authentication behavior needs to be updated to allow the JWT token to be acquired from the initial payload as well as the header.

@jensneuse suggested the following implementation steps on Discord:

Add a configuration option to modify WebSocket authentication behavior to allow authentication via the initial payload.
Allow configuration of the initial payload field where the auth token is located.
Modify websocket.go to consider this configuration.

### Additional context

I have started working on a version of authentication that also looks for the token in the initial payload. I created this feature request to further discuss the implementation steps in more detail.

Thank you! :)

jensneuse:
Hey, thanks for opening this issue.
Can you describe how you'd like to solve this issue?
Ideally, you could start outlining the solution and create a PR that updates the Router config so that we can discuss how to implement this.

alexandra-c:
Here's a draft of my implementation #918
I added a new configuration section for websocket that handles authentication, I was thinking in the future maybe other authentication methods may be implemented... 
It looks something like this: 
```
websocket:
  ...
  authentication:
    from_initial_payload:
      enabled: true
      secret_key: "authorization"
```
Then I separated the initialPayload authentication from the header one, so I defined a new authenticator `websocketInitialPayloadAuthenticator` and renamed the existing one in `httpHeaderAuthenticator` since it was only handling authentication via header.  
Also extracted the JWT decoding part in a separate interface called `jwksTokenDecoder` in order to reuse it in both authenticators.

Now the problem is that in `handleUpgradeRequest` method from `websocket.go` file, we are checking access control before upgrading the connection, which for the auth via initial payload that's too soon, we won't have the initial payload. For ws auth via initial payload, the checking must be done later on, after `handler.Initialize()` is called, hence the new `WebSocketAuthenticationConfiguration` section, and then set tehn initialPayload to the request context so we can use it the authenticator.  This probably can be made in a better way...I'm open to suggestions.

[...]

PR #918, opened 2024-07-04
feat: handle websocket authentication via initial payload

alexandra-c:
Fixing #915

- New `websocketInitialPayloadAuthenticator`:

Introduced a new `authentication.Authenticator` implementation called `websocketInitialPayloadAuthenticator`. This handles authentication specifically via the initial payload of a WebSocket connection.

- Refactoring and Renaming:

The existing `jwksAuthenticator` has been refactored and renamed to `httpHeaderAuthenticator` to better reflect its functionality, which is to authenticate solely through HTTP headers.

- JWT Handling Code Extraction:

The JWT handling logic has been extracted into a separate interface. This change allows the JWT processing code to be reused across both the `websocketInitialPayloadAuthenticator` and `httpHeaderAuthenticator`.

- WebSocket Authentication Configuration:

A new configuration section for WebSocket authentication has been added. This section allows the configuration of authentication via the initial payload and includes an option to forward the token from the initial payload to the request header (`export_token`).

 ```
"WebSocket": {
    [...]
    "Authentication": {
      "FromInitialPayload": {
        "Enabled": false,
        "Key": "Authorization",
        "ExportToken": {
          "Enabled": true,
          "HeaderKey": "Authorization"
        }
      }
    }
}
```

## Motivation and Context

The Cosmo router's built-in authentication previously only checked for JWT tokens in the request header. However, when making WebSocket requests from a browser, headers cannot be modified, meaning the token must be sent via URL Query, Cookie, or the initial payload.

This limitation caused issues when the router's configuration required authentication (`require_authentication: true`), resulting in WebSocket requests being unauthorized and leading to subscription failures.

To address this, the WebSocket authentication behavior of the Cosmo router has been updated. The JWT token can now be retrieved from the initial payload, in addition to the request header, thereby improving the router's flexibility and compatibility with WebSocket connections.

## TODO

- [x] Documentation is changed or added on [https://app.gitbook.com/](https://app.gitbook.com/)

### CODE REVIEW

**Comment by jensneuse** [comment]
@alexandra-c if a client authenticates via initial payload, is this then forwarded as initial payload to a subgraph subscription, via header, or both?

Subsequently, what about sub requests via http to other subgraphs, e.g. to join data to the subscription. In this case I would expect that we're able to map the initial payload field to a subgraph request header. This might be Authorization, but some services, e.g. on AWS cannot use this header name, so it needs to be configurable.

Have you thought about the two use cases?

**Comment by alexandra-c** [comment]
> @alexandra-c if a client authenticates via initial payload, is this then forwarded as initial payload to a subgraph subscription, via header, or both?

The initial payload is being sent to the subgraphs by default, so it will only be on the initial payload since right now no one is copying it to the header.

> Subsequently, what about sub requests via http to other subgraphs, e.g. to join data to the subscription. In this case I would expect that we're able to map the initial payload field to a subgraph request header. This might be Authorization, but some services, e.g. on AWS cannot use this header name, so it needs to be configurable.
> 
> Have you thought about the two use cases?

I haven't really, I tested it right now and indeed we might need to copy the token to the headers also, for the sub requests. Not sure how the configuration should look like or where to put it... I think the header where we'd copy the token should be the same we configured  in the `authentication` section, but there we actually have an array.

How do you suggest I should implement this?

File: router/pkg/config/config.schema.json [hunk 1]
```diff
@@ -176,6 +172,30 @@
...
+          "type": "object",
+          "description": "The configuration used to authenticate the WebSocket connections.",
+          "additionalProperties": false,
+          "properties": {
+            "from_initial_payload": {
+              "type": "object",
+              "description": "The configuration used to tell the Router to look for the JWT (JSON Web Token) in the initial payload of the WebSocket Connection.",
+              "additionalProperties": false,
+              "properties": {
+                "secret_key": {
```
jensneuse, line 186: I'd just call it "key".

router/pkg/config/config.schema.json
```diff
@@ -176,6 +172,30 @@
...
+                  "type": "string",
+                  "default": "Authorization",
+                  "description": "The name of the property holding the JWT (JSON Web Token) value. The default value is 'Authorization'"
+                },
+                "enabled": {
+                  "type": "boolean",
+                  "description": "This configuration indicates if the Router should look for the JWT (JSON Web Token) in the initial payload of the WebSocket Connection. The default value is 'false'.",
+                  "default": false
+                }
```
jensneuse, line 195: We could add a property "export_token_to_client_request_header" with the following properties:
- enabled
- header_name (the header which we export the token into)

With this config, we could enable header forwarding in the engine, which would allow us to forward the token extracted from initial payload to sub-requests.

WDYT?

alexandra-c: sounds good 👍

alexandra-c: It seems like "header_name" is a reserved word or something because it did not let me use it, so I used "header_key" instead.

alexandra-c:
Hey @jensneuse, I think I finished the implementation here, can you take a look? 😊

jensneuse (changes requested):

router-tests/websocket_test.go
```diff
@@ -272,6 +280,121 @@ func TestWebSockets(t *testing.T) {
...
+			})
+			require.NoError(t, err)
+
+			var res testenv.WebSocketMessage
+			err = conn.ReadJSON(&res)
+			require.NoError(t, err)
+			require.Equal(t, "error", res.Type)
+			payload, err := json.Marshal(res.Payload)
+			require.NoError(t, err)
+			require.JSONEq(t, `[{"message":"unauthorized"}]`, string(payload))
```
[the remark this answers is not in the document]
alexandra-c: It won't close the ws connection. Do we want to close it?

router-tests/websocket_test.go
```diff
@@ -272,6 +280,121 @@ func TestWebSockets(t *testing.T) {
...
+						FromInitialPayload: config.InitialPayloadAuthenticationConfiguration{
+							Enabled: true,
+						},
+					},
+				}),
+			},
+		}, func(t *testing.T, xEnv *testenv.Environment) {
+			token, err := authServer.Token(nil)
+			require.NoError(t, err)
+			initialPayload := []byte(`{"Authorization":"Bearer ` + token + `"}`)
```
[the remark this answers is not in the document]
alexandra-c: I think is required. I used the same logic from header authenticator,  "Bearer" prefix is used as a default. 
https://github.com/wundergraph/cosmo/blob/45eeb141aac7bae702ef6215a74209c7ba676887/router/pkg/authentication/http_header_authenticator.go#L79

router-tests/websocket_test.go
```diff
@@ -272,6 +280,121 @@ func TestWebSockets(t *testing.T) {
...
+			require.NoError(t, err)
+			require.Equal(t, "next", res.Type)
+			require.Equal(t, "1", res.ID)
+			require.JSONEq(t, `{"data":{"employeeUpdated":{"id":3}}}`, string(res.Payload))
+
+			require.NoError(t, conn.Close())
+			xEnv.WaitForSubscriptionCount(0, time.Second*5)
+		})
+	})
+	t.Run("subscription with authorization via initial payload reject", func(t *testing.T) {
```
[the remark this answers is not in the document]
alexandra-c: sure, will do.

router-tests/websocket_test.go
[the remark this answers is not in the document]
jensneuse: @StarpTech do we want to close a WS connection when auth via initial payload failed? I'd say yes, we want to close it. What's your opinion?

router-tests/testenv/testenv.go
```diff
@@ -693,6 +693,12 @@ func configureRouter(listenerAddr string, testConfig *Config, routerConfig *node
 				},
 			},
 			ForwardInitialPayload: true,
+			Authentication: config.WebSocketAuthenticationConfiguration{
```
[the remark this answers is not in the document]
alexandra-c: yes, I think it will probably break other tests, I have to fix it but I'm having a hard time running the dev-setup since I'm working on Windows, and the make commands seems to be compatible only with linux or IOS. Do you have any suggestion how can I make it work on Windows?

router-tests/websocket_test.go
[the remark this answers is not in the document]
StarpTech: The user could be able to reuse the connection for a retry in case of the IDP has changed but I think this use case is rare and we should favor better resource utilization. I also favor closing it.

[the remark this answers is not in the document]
alexandra-c: I'm a bit confused about how to update my code to correctly close the ws and also how to update my test to assert that the ws gets closed. @jensneuse or @StarpTech, can you please help me with some ideas? Thanks :)

router-tests/testenv/testenv.go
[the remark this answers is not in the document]
alexandra-c: Managed to run them in a virtual machine.

router-tests/websocket_test.go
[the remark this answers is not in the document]
jensneuse: > The user could be able to reuse the connection for a retry in case of the IDP has changed but I think this use case is rare and we should favor better resource utilization. I also favor closing it.

Actually I think that's not correct. There can only be one single "initial payload" from the very first message sent by the client.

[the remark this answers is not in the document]
jensneuse: > I'm a bit confused about how to update my code to correctly close the ws and also how to update my test to assert that the ws gets closed. @jensneuse or @StarpTech, can you please help me with some ideas? Thanks :)

If you don't yet have a ws conn handler, you can close it like this:
https://github.com/wundergraph/cosmo/blob/cc435bef3a20f72c03c04f54328e63dce410565c/router/core/websocket.go#L280

If you have a ws conn handler initialized already, you can close it like this:
https://github.com/wundergraph/cosmo/blob/cc435bef3a20f72c03c04f54328e63dce410565c/router/core/websocket.go#L306

In case of epoll, the conn needs to be removed:
https://github.com/wundergraph/cosmo/blob/cc435bef3a20f72c03c04f54328e63dce410565c/router/core/websocket.go#L460

In case of non-epoll, just close the conn/handler, nothing else needs to be done.

For testing, you can do multiple things.

e.g. try reading from client conn and assume you'll get an error:
https://github.com/wundergraph/cosmo/blob/cc435bef3a20f72c03c04f54328e63dce410565c/router-tests/websocket_test.go#L338

You can also wait for connection count to be zero:
https://github.com/wundergraph/cosmo/blob/cc435bef3a20f72c03c04f54328e63dce410565c/router-tests/nats_events_test.go#L84

Let me know when you think all tests pass, then I can enable the CI.
If you need further help, let me know.

jensneuse:
We're taking a look this week.

router-tests/websocket_test.go
```diff
@@ -272,6 +280,163 @@ func TestWebSockets(t *testing.T) {
...
+
+			var res testenv.WebSocketMessage
+			err = conn.ReadJSON(&res)
+			require.NoError(t, err)
+			require.Equal(t, "error", res.Type)
+			payload, err := json.Marshal(res.Payload)
+			require.NoError(t, err)
+			require.JSONEq(t, `[{"message":"unauthorized"}]`, string(payload))
+
+			xEnv.WaitForSubscriptionCount(0, time.Second*5)
```
jensneuse, line 387: after this line, can you run `xEnv.WaitForConnectionCount(0,time.Second*5)`

router-tests/websocket_test.go
```diff
@@ -272,6 +280,163 @@ func TestWebSockets(t *testing.T) {
...
```
jensneuse, line 436: after this line, can you run `xEnv.WaitForConnectionCount(0,time.Second*5)`

router/core/websocket.go
```diff
@@ -308,6 +311,44 @@ func (h *WebsocketHandler) handleUpgradeRequest(w http.ResponseWriter, r *http.R
...
+		}
+		handler.r = validatedReq
+
+		// Export the token from the initial payload to the request header
+		if fromInitialPayloadConfig.ExportTokenToRequestHeader.Enabled {
+			var initialPayloadMap map[string]interface{}
+			err := json.Unmarshal(handler.initialPayload, &initialPayloadMap)
+			if err != nil {
+				requestLogger.Error("Error parsing initial payload: %v", zap.Error(err))
+				handler.writeErrorMessage(requestID, err)
```
jensneuse, line 340: should we do `handler.Close()` here?

router/core/websocket.go
```diff
@@ -308,6 +311,44 @@ func (h *WebsocketHandler) handleUpgradeRequest(w http.ResponseWriter, r *http.R
...
+				return
+			}
+			jwtToken, ok := initialPayloadMap[fromInitialPayloadConfig.Key].(string)
+			if !ok {
+				requestLogger.Error("JWT token is not a string")
```
jensneuse, line 345: should we do `handler.Close()` here?

jensneuse (changes requested):

jensneuse:
@alexandra-c overall looks good to me, just some minor things

router/core/websocket.go
alexandra-c: yes, you're right, since the initial payload or the token is invalid I don't see why the web socket should remain open.

alexandra-c:
> @alexandra-c overall looks good to me, just some minor things

thanks! I implemented your suggestion, can we run the test workflows again?

router/core/websocket.go
jensneuse: I'd also add writing an error: `handler.writeErrorMessage(requestID, err)`

jensneuse: Can you add a test that triggers this code path and expects an error?

jensneuse:
> > @alexandra-c overall looks good to me, just some minor things
> 
> thanks! I implemented your suggestion, can we run the test workflows again?

I found one more thing. Can you add this? Then I'll run it again.

router/pkg/authentication/jwks_token_decoder.go
```diff
@@ -0,0 +1,43 @@
...
+func (j *jwksTokenDecoder) Decode(tokenString string) (Claims, error) {
+	token, err := jwt.Parse(tokenString, j.jwks.Keyfunc)
+	if err != nil {
+		return nil, fmt.Errorf("could not validate token: %w", err)
+	}
+	claims := token.Claims.(jwt.MapClaims)
+	return Claims(claims), nil
+}
+
+func NewJwksTokenDecoder(url string, refreshInterval time.Duration) (tokenDecoder, error) {
```
StarpTech, line 31: `tokenDecoder` should be public.

router/pkg/authentication/jwks_token_decoder.go
```diff
@@ -0,0 +1,43 @@
...
+
+	jwks, err := keyfunc.Get(url, keyfunc.Options{
```
StarpTech, line 33: This starts a goroutine which needs to be cleaned up `jwks.EndBackground()` across graph hot reloads and server shutdown. Implement a close() method.

router/pkg/authentication/initial_payload_authenticator.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+	secretKey := strings.ToLower(a.key)
+	for key, tokenString := range initialPayloadMap {
+		if strings.ToLower(key) == secretKey {
+			authorization, ok := tokenString.(string)
+			if !ok {
+				errs = errors.Join(errs, fmt.Errorf("JWT token is not a string"))
+				continue
+			}
+			for _, prefix := range a.headerValuePrefixes {
```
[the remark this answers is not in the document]
alexandra-c: I simply mirrored the behavior from the HTTP header authenticator, in my opinion both should utilize the same 'auth.JWKS.HeaderValuePrefixes' configuration. I don't see any reason to introduce a separate configuration for the initial payload authentication prefix or why they should be different.

router/pkg/config/config.schema.json
```diff
@@ -323,6 +323,47 @@
...
+                  "properties": {
+                    "enabled": {
+                      "type": "boolean",
+                      "description": "This configuration indicates if the Router will export the JWT (JSON Web Token) to the client request header. The default value is 'false'.",
+                      "default": true
+                    },
+                    "header_key": {
+                      "type": "string",
+                      "description": "The name of the header property that will have the JWT (JSON Web Token) value. The default value is 'Authorization'",
+                      "default": "Authorization"
```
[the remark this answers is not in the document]
alexandra-c: Sorry, I don't quite understand what do you refer to by subscription deduplication logic.

router/core/websocket.go
alexandra-c: Actually, looking closer, this code never gets executed, the websocketInitialPayloadAuthenticator already validates the initial payload and the token itself, in its `Authenticate` function. If either of them is not valid it will log the error(s), send an error message and then closes the websocket, [here is the code](https://github.com/osstotalsoft/cosmo/blob/4e9b5784666688e919c1edb219788c96e94675c7/router/core/websocket.go#L322-L330). 

Here is the [payload validation check](https://github.com/osstotalsoft/cosmo/blob/4e9b5784666688e919c1edb219788c96e94675c7/router/pkg/authentication/initial_payload_authenticator.go#L39-L43) and here the [token validation check](https://github.com/osstotalsoft/cosmo/blob/4e9b5784666688e919c1edb219788c96e94675c7/router/pkg/authentication/initial_payload_authenticator.go#L39-L43)

If it's ok with you, I would remove [both "if" statements](https://github.com/osstotalsoft/cosmo/blob/4e9b5784666688e919c1edb219788c96e94675c7/router/core/websocket.go#L336-L349) from `handleUpgradeRequest`.  What do you think?

router/pkg/config/config.schema.json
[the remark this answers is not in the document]
StarpTech: You couldn't sorry. If you step into `AsyncResolveGraphQLSubscription` you see that we built a unique request ID. This ID is used to deduplicate connections to the subgraphs.

At line https://github.com/wundergraph/cosmo/blob/5461bb5a529decd51a1b22be0a5301936b8ad392/router/core/factoryresolver.go#L317 we build the list of header to respect for deduplication. We pass it to `graphql_datasource.NewConfiguration` as part of the subscription datasource. The graphql datasource in the engine will then ensure to respect this list when calculating the unique ID. When we allow to forward the data through a custom header, we need to add this header to the list.

router/pkg/authentication/initial_payload_authenticator.go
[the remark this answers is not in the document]
StarpTech: Ok, thanks for clarification.

router/core/websocket.go
```diff
@@ -308,6 +311,35 @@ func (h *WebsocketHandler) handleUpgradeRequest(w http.ResponseWriter, r *http.R
...
+			handler.writeErrorMessage(requestID, err)
+			handler.Close()
+			return
+		}
+		handler.r = validatedReq
+
+		// Export the token from the initial payload to the request header
+		if fromInitialPayloadConfig.ExportToken.Enabled {
+			var initialPayloadMap map[string]interface{}
+			json.Unmarshal(handler.initialPayload, &initialPayloadMap)
```
StarpTech, line 337: The previous discussion was closed without a comment. Why is the err not handled?

StarpTech, line 337: Please use `"github.com/goccy/go-json"` for json handling.

router/core/websocket.go
```diff
@@ -308,6 +311,35 @@ func (h *WebsocketHandler) handleUpgradeRequest(w http.ResponseWriter, r *http.R
...
+			jwtToken, _ := initialPayloadMap[fromInitialPayloadConfig.Key].(string)
```
StarpTech, line 338: Please handle the type assertion gracefully.

StarpTech (changes requested):

router/pkg/config/config.schema.json
```diff
@@ -323,6 +323,47 @@
...
```
StarpTech, line 360: Why Uppercase? lowercase is defacto standard in json.

router/core/websocket.go
alexandra-c: It was discussed here https://github.com/wundergraph/cosmo/pull/918#discussion_r1709334034 , the type assertion is already handled in the previous lines of code.

router/pkg/config/config.schema.json
StarpTech: So it already existed? If so, I'm fine with it to be backwards compatible.

router/core/websocket.go
StarpTech: The comment points to old code. Please check for the error. We encourage defensive programming and handling every possible error.

router/pkg/config/config.schema.json
```diff
@@ -323,6 +323,47 @@
...
+                  "description": "The name of the property holding the JWT (JSON Web Token) value. The default value is 'Authorization'"
+                },
+                "enabled": {
+                  "type": "boolean",
+                  "description": "This configuration indicates if the Router should look for the JWT (JSON Web Token) in the initial payload of the WebSocket Connection. The default value is 'false'.",
+                  "default": false
+                },
+                "export_token": {
+                  "type": "object",
+                  "description": "The configuration used to export the JWT (JSON Web Token) to the client request header.",
```
StarpTech, line 349: Please add a note here that when a different value than "Authorization" is used, the header needs to be added to the list of propagated headers.

router/pkg/config/config.schema.json
alexandra-c: Yes, 'Authorization' is used as a default value in all the configuration in general.

router/core/websocket.go
alexandra-c: Ok, I reverted to the old code.

router/pkg/authentication/jwks_token_decoder.go
alexandra-c: Can you give me a few more details regarding to where exactly the close() method should be called? I don't know exactly where the hot reloads or the server shutdown code is. 

Is The Close() method ok like this? 
https://github.com/wundergraph/cosmo/blob/a5653293018df745d9f9b99bf98f7b2d6fde9392/router/pkg/authentication/jwks_token_decoder.go#L45-L47

router/pkg/config/config.schema.json
StarpTech: Ok, thanks

router/pkg/authentication/jwks_token_decoder.go
StarpTech: The decoder is used router wide correct? In that case, you only need to call it [here](https://github.com/wundergraph/cosmo/blob/main/router/core/router.go#L1103).

router/pkg/authentication/jwks_token_decoder.go
```diff
@@ -0,0 +1,47 @@
+package authentication
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/MicahParks/keyfunc/v2"
```
StarpTech, line 7: Did you consider upgrading to v3?

https://github.com/MicahParks/keyfunc?tab=readme-ov-file#additional-features

StarpTech (changes requested):

router/pkg/authentication/jwks_token_decoder.go
alexandra-c: no, I'll upgrade to v3.

alexandra-c: I gave it a shot, but it turned out to be more complicated than expected. Do you still want me to do the upgrade?

router/pkg/authentication/jwks_token_decoder.go
alexandra-c: done

router/pkg/authentication/jwks_token_decoder.go
StarpTech: I'm ok with v2 if it well tested.

StarpTech (approved):
Great work and endurance 😄 Please write a few lines about the feature including edge cases. We need this to extend the docs. Alternatively, you can also contribute them [here](https://github.com/wundergraph/cosmo-docs/tree/main/router/subscriptions)

alexandra-c:
> Great work and endurance 😄 Please write a few lines about the feature including edge cases. We need this to extend the docs. Alternatively, you can also contribute them [here](https://github.com/wundergraph/cosmo-docs/tree/main/router/subscriptions)

Thank you! This was quite a journey 😁. It was actually my first time coding in Go—I even took a quick workshop just to tackle this PR. 😄 
I'll follow up with another PR to update the documentation as well.

alexandra-c:
Hey @jensneuse 👋, could you take a look again to see if any further adjustments are needed? Thanks!

router-tests/websocket_test.go
```diff
@@ -272,6 +280,165 @@ func TestWebSockets(t *testing.T) {
...
+			xEnv.WaitForConnectionCount(0, time.Second*5)
+			require.NoError(t, conn.Close())
+		})
+	})
+	t.Run("subscription with authorization via initial payload invalid token with reject", func(t *testing.T) {
```
jensneuse, line 392: it says "with reject" but reject is false. did you mean "without reject"?

router/pkg/config/config.schema.json
```diff
@@ -323,6 +323,47 @@
...
+              "type": "object",
+              "description": "The configuration used to tell the Router to look for the JWT (JSON Web Token) in the initial payload of the WebSocket Connection.",
+              "additionalProperties": false,
+              "properties": {
+                "key": {
+                  "type": "string",
+                  "default": "Authorization",
+                  "description": "The name of the property holding the JWT (JSON Web Token) value. The default value is 'Authorization'"
+                },
+                "enabled": {
```
jensneuse, line 342: I'd move enabled above key.

jensneuse (approved):

jensneuse:
> Hey @jensneuse 👋, could you take a look again to see if any further adjustments are needed? Thanks!

Hey @alexandra-c! Sorry that it took a bit of time.
Well done, great work!

I've added 2 minor comments which you could quickly address while merging main.
Give me a ping and we can run tests + merge.

alexandra-c:
> > Hey @jensneuse 👋, could you take a look again to see if any further adjustments are needed? Thanks!
> 
> Hey @alexandra-c! Sorry that it took a bit of time. Well done, great work!

Thank you! 😊

> I've added 2 minor comments which you could quickly address while merging main. Give me a ping and we can run tests + merge.

I updated the branch and included the small changes you suggested. 👍

PR #920, opened 2024-07-05
chore: add test case to demo to test throughput with big data

jensneuse:


devsergiy (approved):

router/core/websocket.go
```diff
@@ -776,7 +776,7 @@ func (h *WebSocketConnectionHandler) executeSubscription(msg *wsproto.Message, i
 			h.logger.Warn("Resolving GraphQL response", zap.Error(err))
 			buf := pool.GetBytesBuffer()
```
StarpTech, line 777: This can be removed.

router/core/websocket.go
```diff
@@ -786,7 +786,7 @@ func (h *WebSocketConnectionHandler) executeSubscription(msg *wsproto.Message, i
 			h.logger.Warn("Resolving GraphQL subscription", zap.Error(err))
 			buf := pool.GetBytesBuffer()
```
StarpTech, line 787: Here as well

router/pkg/config/config.go
```diff
@@ -210,7 +210,7 @@ type EngineExecutionConfiguration struct {
 	EnableSingleFlight                     bool                     `default:"true" envconfig:"ENGINE_ENABLE_SINGLE_FLIGHT" yaml:"enable_single_flight"`
 	EnableRequestTracing                   bool                     `default:"true" envconfig:"ENGINE_ENABLE_REQUEST_TRACING" yaml:"enable_request_tracing"`
 	EnableExecutionPlanCacheResponseHeader bool                     `default:"false" envconfig:"ENGINE_ENABLE_EXECUTION_PLAN_CACHE_RESPONSE_HEADER" yaml:"enable_execution_plan_cache_response_header"`
-	MaxConcurrentResolvers                 int                      `default:"1024" envconfig:"ENGINE_MAX_CONCURRENT_RESOLVERS" yaml:"max_concurrent_resolvers,omitempty"`
```
StarpTech, line 213: What's the reason behind this change? Should we add the learning to the docs?

StarpTech (changes requested):

router/pkg/config/config.go
jensneuse: I have measured extensively. Above 32, it's usually higher memory usage but slower RPS and slower P95.
I'll write a guide on performance. I've got a ticket for it already.

StarpTech: Awesome!

router/core/graphql_handler.go
```diff
@@ -292,6 +280,7 @@ func (h *GraphQLHandler) WriteError(ctx *resolve.Context, err error, res *resolv
 	case errorTypeUnauthorized:
 		response.Errors[0].Message = "Unauthorized"
 		if h.authorizer.HasResponseExtensionData(ctx) {
+			buf := bytes.NewBuffer(make([]byte, 0, 1024))
```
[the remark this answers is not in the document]
jensneuse: We're not using it enough to justify it. It's not a hot path. We could, but I don't think it makes a big difference.

StarpTech (approved):
LGTM

PR #923, opened 2024-07-05
fix: router config hot reload

StarpTech:
## TODO


router/core/websocket.go
```diff
@@ -110,14 +120,6 @@ func NewWebsocketMiddleware(ctx context.Context, opts WebsocketMiddlewareOptions
 			handler.forwardQueryParamsConfig.withStaticAllowList = len(handler.forwardQueryParamsConfig.staticAllowList) > 0
 			handler.forwardQueryParamsConfig.withRegexAllowList = len(handler.forwardQueryParamsConfig.regexAllowList) > 0
 		}
-		handler.handlerPool = pond.New(
```
StarpTech, line 113: @jensneuse Not sure, why we have created a worker pool per websocket request. It seems overkill.

router/internal/cdn/persisted_operation.go
```diff
@@ -224,3 +224,7 @@ func NewPersistentOperationClient(endpoint string, token string, opts Persistent
 		},
 	}, nil
 }
+
+func (cdn *PersistentOperationClient) Close() {
```
jensneuse, line 228: Can you rename it to "PersistedOperationClient"?
This name is just wrong.. (I know you didn't chose it)

router/core/websocket.go
jensneuse: Actually I think we should move the complete handler init up.
We're also creating the ePoll instance within the middleware loop which seems odd.

.run/router.run.xml
```diff
@@ -2,7 +2,6 @@
   <configuration default="false" name="router" type="GoApplicationRunConfiguration" factoryName="Go Application">
     <module name="cosmo" />
     <working_directory value="$PROJECT_DIR$/router" />
-    <go_parameters value="-race" />
```
jensneuse, line 5: Why remove race detector?

router/core/server.go
```diff
@@ -167,6 +170,14 @@ func (s *server) buildMux(ctx context.Context,
 		planCache = NewNoopExecutionPlanCache()
 	}
 
+	// Close the plan cache when the context is done.
+	// This is happening after the server is shutdown.
+	// This will stop all goroutines and free up resources
+	go func() {
+		<-ctx.Done()
+		planCache.Close()
```
jensneuse, line 178: We should also clean up and close the ePoll instance here, right?
What about WebSockets in general? If we update the config, it's possible that the origins can no longer serve the old contract. Should we terminate and cleanup all WebSockets, independent of ePoll or normal connection handling.

jensneuse (changes requested):

router/internal/cdn/persisted_operation.go
StarpTech: Done

router/core/server.go
StarpTech: Epoll is closed on context cancellation (introduced in this PR) and cleaned up in a defer.

.run/router.run.xml
StarpTech: Reverted after discussion.

router/core/websocket.go
StarpTech: Great observation. Fixed.

router/core/server.go
StarpTech: In the future, we should wait until all resources has been shutdown before we restart the server. For now this is done asynchronously but sufficient.

router/core/graph_server.go
```diff
@@ -69,10 +79,169 @@ type (
...
+		return br.NewWriterLevel(w, level)
+	})
+
+	/**
+	* A group where we can selectively apply middlewares to the graphql endpoint
+	 */
+	httpRouter.Group(func(cr chi.Router) {
+
+		// We are applying it conditionally because brotli compressing the 3MB playground is very slow
+		cr.Use(middleware.Compress(5, CustomCompressibleContentTypes...))
```
jensneuse, line 213: Can we disable brotli by default?

router/core/http_server.go
```diff
@@ -0,0 +1,134 @@
...
+		mu:          sync.RWMutex{},
+		healthcheck: opts.healthcheck,
+		baseURL:     opts.baseURL,
+	}
+
+	httpServer.Handler = http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// Multiple requests can read the handler at the same time, but only one goroutine can write it.
+		// When swapping the graph server there might be in-flight requests that are still being processed
+		// but this is tolerable because we are waiting for them to finish before shutting down the old server.
+		n.mu.RLock()
```
jensneuse, line 62: Have you thought about this?

```go
n.mu.RLock()
handler := n.handler
n.mu.RUnlock()
handler.ServeHTTP(w, r)
```

This should be memory safe but allows to swap the handler even if traffic is being served on the old handler.
With your approach, it will not be possible to acquire a Lock() while the RLock() was not released, which can take a while.

jensneuse (changes requested):

router/core/graph_server.go
StarpTech: I would like to remove support in a separate PR. The current implementation is too slow to be used in production.

router/core/http_server.go
StarpTech: Your assumption doesn't seem correct. The std library says:

> If any goroutine calls Lock while the lock is already held by one or more readers, concurrent calls to RLock will block until the writer has acquired (and released) the lock, to ensure that the lock eventually becomes available to the writer. Note that this prohibits recursive read-locking

When calling Lock() we won't let any new requests obtain a read lock until the write lock was released. Existing RLock() aren't affected. Because we just change a variable, the impact is minimal.

StarpTech: Example https://go.dev/play/p/LDyzDIfCoNx

StarpTech: Fake your own tests 😄 I was wrong. Good catch!

jensneuse (approved):

Issue #907, opened 2024-07-02
"wgc subgraph list' only logs 8192 characters


alumpe:
### Component(s)

cli

### Component version

0.57.3

### wgc version

0.57.3

### What happened?

We are using the wgc subgraph list command in one of our internal tools. We haven't had any issues until we tried to run the tool on the MacBook of my colleague. Since we are parsing the raw JSON output of `npx wgc subgraph list -n ${namespace} -r` we always got the error `Unexpected token in JSON at position 8192.`. We dug deeper and it seems the bug lies within the asynchronous behavior of `console.log` in [these 2 lines](https://github.com/wundergraph/cosmo/blob/main/cli/src/commands/subgraph/commands/list.ts#L63-L64) of code. We inserted a `setTimeout` and also tested it with `fs.writeSync(1, JSON.stringify(xxx))` and this way it correctly outputs the whole JSON before exiting the process. We can only assume that the reason for this bug is the async behavior of `console.log` for large outputs and the immediate `process.exit()`.

If you have a suggested way of fixing this issue, perhaps using `fs.writeSync()`, I'd be willing to create a PR.

## Environment
Laptop model: MacBook Pro M3 FN4CG9RWH1
OS: Sonoma 14.5
Package Manager: npm

alumpe:
Hey @StarpTech, I represent [gastromatic](https://github.com/gastromatic), I believe some of my colleagues already communicated closely with you in the past. If I find time this week, I will make a PR. I thought about creating a small wrapper function a la `logToConsole(text) => {}` that can be used in every place where syncronous logging is important.

StarpTech:
Hi @alumpe, could you provide a reproducible example? I can't test the asynchronous behaviour. I generated a huge JSON object with https://json-generator.com/, printed it and parsed it with jq. It never failed. I see no reason why the cli should behave differently in this case. I'm fine with your solution but before we need a way to test it.

```js
const a = ...
console.log(JSON.stringify(a));
process.exit(0);
```
```bash
node test.js | jq
```

I'm on a Mac as well.

AndreasZeissner:
Hi @alumpe, 

is this still relevant for you? 
I tried to reproduce the issue even by fetching and pushing data through console.log and piping it's outputs.
Unfortunately no chance.  

```
Unexpected token in JSON at position 8192.
``` 

[...]

PR #930, opened 2024-07-10
fix(cli): write large data synchronously to console before exiting process

alumpe:
## Motivation and Context

**Problem:** We are parsing the raw JSON output of `npx wgc subgraph list -n ${namespace} -r` and we got the error `Unexpected token in JSON at position 8192.`. We dug deeper and it seems the bug lies within the asynchronous behavior of `console.log` and its different implementations across different operating systems (bug was happening on MacOS).

This PR fixes this problem, further described in https://github.com/wundergraph/cosmo/issues/907, by using a synchronous method of writing large data to the console.
I changed `console.log` to `writeToConsoleSync` helper function on several places where raw or large data was written to console.

Feel free to comment or suggest any changes.

## TODO


cli/src/utils.ts
```diff
@@ -288,3 +288,10 @@ export const validateSubscriptionProtocols = ({
     }
   }
 };
+
+/**
+ * Synchronously writes to the console.
+ */
+export const writeToConsole = (message: string) => {
```
StarpTech, line 295: Please express in the same that the operation is synchronous e.g. `writeToConsoleSync`

cli/src/utils.ts
```diff
@@ -288,3 +288,10 @@ export const validateSubscriptionProtocols = ({
     }
   }
 };
+
+/**
+ * Synchronously writes to the console.
```
StarpTech, line 293: Add a comment why this api is needed.

StarpTech (changes requested):

PR #937, opened 2024-07-13
feat: improve JSON parsing & merging in the engine

jensneuse:


router/core/transport.go
```diff
@@ -189,68 +190,42 @@ func (ct *CustomTransport) allowSingleFlight(req *http.Request) bool {
...
 			return nil, err
 		}
-		buf.Reset()
+		buf := getBuffer()
+		defer releaseBuffer(buf)
 		_, err = buf.ReadFrom(res.Body)
 		if err != nil {
 			return nil, err
 		}
+		cp := make([]byte, buf.Len())
```
StarpTech, line 224: When we need to alloc the buf anyway, we can also skip the pool.

router/core/transport.go
```diff
@@ -189,68 +190,42 @@ func (ct *CustomTransport) allowSingleFlight(req *http.Request) bool {
...
-	v, err, shared := ct.sf.Do(key, func() (interface{}, error) {
+	v, err, shared := ct.sf.Do(ct.singleFlightKey(req), func() (interface{}, error) {
 		res, err := ct.roundTripper.RoundTrip(req)
 		if err != nil {
 			return nil, err
 		}
-		buf.Reset()
+		buf := getBuffer()
+		defer releaseBuffer(buf)
 		_, err = buf.ReadFrom(res.Body)
```
StarpTech, line 220: Luckily, file uploads are mutations and we don't allow sf. Could you add a comment about this here?

StarpTech (changes requested):

router/core/transport.go
```diff
@@ -280,6 +255,30 @@ func (ct *CustomTransport) roundTripSingleFlight(req *http.Request) (*http.Respo
 	return res, nil
 }
 
+func (ct *CustomTransport) singleFlightKey(req *http.Request) string {
+	keyGen := pool.Hash64.Get()
+	defer pool.Hash64.Put(keyGen)
+
+	if bodyHash, ok := httpclient.BodyHashFromContext(req.Context()); ok {
+		_, _ = keyGen.Write([]byte(strconv.FormatUint(bodyHash, 10)))
```
StarpTech, line 263: Use keygen.WriteString

StarpTech (changes requested):

StarpTech (approved):
LGTM

PR #941, opened 2024-07-17
fix: update text snippet in messages_en.properties

wunderhawk:
Added "managed service" to match the exact legal name of the terms document in the consent modal.


StarpTech (approved):
LGTM

PR #949, opened 2024-07-19
feat: implement normalization cache for non-persisted operations

jensneuse:


router/core/operation_processor.go
```diff
@@ -227,6 +256,8 @@ func (o *OperationKit) Parse(ctx context.Context, clientInfo *ClientInfo) error
 		if err != nil {
 			return err
 		}
+		// it's important to make a copy of the persisted operation data, because it's used in the cache
+		// we might modify it later, so we don't want to modify the cached data
 		o.parsedOperation.Request.Query = string(persistedOperationData)
```
StarpTech, line 261: `unsafebytes.BytesToString(key)`

router/pkg/config/config.schema.json
```diff
@@ -1271,7 +1275,7 @@
         },
         "execution_plan_cache_size": {
           "type": "integer",
-          "default": 10000,
+          "default": 1024,
```
StarpTech, line 1278: Changing default can be unpredictable for customers. How likely is it that this can have an impact?

StarpTech (changes requested):

router/core/operation_processor.go
jensneuse: Did you read the comment? above :D

router/pkg/config/config.schema.json
jensneuse: 1024 plans cached should be more than enough for most setups. I don't expect any negative impacts.

StarpTech (approved):
LGTM

PR #950, opened 2024-07-21
fix: optimize max concurrency for medium, high load, don't retry on 2xx status codes

StarpTech:
## Motivation and Context

`ENGINE_MAX_CONCURRENT_RESOLVERS` has been chosen too low (32). The latency for medium and high load scenarios is increased a lot because max concurrency is implemented as a semaphore and blocks new requests. This PR increase the limit to `256`.

### Test case

The settings has been validated with the following scenario.

400 VUs (Virtual users) over 30 seconds

- ENGINE_MAX_CONCURRENT_RESOLVERS: 256
- Artificial subgraph latency: 200ms
- Subgraph response size: 200KB

Memory consumption: ~300MB

Additionally, I've found out that we retry subgraph requests when the subgraph returns a different code than 2xx than 200 OK.

## TODO

### DIFF
--- a/router/internal/retrytransport/retry_transport.go
+++ b/router/internal/retrytransport/retry_transport.go
@@ -59,8 +59,8 @@ func NewRetryHTTPTransport(roundTripper http.RoundTripper, retryOptions RetryOpt
 func (rt *RetryHTTPTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 
 	resp, err := rt.RoundTripper.RoundTrip(req)
-	// Short circuit if the request was successful
-	if err == nil && resp.StatusCode == http.StatusOK {
+	// Short circuit if the request was successful.
+	if err == nil && isResponseOK(resp) {
 		return resp, nil
 	}
 
@@ -92,13 +92,19 @@ func (rt *RetryHTTPTransport) RoundTrip(req *http.Request) (*http.Response, erro
 		resp, err = rt.RoundTripper.RoundTrip(req)
 
 		// Short circuit if the request was successful
-		if err == nil && resp.StatusCode == http.StatusOK {
+		if err == nil && isResponseOK(resp) {
 			return resp, nil
 		}
 
 	}
 
 	return resp, err
+}
+
+func isResponseOK(resp *http.Response) bool {
+	// Ensure we don't wait for no reason when subgraphs don't behave
+	// spec-compliant and returns a different status code than 200.
+	return resp.StatusCode >= http.StatusOK && resp.StatusCode < http.StatusMultipleChoices
 }
 
 func IsRetryableError(err error, resp *http.Response) bool {
--- a/router/internal/retrytransport/retry_transport.go
+++ b/router/internal/retrytransport/retry_transport.go
@@ -104,7 +104,7 @@ func (rt *RetryHTTPTransport) RoundTrip(req *http.Request) (*http.Response, erro
 func isResponseOK(resp *http.Response) bool {
 	// Ensure we don't wait for no reason when subgraphs don't behave
 	// spec-compliant and returns a different status code than 200.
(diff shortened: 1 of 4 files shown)


router/internal/retrytransport/retry_transport.go
```diff
@@ -101,6 +101,12 @@ func (rt *RetryHTTPTransport) RoundTrip(req *http.Request) (*http.Response, erro
 	return resp, err
 }
 
+func isResponseOK(resp *http.Response) bool {
+	// Ensure we don't wait for no reason when subgraphs don't behave
+	// spec-compliant and returns a different status code than 200.
+	return resp.StatusCode >= http.StatusOK && resp.StatusCode < http.StatusMultipleChoices
```
Aenimus, line 107: Can you add a comment that this is all status 2XX?

Maybe a better name of the function is `isResponse2XX`?

router/pkg/config/config.go
```diff
@@ -211,7 +211,7 @@ type EngineExecutionConfiguration struct {
 	EnableSingleFlight                     bool                     `default:"true" envconfig:"ENGINE_ENABLE_SINGLE_FLIGHT" yaml:"enable_single_flight"`
 	EnableRequestTracing                   bool                     `default:"true" envconfig:"ENGINE_ENABLE_REQUEST_TRACING" yaml:"enable_request_tracing"`
 	EnableExecutionPlanCacheResponseHeader bool                     `default:"false" envconfig:"ENGINE_ENABLE_EXECUTION_PLAN_CACHE_RESPONSE_HEADER" yaml:"enable_execution_plan_cache_response_header"`
-	MaxConcurrentResolvers                 int                      `default:"32" envconfig:"ENGINE_MAX_CONCURRENT_RESOLVERS" yaml:"max_concurrent_resolvers,omitempty"`
+	MaxConcurrentResolvers                 int                      `default:"256" envconfig:"ENGINE_MAX_CONCURRENT_RESOLVERS" yaml:"max_concurrent_resolvers,omitempty"`
```
Aenimus, line 214: This seems like an extreme change?

StarpTech: Please read the PR description.

router/internal/retrytransport/retry_transport.go
StarpTech: I'm going to use numbers.

Aenimus: For me, response OK is 200 and 200 only. The name is misleading. I'd choose `isResponse2XX`, which is what the function actually does. But I won't push it further. :D

Aenimus:
@StarpTech resolver test still looking for 32

Aenimus (approved):
LGTM

router/internal/retrytransport/retry_transport.go
```diff
@@ -59,8 +59,8 @@ func NewRetryHTTPTransport(roundTripper http.RoundTripper, retryOptions RetryOpt
 func (rt *RetryHTTPTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 
 	resp, err := rt.RoundTripper.RoundTrip(req)
-	// Short circuit if the request was successful
-	if err == nil && resp.StatusCode == http.StatusOK {
+	// Short circuit if the request was successful.
+	if err == nil && isResponseOK(resp) {
```
jensneuse, line 63: Do we want to retry 4xx responses? e.g. bad request, unauthorized, etc..? (I don't think so)

jensneuse (approved):

router/internal/retrytransport/retry_transport.go
StarpTech: Correct, only on network and retryable 4xx and 5xx codes. List: https://github.com/wundergraph/cosmo/blob/cbc103d2e2afe32c30d925d3ab04ddf72e614b99/router/internal/retrytransport/retry_transport.go#L26

PR #956, opened 2024-07-23
fix: variable list coercion with normalization cache

jensneuse:


router/core/graphql_prehandler.go
```diff
@@ -242,10 +249,17 @@ func (h *PreHandler) Handler(next http.Handler) http.Handler {
 
 				writeOperationError(r, w, requestLogger, err)
 				h.releaseBodyReadBuffer(buf)
+				operationReadSpan.End()
 				return
 			}
 		}
 
+		operationReadSpan.SetAttributes(
+			otel.WgGraphQLRequestSizeBytes.Int(len(body)),
```
StarpTech, line 258: There is standardized OTEL attribute for request content bytes

```go
semconv17 "go.opentelemetry.io/otel/semconv/v1.17.0"
semconv17.HTTPResponseContentLength()
```

router/core/graphql_prehandler.go
```diff
@@ -242,10 +249,17 @@ func (h *PreHandler) Handler(next http.Handler) http.Handler {
 
 				writeOperationError(r, w, requestLogger, err)
 				h.releaseBodyReadBuffer(buf)
+				operationReadSpan.End()
```
StarpTech, line 252: Attributes aren't added once the span has ended. We need to preserve as much data as possible.

router/core/graphql_prehandler.go
```diff
@@ -200,6 +200,11 @@ func (h *PreHandler) Handler(next http.Handler) http.Handler {
 		// AFTER we're done with body (retrieved from parser.ReadBody())
 		buf := h.getBodyReadBuffer(r.ContentLength)
 
+		_, operationReadSpan := h.tracer.Start(r.Context(), "Operation - Read",
```
StarpTech, line 203: I'd trace a regular query from a file upload request. This makes investigations easier.

router/core/graphql_prehandler.go
```diff
@@ -242,10 +249,17 @@ func (h *PreHandler) Handler(next http.Handler) http.Handler {
...
 				writeOperationError(r, w, requestLogger, err)
 				h.releaseBodyReadBuffer(buf)
+				operationReadSpan.End()
 				return
 			}
 		}
 
+		operationReadSpan.SetAttributes(
+			otel.WgGraphQLRequestSizeBytes.Int(len(body)),
+			otel.WgMultipartFilesCount.Int(len(files)),
```
StarpTech, line 259: Please check if there is already a known otel attribute for that. If not we can invent it but we should keep it free from `wg` prefix because it is HTTP related. Maybe `http.request.files_count`

StarpTech (changes requested):

router/core/graphql_prehandler.go
jensneuse: This is not exactly accurate I was thinking. Independent of multipart or not, we don't want to know the http content length but really just the size of the GraphQL Request, which is also not exactly the body, because of multipart, but only the "GraphQL Request" part.

router/core/graphql_prehandler.go
jensneuse: We're returning after this (error case). What would you do differently?

router/core/graphql_prehandler.go
StarpTech: We already know how much bytes are read from the root span.

StarpTech (approved):
LGTM

PR #963, opened 2024-07-25
feat(graphqlmetrics): enable prometheus metrics

AndreasZeissner:
## Motivation and Context

This PR introduces exposing metrics on the graphqlmetrics service.
The prometheus export can be enabled by using the following environment variables:

`PROMETHEUS_ENABLED`: enable or disable exposing metrics defaults to false
`PROMETHEUS_LISTEN_ADDR`: the listen address for the prometheus server defaults to 127.0.0.1:8088 to exposing metrics to the public bt accident
`PROMETHEUS_PATH`: the path to expose the metrics on defaults to /metrics

## TODO


graphqlmetrics/core/server.go
```diff
@@ -39,16 +44,47 @@ func NewServer(metricsService graphqlmetricsv1connect.GraphQLMetricsServiceHandl
...
-	}))
+	})
+
+	if s.metricConfig.OpenTelemetry.Enabled {
+		tp, err := s.metricConfig.NewTracerProvider()
+		if err != nil {
+			s.logger.Error("Error creating tracing provider", zap.Error(err))
+		}
+		defer func() {
+			if err := tp.Shutdown(context.Background()); err != nil {
```
StarpTech, line 63: Shutdown logic should be done in `Shutdown` with a proper context.

graphqlmetrics/core/server.go
```diff
@@ -39,16 +44,47 @@ func NewServer(metricsService graphqlmetricsv1connect.GraphQLMetricsServiceHandl
...
 	mux.Handle(path, authenticate(s.jwtSecret, s.logger, handler))
 
+	if s.metricConfig.Prometheus.Enabled {
+		mp, registry, err := s.metricConfig.NewPrometheusMeterProvider()
+		if err != nil {
+			s.logger.Error("Failed to create Prometheus exporter", zap.Error(err))
+		}
+
+		defer func() {
+			if err := mp.Shutdown(context.Background()); err != nil {
```
StarpTech, line 80: Same here.

graphqlmetrics/core/server.go
```diff
@@ -84,6 +120,30 @@ func (s *Server) Shutdown(ctx context.Context) error {
 	return nil
 }
 
+func (s *Server) ShutdownPrometheusServer(ctx context.Context) error {
```
StarpTech, line 123: Call them in `Shutdown` and make them private.

graphqlmetrics/core/telemetry_test.go
```diff
@@ -0,0 +1,85 @@
...
+			wantErr: true,
+		},
+	}
+
+	for _, tt := range tests {
+		// as the server setup is all encapsulated within main,
+		// at least check, if the new code is covered and if the
+		// handlers for the prometheus server are registered
+		go func() {
+			if err := tt.svr.StartPrometheusServer(); err != nil {
```
StarpTech, line 64: Please add an integration test that actually test the server implementation. Additionally, we want to validate in the tests that proper metric datapoints has been exposed.

```go
    // 1. Setup the server
	svr := core.NewServer(
		ms,
	   core.WithMetrics(metricsConfig),
		core.WithListenAddr(cfg.ListenAddr),
	)

// 2. Make a request to the buf service
// 3. Validate metrics on the prometheus endpoint
```

graphqlmetrics/pkg/telemetry/metrics_prometheus.go
```diff
@@ -0,0 +1,79 @@
...
+	// Only available on Linux and Windows systems
+	registry.MustRegister(collectors.NewProcessCollector(collectors.ProcessCollectorOpts{}))
+
+	promExporter, err := otelprom.New(
+		otelprom.WithoutUnits(),
+		otelprom.WithRegisterer(registry),
+	)
+
+	if err != nil {
+		panic(err)
```
StarpTech, line 66: Don't use panic for err handling. Propagate the err so it is possible to handle it and to add more context.

graphqlmetrics/pkg/telemetry/tracing.go
```diff
@@ -0,0 +1,70 @@
...
+
+	"go.opentelemetry.io/otel/exporters/otlp/otlptrace"
+	"go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc"
+)
+
+const (
+	// WithEndpoint respects the following environment variables:
+	// OTEL_EXPORTER_OTLP_ENDPOINT
+	// OTEL_EXPORTER_OTLP_TRACES_ENDPOINT
+	DefaultGrpcTelemetryEndpoint = "cosmo-otel.wundergraph.com"
```
StarpTech, line 20: This endpoint is intended for customers routers. We should not give the impression that we can use it for platform monitoring.

graphqlmetrics/pkg/telemetry/tracing.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+func (c *Config) NewTracerProvider() (*sdktrace.TracerProvider, error) {
+	exporter, err := newTraceExporter()
+	if err != nil {
+		return nil, err
+	}
+
+	resources, err := resource.New(
+		context.Background(),
```
StarpTech, line 51: `context.Background()` is rarely used. In most of the cases, you should allow to pass a custom context to allow cancellations.

graphqlmetrics/pkg/telemetry/metrics_prometheus.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	mp := sdkmetric.NewMeterProvider(
+		sdkmetric.WithReader(promExporter),
+		sdkmetric.WithResource(resource.NewWithAttributes(
+			semconv.SchemaURL,
+			semconv.ServiceNameKey.String(c.Name),
```
StarpTech, line 73: Please also track the version with `	resource.WithAttributes(semconv.ServiceVersionKey.String(c.Version))`

graphqlmetrics/pkg/telemetry/tracing.go
```diff
@@ -0,0 +1,70 @@
...
+)
+
+const (
+	// WithEndpoint respects the following environment variables:
+	// OTEL_EXPORTER_OTLP_ENDPOINT
+	// OTEL_EXPORTER_OTLP_TRACES_ENDPOINT
+	DefaultGrpcTelemetryEndpoint = "cosmo-otel.wundergraph.com"
+)
+
+func buildResourceOptions() resource.Option {
```
StarpTech, line 23: We don't have to duplicate those information when they are already cover above with standardized attributes.

graphqlmetrics/pkg/telemetry/tracing.go
```diff
@@ -0,0 +1,70 @@
...
+func newOtlpTraceGrpcClient() otlptrace.Client {
+	return otlptracegrpc.NewClient(
+		otlptracegrpc.WithEndpoint(DefaultGrpcTelemetryEndpoint),
+		otlptracegrpc.WithCompressor("gzip"),
+	)
+}
+
+func newTraceExporter() (*otlptrace.Exporter, error) {
+	return otlptrace.New(
+		context.Background(),
```
StarpTech, line 39: Same here with the context.

graphqlmetrics/pkg/telemetry/metrics_prometheus.go
```diff
@@ -0,0 +1,79 @@
...
+		otelprom.WithRegisterer(registry),
+	)
+
+	if err != nil {
+		panic(err)
+	}
+
+	mp := sdkmetric.NewMeterProvider(
+		sdkmetric.WithReader(promExporter),
+		sdkmetric.WithResource(resource.NewWithAttributes(
```
StarpTech, line 71: Please attach those resource attribute presets as well

```go
	resource.WithProcessPID(),
		resource.WithOSType(),
		resource.WithTelemetrySDK(),
		resource.WithHost(),
```

graphqlmetrics/pkg/telemetry/metrics_prometheus.go
```diff
@@ -0,0 +1,79 @@
...
+			semconv.SchemaURL,
+			semconv.ServiceNameKey.String(c.Name),
+		)),
+	)
+
+	otel.SetMeterProvider(mp)
```
StarpTech, line 77: This will set a global meter. This can be an issue with concurrent tests. We only work with dedicated meters.

graphqlmetrics/core/server.go
StarpTech, line 1: Please add at least a metric that adds more information than the process stats. A good metric could be the total amount of requests for every customer.

StarpTech (changes requested):

graphqlmetrics/pkg/telemetry/tracing.go
AndreasZeissner: I would rather remove the tracing related implementations for now and bring them in, when tracing is integrated.

graphqlmetrics/core/server.go
AndreasZeissner: Would it be ok to use the same context as here as the server itself?

graphqlmetrics/core/telemetry_test.go
AndreasZeissner: As far as I see, in graphqlmetrics, there are only Integration Tests.

Is this ok to only have those here e.g. mocking the MetricsService seemed odd to me in the first place already to get some coverage here.

graphqlmetrics/core/server.go
AndreasZeissner: I moved the shutdown together, seems to work fine to reuse the context.

Let me know if this might be an issue.

graphqlmetrics/pkg/telemetry/tracing.go
AndreasZeissner: I removed tracing related references for now to split up metrics and tracing properly.

graphqlmetrics/pkg/telemetry/interceptor.go
```diff
@@ -0,0 +1,37 @@
...
+	WgOperationName     = attribute.Key("wg.operation.name")
+	WgOperationMethod   = attribute.Key("wg.operation.method")
+	WgOperationProtocol = attribute.Key("wg.operation.protocol")
+)
+
+func defaultAttributes(req connect.AnyRequest) []attribute.KeyValue {
+	var attributes []attribute.KeyValue
+
+	attributes = append(attributes, WgOperationName.String(req.Spec().Procedure))
+	attributes = append(attributes, WgOperationMethod.String(req.HTTPMethod()))
```
StarpTech, line 20: There are standardized attributes for basic http/net attributes.

graphqlmetrics/pkg/telemetry/config.go
```diff
@@ -0,0 +1,58 @@
...
+// Config represents the configuration for the agent.
+type Config struct {
+	// Name represents the service name for metrics. The default value is cosmo-router.
+	Name string
+	// Version represents the service version for metrics. The default value is dev.
+	Version string
+	// Prometheus includes the Prometheus configuration
+	Prometheus PrometheusConfig
+	// CustomPrometheusMetrics to be collected
+	CustomMetrics CustomMetrics
```
StarpTech, line 42: Is this for testing?

graphqlmetrics/core/server.go
```diff
@@ -32,21 +44,53 @@ func NewServer(metricsService graphqlmetricsv1connect.GraphQLMetricsServiceHandl
...
 	)
-	mux.Handle("/health", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+
+	healthHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		w.WriteHeader(http.StatusOK)
-	}))
+	})
+
+	mux.Handle("/health", healthHandler)
 	mux.Handle(path, authenticate(s.jwtSecret, s.logger, handler))
```
StarpTech, line 94: Unfortunately, I have seen this too late. Instead of using an interceptor we can track the metrics in a regular handler. We would also get access to the JWT which contains the organization and federated graph id. That would make the metrics much more insightful.

StarpTech (changes requested):

graphqlmetrics/core/telemetry_test.go
```diff
@@ -0,0 +1,283 @@
...
+				err := svr.Shutdown(ctx)
+				if err != nil {
+					t.Fatalf("Failed to shut down server: %v", err)
+				}
+			}()
+			time.Sleep(2 * time.Second)
+
+			endpoint := fmt.Sprintf("http://%s%s", tt.prom.ListenAddr, tt.prom.Path)
+			resp, err := http.Get(endpoint)
+
```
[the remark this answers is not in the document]
AndreasZeissner: Calling it earlier here is not possible as resp could be nil.
I could change the test and split it up in two tests. 

The intention here was to express the two possibilities, either there will be a response, or it will be nil as there was no connection possible.

graphqlmetrics/core/server.go
```diff
@@ -60,13 +104,24 @@ func (s *Server) bootstrap() {
 	}
 }
 
-func (s *Server) Start() error {
+func (s *Server) Start(stop context.CancelFunc) error {
```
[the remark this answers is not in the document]
AndreasZeissner: What would be the way to go here?
The cancel function is only called within the main servers go routine. Even when the prometheus listener dies the context of the main server is not cancelled, isn't it?

graphqlmetrics/pkg/telemetry/config.go
AndreasZeissner: I am not sure what you mean here.

I updated some comments, it's not reflected here in the PR view, I hope this makes more sense:

https://github.com/wundergraph/cosmo/commit/bd2a30ff4c577f373677a49888a398bddd66a961

graphqlmetrics/core/server.go
AndreasZeissner: Wouldn't it be possible to pass the parent context to the interceptor and the pull out the same information?

graphqlmetrics/core/telemetry_test.go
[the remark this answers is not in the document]
StarpTech: You're right. Let me rephrase.

```go
if err != nil {
// err handling
}
defer resp.Body.Close()

// response handling
```

graphqlmetrics/core/server.go
StarpTech: Yes, when you get access to the request context.

graphqlmetrics/core/telemetry_test.go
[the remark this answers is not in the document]
AndreasZeissner: The test case does not allow this as in once case resp can be nil and the error branch is a valid case here. 

"prometheus was not enabled and no response should have been able to be made and is therefore nil". 

I get your point, I moved the defer to the earliest point, where I know, there is something to close. I will let that sink in and think of splitting it up in two tests.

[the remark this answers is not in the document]
StarpTech: Ok

graphqlmetrics/core/server.go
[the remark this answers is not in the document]
StarpTech: Just don't pass the cancelFunc here and handle at the place where Start is called.

graphqlmetrics/pkg/telemetry/interceptor.go
AndreasZeissner: I updated them to use the GRPC Connect specific one, you can see this in the tests.
I would say this makes sense now.

graphqlmetrics/core/server.go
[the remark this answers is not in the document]
AndreasZeissner: Thanks.

graphqlmetrics/core/telemetry_test.go
```diff
@@ -0,0 +1,283 @@
...
+					Prometheus: tt.prom,
+				}))
+
+			go func() {
+				if err := svr.Start(stop); err != nil && !errors.Is(err, http.ErrServerClosed) {
+					t.Logf("failed starting main server")
+				}
+			}()
+
+			defer func() {
```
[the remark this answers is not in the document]
AndreasZeissner: Unfortunately this seems not to cleanup the server reliable. 

I would leave it like it is as it keeps the test run predictable even when restarting the server in between runs.

graphqlmetrics/core/server.go
AndreasZeissner: Works with the interceptor. 

The claims could be unset, this shouldn't happen as the  validation happens already before, but in case the claims are empty, I pass back a `INVALID_ARGUMENT`. Let me know about your opinion.

StarpTech: Sounds good

graphqlmetrics/core/telemetry_test.go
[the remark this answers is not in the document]
StarpTech: ok

StarpTech (approved):
LGTM

PR #971, opened 2024-07-28
feat: implement s3 provider for config and persistent operations

StarpTech:
## Motivation and Context

```yaml
storage_providers:
  s3:
  - id: "s3"
    endpoint: "localhost:10000"
    bucket: "cosmo"
    access_key: "key"
    secret_key: "secret"
    secure: false

persisted_operations:
  cache:
    size: 100MB
  storage:
    provider_id: s3
    object_prefix: "prefix/to/operations"

execution_config:
  storage:
    provider_id: s3
    object_path: "path/to/router/config/latest.json"
```

## TODO


router/core/router.go
```diff
@@ -751,6 +745,172 @@ func (r *Router) bootstrap(ctx context.Context) error {
...
+	}
+
+	for _, provider := range r.storageProviders.CDN {
+		if _, ok := cdnProviders[provider.ID]; ok {
+			return fmt.Errorf("duplicate cdn storage provider with id '%s'", provider.ID)
+		}
+		cdnProviders[provider.ID] = provider
+	}
+
+	if r.graphApiToken != "" {
```
jensneuse, line 765: do I need a graphApiToken if I'm just using S3?

router/core/router.go
```diff
@@ -751,6 +745,172 @@ func (r *Router) bootstrap(ctx context.Context) error {
...
+			})
+			if err != nil {
+				return err
+			}
+			client = c
+
+			r.logger.Info("Fetching persisted operations from CDN",
+				zap.String("providerID", provider.ID),
+			)
+		} else if provider, ok := s3Providers[r.persistedOperationsConfig.Storage.ProviderID]; ok {
```
jensneuse, line 783: Can you rewrite this without else if, e.g. using functions?

router/core/router.go
```diff
@@ -751,6 +745,172 @@ func (r *Router) bootstrap(ctx context.Context) error {
...
+			return err
+		}
+		r.persistedOperationClient = c
+	}
+
+	var client routerconfig.Client
+
+	if r.configPoller == nil && r.routerConfigPollerConfig != nil {
+
+		if provider, ok := cdnProviders[r.routerConfigPollerConfig.Storage.ProviderID]; ok {
```
jensneuse, line 843: can you rewrite the nested if if else, e.g. using functions?

router/internal/persistedoperation/cache.go
```diff
@@ -0,0 +1,31 @@
...
+)
+
+type OperationsCache struct {
+	// cache is the backing store for the in-memory cache. Note
+	// that if the cache is disabled, this will be nil
+	Cache *ristretto.Cache[string, []byte]
+}
+
+func (c *OperationsCache) key(clientName string, operationHash []byte) string {
+	return clientName + unsafebytes.BytesToString(operationHash)
```
jensneuse, line 19: I think it would be better to use uint64 as key, using xxhash to generate it.
You're concatenating clientName with an unsafe string, which allocates, but it's also dangerous because the operationHash must not be changed.

router/internal/persistedoperation/cache.go
```diff
@@ -0,0 +1,31 @@
...
+
+func (c *OperationsCache) Get(clientName string, operationHash string) []byte {
+	// Since we're returning nil when the item is not found, we don't need to
+	// check the return value from the cache nor the type assertion
+	item, _ := c.Cache.Get(c.key(clientName, unsafebytes.StringToBytes(operationHash)))
+	return item
+}
+
+func (c *OperationsCache) Set(clientName, operationHash string, operationBody []byte) {
+	c.Cache.Set(c.key(clientName, unsafebytes.StringToBytes(operationHash)), operationBody, int64(len(operationBody)))
```
jensneuse, line 30: I don't think it's a good idea to use an unsafe string as key.
Use xxhash and uint64 instead, which is safe and fast.

router/internal/persistedoperation/cdn/client.go
```diff
@@ -0,0 +1,171 @@
...
+		return nil, errors.New("could not read the response body. " + err.Error())
+	}
+
+	var po persistedoperation.PersistedOperation
+	err = json.Unmarshal(body, &po)
+	if err != nil {
+		return nil, err
+	}
+
+	return []byte(po.Body), nil
```
jensneuse, line 135: If po.Body is never written to and only being read, we could reduce allocs by converting it unsafe to byte slice, otherwise keep it as is.

jensneuse (changes requested):

router/core/router.go
StarpTech: Please elaborate.

router/core/router.go
StarpTech: Good catch

StarpTech: Adressed

router/internal/persistedoperation/cache.go
StarpTech: Fixed

router/internal/persistedoperation/cache.go
StarpTech: Fixed, byte conversion was actually not necessary

jensneuse (approved):

PR #973, opened 2024-07-28
chore: add rfc for previews

JivusAyrus:
## TODO


rfc/previews-using-fetaure-flags.md
```diff
@@ -0,0 +1,306 @@
...
+    types: [labeled, synchronize, closed, opened, reopened]
+
+jobs:
+  create:
+    runs-on: ubuntu-latest
+    if: github.event.pull_request.action == 'opened' || github.event.pull_request.action == 'reopened'
+    steps:
+      # 1. Create & Publish feature subgraph based on cosmo.yaml
+      # 2. Create a new feature flag
+      - uses: wundergraph/feature-flags@main
```
jensneuse, line 41: do we call this "feature flags", "previews", or something more generic, like "ci" or "continous-integration-toolkit"

rfc/previews-using-fetaure-flags.md
```diff
@@ -0,0 +1,306 @@
...
+
+#### Destroy Job
+
+- This job will destroy the feature flag and the feature subgraphs used for the preview of this PR.
+- This job will only run when the PR is closed/merged.
+
+
+### Limitations
+
+- The comso.yaml file should not be changed after the PR is opened, if changed for the preview to work as expected, the PR has to be closed and reopened.
```
jensneuse, line 120: Is it possible to detect this and raise a CI error, telling the user to close and re-open the PR?
or, maybe even easier, if you detect this, delete the FF and recreate it?

rfc/previews-using-fetaure-flags.md
```diff
@@ -0,0 +1,306 @@
...
+- Make sure the labels of the feature flag match that of the federated graph for which the preview is being created.
+- Make sure that the subgraphs being changed are a part of the federated graph.
+- Make sure that the changed subgraphs are a part of the config, if not present the featur subgraph for that subgraph will not be created.
+- All the subgraphs in the config have to be deployed and present on Cosmo.
+- The `cosmo.yaml` file should be present in the `.github` directory before the PR is labeled with `cosmo-preview`.
+- All the subgraphs changed in this PR should be deployed at the proposed routing_url in the CI for the preview to work.
+
+### Implementation
+
+The implementation is the same as Proposal-1, the only difference is that the feature flag is created when the PR is labeled with `cosmo-preview` and destroyed when the PR is unlabeled with `cosmo-preview` or when the pr is closed.
```
jensneuse, line 205: Can you add this to the .github/cosmo.yaml as an option to activate it?
Alternatively, is there another way to configure this per repo?

rfc/previews-using-fetaure-flags.md
```diff
@@ -0,0 +1,306 @@
...
+  create:
+    runs-on: ubuntu-latest
+    if: github.event.pull_request.action == 'opened' || github.event.pull_request.action == 'reopened'
+    steps:
+      # 1. Create & Publish feature subgraph based on cosmo.yaml
+      # 2. Create a new feature flag
+      - uses: wundergraph/feature-flags@main
+        id: ff
+        with:
+          namespace: "production"
```
jensneuse, line 240: It doesn't really look simpler to me.
We can also provide a JSON Schema for `.github/comso.yaml` to make the configuration easier

rfc/previews-using-fetaure-flags.md
JivusAyrus: Do you mean the name of the action?

rfc/previews-using-fetaure-flags.md
JivusAyrus: We should be able to provide that error.

Regarding deleting the FF, we will have to do a little more sometimes, like if they change the namespace then we will have to delete everything and then provision everything.

rfc/previews-using-fetaure-flags.md
JivusAyrus: - Can you add this to the .github/cosmo.yaml as an option to activate it?
Even this can be done, by adding an enable option in the comso.yaml, but feel adding the label is much more convenient

rfc/previews-using-fetaure-flags.md
JivusAyrus: Do you mean configuring using the 3rd proposal is not simple?

JivusAyrus: Regarding the JSON Schema for `.github/comso.yaml`, we can provide that

rfc/previews-using-fetaure-flags.md
jensneuse: yes

rfc/previews-using-fetaure-flags.md
jensneuse: This sounds dangerous.
I think it's better to just raise an error.

rfc/previews-using-fetaure-flags.md
jensneuse: You're not following my idea I think.
My idea was to enable or disable the "label-mode" in the `.github/cosmo.yaml` config.
Does this make sense? Is it technically possible?
Can the user decide to opt into this in another way?

rfc/previews-using-fetaure-flags.md
jensneuse: > Do you mean configuring using the 3rd proposal is not simple?

I don't think it's simpler than using `.github/cosmo.yaml`, you still have to configure a lot of things, just in a different place. So it's maybe marginally simpler with the downsides of not being able to handle multiple subgraphs.

rfc/previews-using-fetaure-flags.md
JivusAyrus: Yep we can do that

rfc/previews-using-fetaure-flags.md
JivusAyrus: yep

rfc/previews-using-fetaure-flags.md
JivusAyrus: Yeah, the name can be anything we decide

PR #978, opened 2024-07-29
fix: broken pagination on page size change and subgraphs table with many items

thisisnithin:
Changing page size when page number is greater than expected leads to empty results. Reset page number on page size change


StarpTech (approved):
LGTM

PR #985, opened 2024-07-30
fix: routing url in router playground

### LINKED ISSUE #982: Playground querying localhost on router v0.99+
Since router 0.99 which introduced the split of the playground in its own package, the router playground does not send queries to the hostname exposing the router (e.g. my-router-domain.com) but rather to localhost:3002. Is there a new env var that must be set ?

thisisnithin:
## Motivation and Context

The playground was built with local envs. Also updated workflow to catch it. Closes #982 

## TODO


StarpTech (approved):
Lgtm

PR #990, opened 2024-08-01
chore: improve handling and error reporting of exported default varia…

jensneuse:


thisisnithin (approved):

router/core/operation_processor.go
```diff
@@ -569,6 +575,17 @@ func (o *OperationKit) normalizeNonPersistedOperation() (cached bool, err error)
 	return false, nil
 }
 
+func (o *OperationKit) nullifyExportedVariables(variables []byte) []byte {
+	if variables == nil {
+		return nil
+	}
+	// min meaningful variables = {"a":1}
+	if len(variables) < 7 {
+		return nil
```
StarpTech, line 584: Why do we return nil when we have variables? Could you add more insights into the comment?

router/core/operation_processor.go
```diff
@@ -616,6 +633,13 @@ func (o *OperationKit) loadPersistedOperationFromCache() (ok bool, err error) {
 
 // populateDefaultVariablesFromExportedDefaults iterates through the exported default variables and sets missing ones in the variables
 func (o *OperationKit) populateDefaultVariablesFromExportedDefaults(exportedVariables, override []byte) ([]byte, error) {
+	if exportedVariables == nil {
+		return override, nil
+	}
+	if len(exportedVariables) < 7 {
```
StarpTech, line 639: This is duplicated.

StarpTech (changes requested):

router/core/operation_processor.go
jensneuse: added

StarpTech (approved):
LGTM

PR #997, opened 2024-08-01
chore: ensure that go mod is clean in the ci

AndreasZeissner:
## Motivation and Context

Enusre, that the build will break early if go mod was not cleaned up properly

## TODO


.github/workflows/aws-lambda-router-ci.yaml
```diff
@@ -45,6 +45,10 @@ jobs:
       - name: Check if git is not dirty after generating files
         run: git diff --no-ext-diff --exit-code
 
+      - uses: ./.github/actions/go-mod-tidy
```
StarpTech, line 48: Please do this before any go code is run.

StarpTech (changes requested):

.github/workflows/graphqlmetrics-ci.yaml
```diff
@@ -12,6 +12,7 @@ concurrency:
 env:
   CI: true
   INT_TESTS: true
+  GOWORK: 'off'
```
StarpTech, line 15: You would also need to update "make lint" in each go project.

go.work
```diff
@@ -4,9 +4,6 @@ toolchain go1.22.0
 
 use (
 	./aws-lambda-router
-	./composition-go
```
StarpTech, line 7: I tend to keep them for better IDE integration. With the CI check we should be already safe.

StarpTech (changes requested):

go.work
AndreasZeissner: Alright, let's keep it simple for now. Just a check to avoid having unpredictable dependency updates.

.github/workflows/graphqlmetrics-ci.yaml
AndreasZeissner: Thanks for the hint.

PR #1002, opened 2024-08-02
feat: add priority to modules

JivusAyrus:
## TODO


router-tests/modules/module_test.go
```diff
@@ -38,3 +39,103 @@ func TestModuleSetCustomHeader(t *testing.T) {
...
+	cfg := config.Config{
+		Modules: map[string]interface{}{
+			"myModule": module.MyModule{
+				Value: 1,
+			},
+		},
+	}
+
+	modulesW := make(map[string]core.ModuleInfo)
+	modulesW["module1"] = core.ModuleInfo{
```
StarpTech, line 53: Could you suffix the module names with the priority? This would be easy to read.

router-tests/modules/module_test.go
```diff
@@ -38,3 +39,103 @@ func TestModuleSetCustomHeader(t *testing.T) {
 		assert.JSONEq(t, res.Body, `{"data":{"employees":[{"id":1},{"id":2},{"id":3},{"id":4},{"id":5},{"id":7},{"id":8},{"id":10},{"id":11},{"id":12}]}}`)
 	})
 }
+
+func TestSortingModulesByPriority(t *testing.T) {
```
StarpTech, line 43: Priority 0 won't change the previous behavior correct?

router/core/modules.go
```diff
@@ -67,6 +70,32 @@ func RegisterModule(instance Module) {
 	modules[string(mod.ID)] = mod
 }
 
+// Sorts the modules by priority
+func SortModules(modules map[string]ModuleInfo) []ModuleInfo {
```
StarpTech, line 74: We can make private if it is only used in core.

router/core/router.go
```diff
@@ -508,7 +508,9 @@ func (r *Router) listenAndServe(cfg *nodev1.RouterConfig) error {
 }
 
 func (r *Router) initModules(ctx context.Context) error {
-	for _, moduleInfo := range modules {
+	var moduleList = SortModules(modules)
+
+	for _, moduleInfo := range moduleList {
```
StarpTech, line 513: Is this the only place where we access the modules?

StarpTech (changes requested):

router/core/modules.go
JivusAyrus: Used in tests

router/core/router.go
JivusAyrus: No, we access it in RegisterModule, but should not be a problem 
```
if _, ok := modules[string(mod.ID)]; ok {
    panic(fmt.Sprintf("module already registered: %s", mod.ID))
}
modules[string(mod.ID)] = mod
```

router-tests/modules/module_test.go
JivusAyrus: Fixed

router-tests/modules/module_test.go
JivusAyrus: Done

router-tests/modules/module_test.go
StarpTech: Unit tests should be live in the router module.

StarpTech (approved):
LGTM

PR #617, opened 2024-03-10
feat(controlplane): mailer options

d4rkstar:
## Motivation and Context
Added support for all mailer options. Now is possibile to change smtp hostname, port, credentials, tls and secure.
I added because SMTP server is hardcoded and is not possible to change it via env.

## TODO

- [X] Tests or benchmark included
- [ ] Documentation is changed or added on [https://app.gitbook.com/](https://app.gitbook.com/)
- [X] PR title must follow [conventional-commit-standard](https://github.com/wundergraph/wundergraph/blob/main/CONTRIBUTING.md#conventional-commit-standard)

PR #1008, opened 2024-08-05
feat(controlplane): add overriding the mailing options

AndreasZeissner:
## Motivation and Context

This is a follow up to: https://github.com/wundergraph/cosmo/pull/617

It introduces overriding the smtp server in the controlplane by introducing the following environment variables:

`SMTP_HOST:` the host to connect to (no default)   
`SMTP_PORT:`  the port to use (defaults to 587)
`SMTP_SECURE:` enable or disable smtps (defaults to true)
`SMTP_REQUIRE_TLS:` enable or disable using TLS (defaults to true)

## TODO


helm/cosmo/charts/controlplane/templates/deployment.yaml
```diff
@@ -236,6 +236,30 @@ spec:
                 secretKeyRef:
                   name: {{ include "controlplane.secretName" . }}
                   key: smtpUsername
+            - name: SMTP_PORT
```
StarpTech, line 239: Ideally, we set only add variables when the value is set.

StarpTech (changes requested):

helm/cosmo/charts/controlplane/templates/deployment.yaml
StarpTech: Sounds good.

controlplane/src/core/env.schema.ts
```diff
@@ -105,8 +105,26 @@ export const envVariables = z
...
+      .string()
+      .transform((val) => val === 'true')
+      .default('false'),
     SMTP_USERNAME: z.string().optional(),
     SMTP_PASSWORD: z.string().optional(),
+    SMTP_HOST: z.string().optional(),
+    SMTP_PORT: z
+      .string()
+      .optional()
+      .default('') // here to avoid possible type of number | "" | undefined type
```
AndreasZeissner, line 118: @StarpTech you might have a better solution for this.

If I leave away the default, tsc interferes `number | "" | undefined` as possible type as I would have to do something like this:

```
.transform((val) => val && Number.parseInt(val)),     
```

As val has then no explicit type. I could also handle it throughout the interfaces, but it feels bad, as all other types are usually `undefined | string` or `undefined | number` or such.

StarpTech: Do we need a default?

StarpTech (approved):
LGTM

controlplane/src/core/env.schema.ts
StarpTech:
```ts
      SMTP_PORT: z
          .coerce.number().optional()
```

AndreasZeissner: Thank you, this will now avoid the default.

PR #1013, opened 2024-08-06
feat: config file watcher

StarpTech:
## Motivation and Context

```yaml
version: "1"

execution_config: 
  file: 
    path: "router.json"
    watch: true
```

## TODO


router/core/router.go
```diff
@@ -962,7 +980,45 @@ func (r *Router) Start(ctx context.Context) error {
...
+
+				r.logger.Info("Config file changed. Updating server with new config")
+
+				cfg, err := execution_config.SerializeConfigBytes(data)
+				if err != nil {
+					r.logger.Error("Failed to serialize config file", zap.Error(err))
+					return
+				}
+
+				if err := r.newServer(ctx, cfg); err != nil {
```
jensneuse, line 1011: Is it possible to re-use logic from other polling/update mechanisms?

router/core/router.go
```diff
@@ -1050,6 +1106,12 @@ func (r *Router) Shutdown(ctx context.Context) (err error) {
 		ctx = ctxWithTimer
 	}
 
+	if r.configFileWatcher != nil {
+		if subErr := r.configFileWatcher.Close(); subErr != nil {
```
jensneuse, line 1110: Can a return happen before this?
If yes, I'd implement this cleanup with a defer after creating the watcher didn't return an error.
Otherwise ignore me.

jensneuse (approved):

JivusAyrus (approved):

router/core/router.go
StarpTech: We do. The only code we added here is specific to the watcher. The implementation that polls storage providers for updates is already unified.

router/core/router.go
StarpTech: No it can't. All cleanup functions never return and the shutdown function will return joined errors.

router/core/router.go
```diff
@@ -962,7 +980,45 @@ func (r *Router) Start(ctx context.Context) error {
...
+
+				data, err := os.ReadFile(r.executionConfig.Path)
+				if err != nil {
+					r.logger.Error("Failed to read config file", zap.Error(err))
+					return
+				}
+
+				r.logger.Info("Config file changed. Updating server with new config")
+
+				cfg, err := execution_config.SerializeConfigBytes(data)
```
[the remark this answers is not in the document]
StarpTech: haha done.

PR #1063, opened 2024-08-11
chore(RFC): open core module system v1

StarpTech:
## Motivation and Context

Related work: 

- https://github.com/wundergraph/cosmo/pull/1130#issuecomment-2452872727
- https://github.com/wundergraph/cosmo/pull/1125

## TODO


jensneuse:
A few goals I'd like to see covered:
- load my own config from my own config provider
- modify any config after load but before using it for a new mix
- allow the user to customize the transport (we're currently wrapping the default transport, let the user bring their own)
- hook into events like normalization, validation, parse (before, after)
- allow the user to define additional validation rules/steps
- Ideally, calling hooks is built in a way that automatically generates meaningful traces so that we're able to understand if and how much time the Router spends in a hook
- it should be easy to create a custom module / plugin as a go module which can then be imported and configured easily, making it simple to re-use and share custom extensions 

Additional comments:
Should we call it Gateway or Router? We talk a lot about Router, should we stick to this convention?

I'll look more into the RFC and give more specific feedback later/tomorrow.

rfc/custom-modules-v1.md
```diff
@@ -0,0 +1,456 @@
...
+## Introduction
+
+As of today, customers can extend the router with custom modules. These modules can be used to change the behavior of the router, add custom logic, or integrate with other systems. The current module system has several limitations that we want to address with this RFC:
+
+- The current module system is not native to GraphQL. It is based on HTTP middleware and does not provide a GraphQL-specific API.
+- The current module system is inconsistent and hard to use. It does not provide a clear API for developers to intercept and modify GraphQL requests and responses.
+- The current module system does not provide a way to create or modify OpenTelemetry data, logs for different parts of the gateway lifecycle.
+- The current module system does not provide a way to interact with the parsed, normalized, and planned GraphQL operation in order to implement custom logic.
+
+Ultimately, custom modules must be self-contained, composable and testable. They should provide a clear API for developers to interact with the gateway and subgraph lifecycle and implement custom logic without having to understand the internal workings of the router or advanced Go programming concepts.
```
jensneuse, line 27: Should we use the term "gateway" or "Router"?

rfc/custom-modules-v1.md
```diff
@@ -0,0 +1,456 @@
...
+To briefly explain the decision to use Go as the language for the module system, we have chosen Go because it is a simple and easy-to-learn language that is widely used in the infrastructure and cloud-native ecosystem. You can build custom integration on top production-grade SDK without re-implementing them from scratch. It superiors to scripting languages like Rhai or cross-compiling WebAssembly because it can be easily debugged, profiled, and tested with any modern IDE (VsCode, Goland, etc.). Not part of this RFC, are our ambitions to make the workflow as smooth as possible with a CLI tool that can scaffold, test, and deploy custom modules. In the future, custom modules could be published to a central registry and shared with the community. A brief overview of the workflow is provided at the end of this RFC.
```
jensneuse, line 28: If we bundle a custom module as "go module" and publish it on GitHub, sharing modules should be straight forward.

rfc/custom-modules-v1.md
```diff
@@ -0,0 +1,456 @@
...
+```go
+const myModuleID = "myModule"
+
+// MyModule is a simple module that has access to the GraphQL operation and add a header to the response
+// It demonstrates how to use the different handlers to customize the router.
+// It also shows how to use the config file to configure and validate your module config.
+// By default, the config file is located at `config.yaml` in the working directory of the router.
+type MyModule struct {
+	// Properties that are set by the config file are automatically populated based on the `mapstructure` tag
+	// Create a new section under `modules.<name>` in the config file with the same name as your module.
```
jensneuse, line 49: Can you clarify "config file"?
Is this the Router config.yaml?

rfc/custom-modules-v1.md
```diff
@@ -0,0 +1,456 @@
...
+	// Set a header on all origin requests
+	request.Header.Set("myHeader", ctx.GetString("myValue"))
+
+	// Set a custom value on the request context. See OnOriginResponse
+	ctx.Set("myValue", "myValue")
+
+	return request, nil
+}
+
+func (m *MyModule) Middleware(ctx core.RequestContext, next http.Handler) {
```
jensneuse, line 104: Middleware sounds very generic?
Where in the lifecycle is this?
Should we be more granular with the events we'd like to cover? (see my comment on the root PR)

rfc/custom-modules-v1.md
```diff
@@ -0,0 +1,456 @@
...
+	// The active Telemetry instance
+	Telemetry *core.Telemetry
+	// The final GraphQL response with all the information like data, errors, extensions etc.
+	// This is the response that will be sent to the client and can be manipulated or replaced
+	Response *core.GraphQLResponse
+	// The original HTTP response
+	Orignal *http.Response
+}
+
+type GatewayHooks interface {
```
jensneuse, line 210: Gateway or Router?

rfc/custom-modules-v1.md
StarpTech: You're commenting on the old module system.

rfc/custom-modules-v1.md
StarpTech: https://github.com/wundergraph/cosmo/pull/1063/files/eeca1bdc157a78164b22ef4520d31a31ec438a00#r1712985033

rfc/custom-modules-v1.md
StarpTech: I like Gateway. Router is the whole product, gateway represents for me the first layer when a request hits the router. After that, it will be passed to subgraphs.

rfc/custom-modules-v1.md
StarpTech: Yes, this is the idea. See "Module Registration" section.

StarpTech:
> A few goals I'd like to see covered:

Please take a closer look to the proposal. Most of your points has been already addressed. For other points, I do miss a real world use case e.g 

> allow the user to customize the transport ...

The goal of the module system is to enable all sorts of use cases, but we need to be very careful not to expose internals where developers can shoot themselves in the foot.

Other points are valid though and I'm going to address them. Thank you.

rfc/custom-modules-v1.md
jensneuse: I'd remove the old system from the RFC

jensneuse:
> > A few goals I'd like to see covered:
> 
> @jensneuse Please take a closer look to the proposal. Most of your points has been already addressed. For other points, I do miss a real world use case e.g
> 
> > allow the user to customize the transport ...
> 
> The goal of the module system is to enable all sorts of use cases, but we need to be very careful not to expose internals where developers can shoot themselves in the foot.
> 
> Other points are valid though and I'm going to address them. Thank you.

Some companies want to have full control over the transport.
I'd leverage our existing transport architecture, but allow them to override the wrapped transport if they want to.
It's a potential risk, sure, but in such a case they will know what they do.

rfc/custom-modules-v1.md
```diff
@@ -0,0 +1,510 @@
...
+	Name string
+	// The ID of the subgraph
+	ID string
+	// The URL of the subgraph
+	URL *url.URL
+	// The schema of the subgraph
+	Schema *graphql.Schema
+}
+
+type Graph struct {
```
Aenimus, line 60: Graph is too generic IMO. It's a `FederatedGraph`.

StarpTech: We also have monographs. I thought the separation of Graph and Subgraph is clear.

Aenimus: And under the hood, a monograph is a federated graph with a single subgraph.

rfc/custom-modules-v1.md
StarpTech: No strong objection, see my comment on the other thread.

rfc/custom-modules-v1.md
```diff
@@ -0,0 +1,568 @@
...
+
+As powerful as the new module becomes, it is important to move basic and common functionality into the core of the router because building and maintaining custom modules should be a last resort. The router should provide a rich set of features out of the box that cover the most common use cases. Custom modules should be reserved for advanced or highly specific use cases that cannot be achieved with the built-in features of the router. Integration with third-party services, custom authentication, and advanced telemetry are examples of use cases that are well-suited for custom modules.
+
+## Proposal
+
+A developer can implement a custom module by creating a struct that implements one or more of the following interfaces:
+
+- `GatewayHooks`: Provides hooks for the gateway lifecycle, including request and response handling.
+- `SubgraphHooks`: Provides hooks for subgraph requests and responses.
+- `ApplicationHooks`: Provides hooks for the application lifecycle, including startup, shutdown, and error handling.
```
jensneuse, line 40: What about Configuration Hooks?
E.g. when a new config was downloaded, when a new mux is ready to be used (e.g. to warmup a cache), when the mux got swapped..

rfc/custom-modules-v1.md
```diff
@@ -0,0 +1,568 @@
...
+- `AuthenticationHooks`: Provides hooks for authentication and authorization logic.
```
jensneuse, line 41: Should we distinguish Authentication & Authorization into different hooks?
AuthN typically comes at the Gateway level, AuthZ at the Subgraph or Field level.

rfc/custom-modules-v1.md
```diff
@@ -0,0 +1,568 @@
...
+- `TelemetryHooks`: Provides hooks for OpenTelemetry spans and metrics.
```
jensneuse, line 42: Should we divide such hooks into Metrics & Traces or keep them together?

rfc/custom-modules-v1.md
```diff
@@ -0,0 +1,568 @@
...
+- `OperationHooks`: Provides hooks for parsed, normalized, and planned GraphQL operations.
```
jensneuse, line 43: Operation might not be 100% clear. What about "GraphQLOperationHooks"?

rfc/custom-modules-v1.md
jensneuse: Call it Supergraph maybe?

rfc/custom-modules-v1.md
jensneuse: I think Gateway makes sense.

rfc/custom-modules-v1.md
```diff
@@ -0,0 +1,568 @@
...
+	// The active Telemetry instance
+	Telemetry *core.Telemetry
+	// The final GraphQL response with all the information like data, errors, extensions etc.
+	// This is the response that will be sent to the client and can be manipulated or replaced
+	Response *core.GraphQLResponse
+	// The original HTTP response
+	Orignal *http.Response
+}
+
+type GatewayHooks interface {
```
jensneuse, line 101: I would take a slightly different approach to defining interfaces.
Instead of having one big iFace with many funcs, create multiple smaller ones and combine them.

Example:

```go
type GatewayHooks interface {
  GatewayOnRequestHook
  GatewayOnResponseHook
  GatewayOnErrorHook
}

type GatewayOnRequestHook interface {
  OnGatewayRequest(ctx *core.GatewayRequest, err error) error
}

type GatewayOnResponseHook interface {
  OnGatewayResponse(ctx *core.GatewayResponse, err error) error
}

type GatewayOnErrorHook interface {
  OnGatewayError(err error)
}
```

The reasoning is that a user can implement such hooks in a more granular way.
They might need `OnGatewayResponse` but don't want the other functions.
So an implementer only needs to implement what they really need,
and the hook system only needs to call what was implemented.

rfc/custom-modules-v1.md
```diff
@@ -0,0 +1,568 @@
...
+
+type SubgraphRequest struct {
+	// The active subgraph
+	Subgraph *core.Subgraph
+	// The original Gateway request
+	GatewayRequest *core.RouterRequest
+	// The final GraphQL request to the subgraph
+	Request *core.GraphQLRequest
+	// The active Telemetry instance
+	Telemetry *core.Telemetry
```
jensneuse, line 120: What is `core.Telemetry` and what funcs does it expose?
Is this to create spans and metrics?
Does it contain the context to properly propagate spans?

rfc/custom-modules-v1.md
```diff
@@ -0,0 +1,568 @@
...
+	// The original http request to the subgraph
+	Orignal *http.Request
```
jensneuse, line 122: maybe call it HttpRequest or ClientRequest?

rfc/custom-modules-v1.md
```diff
@@ -0,0 +1,568 @@
...
+	Shutdown() error
+	// Module returns the module information and factory function
+	Module() core.ModuleInfo
+}
+
+// SubgraphHooks are called when a subgraph request or response is made.
+// The order is not guaranteed, so the hooks should be idempotent and side-effect free.
+// if state needs to be shared between hooks, it should be stored in the context.
+// We will provide an easy way to share state between hooks.
+type SubgraphHooks interface {
```
jensneuse, line 153: https://github.com/wundergraph/cosmo/pull/1063/files#r1718053098

rfc/custom-modules-v1.md
```diff
@@ -0,0 +1,568 @@
...
+type AuthenticationHooks interface {
+	// OnAuthenticate is called when a gateway request is authenticated
+	// Returning an error will result in a GraphQL error as a response from the gateway.
+	OnAuthenticate(req *core.GatewayRequest, err error) error
+	// OnPostAuthenticate is called after a gateway request is authenticated
+	// Returning an error will result in a GraphQL error as a response from the gateway.
+	OnPostAuthenticate(req *core.GatewayRequest, err error) error
+}
+
+type OperationHooks interface {
```
jensneuse, line 189: rename to GraphQLOperationHooks?
also: https://github.com/wundergraph/cosmo/pull/1063/files#r1718053098

rfc/custom-modules-v1.md
```diff
@@ -0,0 +1,568 @@
...
+	// OnNormalize is called when an operation is normalized
+	// Returning an error will result in a GraphQL error as a response from the gateway.
+	OnOperationNormalize(op *core.Operation, err error) error
```
jensneuse, line 192: Should we follow a pattern of "Pre" + "Post", just to be clear when it's being called?
This question applies to all occurrences of On... and OnPost...

rfc/custom-modules-v1.md
```diff
@@ -0,0 +1,568 @@
...
+	// OnPostNormalize is called after an operation is normalized
+	// Returning an error will result in a GraphQL error as a response from the gateway.
+	OnPostOperationNormalize(op *core.Operation, err error) error
+	// OnPlan is called when an operation is planned
+	// Returning an error will result in a GraphQL error as a response from the gateway.
+	OnOperationPlan(op *core.Operation, err error) error
+	// OnPostOperationPlan is called after an operation is planned
+	// Returning an error will result in a GraphQL error as a response from the gateway.
+	OnPostOperationPlan(op *core.Operation, err error) error
```
jensneuse, line 201: These are missing:
- PreValidate (modify request before validation)
- PostValidate (add additional validation logic)
- PostParseRequestOptions (e.g. to modify trace options or execution options)
- PreReadMultipart (e.g. to handle multipart options or intercept, prevent certain uploads)
- PostReadMultipart (e.g. to modify files or do something with files, like validation)
- PreFetchPersistedOperation (e.g. to do additional validation before fetching a PO)
- PostFetchPersistedOperation (e.g. to do post processing of the PO)

Question regarding the above. Do we allow mutating state? (I think we should)
Or should we distinguish read only hooks an mutating hooks (I don't think so)

In addition, I'd also suggest to add the following hooks to "completely" override the behaviour of certain parts of the request chain.

- OverrideParse (replace existing logic with custom parse logic, implementer gets input and must return parsed operation or error)
- OverrideValidate (replace existing validation logic)
- OverrideNormalize (normalize on your own)
- OverridePlan (plan on your own)
- OverrideFetchPersistedOperation (load PO on your own)

rfc/custom-modules-v1.md
```diff
@@ -0,0 +1,568 @@
...
+
+```go
+type MyModule struct{}
+
+// Ensure that MyModule implements the GatewayHooks interface
+var _ GatewayHooks = (*MyModule)(nil)
+
+func (m *MyModule) OnGatewayResponse(res *core.GatewayResponse, err error) error {
+	// Add custom extensions to the response
+	res.Response.Extensions["myExtension"] = "myValue"
```
jensneuse, line 256: We don't currently expose the response as an object that could be used like this, it's just a byte slice.
This is for performance reasons. We're encoding the response into a byte buffer.
However, it's possible that we would expose "data" as a byte slice, "errors" as a fastjson array, and extensions as a fastjson object. This way, you could have some level of access, probably enough for the required customizations.
I assume that most customizations won't touch data, but errors and extensions, which we can easily expose.

However, if we do it like this, I'd restructure the hook here.
OnGatewayResponse will really "just" contain the encoded response.
We can add another hook, e.g. "PreGraphQLResponseEncoding" which gives you all the parts (data, errors, extensions) before we encode them as a single response.

rfc/custom-modules-v1.md
```diff
@@ -0,0 +1,568 @@
...
+	// an error is returned because the behavior is undefined.
+	
+	ctx.RegisterOriginTransport(&http.Transport{})
+	return nil
+}
+```
+
+## React to specific GraphQL Directives
+
+React to specific GraphQL directives in the subgraph schema and decide if the request should be forwarded to the subgraph. The following example checks if the `@requiresScopes` directive is present in the operation. Another example could be to check if the `@cacheControl` directive is present and set the cache control header on the response.
```
jensneuse, line 421: This is very powerful.

At what specific point in the request chain will such a handler be called?
- Normalization
- Planning
- Validation
- Execution

It could make sense to allow the user to choose.

rfc/custom-modules-v1.md
```diff
@@ -0,0 +1,568 @@
...
+## Custom Module configuration
+
+Custom modules can be configured using a YAML file that is loaded by the router at startup `config.yaml`. We reserve a section in the configuration file for custom modules. Each module can have its own configuration section with custom properties.
+The name of the module in the configuration file must match the name specified in the `ModuleInfo` struct.
+
+```yaml
+# config.yaml
+modules:
+  my_module:
+    value: 42
```
jensneuse, line 350: 👍

rfc/custom-modules-v1.md
StarpTech: I like.

rfc/custom-modules-v1.md
StarpTech: Great idea. Except the Module interface needs to be implemented fully.

rfc/custom-modules-v1.md
StarpTech: HttpRequest is good.

rfc/custom-modules-v1.md
StarpTech: Done

rfc/custom-modules-v1.md
StarpTech: Done

rfc/custom-modules-v1.md
StarpTech: I understand your concerns and I like the fastjson object idea. I think we should try to make the API as intuitive as possible. In the first draft, I'd like to avoid dedicated hooks. We should reiterate on this before the release once we did some performance tests. We could also provide optimized helper functions to get and manipulate extensions and errors that operating on byte level.

rfc/custom-modules-v1.md
StarpTech: My idea was to call them after the GraphQL operation has been fully parsed, normalized and planned. After this custom logic can be executed. The handler is also called on subgraph schema directives. The User can distinguish both cases by looking into the context e.g. when the current subgraph is set.

rfc/custom-modules-v1.md
StarpTech: Most of them have been addressed. The RFC should already give a very good picture of the extensibility of the system. The current hooks are oriented on customer feedback. We can add more hooks any time.

rfc/custom-modules-v1.md
StarpTech: I added a note to the RFC that we will evaluate Pre/Post hooks for certain methods. I don't think it makes sense for all. A strong point for Pre/Post hooks is tracing, where users can precisely measure the time of a hook themselves, but this would require that they can disable built-in tracing otherwise we would generate duplicates.

rfc/custom-modules-v1.md
StarpTech: Great question. I can answer this question once we start to implement the first use cases, but it is likely that we will separate for the same reasons you mentioned by having more fine-grained interfaces.

rfc/custom-modules-v1.md
StarpTech: Yes. I separated them.

rfc/custom-modules-v1.md
StarpTech: We have never called it Supergraph anywhere. Do you still think it is a good idea? I'm also fine with @Aenimus approach to call it FederatedGraph.

rfc/custom-modules-v1.md
StarpTech: Great idea, added in https://github.com/wundergraph/cosmo/pull/1063/commits/cd657efae856d4057b952eaceac504b2cc01e666

rfc/custom-modules-v1.md
StarpTech: Renamed to router

rfc/custom-modules-v1.md
```diff
@@ -0,0 +1,706 @@
...
+- TelemetryHooks: Provides hooks for OpenTelemetry tracing and metrics.
+  - `TelemetrySpanHook`: Called when a span is created.
+  - `TelemetryMetricHook`: Called when a metric is recorded.
+- GraphServerHooks: Provides hooks for the lifecycle of a GraphQL server.
+  - `GraphServerStartHook`: Called when the GraphQL server starts e.g. for the first time or when the schema is updated.
+  - `GraphServerStopHook`: Called when the GraphQL server stops e.g. when the application is shut down or the old schema is replaced.
+- GraphQLOperationHooks: Provides hooks for parsed, normalized, and planned GraphQL operations.
+  - `GraphQLOperationParseHook`: Called when an operation is parsed.
+  - `GraphQLOperationNormalizeHook`: Called when an operation is normalized.
+  - `GraphQLOperationPlanHook`: Called when an operation is planned.
```
kaialang, line 66: Will user be able to use GraphQLOperationPlanHook for monitoring the planCache or a separate GraphqlOperationPlanCacheHook is needed? It could be useful for monitoring the cache hit/miss and maybe even modify the caching policy based on their needs.

rfc/custom-modules-v1.md
```diff
@@ -0,0 +1,706 @@
...
+type RouterRequestHook interface {
+	// OnRouterRequest is called when a request is made to the router and after all GraphQL information is available
+	// Returning an error will result in a GraphQL error being returned to the client.
+	OnRouterRequest(ctx *core.RouterRequest, err error) error
+}
+
+type RouterResponseHook interface {
+	// OnRouterResponse is called before the response is sent to the client
+	// Returning an error will result in a GraphQL error being returned to the client.
+	OnRouterResponse(ctx *core.RouterResponse, err error) error
```
kaialang, line 189: For the err here, can we include a meaninful errorCode or statusCode? It would be helpful in error handling.

rfc/custom-modules-v1.md
```diff
@@ -0,0 +1,706 @@
...
+}
+
+type RouterErrorHook interface {
+	// OnError is called when an error occurs during the router lifecycle
+	OnRouterError(err error)
```
flymedllva, line 194: I'm not sure how to use it yet. I would like to be able to replace the status code in the subgraph responses, to decide which are good and which are not, which should be retrained and which should not. And then modify the router's final response after gluing and downloading everything

flymedllva: This interface will have to be driven to your error type somehow? Or how to change it?

rfc/custom-modules-v1.md
```diff
@@ -0,0 +1,706 @@
...
+	// The original http response from the subgraph
+	HttpResponse *http.Response
+}
+
+// Router Hooks
+
+type RouterRequestHook interface {
+	// OnRouterRequest is called when a request is made to the router and after all GraphQL information is available
+	// Returning an error will result in a GraphQL error being returned to the client.
+	OnRouterRequest(ctx *core.RouterRequest, err error) error
```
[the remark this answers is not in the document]
kaialang: Now I think about it, my solution has the assumption of OnRouterRequest is "PreRouterRequest". Without knowing exact point on OnRouteRequest, this may or may not work.

rfc/custom-modules-v1.md
kaialang: +1 to "Pre" + "Post". It's clearer what time point they're called. 

IMO, tracing should be just another plugin, a default built-in plugin that's built on top of this RFC. They should by default off. User should either implement their own tracing plugin or enable the default.

rfc/custom-modules-v1.md
```diff
@@ -0,0 +1,706 @@
...
+	// Returning a function to be called when the span ends.
+	// This can be used to add custom attributes or events to the span.
+	OnSpan(span *trace.Span) func() // Return a function to be called when the span ends
+}
+
+type TelemetryMetricHook interface {
+	// OnMetric is called when a metric is recorded
+	// Returning an error will result in a telemetry error
+	OnMetric(metric *metric.Metric) error
+}
```
kaialang, line 251: I see OnSpan() and OnMetric() can be useful in some cases. But adding attributes can be achieved by adding keyValue pairs in request context (please correct me if I'm wrong).

This is a different level of abstraction. I would strongly prefer a more flexible hook system, eg, using Pre/Post, that allows to metric/tracing plugins. We can still provide default Prometheus, OTel metrics/tracings by building default plugins on top of the hook system.

StarpTech:
Closed accidentally from bot

rfc/custom-modules-v1.md
```diff
@@ -0,0 +1,706 @@
...
+// Ensure that MyModule implements the RouterResponseHook interface
+var _ RouterResponseHook = (*MyModule)(nil)
+
+func (m *MyModule) OnRouterResponse(res *core.RouterResponse, err error) error {
+	// Add custom extensions to the response
+	res.Response.Extensions["myExtension"] = "myValue"
+
+	// Rewrite errors in the response
+	if len(res.Response.Errors) > 0 {
+		for _, err := range res.Response.Errors {
```
flymedllva, line 370: 1.1 We are interested in changing the message, as I understand we will now have this access.
1.2 It is not clear how we will work with untyped extensions, through fastjson? It's important for us not to change custom extensions that we didn't expect.
2. We would also like to understand which subgraph the error belongs to, so that we can change it with this logic in mind.
3. We want to do ctx.Set, so that we can get the modified errors in other methods (for example, for logging).

rfc/custom-modules-v1.md
```diff
@@ -0,0 +1,706 @@
...
+
+// SubgraphHooks are called when a subgraph request or response is made.
+// The order is not guaranteed, so the hooks should be idempotent and side-effect free.
+// if state needs to be shared between hooks, it should be stored in the context.
+// We will provide an easy way to share state between hooks.
+
+type SubgraphRequestHook interface {
+	// OnSubgraphRequest is called when a subgraph request is made
+	// Returning an error will result in a GraphQL error being returned to the client.
+	OnSubgraphRequest(ctx *core.SubgraphRequest, err error) error
```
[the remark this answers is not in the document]
flymedllva: It seems that what I described will be done in OnRouterResponse? And this method will be triggered after OnRouterResponse?

rfc/custom-modules-v1.md
flymedllva: It seems that what I described will be done in OnRouterResponse https://github.com/wundergraph/cosmo/pull/1063/files#r1778364936? And this method will be triggered after OnRouterResponse?

rfc/custom-modules-v1.md
```diff
@@ -0,0 +1,706 @@
...
+	// The normalization of the operation
+	Normalization *core.Normalization
+}
+
+type RouterRequest struct {
+	// The original GraphQL request with all the information like query, variables, operation name, extensions etc.
+	Request *core.GraphQLRequest
+	// The parsed, normaliazed and planned operation with all the information like name, variables, type, document representation,
+	// client name version, uploaded files, plan, normalization, persisted operation etc.
+	Operation *core.Operation
```
flymedllva, line 127: Somewhere in here I wanted to be able to get
```type Subgraph struct {
	Id string
	Name string
	Url *url.URL
	UrlString string
}
```

So that this information can be used when changing the error.

rfc/custom-modules-v1.md
StarpTech: That is a possible way. The current direction is to provide additional information in the context e.g. if a plan was fetched from the cache could be on the context of the `PostGraphQLOperationPlanHook`

rfc/custom-modules-v1.md
```diff
@@ -0,0 +1,706 @@
...
+	ID string
+	// Version of the graph
+	Version string
+	// Subgraphs of the graph
+	Subgraphs []*core.Subgraph
+	// GraphQL schema of the graph
+	Schema *graphql.Schema
+}
+
+type Operation struct {
```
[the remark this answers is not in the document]
StarpTech: Good question. This RFC tries to approach the overall picture of the system. Everything what was possible in the old system must be possible in the new system just in a different way. The original and normalized query will be available on the operation object as well.

rfc/custom-modules-v1.md
```diff
@@ -0,0 +1,706 @@
...
+## Abstract
+
+This RFC describes an overhaul of the current module system in the router. The new module system is designed to be more flexible and native to GraphQL. It allows developers to hook into the router lifecycle as well as outgoing and incoming requests to subgraphs.
+
+## Introduction
+
+As of today, customers can extend the router with custom modules. These modules can be used to change the behavior of the router, add custom logic, or integrate with other systems. The current module system has several limitations that we want to address with this RFC:
+
+- The current module system is not native to GraphQL. It is based on HTTP middleware and does not provide a GraphQL-specific API.
+- The current module system is inconsistent and hard to use. It does not provide a clear API for developers to intercept and modify GraphQL requests and responses.
```
ghorak-happening, line 23: Hi,

Great stuff here. One thing that I didn't see is the possibility to hook into the EDFS lifecycle.
Would it be possible with the Custom Module System to intercept subscription requests to subgraphs and inject a custom transport, similar to the Kafka or NATS  that you already provide.

I would be particularly interested in adding Redis PubSub support to EDFS.

Thx,
Goran

StarpTech: Hi @ghorak-happening, that's a great idea. The goal of this RFC is to propose an architecture that is flexible enough to add those extensions later. The first version will focus on the fundamentals. I'm going to add your requirements to the list so as not to forget them.

ghorak-happening: That would be great @StarpTech, thank you!

rfc/custom-modules-v1.md
```diff
@@ -0,0 +1,706 @@
...
+
+type AuthorizationHook interface {
+	// OnAuthorize is called when a router request is authorized
+	// Returning an error will result in a GraphQL error as a response from the router.
+	OnAuthorize(ctx *core.RouterRequest, err error) error
+}
+
+// OperationHooks are called when an operation is parsed, normalized, or planned
+
+type GraphQLOperationParseHook interface {
```
janeli1, line 271: Can we have add a GraphQLOperationPreParseHook? I am asking because I think we can use this new module system to implement Persisted Operations feature and PO fetching is done before parsing the operation.

StarpTech:
Closed in favor of the ADR https://github.com/wundergraph/cosmo/pull/1863

Issue #886, opened 2024-06-25
Unable to make full-demo-up


thekevinbrown:
### Component(s)

router, controlplane, studio, composition, otelcollector, graphqlmetrics, helm, cli

### Component version

5bcf14915115d400ea6dc394b5d219e4f6e2eaca

### wgc version

None

### controlplane version

Unsure

### router version

Unsure

### What happened?

If possible, please create a PR with a failing test to illustrate the issue clearly.
Otherwise, please attach a minimum reproduction through a GitHub repository that includes
essential information such as the relevant subgraph SDLs.
Please also make sure that the instructions for the reproduction are clear, tested, and fully accurate.

## Description

## Steps to Reproduce

Run `make full-demo-up`

## Expected Result

Demo starts

## Actual Result

```console
make full-demo-up
docker compose -f docker-compose.full.yml --profile default up --build --remove-orphans --detach && ./scripts/setup-fulldemo.sh
WARN[0000] The "OTEL_AUTH_TOKEN" variable is not set. Defaulting to a blank string. 
WARN[0000] The "OTEL_AUTH_TOKEN" variable is not set. Defaulting to a blank string. 
WARN[0000] The "OTEL_AUTH_TOKEN" variable is not set. Defaulting to a blank string. 
WARN[0000] The "OTEL_AUTH_TOKEN" variable is not set. Defaulting to a blank string. 
WARN[0000] The "ROUTER_TOKEN" variable is not set. Defaulting to a blank string. 
WARN[0000] The "OTEL_AUTH_TOKEN" variable is not set. Defaulting to a blank string. 
WARN[0000] The "OTEL_AUTH_TOKEN" variable is not set. Defaulting to a blank string. 
WARN[0000] /Users/kevin/development/wundergraph/cosmo/docker-compose.full.yml: `version` is obsolete 
[+] Running 90/20
 ✔ seed Pulled                                                                                                                                                                     21.3s 
 ✔ clickhouse-migration Pulled                                                                                                                                                     21.3s 
 ✔ otelcollector Pulled                                                                                                                                                            57.9s 
 ✔ clickhouse Pulled                                                                                                                                                               68.9s 
 ✔ nats Pulled                                                                                                                                                                     64.5s 
 ✔ keycloak Pulled                                                                                                                                                                 82.4s 
 ✔ redis Pulled                                                                                                                                                                    79.3s 
 ✔ studio Pulled                                                                                                                                                                   39.9s 
[...]

PR #1064, opened 2024-08-12
fix:  make full-demo-up failure

chensunny:
fix full-demo-up hang or killed while Start the subgraphs and the router

✅ Closes: https://github.com/wundergraph/cosmo/issues/886

## Motivation and Context

when I  run make full-demo-up,  It will hang or be killed  like this
image
 

It is  useless when  I change the Memory and CPU
image

I think the docker-compose use too much concurrency case any issues.
After I set the param of --parallel 3. It was work well

## TODO


StarpTech:
Hi @chensunny, thank for raising this up. Would you like to create a PR to add this to the README?

PR #1066, opened 2024-08-12
fix: cannot make composition with multiple interfaces for interfaceObject 

AndriiTsok:
I created two tests to indicate the use case I tried to do. I wanted to implement Updatable, Deletable, Shareable interfaces and make its implementation via interfaceObject. Cosmo works fine if Entity implements only 1 of these 3 interfaces... but fails when we do more than 1.

## Motivation and Context

Provided two tests that should help identifying the issue.

## TODO

### DIFF
--- a/composition/tests/resolvability.test.ts
+++ b/composition/tests/resolvability.test.ts
@@ -1202,6 +1202,16 @@ describe('Field resolvability tests', () => {
   });
 });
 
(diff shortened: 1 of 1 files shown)


Aenimus:
Fixed in https://github.com/wundergraph/cosmo/pull/1073/files

Thanks for the report!

PR #1069, opened 2024-08-13
fix: semaphore is not released after panic

StarpTech:
## TODO


router/core/graphql_prehandler.go
```diff
@@ -260,340 +273,304 @@ func (h *PreHandler) Handler(next http.Handler) http.Handler {
...
-			// the kit is bound to the number of CPUs, and we must not hold onto it while doing IO operations
-			operationKit.Free()
-			operationKit = nil
-			finalErr = err
-
 			// Mark the root span of the router as failed, so we can easily identify failed requests
 			rtrace.AttachErrToSpan(routerSpan, err)
 
 			writeOperationError(r, w, requestLogger, err)
-			h.releaseBodyReadBuffer(buf)
```
jensneuse, line 290: Do we still release the buf like before?

router/core/graphql_prehandler.go
```diff
@@ -260,340 +273,304 @@ func (h *PreHandler) Handler(next http.Handler) http.Handler {
...
 			return
 		}
 
-		var skipParse bool
+		// If we have authenticators, we try to authenticate the request
+		if len(h.accessController.authenticators) > 0 {
```
jensneuse, line 298: Can we encapsulate this?

router/core/graphql_prehandler.go
```diff
@@ -260,340 +273,304 @@ func (h *PreHandler) Handler(next http.Handler) http.Handler {
...
+		// the kit must be freed before we're doing io operations
+		// the kit is bound to the number of CPUs, and we must not hold onto it while doing IO operations
+		// it needs to be called inside a defer to ensure it is called in panic situations as well
 
-		if blockedErr := h.operationBlocker.OperationIsBlocked(operationKit.parsedOperation); blockedErr != nil {
-			// the kit must be freed before we're doing io operations
-			// the kit is bound to the number of CPUs, and we must not hold onto it while doing IO operations
+		if operationKit != nil {
 			operationKit.Free()
 			operationKit = nil
```
jensneuse, line 365: is this ineffectful?

router/core/graphql_prehandler.go
```diff
@@ -260,340 +273,304 @@ func (h *PreHandler) Handler(next http.Handler) http.Handler {
...
 
-		attributes = []attribute.KeyValue{
-			otel.WgOperationName.String(operationKit.parsedOperation.Request.OperationName),
-			otel.WgOperationType.String(operationKit.parsedOperation.Type),
-		}
-		if operationKit.parsedOperation.GraphQLRequestExtensions.PersistedQuery != nil &&
-			operationKit.parsedOperation.GraphQLRequestExtensions.PersistedQuery.Sha256Hash != "" {
-			attributes = append(attributes, otel.WgOperationPersistedID.String(operationKit.parsedOperation.GraphQLRequestExtensions.PersistedQuery.Sha256Hash))
-		}
+	if err != nil {
```
jensneuse, line 369: Should this be above the defer? I'd assume that if NewKit returns err, operationKit should be nil, so we don't need to Free it.

router/core/graphql_prehandler.go
```diff
@@ -260,340 +273,304 @@ func (h *PreHandler) Handler(next http.Handler) http.Handler {
...
+		return nil, err
+	}
 
-		routerSpan.SetAttributes(attributes...)
-		metrics.AddAttributes(attributes...)
+	if err := operationKit.UnmarshalOperation(); err != nil {
```
jensneuse, line 373: Did we trace unmarshal before?

router/core/graphql_prehandler.go
```diff
@@ -651,19 +628,26 @@ func (h *PreHandler) parseRequestOptions(r *http.Request, clientInfo *ClientInfo
...
-	if h.enableRequestTracing && clientInfo.WGRequestToken != "" && h.routerPublicKey != nil {
-		_, err := jwt.Parse(clientInfo.WGRequestToken, func(token *jwt.Token) (interface{}, error) {
-			return h.routerPublicKey, nil
-		}, jwt.WithValidMethods([]string{jwt.SigningMethodES256.Name}))
-		if err != nil {
-			requestLogger.Error(fmt.Sprintf("failed to parse request token: %s", err.Error()))
-			return resolve.ExecutionOptions{}, resolve.TraceOptions{}, err
+	// Determine if we should enable request tracing / query plans at all
+	if h.enableRequestTracing {
+		// In dev mode we always enable tracing / query plans
```
jensneuse, line 633: reword to "we always allow", because enabling is determined by the parse funcs

router/core/graphql_prehandler.go
```diff
@@ -651,19 +628,26 @@ func (h *PreHandler) parseRequestOptions(r *http.Request, clientInfo *ClientInfo
...
+		if h.developmentMode {
+			return h.parseRequestExecutionOptions(r), h.parseRequestTraceOptions(r), nil
+		}
+		// If the client has a valid request token and we have a public key from the controlplane
+		if clientInfo.WGRequestToken != "" && h.routerPublicKey != nil {
```
jensneuse, line 638: do we want to move this logic somewhere else and encapsulate it better?

jensneuse (changes requested):
Overall seems good, just some minor comments.

One thing I'd like you to do is add the following 2 tests.

1.

Set max parsers to 1.
Make one request that triggers a panic during parsing.
Assert that you can make a second request.

2. Set max resolvers to 1
Make on request that triggers a panic during resolving.
Assert that you can make a second request.

router/core/graphql_prehandler.go
StarpTech: Yes, same behaviour in `handleOperation`

router/core/graphql_prehandler.go
StarpTech: I like it. Done.

router/core/graphql_prehandler.go
StarpTech: I was also under the same impression. I just copied the code over. I'm going to remove it.

router/core/graphql_prehandler.go
StarpTech: Yes, done.

router/core/graphql_prehandler.go
StarpTech: No

router/core/graphql_prehandler.go
StarpTech: I think it is fine for now.

jensneuse (approved):

PR #1082, opened 2024-08-15
feat: enable sentry usage in studio

AndreasZeissner:
## Motivation and Context

This PR enables using sentry in the studio. 

## TODO


studio/sentry.client.config.ts
```diff
@@ -0,0 +1,54 @@
...
+   */
+  replaysOnErrorSampleRate: parseFloat(
+    process.env.NEXT_PUBLIC_SENTRY_CLIENT_REPLAYS_ON_ERROR_SAMPLE_RATE || "1",
+  ),
+  /**
+   * The sample rate for session-long replays.
+   * 1.0 will record all sessions and 0 will record none.
+   */
+  replaysSessionSampleRate: parseFloat(
+    process.env.NEXT_PUBLIC_SENTRY_CLIENT_REPLAYS_SESSION_SAMPLE_RATE || "1",
```
StarpTech, line 45: Please stick with 0 by default.

studio/sentry.server.config.ts
```diff
@@ -0,0 +1,18 @@
...
+// The config you add here will be used whenever the server handles a request.
+// https://docs.sentry.io/platforms/javascript/guides/nextjs/
+
+import * as Sentry from "@sentry/nextjs";
+
+Sentry.init({
+  dsn: process.env.NEXT_PUBLIC_SENTRY_DSN,
+
+  // Adjust this value in production, or use tracesSampler for greater control
+  tracesSampleRate: parseFloat(process.env.SENTRY_SERVER_SAMPLE_RATE || "1"),
```
StarpTech, line 11: We don't want to have 100% sample rate by default. This can be dangerous in production. Please disable tracing by default.

studio/src/instrumentation.ts
StarpTech, line 1: Is this file loaded by convention?

studio/next.config.mjs
```diff
@@ -81,4 +100,60 @@ const config = {
   },
 };
 
-export default withMarkdoc({ mode: "static" })(config);
+const withOptionalSentryConfig = (org, project, config) =>
```
StarpTech, line 103: What impact does Sentry have on the page sizes / bundle?

studio/Dockerfile
```diff
@@ -19,6 +19,23 @@ COPY studio/.env.docker studio/.env.production
...
+ARG SENTRY_ENABLED
+ARG SENTRY_DEBUG
+ARG SENTRY_AUTH_TOKEN
+ARG SENTRY_ORGANIZATION
+ARG SENTRY_PROJECT
+ARG SENTRY_SEVER_SAMPLE_RATE
+ARG SENTRY_EDGE_SAMPLE_RATE
+
+ENV SENTRY_ENABLED $SENTRY_ENABLED
+ARG SENTRY_DEBUG $SENTRY_DEBUG
```
StarpTech, line 32: ARG intentionally?

studio/next.config.mjs
```diff
@@ -1,11 +1,30 @@
 import withMarkdoc from "@markdoc/next.js";
+import { withSentryConfig } from "@sentry/nextjs";
 import pkg from "./package.json" assert { type: "json" };
 
 const isPreview = process.env.VERCEL_ENV === "preview";
 // Allow it only for development once https://github.com/vercel/next.js/issues/23587 is fixed
 const allowUnsafeEval = true;
 // Report CSP violations to the console instead of blocking them
 const debugCSP = false;
```
StarpTech, line 9: Do we need to adjust our CSP policies to allow sentry as trusted origin? You can test this by running the production build.

studio/src/pages/_error.jsx
```diff
@@ -0,0 +1,17 @@
+import * as Sentry from "@sentry/nextjs";
+import Error from "next/error";
+
+const CustomErrorComponent = (props) => {
```
StarpTech, line 4: What we definitly need to do is to add the orgId, userId  to the sentry context after authentication. You get the authenticated user in `app-provider.tsx`

See: https://docs.sentry.io/platforms/javascript/session-replay/configuration/#identifying-users

studio/src/instrumentation.ts
AndreasZeissner: Yes, it's a nextjs convention: https://nextjs.org/docs/app/building-your-application/optimizing/instrumentation

StarpTech: Ok, thanks

studio/next.config.mjs
AndreasZeissner: on main:
```
132M    .next

7.4M    .next/static
6.9M    .next/static/chunks
1.5M    .next/static/chunks/pages

14M     .next/server
13M     .next/server/pages
14M     .next/standalone/studio/
```
on the branch:
```
221M    .next

8.4M    .next/static
7.3M    .next/static/chunks
1.9M    .next/static/chunks/pages

58M     .next/server
25M     .next/server/pages
58M     .next/standalone/studio
```
on the branch with reactComponentAnnotation disabled:
```
220M    .next/

8.2M    .next/static
7.1M    .next/static/chunks
1.7M    .next/static/chunks/pages

58M     .next/server
25M     .next/server/pages
58M     .next/standalone/studio
```
For the static builds it is approx 1mb. 
For server and standalone things increased approx 4 times

StarpTech: We should do more tests to see how it affects UX. From your numbers This looks significant.

AndreasZeissner: The source maps are doing a huge part here especially for server and standalone.
Without the source maps, which are deleted after uploading them to sentry, it ends up with about 300kb.
```
# on branch
7.7M        .next/static

115M        .next/standalone
22M         .next/server
```

```
# on main
7.4M        .next/static

107M        .next/standalone
14M         .next/server
```

Session Replay brings in a lot:
```
7.9M .next/static
```

What about disabling session replay in the first place?

If the outputs we get from Sentry is already giving us some insights about that something is broken, the session replay could be temporarily enabled when hard to reproduce bugs have been reported.

StarpTech: Ok, that makes sense. Sourcemaps are only generated when sentry is enabled right?

AndreasZeissner: Yes, the handling for sentry is part of the nextjs plugin configuration:
```                                                                                                          
    org: org,                                                                                                
    project: project,                                                                                        
    sourcemaps: {                                                                                            
      disable: true,                                                                                         
      deleteSourcemapsAfterUpload: true,                                                                     
    },                                                                                                       
```
It's currently hardcoded I don't see, that we would not use it at the moment.

StarpTech: Ok, thanks

studio/next.config.mjs
AndreasZeissner: Checked this, it needed at least `blob` for the web worker.
`connect-src` can't be enforced atm.

StarpTech: ok

studio/next.config.mjs
```diff
@@ -81,4 +114,77 @@ const config = {
...
+    telemetry: false,
+
+    // Only print logs for uploading source maps in CI
+    silent: !process.env.CI,
+
+    // For all available options, see:
+    // https://docs.sentry.io/platforms/javascript/guides/nextjs/manual-setup/
+
+    // Upload a larger set of source maps for prettier stack traces (increases build time)
+    widenClientFileUpload: true,
```
StarpTech, line 138: How big is the difference in build time? Do we need it at all?

studio/next.config.mjs
```diff
@@ -81,4 +114,77 @@ const config = {
...
+    },
+
+    // Automatically tree-shake Sentry logger statements to reduce bundle size
+    disableLogger: !sentryDebugEnabled,
+
+    // Enables automatic instrumentation of Vercel Cron Monitors. (Does not yet work with App Router route handlers.)
+    // See the following for more information:
+    // https://docs.sentry.io/product/crons/
+    // https://vercel.com/docs/cron-jobs
+    automaticVercelMonitors: true,
```
StarpTech, line 159: We don't use crons. We can disable it.

studio/next.config.mjs
```diff
@@ -81,4 +114,77 @@ const config = {
...
+      sentryProject,
+      config,
+    );
+  }
+  return config;
+};
+
+const nextConfig = {
+  webpack: (config, { webpack }) => {
+    config.plugins.push(
```
StarpTech, line 175: Please add a comment or make a reference why this is needed.

StarpTech (changes requested):

studio/next.config.mjs
AndreasZeissner: I made a test on a vercel test project. The build without the widen client file upload took 5 seconds longer. I would not say this is representative  but, it seems to not influence the build times significantly.

StarpTech: ok, thanks for testing.

StarpTech (approved):
LGTM

PR #1083, opened 2024-08-15
fix: dont initialize persisted operation client when disabled

flymedllva:
## Motivation and Context

In the new version of the router, panic happens in the `persistedoperation` package when a SIGINT signal is received.

### DIFF
--- a/router/core/router.go
+++ b/router/core/router.go
@@ -857,22 +857,24 @@ func (r *Router) buildClients() error {
 		)
 	}
 
-	// For backwards compatibility with cdn config field
-	cacheSize := r.persistedOperationsConfig.Cache.Size.Uint64()
-	if cacheSize <= 0 {
-		cacheSize = r.cdnConfig.CacheSize.Uint64()
-	}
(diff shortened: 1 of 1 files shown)


StarpTech:
HI @flymedllva, thank you for the fix

StarpTech:
In that case, we also shouldn't create the `persistedOperationClient`. Could you ensure that we don't run the [code](https://github.com/wundergraph/cosmo/blob/1b85ec58eb67ffa3c56e65323fa9383ce7564e2e/router/core/router.go#L860-L875), when the pClient is nil?

flymedllva:
There are a few cases where `r.persistedOperationClient` is set, I made it happen only in those cases. The installation will still depend on `r.persistedOperationsConfig`, `r.cdnConfig`. But this seems to solve the SIGINT problem

If you have other ideas on how to solve this better, they would work for me too, it's just that right now when testing I always catch panic when shutting down

StarpTech (approved):
LGTM

PR #1097, opened 2024-08-19
fix: use uuid for config version in compose command

thisisnithin:
## Motivation and Context

Hash breaks controlplane queries if the router config from compose command is used with router. Use uuid which is expected.

## TODO


cli/src/commands/router/commands/compose.ts
```diff
@@ -139,7 +139,7 @@ export default (opts: BaseCommandOptions) => {
       }),
     });
 
-    routerConfig.version = createHash('sha1').update(routerConfig.toJsonString()).digest('hex');
+    routerConfig.version = randomUUID();
```
StarpTech, line 142: This has the effect that the server is always swapped even when the config hasn't changed. I think this is fine but we should also check in the controlplane if version is valid UUID before we use it in the query to avoid such errors in the future.

StarpTech (changes requested):

cli/src/commands/router/commands/compose.ts
thisisnithin: Done

PR #1109, opened 2024-08-22
fix: unicode escaping by replacing fastjson version

devsergiy:
## TODO


jensneuse:
@devsergiy can you add a test that validates that the behavior is "locked in*?

StarpTech (changes requested):
Please add a test case.

StarpTech:
You can take inspiration from 

- https://github.com/golang/go/blob/master/src/encoding/json/encode_test.go#L784
- https://github.com/goccy/go-json/blob/master/encode_test.go#L928

router-tests/testdata/queries/products.json
```diff
@@ -22,7 +22,8 @@
         "clientLanguages": [
           "RUST",
           "TYPESCRIPT"
-        ]
+        ],
+        "unicode": "😀♥!\u0015\u0010🝥"
```
StarpTech, line 26: This won't reproduce the error. You can use `"\u0015"`

StarpTech (changes requested):

router-tests/testdata/queries/products.json
devsergiy: you should have tried
```
Unicode: "\U0001F600\u2665\u0021\u0015\u0010\U0001F765"
```

StarpTech: 👍

StarpTech: I didn't understand the test in the first place. You extended the test query and updated the fixture. Now, I understand. Thanks.

StarpTech (approved):
LGTM

Issue #905, opened 2024-07-01
Support custom  labels for controlPlane's job 


esparta86:
### Component(s)

controlplane, helm

### Is your feature request related to a problem? Please describe.

My company uses Istio mesh, so we usually use labels in the namespace to inject the sidecar istio proxy in each pod.
 helm installs or upgrades are failing since istio-proxy are injecting in pods from Jobs.
 because the pods don't finish in a completed state.

### Describe the solution you'd like

if jobs support custom labels, I am going to have more control which labels are injecting on Job Pods.
It will also help to identify attributes of objects that are meaningful and relevant.
also I can tell to istio controller to don't inject istio-sidecar

### Describe alternatives you've considered

This is my proposal solution. You will notice that I added a new variable called customLabelsJobsControplane
```
# The otelcollector provides the necessary database schema for the migration
{{- if .Values.global.otelcollector.enabled }}
apiVersion: batch/v1
kind: Job
metadata:
  name: "{{ include "controlplane.fullname" . }}-clickhouse-migration-{{ now | unixEpoch }}"
  labels:
    {{- include "controlplane.labels" . | nindent 4 }}
    {{- if .Values.customLabelsJobsControplane }}{{ toYaml .Values.customLabelsJobsControplane | nindent 4 }}{{- end }}
  annotations:
    # Run migration job before helm install/upgrade
    "helm.sh/hook": post-install,post-upgrade
    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
    # Less priority than the migration job so the secret is created before the migration job
    "helm.sh/hook-weight": "-5"
    # Support for k14s.io. This annotation ensure that the job is run after clickhouse deployment
    {{- if .Values.global.clickhouse.enabled }}
    kapp.k14s.io/change-rule.clickhouse: "upsert after upserting cosmo.apps.clickhouse.wundergraph.com/deployment"
    {{- end }}
    {{- if .Values.global.otelcollector.enabled }}
    kapp.k14s.io/change-rule.otelcollector: "upsert after upserting cosmo.apps.otelcollector.wundergraph.com/deployment"
    {{- end }}
    # Create a new change-group so the controlplane can be deployed after the clickhouse migration
    kapp.k14s.io/change-group: "cosmo.apps.controlplane.wundergraph.com/ch-migration"
spec:
  backoffLimit: 10
  parallelism: 1
  template:
    metadata:
      name: "{{ include "controlplane.fullname" . }}-clickhouse-migration"
[...]

PR #1112, opened 2024-08-23
feat(controlplane): enable custom job labels and pass common labels 

AndreasZeissner:
## TODO


helm/cosmo/values.yaml
```diff
@@ -98,6 +98,51 @@ cdn:
...
+  commonLabels: {}
+
+  # -- Configure jobs to be executed in the control plane
+  jobs:
+    # -- Used to deactivate an organization with a reason and schedule deletion
+    deactivateOrganization:
+      # -- Enables the job to be run
+      enabled: false
+      # -- Adds custom labels to the job
+      customLabels: {}
```
StarpTech, line 111: Two popints

1. I'd prefer just `labels`
2. In https://github.com/wundergraph/cosmo/issues/905 the user wanted to apply labels to jobs to avoid sidecar injection. In the current setup, he would need to copy/paste it to all jobs. It would be great if the `jobs` section would have a `commonLabels` section as well.

AndreasZeissner: Thanks for the feedback: 

1. Thought about this just kept it to note it's something additional, the default labels will still be applied, this might lead to misunderstandings. I will check how this works out with number 2. 

2. I thought about moving stuff e.g. also to the global part. Seed is not really a core function, deleting users or disabling users is part of running cosmo and therefore a job 'core' function. I will check how to split it up for easier handling. (Opinion) Even though I think it's a Helm problem and we should avoid having such logic built into the charts, it's just one of the jobs, each of those might be handled differently and you have to define this transparently for each of them.

StarpTech: 1. So the labels will be merged? `additionalLabels` ?
2. ok, sounds good

AndreasZeissner: Let's go like this:

```
# -- Pass additional labels to all jobs
additionalJobLabels: {}

# -- Configure jobs to be executed in the control plane
jobs:
  # -- Used to deactivate an organization with a reason and schedule deletion
  deactivateOrganization:
    ...
    # -- Adds additional labels to the job
    additionalLabels: {}
    ...
```
So it's clear it get's added.

thisisnithin (approved):

PR #1115, opened 2024-08-25
chore(rfc): operation cache warmer

StarpTech:
## TODO

### DIFF
--- a/rfc/distributed-operation-cache.md
+++ b/rfc/distributed-operation-cache.md
@@ -64,7 +64,7 @@ The Top-N computation is done for a specific time interval e.g. 3-72 hour (confi
 
 ### Cache update process
 
-The router checks periodically for updates of the distributed operation cache. The cache is updated explicitly when the router starts and when the schema changes. The cache is loaded and all operations are pre-planned before the router accepts traffic. The cache is updated in the background and doesn't block the router from accepting traffic.
+The router checks periodically e.g. every 5min for updates of the distributed operation cache. The cache is checked explicitly when the router starts and when the schema changes. The cache is loaded and all operations are pre-planned before the router accepts traffic. The cache is updated in the background and doesn't block the router from accepting traffic.
 
 ### Platform integration
 
--- a/rfc/distributed-operation-cache.md
+++ b/rfc/distributed-operation-cache.md
@@ -37,9 +37,7 @@ The CLI command will add the operations from the file `operations.json` to the d
 [
   // Queries
   {
-    "query": "query { ... }",
-    // Only required for multi operation queries
-    "operationName": "myQuery"
+    "body": "query { ... }"
   },
   // Persisted operation
   {
--- a/rfc/distributed-operation-cache.md
+++ b/rfc/distributed-operation-cache.md
@@ -47,7 +47,7 @@ The CLI command will add the operations from the file `operations.json` to the d
 ]
 ```
 
-The update operation is idempotent and always updates the cache with the latest operations. This doesn't trigger the computation of the Top-N operations which is done periodically by the Cosmo Platform.
+The cli command is idempotent and always updates the cache with the latest operations. This doesn't trigger the computation of the Top-N operations which is done periodically by the Cosmo Platform.
 
 ### Automatic operation computation
 
--- a/rfc/distributed-operation-cache.md
+++ b/rfc/distributed-operation-cache.md
@@ -53,7 +53,11 @@ The update operation is idempotent and always updates the cache with the latest 
 
 ### Automatic operation computation
 
-At the same time, WunderGraph Cosmo is analyzing the incoming traffic based on the OpenTelemetry metrics that each router is sending. The Cosmo Platform computes the Top-N operations for each graph and combines it with the manually added operations. The Top-N operations are then pushed to the distributed operation cache of the graph. On each router startup and schema change, the cache is loaded and all operations are pre-planned before the router accepts traffic. For containerized environments like Kubernetes, users should use the readiness probe to ensure that the router is ready to accept traffic. For schema updates after startup, this process is transparent to the environment and clients because the new Graph schema isn't swapped until the cache is warmed up.
+At the same time, WunderGraph Cosmo is analyzing the incoming traffic based on the OpenTelemetry metrics that each router is sending. The Cosmo Platform computes the Top-N operations for each graph and combines it with the manually added operations. The Top-N operations are then pushed to the distributed operation cache of the graph. On each router startup and schema change, the cache is loaded and all operations are pre-planned before the router accepts traffic.
+
+### Platform integration
+
+For containerized environments like Kubernetes, users should use the readiness probe to ensure that the router is ready to accept traffic. Setting not to small values for the readiness probe timeout is recommended to ensure that the router has enough time to prepare the cache. For schema updates after startup, this process is transparent to the environment and clients because the new Graph schema isn't swapped until the cache is warmed up.
 
 ### Top-N computation
 
--- a/rfc/distributed-operation-cache.md
+++ b/rfc/distributed-operation-cache.md
@@ -87,4 +87,7 @@ version: "1"
 
 cache_warmup:
   enabled: true
+  interval: 5m
 ```
+
+_For this RFC, we only consider support for the WunderGraph Cosmo CDN._
--- a/rfc/distributed-operation-cache.md
+++ b/rfc/distributed-operation-cache.md
@@ -1,0 +1,83 @@
+---
+title: "Distributed Operation Cache"
+author: Dustin Deus
+date: 2024-08-25
+status: Draft
+---
+
+# Distributed Operation Cache
+
+- **Author:** Dustin Deus
+- **Date:** 2024-08-25
+- **Status:** Draft
+
+## Abstract
+
+This RFC describes a new feature to reduce the latency of the system by pre-planning the most expensive and requested operations before the router accepts traffic. We archive this by computing the Top-N GraphQL operations available and making them available to all routers instances.
+
+## Motivation
+
+GraphQL is a powerful tool to query data from a server. However, the flexibility of the query language comes with a cost. The cost is the complexity of the query and how expensive it is to normalize, plan and execute it. While execution performance is predictable and primarily a concern of the underlying subgraphs, the planning phase can be a significant latency contributor. The distributed operation cache aims to reduce this latency by pre-planning the most expensive and requested operations ahead to make it invisible to the user.
+
+# Proposal
+
+The distributed operation cache is semi-automatic and allows the user to push specific operations to the cache but also automatically computes the most expensive and requested operations of the last time frame (configurable). The cache has a fixed size of operations e.g. 100 (configurable) and is shared across all router instances. An operation can be a regular query, subscription, mutation or persisted operation. When the cache capacity is reached, manual operations have a higher priority than automatic operations. This allows users to manage the priority of operations in the cache themselves.
+
+### Pushing operations to the cache
+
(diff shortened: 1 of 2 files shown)


rfc/distributed-operation-cache.md
```diff
@@ -0,0 +1,107 @@
+---
+title: "Distributed Operation Cache"
```
jensneuse, line 2: I'd call this "Cache Warming" or similar.
This is not about caching and it's not distributed,
which might be misunderstood as we're planning to add "response" caching, which serves a completely different purpose.

The purpose of this RFC is solely to warm up all existing caches (normalization, validation, planning)

rfc/distributed-operation-cache.md
```diff
@@ -0,0 +1,107 @@
...
+# Proposal
+
+The distributed operation cache is semi-automatic and allows the user to push specific operations to the cache but also automatically computes the most expensive and requested operations of the last time frame (configurable). The cache has a fixed size of operations e.g. 100 (configurable) and is shared across all router instances. An operation can be a regular query, subscription, mutation or persisted operation. When the cache capacity is reached, manual operations have a higher priority than automatic operations. This allows users to manage the priority of operations in the cache themselves.
+
+### Pushing operations to the cache
+
+The User can push individual operations to the distributed operation cache by using the CLI:
+
+```bash
+wgc router cache add -g mygraph operations.json
```
jensneuse, line 31: Should this be under the namespace "router"?
At which level do we apply cache warming?
Org? Namespace? Graph? Router?

As an alternative, you could consider creating a "cache-warming-bucket", which you can then use to add operations, and which can be used in the router config, e.g.:

```yaml
version: "1"

cache_warmup:
  enabled: true
  interval: 5m
  bucket: myBucket
```

I'm not sure though if this is the right direction.
The cache warming bucket is somewhat closely related to a Graph because the containing operations need to be compatible with the client schema of the graph.

rfc/distributed-operation-cache.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Abstract
+
+This RFC describes a new feature to reduce the latency of the system by pre-planning the most expensive and requested operations before the router accepts traffic. We achieve this by computing the Top-N GraphQL operations available and making them available to all routers instances before they accept traffic.
+
+## Motivation
+
+GraphQL is a powerful tool to query data from a server. However, the flexibility of the query language comes with a cost. The cost is the complexity of the query and how expensive it is to normalize, plan and execute it. While execution performance is primarily a concern of the underlying subgraphs, the planning phase can be a unpredictable and significant latency contributor. The distributed operation cache aims to reduce this latency by pre-planning the most expensive and requested operations ahead to make it invisible to the user.
+
+# Proposal
```
jensneuse, line 22: One general comment: Cache warming should allow for Operations to be incompatible with the client schema.
This could result in a warning, but should not raise an error.

rfc/distributed-operation-cache.md
```diff
@@ -0,0 +1,107 @@
...
+  },
+  // Persisted operation
+  {
+    "sha256Hash": "1234567890",
+    "body": "query { ... }",
+  }
+]
+```
+
+The cli command is idempotent and always updates the cache with the latest operations. This doesn't trigger the computation of the Top-N operations which is done periodically by the Cosmo Platform.
```
jensneuse, line 50: Another aspect to think about is pre-compiling operation plans.
We could make plans serializable, generate and serialize them at composition time.
Once the router wants to "load" them, it doesn't have to plan everything but can load the plan much faster.

rfc/distributed-operation-cache.md
StarpTech: I see where the confusing is coming from. With `distributed` I mean the mechanism is applied to all router instances and each

rfc/distributed-operation-cache.md
StarpTech: Done.

rfc/distributed-operation-cache.md
StarpTech: Renamed

rfc/operation-cache-warmer.md
```diff
@@ -0,0 +1,107 @@
...
+# Proposal
+
+The distributed operation cache is semi-automatic and allows the user to push specific operations to the cache but also automatically computes the most expensive and requested operations of the last time frame (configurable). The cache has a fixed size of operations e.g. 100 (configurable) and is shared across all router instances. An operation can be a regular query, subscription, mutation or persisted operation. When the cache capacity is reached, manual operations have a higher priority than automatic operations. This allows users to manage the priority of operations in the cache themselves. It is possible that operations aren't compatible with all future schema changes. In that case, the operation is removed from the cache.
+
+### Pushing operations to the cache
+
+The User can push individual operations to the operation cache by using the CLI:
+
+```bash
+wgc federated-graph operation-cache add --graph mygraph --file operations.json
```
vasylruban-nw, line 31: Curious about manual cache invalidation.

What will happen in the following scenario:

First, we run:

```
wgc federated-graph operation-cache add --graph mygraph --file operationA.json
```

Then we run:
```
wgc federated-graph operation-cache add --graph mygraph --file operationB.json
```

What will we have in the cache after the second operation? Will it be just Operation B or both Operation A and Operation B? 

Curious, because if it is both Operation A and Operation B, the cache eventually will be filled with the manual operation and there will not be space for automatic operations unless we can invalidate it explicitly.

vasylruban-nw, line 31: I wonder if this command could be executed automatically (like part of CI/CD) or if we have to call it manually?

rfc/operation-cache-warmer.md
```diff
@@ -0,0 +1,107 @@
...
+
+The router checks periodically e.g. every 5min for updates of the operation cache. The cache is checked explicitly when the router starts and when the schema changes. The cache is loaded and all operations are pre-planned before the router accepts traffic. The cache is updated in the background and doesn't block the router from accepting traffic.
+
+### Platform integration
+
+For containerized environments like Kubernetes, users should use the readiness probe to ensure that the router is ready to accept traffic. Setting not to small values for the readiness probe timeout is recommended to ensure that the router has enough time to prepare the cache. For schema updates after startup, this process is non-blocking because the new graph schema isn't swapped until the cache is warmed up.
+
+### Cosmo UI integration
+
+A User can disable the operation cache in the Cosmo UI. The User can see the current operations in the cache and remove them if necessary. The User can also see the current status of the cache and the last computation time.
```
vasylruban-nw, line 89: Dumb question: 

what will happen if we disable the operation cache in the Cosmo UI but keep `cache_warmup` enabled in the router configuration?

vasylruban-nw (commented):
Thank you so much for drafting this proposal, it looks great, can't wait to see it live! 

Left a few questions I had during reading.

rfc/operation-cache-warmer.md
```diff
@@ -0,0 +1,107 @@
...
+At the same time, WunderGraph Cosmo is analyzing the incoming traffic based on the OpenTelemetry metrics that each router is sending. The Cosmo Platform computes the Top-N operations for each graph and combines it with the manually added operations. The Top-N operations are then pushed to the operation cache of the graph.
+
+### Top-N computation
+
+The Top-N computation is based on the following metrics:
+
+- Total operation pre-execution time: Normalization, Validation, Planning
+- Total request count
+
+The Top-N computation is done for a specific time interval e.g. 3-72 hour (configurable). The operations are sorted by the pre-execution time and request count. The Top-N operations are then pushed to the operation cache. Manual operations have a higher priority than automatic operations. This means when the cache capacity is reached, manual operations are moved to the cache first and automatic operations are removed.
```
joshlevinson, line 63: How will Top-N prioritize between request count and pre-execution time? I could see wanting to prioritize slowest planned queries and then sort by request count or vice versa.

rfc/operation-cache-warmer.md
```diff
@@ -0,0 +1,107 @@
...
+
+Alternatively, the user can add three more manual operations to the cache until the cache capacity is reached. This has the effect that no automatic operations can be added to the cache. In that case, we assume that the user knows better which operations are important.
+
+### Cache update process
+
+The router checks periodically e.g. every 5min for updates of the operation cache. The cache is checked explicitly when the router starts and when the schema changes. The cache is loaded and all operations are pre-planned before the router accepts traffic. The cache is updated in the background and doesn't block the router from accepting traffic.
+
+### Platform integration
+
+For containerized environments like Kubernetes, users should use the readiness probe to ensure that the router is ready to accept traffic. Setting not to small values for the readiness probe timeout is recommended to ensure that the router has enough time to prepare the cache. For schema updates after startup, this process is non-blocking because the new graph schema isn't swapped until the cache is warmed up.
```
vasylruban-nw, line 85: To double-check: 
`/health/ready` will be successful only when the router is warmed up, right?

StarpTech: That's right!

rfc/operation-cache-warmer.md
StarpTech: The ultimate goal is to prepare operations in advance, focusing on those that take the most time to plan. We will only sort by request count when two items have the same pre-execution time.

rfc/operation-cache-warmer.md
StarpTech: In that case, Cosmo would no longer compute the latest TopN operations for you and your router will fetch at some point an outdated list of operations.

rfc/operation-cache-warmer.md
StarpTech: Yes, the idea is to run this locally as well as part of the CI/CD process.

StarpTech: This is right. Both operations will be in the "batch". This is the part where a customer can manage the cache themself and is responsible of it. At some point, we could also introduce a percentage limit about how much space can be reserved for manual pushed operations. In the first version, we want to keep it simple.

jensneuse: Good point. I'd recommend to make the operation idempotent so that we can run this over and over again on each deployment without having duplicates.

rfc/operation-cache-warmer.md
jensneuse: I'd say that in this case, the Router would try to fetch operations for warmup, but the CP won't return anything, so nothing will be warmed up.

rfc/operation-cache-warmer.md
jensneuse: I think @joshlevinson is right in that we should prioritize by planning time and then request count.

rfc/operation-cache-warmer.md
StarpTech: The latest batch of operations has been pushed on the S3. The router will fetch for it on startup and schema changes. If nothing changes, the router will always fetch the latest available batch. There is no interaction with the controlplane. This is one possibility. It would be the "best-effort" approach because there might be operations can will benefit from the stale cache. Another solution is to delete the cache after the user has disabled the functionality, in that case the router won't find any artifact and skip the warm up.

rfc/operation-cache-warmer.md
StarpTech: That was the idea. Operations are identified by the hash and appended to the "batch" until the cache is full. A user can repush the changes in another pipeline run to re-prioritize it.

rfc/operation-cache-warmer.md
jensneuse: I think it's reasonable to delete the warming data in S3 when the feature is disabled.

StarpTech: I'm fine with both ways. 👍

StarpTech:
The feature has been implement and released.

PR #1124, opened 2024-08-26
feat: enable proxy handling for native fetch and connectrpc calls

AndreasZeissner:
## Motivation and Context

Setting the proxy relies on undici for now undici does not respect the HTTP(S)_PROXY vars, it's planned to build it in from v23 on by default:

https://github.com/nodejs/node/issues/43187#issuecomment-2090243675

> My current plan is to add support for the HTTP_PROXY env variable
behind a flag in v22 (and possibly v20), then unflag in v23.

Using global dispatcher is the current way to go for native fetch,
      therefore undici needs to be added.

For connectrpc the agent will be passed down the chain:

```
      /**
       * Options passed to the request() call of the Node.js built-in
       * http or https module.
       */
      nodeOptions?:
        | Omit<http.RequestOptions, "signal">
        | Omit<https.RequestOptions, "signal">;
```
 
proxy

Depending on the proxy, setting: `export NODE_TLS_REJECT_UNAUTHORIZED=0` might be needed (e.g. calling the wgc npm "update" command)

## TODO


cli/src/core/client/client.ts
```diff
@@ -20,7 +22,9 @@ export const CreateClient = (opts: ClientOptions): Client => {
 
     // You have to tell the Node.js http API which HTTP version to use.
     httpVersion: '1.1',
-
+    nodeOptions: {
+      agent: opts.proxyUrl ? new HttpsProxyAgent(opts.proxyUrl) : undefined,
```
Aenimus, line 26: Is it OK that this will propagate an agent property no matter what (even if it's undefined)?

If you don't want to propagate the agent, you might consider ```...(opts.proxyUrl ? { agent: new HttpsProxyAgent(opts.proxyUrl) : {})```

Sometimes this matters.

cli/test/proxy.test.ts
```diff
@@ -0,0 +1,103 @@
...
+  vi.resetAllMocks();
+
+  delete process.env.HTTPS_PROXY;
+  delete process.env.HTTP_PROXY;
+};
+
+beforeEach(prepare);
+
+describe('fetch when the HTTP_PROXY variable is set', () => {
+  test('when using HTTP_PROXY the ProxyAgent for native fetch was configured', async () => {
```
Aenimus, line 38: `test` should read like english, e.g., `test('that when...')`.

We don't tend to use `it`, but if it were `it`: `it('works when...')`

Issue #983, opened 2024-07-30
S3StorageURL Authorization header is malformed; the region 'auto' is wrong.


Sam-tesouro:
### Component(s)

controlplane

### Component version

controlplane@0.97.1

### wgc version

wgc@0.58.5

### controlplane version

controlplane@0.97.1

### router version

N/A

## Description
Targeting s3 for CDN and ControlPlane with URL format https://ID:SECRET@s3.[REGION].amazonaws.com/[BUCKET-NAME] leads to

`"stack":"AuthorizationHeaderMalformed: The authorization header is malformed; the region 'auto' is wrong; expecting '[REGION]'"`

## Steps to Reproduce
https://github.com/Sam-tesouro/cosmo-controlplane-s3-url-issue/tree/main

## Expected Result
Successfully list buckets

## Actual Result
```
node_modules/@smithy/smithy-client/dist-cjs/index.js:839
  const response = new exceptionCtor({
                   ^
AuthorizationHeaderMalformed: The authorization header is malformed; the region 'auto' is wrong; expecting 'us-east-1'
    at throwDefaultError
```

## Environment
OS: (e.g., "Ubuntu 20.04")
Package Manager: pnpm, npm, yarn, etc
Compiler(if manually compiled): (e.g., "go 14.2")

### Additional context

The s3 client is forcing [path style urls](https://github.com/wundergraph/cosmo/blob/9cf37161de151a719c752f29572819b89a265063/controlplane/src/core/build-server.ts#L327) which were originally slated for deprecation 4 years ago. https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/ 

Is there any reason why you aren't supporting [Virtual-hosted–style requests](https://docs.aws.amazon.com/AmazonS3/latest/userguide/VirtualHosting.html#virtual-hosted-style-access)?

Regardless I suspect I am just missing the correct incantation for an s3 storage url for Cosmo. After trying many different permutations I am hoping it's simple and I can save others the headache in the future with this issue!

Thank you!

AndreasZeissner:
Hi @Sam-tesouro,

I wanted to let you know that we’ve recently added support for AWS S3 URLs.

You might want to have a look here:

helm-cosmo@v0.11.0:
[GitHub Release](https://github.com/wundergraph/cosmo/releases/tag/helm-cosmo%400.11.0)
[...]

PR #1135, opened 2024-09-02
feat: make cosmo usable with aws s3 

AndreasZeissner:
## Motivation and Context

This enables virtual-host based s3 paths to be used for the `S3_STORAGE_URL` in the controlplane and the cdn.
When using aws ensure to set the `S3_REGION` e.g.:

`S3_REGION`:  us-east-1
`S3_STORAGE_URL`: https://cosmo-bucket.s3.amazonaws.com

By default, the username and password is taken from the url itself.

e.g.: https://username:password@minio:9000/cosmo

This is done for backwards compatibility and can be used by default.

When needed it's possible to pass: `S3_ACCESS_KEY_ID` and  `S3_SECRET_ACCESS_KEY` for credentials management. 
The `S3_ENDPOINT` is usually built out of the url.orgin but can be passed if needed.

rel: https://github.com/wundergraph/cosmo/issues/983

## TODO


controlplane/src/core/util.ts
```diff
@@ -371,3 +373,46 @@ export function getValueOrDefault<K, V>(map: Map<K, V>, key: K, constructor: ()
...
+
+  if (isVirtualHostStyleUrl(url)) {
+    return url.hostname.split('.')[0];
+  }
+
+  // path based style
+  return url.pathname.slice(1);
+}
+
+export function isVirtualHostStyleUrl(url: URL) {
```
thisisnithin, line 416: Should we put these into `shared` package?

thisisnithin (changes requested):

controlplane/src/core/util.ts
AndreasZeissner: Ah yes we should, thanks for the hint, will move it.

thisisnithin (approved):

PR #1141, opened 2024-09-04
chore: improve mem usage for attributes, single flight

jensneuse:


go.work
```diff
@@ -12,3 +12,4 @@ use (
 )
 
 //replace github.com/wundergraph/graphql-go-tools/v2 => ../graphql-go-tools/v2
+//replace github.com/wundergraph/astjson => ../astjson
```
Aenimus, line 15: Did you intend to leave this in?

jensneuse: yes, I think it's useful in case someone wants to work on astjson, similarly to go tools

PR #1143, opened 2024-09-04
feat:  Add structured logging testing

df-wg:
## Motivation and Context
As we add in further structured logging, introducing a way to validate logs is critical to testing the full scope of the code. This code is inspired by the work of @Noroth

Completes ENG-5611

## TODO

### DIFF
--- a/router-tests/integration_test.go
+++ b/router-tests/integration_test.go
@@ -4,8 +4,6 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
-	"go.uber.org/zap"
-	"go.uber.org/zap/zapcore"
 	"io"
 	"math/rand"
 	"net/http"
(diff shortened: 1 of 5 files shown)


router-tests/integration_test.go
```diff
@@ -365,6 +367,25 @@ func prettifyJSON(t *testing.T, jsonStr string) string {
 	return res.String()
 }
 
+func TestQueryWithLogging(t *testing.T) {
```
StarpTech, line 370: Please put this in a separate file for more visibility and better test structure.

router-tests/integration_test.go
```diff
@@ -365,6 +367,25 @@ func prettifyJSON(t *testing.T, jsonStr string) string {
...
+		Enabled:  true,
+		LogLevel: zapcore.InfoLevel,
+	}}, func(t *testing.T, xEnv *testenv.Environment) {
+		res := xEnv.MakeGraphQLRequestOK(testenv.GraphQLRequest{
+			Query: `{ employees { id } }`,
+		})
+		require.JSONEq(t, employeesIDData, res.Body)
+		logEntries := xEnv.Observer().All()
+		require.Len(t, logEntries, 12)
+		natsLogs := xEnv.Observer().FilterMessageSnippet("Nats").All()
```
StarpTech, line 382: Why do we produce nats logs when nats isn't enabled? 🤔

router-tests/integration_test.go
StarpTech, line 370: There are defaults logs that are printed when starting the router in INFO level. It would be a great test case to cover.

router-tests/integration_test.go
```diff
@@ -365,6 +367,25 @@ func prettifyJSON(t *testing.T, jsonStr string) string {
...
+		require.Len(t, natsLogs, 4)
+		providerIDFields := xEnv.Observer().FilterField(zap.String("providerID", "default")).All()
+		require.Len(t, providerIDFields, 1)
```
StarpTech, line 385: It would be great to verify the request logs after the request.

StarpTech (changes requested):

router-tests/integration_test.go
df-wg: They are all logs of `Nats Event Source enabled`, so I'm assuming it's from the config

StarpTech (approved):
LGTM

PR #1153, opened 2024-09-09
feat: Add Max Query Depth

df-wg:
## Motivation and Context
A security concern raised is that users can submit massive queries that would cause memory issues or degradation of the service. This PR implements an optional security configuration, `maxQueryDepth`, that users can use to configure a max query depth that would be allowed. 

```
# config.yaml

security:
  max_query_depth: 7
```

## TODO


router/core/graphql_prehandler.go
```diff
@@ -597,6 +602,36 @@ func (h *PreHandler) handleOperation(req *http.Request, buf *bytes.Buffer, httpO
...
+		_, maxQueryDepthSpan := h.tracer.Start(req.Context(), "Operation - Checking Query Depth",
+			trace.WithSpanKind(trace.SpanKindInternal),
+			trace.WithAttributes(otel.WgEngineRequestTracingEnabled.Bool(httpOperation.traceOptions.Enable)),
+			trace.WithAttributes(attributes...),
+		)
+
+		operation := opContext.preparedPlan.operationDocument
+		definition := opContext.preparedPlan.schemaDocument
+		report := operationreport.Report{}
+		globalComplexityResult, _ := operation_complexity.CalculateOperationComplexity(operation, definition, &report)
```
StarpTech, line 619: We do parse and plan the operation only once. We should do this for complexity as well.

router-tests/integration_test.go
```diff
@@ -965,6 +965,52 @@ func TestBlockNonPersistedOperations(t *testing.T) {
 	})
 }
 
+func TestMaxQueryDepth(t *testing.T) {
```
StarpTech, line 968: This should also work for persisted operation. Please add a test for this.

router/core/graphql_prehandler.go
```diff
@@ -597,6 +602,36 @@ func (h *PreHandler) handleOperation(req *http.Request, buf *bytes.Buffer, httpO
 
 	httpOperation.traceTimings.EndPlanning()
 
+	/**
+	* Validate that the planned query doesn't exceed the maximum query depth configured
+	 */
+
+	if h.maxQueryDepth > 0 {
+		_, maxQueryDepthSpan := h.tracer.Start(req.Context(), "Operation - Checking Query Depth",
```
StarpTech, line 610: I'm open for this but in my opinion it is not a good idea to create a span for this prematurely. You can add the error to the root span.

router/core/graphql_prehandler.go
```diff
@@ -597,6 +602,36 @@ func (h *PreHandler) handleOperation(req *http.Request, buf *bytes.Buffer, httpO
...
+			trace.WithAttributes(attributes...),
+		)
+
+		operation := opContext.preparedPlan.operationDocument
+		definition := opContext.preparedPlan.schemaDocument
+		report := operationreport.Report{}
+		globalComplexityResult, _ := operation_complexity.CalculateOperationComplexity(operation, definition, &report)
+
+		if globalComplexityResult.Depth > h.maxQueryDepth {
+			err := &httpGraphqlError{
```
StarpTech, line 622: You also need to the set the finErr so the root span is marked as failed.

StarpTech (changes requested):

router/core/graphql_prehandler.go
df-wg: Any error returned from this function is set as the finErr

router/core/graph_server.go
```diff
@@ -393,6 +397,18 @@ func (s *graphServer) buildGraphMux(ctx context.Context,
 		}
 	}
 
+	if s.securityConfiguration.MaxQueryDepth > 0 {
+		queryDepthCacheConfig := &ristretto.Config[uint64, bool]{
+			MaxCost:     1024,
```
df-wg, line 402: This currently is a magic number. Should I add the ability to configure the cache size to the config? If so, I'll do that there, similar to how the validation cache sizing works

router/core/graphql_prehandler.go
```diff
@@ -597,6 +600,15 @@ func (h *PreHandler) handleOperation(req *http.Request, buf *bytes.Buffer, httpO
 
 	httpOperation.traceTimings.EndPlanning()
 
+	/**
+	* Validate that the planned query doesn't exceed the maximum query depth configured
+	 */
+
+	queryDepthErr := operationKit.ValidateQueryDepth(h.maxQueryDepth, opContext.preparedPlan.operationDocument, opContext.preparedPlan.schemaDocument)
```
df-wg, line 607: It's difficult to test the caching, because there's no span for us to set it as an attribute (which is how we test the validation-cache). Should I add a span for that, or should we expose it as a header, like we do with the Execution Plan Cache? 

https://github.com/wundergraph/cosmo/blob/06fd0e93bdeff2e9cecf7e39cf9db5c8e219a8df/router/core/graphql_handler.go#L39

router/pkg/config/config.schema.json
```diff
@@ -1365,6 +1365,11 @@
           "type": "boolean",
           "default": false,
           "description": "Block non-persisted Operations. If the value is true, the non-persisted operations are blocked."
+        },
+        "max_query_depth": {
+          "type": "integer",
+          "description": "The max depth limit for query. This limit prevents infinite querying, and also limits the size of the data returned. If the limit is 0, this limit isn't applied.",
```
jensneuse, line 1371: Can you describe more specifically what depth means?

router/core/graphql_prehandler.go
jensneuse: Either is fine

router/core/graph_server.go
jensneuse: Please make it configurable, yes.
1024 as default sounds ok.

router-tests/integration_test.go
```diff
@@ -965,6 +965,72 @@ func TestBlockNonPersistedOperations(t *testing.T) {
...
+			ModifySecurityConfiguration: func(securityConfiguration *config.SecurityConfiguration) {
+				securityConfiguration.MaxQueryDepth = 2
+			},
+		}, func(t *testing.T, xEnv *testenv.Environment) {
+			header := make(http.Header)
+			header.Add("graphql-client-name", "my-client")
+			res, _ := xEnv.MakeGraphQLRequestOverGET(testenv.GraphQLRequest{
+				OperationName: []byte(`Find`),
+				Variables:     []byte(`{"criteria":  {"nationality":  "GERMAN"   }}`),
+				Extensions:    []byte(`{"persistedQuery": {"version": 1, "sha256Hash": "e33580cf6276de9a75fb3b1c4b7580fec2a1c8facd13f3487bf6c7c3f854f7e3"}}`),
```
jensneuse, line 1025: I think we should add a boolean flag that disables depth limiting for persisted operations.
The reasoning behind this could be that persisted operations are stored intentionally.
As such, the user might want to skip depth limiting for such cases.
However, it's also possible that the platform team doesn't trust their users, so they might still want to enable depth limiting for POs.

jensneuse (approved):
Overall looks good. Just a few minor remarks.

router/core/graphql_prehandler.go
StarpTech: Great, I couldn't see this in the diff.

router/core/graphql_prehandler.go
StarpTech, line 607: Depth analysis should be done before we planning to prevent attacks and computation.

StarpTech (changes requested):

router/core/graphql_prehandler.go
df-wg: @StarpTech how can I get the operation/query document before the plan? Currently I'm getting the `*ast.Documents` necessary from the `opContext`

StarpTech: You would need to parse it before https://github.com/wundergraph/cosmo/blob/41cbcad7e28b500b9ad08d92d3327694a579bd4c/router/core/operation_planner.go#L66 and the routerSchema is available globally.

jensneuse: no, it's already parsed
take a look at how normalization or other validation steps use the operation document

StarpTech: Right, it is already populated in `operationKit.kit.doc` after parsing.

router/pkg/config/config.go
```diff
@@ -230,9 +230,12 @@ type EngineExecutionConfiguration struct {
...
-	BlockSubscriptions          bool `yaml:"block_subscriptions" envDefault:"false" env:"SECURITY_BLOCK_SUBSCRIPTIONS"`
-	BlockNonPersistedOperations bool `yaml:"block_non_persisted_operations" envDefault:"false" env:"SECURITY_BLOCK_NON_PERSISTED_OPERATIONS"`
+	BlockMutations                       bool  `yaml:"block_mutations" envDefault:"false" env:"SECURITY_BLOCK_MUTATIONS"`
+	BlockSubscriptions                   bool  `yaml:"block_subscriptions" envDefault:"false" env:"SECURITY_BLOCK_SUBSCRIPTIONS"`
+	BlockNonPersistedOperations          bool  `yaml:"block_non_persisted_operations" envDefault:"false" env:"SECURITY_BLOCK_NON_PERSISTED_OPERATIONS"`
+	MaxQueryDepth                        int   `yaml:"max_query_depth,omitempty" envDefault:"0" env:"SECURITY_MAX_QUERY_DEPTH"`
+	QueryDepthCacheSize                  int64 `yaml:"query_depth_cache_size,omitempty" envDefault:"1024" env:"SECURITY_QUERY_DEPTH_CACHE_SIZE"`
+	DisableDepthLimitPersistedOperations bool  `yaml:"disable_depth_limit_persisted_operations,omitempty" envDefault:"false" env:"SECURITY_DISABLE_DEPTH_LIMIT_PERSISTED_OPERATIONS"`
```
df-wg, line 238: I'm not set on this config value name, it seems both long and a bit confusing; I'd welcome a clearer alternative!

StarpTech: I agree, good call. What about? This also gives room for extensibility.

```yaml
depth_limit:
   enabled: true
   limit: 7
   persisted_operations: false
```

df-wg: @jensneuse  
Makes sense to group all of the configuration once there are a few elements. Should we have an `enabled` flag at all? I was previously just using `limit` to signify if it's enabled, if it's set to a value about 0. 

I'd also have the cache size in that configuration

StarpTech: We use `enabled` everywhere to give the developer an option to disable parts of the config without removing the code.

jensneuse: Yeah, we kinda made it a pattern to have an object with "enabled".

jensneuse (approved):

router/pkg/otel/attributes.go
```diff
@@ -33,6 +33,8 @@ const (
 	WgNormalizationCacheHit            = attribute.Key("wg.engine.normalization_cache_hit")
 	WgValidationCacheHit               = attribute.Key("wg.engine.validation_cache_hit")
 	WgVariablesValidationSkipped       = attribute.Key("wg.engine.variables_validation_skipped")
+	WgQueryDepth                       = attribute.Key("wg.engine.query_depth")
```
StarpTech, line 36: I think `wg.operation.query_depth` is more appropriate in that case.

StarpTech (changes requested):

StarpTech (approved):
LGTM

PR #1155, opened 2024-09-09
feat: add response header propagation

jensneuse:


StarpTech (changes requested):
Quick feedback on the existing functionality: The cache header implementation needs an extensive testsuite. I suggest to split header propagation and cache header handling in two distinct core modules and different configuration. This would improve maintainability and increase awareness of the features. Caching is a complex topic and I expect more customizability in the future here (minTTL, default etc...). A dedicated config section would be great here and allow us to add more in the future.

```yaml
http_cache:
  enabled: true
  defaultTTL: 30m
  minTTL: 5m
```

router/core/header_propagation.go
```diff
@@ -0,0 +1,475 @@
...
+		RespHeaders:            res.Header,
+		RespStatusCode:         res.StatusCode,
+		RespExpiresHeader:      expiresHeader,
+		RespDateHeader:         dateHeader,
+		RespLastModifiedHeader: lastModifiedHeader,
+
+		//CacheIsPrivate: false,
+
+		ReqDirectives: reqDir,
+		ReqHeaders:    res.Request.Header,
```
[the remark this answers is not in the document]
StarpTech: Nevermind, res.Request is the request.

hganganna:
@jensneuse We have the requirement to propagate response header propagation at RCCL ( Royal Caribbean ). We really appreciate the fix coming in. Please let me know timeline on this, so that we can plan accordingly.

router/pkg/config/config.go
```diff
@@ -191,6 +192,36 @@ type RequestHeaderRule struct {
...
+	// Named is the exact header name to match
+	Named string `yaml:"named"`
+	// Rename renames the header's key to the provided value
+	Rename string `yaml:"rename,omitempty"`
+	// Default is the default value to set if the header is not present
+	Default string `yaml:"default"`
+	// Algorithm is the algorithm to use when multiple headers are present
+	Algorithm ResponseHeaderRuleAlgorithm `yaml:"algorithm,omitempty"`
+	// OnlyEnableOnClientRequestGet only propagates the header if the client request is a GET request
+	OnlyEnableOnClientRequestGet bool `yaml:"only_enable_on_client_request_get,omitempty"`
```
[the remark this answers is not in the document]
df-wg: @jensneuse  @StarpTech what should I do here, should they be able to specify the propagation of the rule based on the HTTP method? Only for Gets?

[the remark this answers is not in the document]
jensneuse: From my POV, it only makes sense to propagate Cache Control Headers when the client request method is GET, which our Router already supports. However, it might be the case that someone uses a proxy, like Cloudflare Worker to rewrite POST to GET, so we could also allow them to pass CC headers on non GET. But the default should be to only do it for GET. Alternatively, we could also allow it for all methods and limit a rule to certain methods if specified.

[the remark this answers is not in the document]
StarpTech: Sounds reasonable.

router/core/header_rule_engine.go
```diff
@@ -37,158 +46,472 @@ var (
...
+	if propagation.previousCacheControl == nil {
+		if rule.Default != "" {
+			propagation.previousCacheControl = defaultCacheControlObj
+			propagation.header.Set(cacheControlKey, rule.Default)
+		} else {
+			propagation.previousCacheControl = obj
+			propagation.header.Set(cacheControlKey, res.Header.Get(cacheControlKey))
+			return
+		}
+	} else if rule.Default != "" && isMoreRestrictive(defaultCacheControlObj, propagation.previousCacheControl) {
```
StarpTech, line 446: Maybe I miss something but the cache library has an api to determine the cacheability and expiration time of an object.

```go
cachedirective.CachableObject(obj, &rv)
cachedirective.ExpirationObject(obj, &rv)
```
Don't we only need to compare those values to calculate the more restrictive object?

StarpTech (changes requested):

router/core/header_rule_engine.go
df-wg: @StarpTech That's true, but the cache library doesn't have a direct api to compare two cachable objects, which is why I defined that function. More, when I was debugging it I haven't been able to find a case where the `ObjectResults` actually have relevant information (the Warnings and Reasons are always null in all of our test cases)

router-tests/header_propagation_test.go
```diff
@@ -0,0 +1,521 @@
...
+									Response: []*config.ResponseHeaderRule{
+										getRule(alg, ""),
+									},
+								},
+								"hobbies": {
+									Response: []*config.ResponseHeaderRule{
+										{
+											Operation: config.HeaderRuleOperationPropagate,
+											Algorithm: config.ResponseHeaderRuleAlgorithmMostRestrictiveCacheControl,
+											Default:   "max-age=300",
```
jensneuse, line 470: Please add 3 tests:
1. default no cache, one subgraph returns no headers -> no cache
2. default no cache, both subgraphs return max age, use most restrictive max age
3. default max age 300, both subgraphs return max age 60 -> max age 60

router/core/header_rule_engine.go
```diff
@@ -37,158 +46,472 @@ var (
...
+				request.Header.Set(name, ctx.Request().Header.Get(name))
 			}
 		}
 	}
+}
 
-	return request, nil
+func (h *HeaderPropagation) applyResponseRuleMostRestrictiveCacheControl(res *http.Response, propagation *responseHeaderPropagation, rule *config.ResponseHeaderRule) {
+	ctx := res.Request.Context()
+	tracer := rtrace.TracerFromContext(ctx)
```
jensneuse, line 387: I think we should skip this for mutations.
You can test this via:
```go
resolve.SingleFlightDisallowed(ctx)
```

jensneuse: SingleFlightDisallowed == mutation (naming could be improved)

router-tests/header_propagation_test.go
df-wg: I added them. however, in your case #2 (no cache, while subgraphs return max age), there the cache header will be no-cache, because it's the most restrictive cache option.

Issue #1156, opened 2024-09-09
feat: Make the maximum recycleable parser cache size configurable


jfroundjian:
### Component(s)

router

### Is your feature request related to a problem? Please describe.

Problem:

One of our federated queries is for loading a batch of data >1500 items all with >100 fields. This often results in larger JSON responses that the Resolver ends up parsing. After switching to ParserPools in fastjson (sometime after Router .88.0) we noticed unbounded memory growth when performing a relatively large number of these batch requests (> 150 min) often resulting in unwieldy heap sizes and a struggling GC.  This was not a problem for smaller requests (same Query but with <200 items). 

It turns out that the issue is for our given queries, the ParserPool ends up retaining Parser objects which have allocated arrays of 2^17 (131,072) astjson.Value objects. This makes the Router retain an unusually large Heap that can not be reduce by the GC because the Parsers are being actively used to resolve GraphQL responses. 

I've attached a few pprof dumps from v 0.88.0 and v 0.107.0
[v88.pb.gz](https://github.com/user-attachments/files/16933482/v88.pb.gz)
[v107.pb.gz](https://github.com/user-attachments/files/16933483/v107.pb.gz)

### Describe the solution you'd like

I'm proposing and have implemented a somewhat simple solution to the problem to discard Parser objects that have unusually large caches so that the GC can eventually release that memory back to the Program. 

Additionally, I propose making this a configurable property of the Router and Graphql-go-tools so that the default behavior will continue as is and this value can be set on a per need basis.

I've already tested this locally and it works well for our use case. Here are the 3 PRs I'm proposing:

1. astjson: Add a PutIfSizeLessThan method to ParserPool
https://github.com/wundergraph/astjson/pull/2

2. graphql-go-tools: Add ResolverOption for MaxRecyclableParserSize and use on Resolvable.Reset()
https://github.com/wundergraph/graphql-go-tools/pull/881

3. router: Add ResolverMaxRecyclableParserSize configuration and pass to Resolver on creation
https://github.com/wundergraph/cosmo/pull/1157

jensneuse:
@jfroundjian thanks for raising this, it sounds very reasonable.
I mostly like your proposal, although I'd suggest that we set a default max value to automatically cleanup the heap based on a "reasonable" number of cache objects.

Do you have any suggestion on what this number could be?
[...]

PR #1157, opened 2024-09-09
feat: make the maximum recycleable parser cache size configurable

jfroundjian:
## Motivation and Context
https://github.com/wundergraph/cosmo/issues/1156

## TODO

### DIFF
--- a/router/pkg/config/config.go
+++ b/router/pkg/config/config.go
@@ -227,7 +227,7 @@ type EngineExecutionConfiguration struct {
 	ParseKitPoolSize                       int                      `envDefault:"16" env:"ENGINE_PARSEKIT_POOL_SIZE" yaml:"parsekit_pool_size,omitempty"`
 	EnableValidationCache                  bool                     `envDefault:"true" env:"ENGINE_ENABLE_VALIDATION_CACHE" yaml:"enable_validation_cache"`
 	ValidationCacheSize                    int64                    `envDefault:"1024" env:"ENGINE_VALIDATION_CACHE_SIZE" yaml:"validation_cache_size,omitempty"`
-	ResolverMaxRecyclableParserSize        int                      `envDefault:"0" env:"ENGINE_RESOLVER_MAX_RECYCLABLE_PARSER_SIZE" yaml:"resolver_max_recyclable_parser_size,omitempty"`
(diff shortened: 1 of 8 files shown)


router/pkg/config/fixtures/full.yaml
```diff
@@ -212,6 +212,7 @@ engine:
   epoll_kqueue_conn_buffer_size: 128
   websocket_read_timeout: "5s"
   execution_plan_cache_size: 1024
+  resolver_max_recyclable_parser_size: 32768
```
jensneuse, line 215: Where is this number coming from?

router/pkg/config/config.go
```diff
@@ -227,6 +227,7 @@ type EngineExecutionConfiguration struct {
 	ParseKitPoolSize                       int                      `envDefault:"16" env:"ENGINE_PARSEKIT_POOL_SIZE" yaml:"parsekit_pool_size,omitempty"`
 	EnableValidationCache                  bool                     `envDefault:"true" env:"ENGINE_ENABLE_VALIDATION_CACHE" yaml:"enable_validation_cache"`
 	ValidationCacheSize                    int64                    `envDefault:"1024" env:"ENGINE_VALIDATION_CACHE_SIZE" yaml:"validation_cache_size,omitempty"`
+	ResolverMaxRecyclableParserSize        int                      `envDefault:"0" env:"ENGINE_RESOLVER_MAX_RECYCLABLE_PARSER_SIZE" yaml:"resolver_max_recyclable_parser_size,omitempty"`
```
jensneuse, line 230: I don't think we should keep this to be 0.

jensneuse (changes requested):

router/pkg/config/fixtures/full.yaml
jfroundjian: This is the number of cached json values being parsed by the Parser. The Parser itself doubles the size of the underlying array every time it reaches the limit so a power of 2 is usually a good threshold imho.

router/pkg/config/config.schema.json
```diff
@@ -1550,6 +1550,11 @@
           "type": "integer",
           "default": 1024,
           "description": "The size of the validation cache."
+        },
+        "resolver_max_recyclable_parser_size": {
```
jensneuse, line 1554: Can you move this below the "parsekit_pool_size" item?

jensneuse (changes requested):

jensneuse (approved):

PR #1164, opened 2024-09-11
feat: subgraph error propagation improvements

StarpTech:
## Motivation and Context

- Add additional options to configure error propagation:
  - `default_extension_code`: (Default: `DOWNSTREAM_SERVICE_ERROR`)The default extension code. The default extension code is used to specify the default code for the Subgraph errors when the code is not present.
  - `allowed_extension_fields`:  (Default: `"code"`) The allowed extension fields. The allowed extension fields are used to specify which fields of the Subgraph errors are allowed to be propagated to the client. In development mode, we additionally allow `stacktrace` to be part of the extensions.
  - `attach_service_name`: (Default: `true`) Attach the service name to each Subgraph error. If the value is true (default), the service name will be attached to the Subgraph errors.
- Set proper `application/json` content type in errors that aren't written by the engine.
- Don't set `data:null` for errors that are happening before query execution. This compliant with the [spec](https://spec.graphql.org/draft/#sec-Data).
- Use subgraph name in error messages.
- Deprecate `subgraph_error_propagation.enabled` and make it true by default.

Requires: https://github.com/wundergraph/graphql-go-tools/pull/883

The following configuration can be used to get the most common propagation behaviour:
```yaml
subgraph_error_propagation:
  mode: pass-through
  omit_locations: false
  default_extension_code: DOWNSTREAM_SERVICE_ERROR
  attach_service_name: true
  allowed_extension_fields:
    - "code"
```

## Example

```json
{
    "errors": [
        {
            "message": "error resolving RootFieldThrowsError for Employee 12",
            "path": [
                "employees",
                9,
                "rootFieldThrowsError"
            ],
            "extensions": {
                "code": "ERROR_CODE",
                "serviceName": "employees",
                "statusCode": 200
            }
        }
    ],
    "data": null
}
```

## TODO


router/core/errors.go
```diff
@@ -122,10 +122,14 @@ func propagateSubgraphErrors(ctx *resolve.Context, logger *zap.Logger) {
 // It accepts a graphqlerrors.RequestErrors object and writes it to the response based on the GraphQL spec.
 func writeRequestErrors(r *http.Request, w http.ResponseWriter, statusCode int, requestErrors graphqlerrors.RequestErrors, requestLogger *zap.Logger) {
 	if requestErrors != nil {
-		if statusCode != 0 {
-			w.WriteHeader(statusCode)
-		}
+
 		if r.URL.Query().Has("wg_sse") {
```
jensneuse, line 126: did you not want to fix this? =)

StarpTech: Yes, but in a separate PR. This is already too big.

demo/pkg/subgraphs/employees/subgraph/schema.resolvers.go
```diff
@@ -10,13 +10,23 @@ import (
...
 )
 
 // RootFieldThrowsError is the resolver for the rootFieldThrowsError field.
 func (r *employeeResolver) RootFieldThrowsError(ctx context.Context, obj *model.Employee) (*string, error) {
-	return nil, fmt.Errorf("error resolving RootFieldThrowsError for Employee %d", obj.ID)
+	graphql.AddError(ctx, &gqlerror.Error{
+		Path:    graphql.GetPath(ctx),
+		Message: fmt.Sprintf("error resolving RootFieldThrowsError for Employee %d", obj.ID),
+		Extensions: map[string]interface{}{
+			"code": "ERROR_CODE",
```
Aenimus, line 24: Is this intended or debug code?

router-tests/feature_flags_test.go
```diff
@@ -20,7 +20,7 @@ func TestFeatureFlags(t *testing.T) {
 				Query: `{ employees { id productCount } }`,
 			})
 			require.Empty(t, res.Response.Header.Get("X-Feature-Flag"))
-			require.JSONEq(t, `{"errors":[{"message":"field: productCount not defined on type: Employee","path":["query","employees","productCount"]}],"data":null}`, res.Body)
```
Aenimus, line 23: So if there is a propagated error, the data shouldn't be null—the property should not exist at all?

StarpTech: All errors before query execution e.g parsing, validation should not contain data property. https://spec.graphql.org/draft/#sec-Data

Aenimus (approved):
LGTM!

PR #1169, opened 2024-09-13
chore: setup connect-go proto files

AndreasZeissner:
## Motivation and Context

This is used for https://github.com/wundergraph/terraform-provider-cosmo.

The package can be used like this: 

```
go get github.com/wundergraph/cosmo@638540986b469dxxxxxxxxxxxxxxxxxx
```

## TODO


StarpTech (changes requested):
How do we detected outdated artifacts? Is it running when the controlplane changes?

AndreasZeissner:
Our pipelines will fail when the git directory is dirty. This happens on router builds at least automatically.

Didn't saw this as an artifact, clients using it have to ensure those are up to date which is at the moment the terraform-provider which needs to adapt to whatever changes here.

StarpTech:
Does it mean a workflow will fail when the code becomes outdated in this repo?

AndreasZeissner:
Not necessarily in the terraform provider. It will fail there when e2e tests are failing which are running against latest cosmo. 

I think it's a good idea to keep the files updated in general.
Nevertheless I think no one will do unless you need to, or the pipeline fails. This would be  latest in the router builds as they are running the generate-go command and do the dirty check, even though it's unrelated.

The controlpalne ci seems to be the only place where this would make sense, even when publishing it somewhere later.

Are there any considerations on reusing things on the router?

buf.connect-go.go.gen.yaml
```diff
@@ -0,0 +1,16 @@
+version: v2
+managed:
+  enabled: true
+  disable:
+    - file_option: go_package
+      module: buf.build/googleapis/googleapis
+  override:
+    - file_option: go_package_prefix
+      value: github.com/wundergraph/cosmo/connect-go
```
StarpTech, line 9: Let's stick with the convention of "gen". We do this for router and collector.

StarpTech (approved):
LGTM

PR #1170, opened 2024-09-13
fix: use appVersion by default, optimize keycloak startup time

StarpTech:
## Motivation and Context

This forces an image upgrade on the cluster when someone updates the helm chart. In general, we should always use immutable tags to ensure that the specific helm chart ist compatible.

Additionally, this PR fixes Keycloak because the helm chart enforces too slow resource constraints. This PR optimize keycloak startup performance by a prebuilding the keycloak distribution.

## TODO


AndreasZeissner (approved):
lgtm

PR #1196, opened 2024-09-19
feat: implement set header

df-wg:
## Motivation and Context

This PR expands our header propagation abilities, by allowing users to define a value to set on all request/response headers in the format: 

```
headers:
  all: # Header rules for all subgraphs
    request:
      - op: "set"
        name: "X-API-Key"
        value: ${SECRET_FOR_SUBGRAPH}
```

The `set` operation works both for requests and responses, and it can be applied globally (to all requests on the router), or only on a specific subgraph. 

## TODO

### DIFF
--- a/router/pkg/config/config.schema.json
+++ b/router/pkg/config/config.schema.json
@@ -1757,7 +1757,7 @@
         },
         "value": {
           "type": "string",
-          "examples": ["${env.SECRET_FOR_SUBGRAPH}"],
+          "examples": ["My-Secret-Value"],
           "description": "The value to set for the header. This can include environment variables."
         }
       },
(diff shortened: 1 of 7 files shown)


router/pkg/config/config.schema.json
```diff
@@ -1727,6 +1739,29 @@
...
+          "description": "The 'set' operation is used to set a specific header value."
+        },
+        "name": {
+          "type": "string",
+          "examples": ["X-API-Key"],
+          "description": "The name of the header to set."
+        },
+        "value": {
+          "type": "string",
+          "examples": ["${env.SECRET_FOR_SUBGRAPH}"],
```
jensneuse, line 1760: `env.` might be misleading

jensneuse (approved):
one minor comment

PR #1204, opened 2024-09-22
feat: enable datadog trace propagation

df-wg:
## Motivation and Context

This PR adds datadog trace propagation for users, which will set subgraph datadog headers (`x-datadog-trace-id`, `x-datadog-parent-id` and `x-datadog-sampling-priority`) based on the trace id/span id. 

In order to enable it, users can add 
```
telemetry: 
  tracing: 
    propagation:
      datadog: true
```

to their configuration
## TODO

### DIFF
--- a/router-tests/telemetry_test.go
+++ b/router-tests/telemetry_test.go
@@ -2271,6 +2271,10 @@ Downstream errors:
 	t.Run("Datadog Propagation", func(t *testing.T) {
 		var (
 			datadogTraceId = "9532127138774266268"
+			testPropConfig = config.PropagationConfig{
+				TraceContext: true,
+				Datadog:      true,
+			}
 		)
 
 		t.Run("Datadog headers are propagated if enabled", func(t *testing.T) {
@@ -2279,8 +2283,8 @@ Downstream errors:
 			exporter := tracetest.NewInMemoryExporter(t)
 
 			testenv.Run(t, &testenv.Config{
-				TraceExporter:    exporter,
-				DatadogTelemetry: true,
+				TraceExporter:     exporter,
+				PropagationConfig: testPropConfig,
 				Subgraphs: testenv.SubgraphsConfig{
 					Employees: testenv.SubgraphConfig{
 						Middleware: func(handler http.Handler) http.Handler {
@@ -2310,8 +2314,8 @@ Downstream errors:
 			exporter := tracetest.NewInMemoryExporter(t)
 
 			testenv.Run(t, &testenv.Config{
-				TraceExporter:    exporter,
-				DatadogTelemetry: true,
+				TraceExporter:     exporter,
+				PropagationConfig: testPropConfig,
 				Subgraphs: testenv.SubgraphsConfig{
 					Employees: testenv.SubgraphConfig{
 						Middleware: func(handler http.Handler) http.Handler {
@@ -2341,8 +2345,8 @@ Downstream errors:
 			exporter := tracetest.NewInMemoryExporter(t)
 
 			testenv.Run(t, &testenv.Config{
-				TraceExporter:    exporter,
-				DatadogTelemetry: true,
+				TraceExporter:     exporter,
+				PropagationConfig: testPropConfig,
 				Subgraphs: testenv.SubgraphsConfig{
 					Employees: testenv.SubgraphConfig{
 						Middleware: func(handler http.Handler) http.Handler {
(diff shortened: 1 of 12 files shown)


router-tests/telemetry_test.go
```diff
@@ -2267,4 +2267,103 @@ Downstream errors:
 			require.Equal(t, "exception", events[0].Name)
 		})
 	})
+
+	t.Run("Datadog Propagation", func(t *testing.T) {
```
StarpTech, line 2271: Please also add tests where the client sends DD trace headers.

StarpTech (changes requested):

router-tests/telemetry_test.go
df-wg: added

router-tests/telemetry_test.go
```diff
@@ -2267,4 +2267,136 @@ Downstream errors:
...
+			}, func(t *testing.T, xEnv *testenv.Environment) {
+				res := xEnv.MakeGraphQLRequestOK(testenv.GraphQLRequest{
+					Query: `query myQuery { employees { id } }`,
+					Header: map[string][]string{
+						"x-datadog-trace-id":          {datadogTraceId},
+						"x-datadog-parent-id":         {"12345"},
+						"x-datadog-sampling-priority": {"1"},
+					},
+				})
+				require.JSONEq(t, employeesIDData, res.Body)
```
StarpTech, line 2367: The only missing piece now, is to validate if the exported trace and span IDs are correct. You can test it very easily.

StarpTech (changes requested):

router-tests/telemetry_test.go
df-wg: @StarpTech what do you mean by that? What test were you thinking of?

df-wg: I'm able to validate that the trace ID is correct, but at some point in the process, it looks like the span ID gets overwritten so I'm having difficulty validating that. What do you recommend? I've tried stepping through with the debugger, I'm not sure how long to spend on it

StarpTech: Hi @df-wg, I meant to validate the traceId and spanId exporter by the router. It's correct that a different span is generated because it's a different unit of work.

df-wg: @StarpTech I'm doing that here: https://github.com/wundergraph/cosmo/pull/1204/files#diff-e5f4b837be23c8cf9ed426c1bb3384f71ca32f1e63eba160ae024d5f99ad6220R2300

Unless I'm misunderstanding

router/pkg/config/config.schema.json
```diff
@@ -549,6 +549,10 @@
                   "type": "boolean",
                   "description": "Enable the B3 propagation. See https://github.com/openzipkin/b3-propagation (zipkin) for more information."
                 },
+                "datadog": {
+                  "type": "boolean",
+                  "description": "Enable the Datadog propagation, using https://github.com/tonglil/opentelemetry-go-datadog-propagator."
```
StarpTech, line 554: Let's not expose internals. People won't get in touch with it.

router-tests/testenv/testenv.go
```diff
@@ -126,6 +126,7 @@ type Config struct {
 	PrometheusRegistry                 *prometheus.Registry
 	ShutdownDelay                      time.Duration
 	NoRetryClient                      bool
+	DatadogTelemetry                   bool
```
StarpTech, line 129: Can we make it generic so we don't have to introduce a flag per propagation method?

router-tests/telemetry_test.go
```diff
@@ -2267,4 +2267,141 @@ Downstream errors:
...
+						"x-datadog-parent-id":         {"6023947403358210776"},
+						"x-datadog-sampling-priority": {"1"},
+					},
+				})
+				require.JSONEq(t, employeesIDData, res.Body)
+
+				sn := exporter.GetSpans().Snapshots()
+				require.GreaterOrEqual(t, len(sn), 1)
+				require.Equal(t, "00000000000000008448eb211c80319c", sn[0].SpanContext().TraceID().String())
+				//require.Equal(t, "6179b4f63c68cdfd", sn[0].SpanContext().SpanID().String())
```
StarpTech, line 2372: Please remove if not used.

StarpTech (changes requested):

StarpTech (approved):
LGTM

PR #1205, opened 2024-09-22
feat: add apollo compatibility mode with support for valueCompletion

jensneuse:


router/pkg/config/config.schema.json
```diff
@@ -1643,6 +1643,23 @@
...
+    },
+    "apollo_compatibility_flags": {
+      "type": "object",
+      "description": "To enable full compatibility with Apollo Federation, Apollo Gateway and Apollo Router, you can enable certain compatibility flags, allowing you to use Cosmo Router as a drop-in replacement for Apollo.",
+      "additionalProperties": false,
+      "properties": {
+        "enable_all": {
+          "type": "boolean",
+          "default": false,
+          "description": "Enable all Apollo compatibility flags. If the value is true, all Apollo compatibility flags are enabled."
```
StarpTech, line 1655: Can individual flags still be toggled or will `enable_all` enforce true?

StarpTech (changes requested):

router/pkg/config/config.go
```diff
@@ -579,6 +580,11 @@ type PersistedOperationsConfig struct {
 	Storage PersistedOperationsStorageConfig `yaml:"storage"`
 }
 
+type ApolloCompatibilityFlags struct {
+	EnableAll       bool `yaml:"enable_all" envDefault:"false" env:"APOLLO_COMPATIBILITY_ENABLE_ALL"`
+	ValueCompletion bool `yaml:"value_completion" envDefault:"false" env:"APOLLO_COMPATIBILITY_VALUE_COMPLETION"`
```
StarpTech, line 585: Do you think it makes sense to have an object here instead of a boolean? This would keep it open for extensions. In the future, we might have flags with additional configuration.

router/pkg/config/config.schema.json
jensneuse: enable all will enforce true to enable all individual features

PR #1226, opened 2024-09-29
fix(websocket): avoid data race on client request

StarpTech:
## Motivation and Context

This PR ensures that we don't modify the client request concurrently in multiple goroutines. We copy the request before registering the subscription. This change doesn't come with a big performance penalty because we were already forced to clone the headers. Furthermore, I moved URL parsing out of `executeSubscription` to do it only once in the websocket handler.

## TODO

### DIFF
--- a/router/core/websocket.go
+++ b/router/core/websocket.go
@@ -6,6 +6,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/buger/jsonparser"
+	"github.com/wundergraph/graphql-go-tools/v2/pkg/engine/plan"
 	"net"
 	"net/http"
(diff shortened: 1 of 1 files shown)


df-wg (approved):
LGTM. So to clarify, the main fix is in parsing the preHandler.parseRequestOptions earlier, so that it's handed down in the stack-trace and not done across goroutines, as well as explicitly cloning the `request`?

StarpTech:
> LGTM. So to clarify, the main fix is in parsing the preHandler.parseRequestOptions earlier, so that it's handed down in the stack-trace and not done across goroutines, as well as explicitly cloning the request?

Yes

PR #1227, opened 2024-09-29
feat: create subsciption over multipart

df-wg:
## Motivation and Context

One of the methods that Apollo recommends using is HTTP Multipart for subscriptions, as described [here](https://www.apollographql.com/docs/router/executing-operations/subscription-multipart-protocol) and [here](https://www.apollographql.com/docs/router/executing-operations/subscription-support#trying-subscriptions-with-curl)

This PR enables the router to follow the same protocol, so that it can respond in the format: 
```
--graphql
content-type: application/json

{}
--graphql
content-type: application/json

{"payload":{"data":{"productPriceChanged":{"name":"Croissant","price":400,"reviews":[{"score":5}]}}}}
--graphql
content-type: application/json

{"payload":{"data":{"productPriceChanged":{"name":"Croissant","price":375,"reviews":[{"score":5}]}}}}
--graphql
content-type: application/json

{"payload":{"data":{"productPriceChanged":{"name":"Croissant","price":425,"reviews":[{"score":5}]}}}}
--graphql--
```

## TODO


router-tests/nats_events_test.go
```diff
@@ -256,6 +256,77 @@ func TestNatsEvents(t *testing.T) {
...
+				require.Equal(t, http.StatusOK, resp.StatusCode)
+				defer resp.Body.Close()
+
+				reader := bufio.NewReader(resp.Body)
+
+				// Read the first part
+				assertMultipartPrefix(reader)
+				assertLineEquals(reader, "{\"payload\":{\"data\":{\"employeeUpdated\":{\"id\":3,\"details\":{\"forename\":\"Stefan\",\"surname\":\"Avram\"}}}}}")
+				assertMultipartPrefix(reader)
+				assertLineEquals(reader, "{\"payload\":{\"data\":{\"employeeUpdated\":{\"id\":3,\"details\":{\"forename\":\"Stefan\",\"surname\":\"Avram\"}}}}}")
```
df-wg, line 304: I was unable to find a way to close the subscription, and verify that it would send the `--graphql--` closing statement. Is there another example in the tests you can point me to?

df-wg (commented):
Should I manually implement the heartbearts feature? https://www.apollographql.com/docs/router/executing-operations/subscription-multipart-protocol#heartbeats

router-tests/nats_events_test.go
```diff
@@ -256,6 +256,121 @@ func TestNatsEvents(t *testing.T) {
...
+
+		assertMultipartPrefix := func(reader *bufio.Reader) {
+			assertLineEquals(reader, "--graphql")
+			assertLineEquals(reader, "Content-Type: application/json")
+			assertLineEquals(reader, "")
+		}
+
+		heartbeatInterval := 5 * time.Second
+
+		t.Run("subscribe with multipart responses", func(t *testing.T) {
```
StarpTech, line 274: We also need tests for regular subscriptions. This test addresses EDFS.

router-tests/nats_events_test.go
```diff
@@ -256,6 +256,121 @@ func TestNatsEvents(t *testing.T) {
...
+
+				var client http.Client
+				go func() {
+					client = http.Client{
+						Timeout: time.Second * 100,
+					}
+					req, err := http.NewRequest(http.MethodPost, xEnv.GraphQLMultipartResponsesURL(), bytes.NewReader(subscribePayload))
+					require.NoError(t, err)
+
+					req.Header.Set("Content-Type", "application/json")
```
StarpTech, line 292: According to the spec, the response should include `Transfer-Encoding: chunked` header for HTTP2 this header is disallowed.

Ref: https://www.apollographql.com/docs/router/executing-operations/subscription-multipart-protocol/

router/core/flushwriter.go
```diff
@@ -73,14 +100,48 @@ func (f *HttpFlushWriter) Flush() (err error) {
...
 	}
 	f.flusher.Flush()
 	return nil
 }
 
+func (f *HttpFlushWriter) StartHeartbeat() {
+	if f.multipart {
+		f.ticker = time.NewTicker(multipartHeartbeatInterval * time.Second)
+		go func() {
+			for {
```
StarpTech, line 119: It's more idiomatic when we release resources here. We can stop the ticker with `defer ticker.Stop()`

router/core/flushwriter.go
```diff
@@ -95,29 +156,40 @@ func GetSubscriptionResponseWriter(ctx *resolve.Context, variables []byte, r *ht
...
+		flushWriter.StartHeartbeat()
+	}
+
 	return ctx, flushWriter, true
 }
 
-func setSubscriptionHeaders(w http.ResponseWriter) {
-	w.Header().Set("Content-Type", "text/event-stream")
+func setSubscriptionHeaders(wgParams WgRequestParams, w http.ResponseWriter) {
+	if wgParams.SubscribeOnce {
```
StarpTech, line 183: I think we have no test for this case.

router/core/flushwriter.go
```diff
@@ -95,29 +156,40 @@ func GetSubscriptionResponseWriter(ctx *resolve.Context, variables []byte, r *ht
...
+		return
+	}
+
+	if wgParams.UseMultipart {
+		w.Header().Set("Content-Type", multipartContentType)
+	} else {
+		w.Header().Set("Content-Type", "text/event-stream")
```
StarpTech, line 190: We should set header only when we negotiated with the client the SSE protocol.

router-tests/testenv/testenv.go
```diff
@@ -1202,6 +1202,13 @@ func (e *Environment) GraphQLServeSentEventsURL() string {
 	return u.String()
 }
 
+func (e *Environment) GraphQLMultipartResponsesURL() string {
+	u, err := url.Parse(e.GraphQLRequestURL())
+	require.NoError(e.t, err)
+	u.RawQuery = core.WgMultipartParam
```
StarpTech, line 1208: This is not the way how protocol negotiation is done. The user sends the `Accept: multipart/mixed;subscriptionSpec="1.0", application/json` header to use subscriptions over multipart.

router/core/flushwriter.go
```diff
@@ -57,12 +77,19 @@ func (f *HttpFlushWriter) Flush() (err error) {
...
+	flushBreak := GetWriterPrefix(f.sse, f.multipart)
+	if flushBreak != "" {
+		_, err = f.writer.Write([]byte(flushBreak))
 		if err != nil {
 			return err
 		}
 	}
+
+	if f.multipart && len(resp) > 0 {
+		// Per the Apollo docs, multipart messages are supposed to be json, wrapped in ` "payload"`
```
StarpTech, line 89: It's not common, but consider that when the Accept does not ask for `application/json` the request should be aborted early.

router/core/flushwriter.go
```diff
@@ -57,12 +77,19 @@ func (f *HttpFlushWriter) Flush() (err error) {
...
+		resp = append([]byte(`{"payload":`), resp...)
```
StarpTech, line 90: For this type of operations, I'd use our astjson library. This ensure that a valid value is constructed without newlines.

StarpTech (changes requested):

router/core/flushwriter.go
```diff
@@ -109,6 +111,9 @@ func (f *HttpFlushWriter) Flush() (err error) {
 	if err != nil {
 		return err
 	}
+
+	f.mu.Lock()
```
StarpTech, line 115: @df-wg Did you figure out why this is necessary or is there even an alternative approach?

router/core/flushwriter.go
```diff
@@ -21,8 +31,11 @@ type HttpFlushWriter struct {
 	flusher       http.Flusher
 	subscribeOnce bool
 	sse           bool
+	multipart     bool
```
StarpTech, line 34: We should start to encapsulate SSE and multipart in different writer implementations.

jensneuse (approved):

router/core/flushwriter.go
```diff
@@ -3,15 +3,24 @@ package core
...
 const (
-	WgPrefix             = "wg_"
-	WgSseParam           = WgPrefix + "sse"
-	WgSubscribeOnceParam = WgPrefix + "subscribe_once"
+	WgPrefix                   = "wg_"
+	WgSseParam                 = WgPrefix + "sse"
+	WgSubscribeOnceParam       = WgPrefix + "subscribe_once"
+	multipartBoundary          = "graphql"
+	multipartHeartbeatInterval = 5 * time.Second
+	multipartAcceptHeader      = "multipart/mixed;subscriptionSpec=\"1.0\", application/json"
```
StarpTech, line 20: You can do a equal match here. Client could use spaces and it would be still valid. You have to parse the header with https://pkg.go.dev/mime

StarpTech (changes requested):

StarpTech (approved):
LGTM

PR #1234, opened 2024-10-01
feat: add option to add response header with trace id

JivusAyrus:
## TODO


router/core/graphql_handler.go
```diff
@@ -183,6 +186,7 @@ func (h *GraphQLHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	switch p := operationCtx.preparedPlan.preparedPlan.(type) {
 	case *plan.SynchronousResponsePlan:
 		w.Header().Set("Content-Type", "application/json")
+		h.setTraceResponseHeader(w, graphqlExecutionSpan)
```
StarpTech, line 189: Instead of setting this at specific places it would be more reliable if we would do this in a separate http middleware after the trace middleware. In that case, we ensure that this header is set in all cases.

This could be a good place https://github.com/wundergraph/cosmo/blob/5461bb5a529decd51a1b22be0a5301936b8ad392/router/pkg/trace/middleware.go#L41

StarpTech (changes requested):

StarpTech (approved):
LGTM

PR #1236, opened 2024-10-02
feat: add apollo compatibility flag to truncate floats

jensneuse:


demo/pkg/subgraphs/test1/subgraph/schema.resolvers.go
```diff
@@ -247,12 +252,6 @@ func (r *Resolver) Subscription() generated.SubscriptionResolver { return &subsc
...
 
-// !!! WARNING !!!
-// The code below was going to be deleted when updating resolvers. It has been copied here so you have
-// one last chance to move it out of harms way if you want. There are two reasons this happens:
-//   - When renaming or deleting a resolver the old code will be put in here. You can safely delete
-//     it when you're done.
-//   - You have helper methods in this file. Move them out to keep these resolver files clean.
 var (
 	aBigObject = &model.ABigObject{
 		AFieldOnABigObject: "a field on a big object - lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur excepteur sint occaecat cupidatat non proident sunt in culpa qui officia deserunt mollit anim id est laborum",
```
Aenimus, line 257: This doesn't look right. It looks like someone committed resolver changes that are not in the schema. Does that var still serve a purpose?

router-tests/apollo_compatibility_test.go
```diff
@@ -68,4 +68,79 @@ func TestApolloCompatibility(t *testing.T) {
...
+			},
+		}, func(t *testing.T, xEnv *testenv.Environment) {
+			res, err := xEnv.MakeGraphQLRequest(testenv.GraphQLRequest{
+				Query:     `query FloatQuery($arg: Float) { floatField(arg: $arg) }`,
+				Variables: json.RawMessage(`{"arg":1}`),
+			})
+			require.NoError(t, err)
+			require.Equal(t, http.StatusOK, res.Response.StatusCode)
+			require.Equal(t, `{"data":{"floatField":1}}`, res.Body)
+		})
```
Aenimus, line 144: Please add:
Test for `1.1` that returns `1.1` regardless of the value of`ApolloCompatibilityTruncateFloats`.
Test that string "1.1" returns `1.1` regardless of the value of `ApolloCompatibilityTruncateFloats`.
Test that string "1.0" returns `1` if `ApolloCompatibilityTruncateFloats` is `true`.

jensneuse: I test these cases in go tools already, here I test that the config has an effect.

router/pkg/config/config.schema.json
```diff
@@ -1706,6 +1706,17 @@
               "default": false
             }
           }
+        },
+        "truncate_floats":{
+          "type": "object",
+          "description": "Truncate floats like 1.0 to 1, 2.0 to 2, etc.",
```
Aenimus, line 1712: It should be clear that it only truncates `.0`. It does not truncate `1.1` for example.

Aenimus (changes requested):
See comments please.

demo/pkg/subgraphs/test1/subgraph/schema.resolvers.go
jensneuse: It's correct. Someone forgot to clean it up.

router-tests/apollo_compatibility_test.go
Aenimus: I see no good reason not to include the tests I suggested above in Cosmo.

router/pkg/config/config.schema.json
jensneuse: updated

Aenimus (approved):
LGTM!

Issue #1238, opened 2024-10-03
Panic when trying to use initialPayload ExportToken config


ElliottZeroFlucs:
### Component(s)

router

### Component version

0.116.3

### wgc version

0.66.2

### controlplane version

0.110.0

### router version

0.116.3

## Description
When trying to use subscriptions we need to use the `fromInitialPayloadConfig.ExportToken.Enabled` config to export the token from the initial payload to the request header. We do not need the `accessController.Access` check and haven't configured any authenticators. We would expect that if authentication was not enabled and no authenticators provided the token export should still be able to run.

We have found that there is a missing nil check for the accessController and so there is a panic, nil pointer dereference error.

Further up in the `handleUpgradeRequest` you can see the nil check is in place.
https://github.com/wundergraph/cosmo/blob/main/router/core/websocket.go#L236

Here the nil check is missing and this is what causes the panic. With the correct nil check our use case works as expected.
https://github.com/wundergraph/cosmo/blob/main/router/core/websocket.go#L343

## Steps to Reproduce
1. Setup a router with acess to a subgraph containing a subscription.
2. Setup websocket config with the from_initial_payload.export_token enabled. (Don't setup any authenticators)
3. Run the router.
4. Create a subscription in the playground.
5. The panic will occur.

## Expected Result
The subscription should be created successfully.

## Actual Result
The router panics.

## Environment
OS: (e.g., "MacOS 13.5")
Package Manager: pnpm
Compiler(if manually compiled): "go 1.23.1"

### Router configuration

```yaml
# yaml-language-server: $schema=./pkg/config/config.schema.json

# See pkg/config/config.go for the full list of configuration options.
# This file is used to debugging purposes only.

version: "1"

router_config_path: "./__schemas/config.json"

engine:
  max_concurrent_resolvers: 1024
  debug:
    report_websocket_connections: true
    report_memory_usage: true

websocket:
  enabled: true
  forward_initial_payload: true
  authentication:
    from_initial_payload:
      enabled: true
      key: Authorization
      export_token:
        enabled: true
        header_key: Authorization

events:
  providers:
    nats:
      - id: default
        url: "nats://localhost:4222"
      - id: my-nats
        url: "nats://localhost:4222"
    kafka:
      - id: my-kafka
        brokers:
          - "localhost:9092"

headers:
  all:
    request:
      - op: "propagate"            
        named: Authorization
```

### Log output

```shell
[...]

PR #1239, opened 2024-10-03
fix: fix panic when accessController is nil in handleUpgradeRequest

ElliottZeroFlucs:
## Motivation and Context

Issue is described in detail here https://github.com/wundergraph/cosmo/issues/1238

I have added the nil check and written a test that fails before the change and succeeds after.

## TODO


StarpTech:
Hi @ElliottZeroFlucs, thanks for the fix with a test 👍

StarpTech (approved):
LGTM

PR #1260, opened 2024-10-10
fix: ensure sse is spec compliant

df-wg:
## Motivation and Context

The previous implementation of SSE worked using a query param (`wg_sse`), which isn't spec compliant. This parses the value of the MIME type in Accept header, which is per the spec
## TODO


jensneuse (approved):

StarpTech:
Please don't forget to document the new approach in the [docs](https://cosmo-docs.wundergraph.com/router/subscriptions/server-sent-events-sse) and remove the old one. We can remove it after a grace period.

df-wg:
Encountered https://github.com/99designs/gqlgen/issues/3328 while verifying the PR, which is why it took so long to close this

PR #1267, opened 2024-10-13
feat: custom metric attributes

StarpTech:
## Custom Metric Attributes

This PR adds the ability to define custom metric attributes. Those attributes can be extracted from a header or request context. If extracted from a header, it is guaranteed that the attribute is added to all metrics down the request and response. Context fields that are calculated after the response e..g `graphql_error_codes` are added only to metrics that are measured at the same stage or later. This is valid e.g. for request / response metrics but invalid for `router.graphql.operation.planning_time` metric.

```yaml
telemetry:
  metrics:
    attributes:
      - key: "X-New-Attribute"
        default: "foo" # when value could not be determined
        value_from:
          request_header: "X-Request-ID"
      - key: "error_codes"
        value_from:
          context_field: graphql_error_codes
```

In OTLP, slice attributes are supported and encoded as JSON values. In Prometheus, this isn't possible and would result in non-queryable metrics. To circumvent that limitation, we emit a new metric series per slice element.

During the implementation, it was necessary to refactor the way we organize base and metric attributes. This PR comes with significant improvements in this area as well.

I also removed span creation for websockets entirely. Previously, we have created spans in the operation processor.

## New `router.graphql.operation.planning_time` metric

A new metric has been added to measure the planning time of graphql operations. Beside default attributes, we also expose if we plan was loaded from the cache.

### Breaking changes

The attribute `wg.subgraph.error.extended_code` is no longer added to metrics by default to avoid collision with custom attributes. In order to get the same behaviour you can use the following config:

```yaml
telemetry:
  metrics:
    attributes:
      - key: "wg.subgraph.error.extended_code"
        value_from:
          context_field: graphql_error_codes
```

## TODO

- [X] Tests or benchmark included


StarpTech:
Remaining tasks: Measure performance.

StarpTech:
> Remaining tasks: Measure performance.
I couldn't measure any significant performance differences. Alloc space has been reduced.

router-tests/kafka_events_test.go
```diff
@@ -116,7 +116,7 @@ func TestKafkaEvents(t *testing.T) {
 
 			go func() {
 				wg.Wait()
-				require.NoError(t, client.Close())
```
StarpTech, line 119: Produced "use of closed connection" issues which can be ignored because in some conditions the connection has been already closed which is fine.

router/pkg/metric/metric_store.go
```diff
@@ -69,17 +79,19 @@ type (
...
 
 	Provider interface {
 		MeasureInFlight(ctx context.Context, attr ...attribute.KeyValue) func()
 		MeasureRequestCount(ctx context.Context, attr ...attribute.KeyValue)
 		MeasureRequestSize(ctx context.Context, contentLength int64, attr ...attribute.KeyValue)
 		MeasureResponseSize(ctx context.Context, size int64, attr ...attribute.KeyValue)
-		MeasureLatency(ctx context.Context, requestStartTime time.Time, attr ...attribute.KeyValue)
+		MeasureLatency(ctx context.Context, latency time.Duration, attr ...attribute.KeyValue)
 		MeasureRequestError(ctx context.Context, attr ...attribute.KeyValue)
+
```
df-wg, line 92: nit: remove the extra lines around this func

router/pkg/metric/prom_metric_store.go
```diff
@@ -47,89 +45,170 @@ func NewPromMetricStore(logger *zap.Logger, meterProvider *metric.MeterProvider,
...
+			// String Slice attributes have to be exploded into multiple metrics with different label values in Prometheus.
+			// This is because Prometheus does not support multi-value labels.
+			for _, attr := range dimensionsSet {
+				for _, v := range attr.Value.AsStringSlice() {
+					c.Add(ctx, 1, otelmetric.WithAttributes(append(otherAttributes, attribute.KeyValue{
+						Key:   attr.Key,
+						Value: attribute.StringValue(v),
+					})...))
+				}
+			}
```
df-wg, line 65: Can you pull this out into a function that we call with a value parameter? This double loop is repeated a few times throughout

df-wg (approved):
Overall, this looks really great! I have two cleanup comments, but it also looks fine as is

df-wg (approved):
That's awesome. LGTM 🥳

thisisnithin (approved):

PR #1289, opened 2024-10-19
fix(websocket): set op name and operation

StarpTech:
## Motivation and Context

This PR fixes a websocket regression for custom modules. An integration test has been added to prevent it in the future.

## Checklist

### DIFF
--- a/router/core/websocket.go
+++ b/router/core/websocket.go
@@ -759,9 +759,6 @@ func (h *WebSocketConnectionHandler) parseAndPlan(payload []byte) (*ParsedOperat
 	defer operationKit.Free()
 
 	opContext := &operationContext{
-		name:       operationKit.parsedOperation.Request.OperationName,
-		opType:     operationKit.parsedOperation.Type,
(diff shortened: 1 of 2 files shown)


router/core/websocket.go
```diff
@@ -792,6 +789,9 @@ func (h *WebSocketConnectionHandler) parseAndPlan(payload []byte) (*ParsedOperat
 		opContext.parsingTime = time.Since(startParsing)
 	}
 
+	opContext.name = operationKit.parsedOperation.Request.OperationName
```
jensneuse, line 792: I'd use strings.clone or similar here to make a dedicated copy of the name. Operationkit is re-used.

jensneuse (approved):

router/core/websocket.go
StarpTech: We already ensure [here](https://github.com/wundergraph/cosmo/blob/f6a4224a2370e8eb6e36598a22f60a3eee83f055/router/core/operation_processor.go#L489) that we copy the value.

PR #1294, opened 2024-10-21
fix: escape characters in clickhouse query params

thisisnithin:
## Motivation and Context

Clickhouse query fails if special characters are present. We need to escape them

## Checklist


controlplane/src/core/repositories/analytics/AnalyticsRequestViewRepository.ts
```diff
@@ -721,6 +731,8 @@ export class AnalyticsRequestViewRepository {
 
     whereSql += scopedSql;
 
+    this.escapeStrings(coercedQueryParams);
```
StarpTech, line 734: There could be other places too right? Ideally, we can fix this on the driver level but since we only pass strings this seems to be impossible. We should think about a better abstraction in the future.

StarpTech (changes requested):

controlplane/src/core/repositories/analytics/AnalyticsRequestViewRepository.ts
thisisnithin: Fixed it in other places too

StarpTech (approved):
LGTM

PR #1302, opened 2024-10-23
feat: custom scripts

### LINKED ISSUE #1185: New server response format when not 200 code in playground looks unreadable
### Component(s)

studio

I extremely dislike the way it now shows responses not having a 200 code

Screenshot 2024-09-18 at 12 24 37

1. it's unreadable, although the idea to show what came over HTTP is understandable, but it's better to add some custom fields to the standard GraphQL response structure. Now I don't understand how to try to read responseBody during debugging, and other fields don't give useful load
2. We subgraphs return 4** error codes in case of non-significant problems, 5** in case of unknown server errors. At the same time they describe the standard GraphQL response structure, deviations from this logic are possible only in case of some serious problems, but it seems Playground should be used mainly for debugging, why should I deviate from this structure trying to invent another one?

- Return to the old format
- Extend the standard GraphQL answer with fields that you think should be useful for non-200 codes
- Do not change the structure of the response display for non-200 codes

_No response_

### Additional context

_No response_

thisisnithin:
## Motivation and Context

Fixes ENG-5767
Closes #1185 

## Checklist


StarpTech (approved):
LGTM

PR #1311, opened 2024-10-26
feat: extend apollo compatible error support

Aenimus:
## Checklist


Aenimus:
Relies on https://github.com/wundergraph/graphql-go-tools/pull/939

router/core/errors.go
```diff
@@ -175,48 +175,49 @@ func propagateSubgraphErrors(ctx *resolve.Context) {
...
-		} else if wgRequestParams.UseMultipart {
-			// Handle multipart error response
-			if err := writeMultipartError(w, requestErrors, requestLogger); err != nil {
-				if requestLogger != nil {
-					requestLogger.Error("error writing multipart response", zap.Error(err))
-				}
-			}
-			return
-		}
+	if requestErrors == nil {
```
Aenimus, line 178: I just removed nesting by returning early. Previously, the whole block was wrapped in `err != nil`

router/core/graphql_prehandler.go
```diff
@@ -315,11 +317,29 @@ func (h *PreHandler) Handler(next http.Handler) http.Handler {
...
+				report := reportErr.Report()
+				for _, externalErr := range report.ExternalErrors {
+					if externalErr.ApolloCompatibilityError != nil {
+						outputErr = externalErr.ApolloCompatibilityError
+						break
+					}
+				}
+			}
+
+			requestContext.error = outputErr
```
StarpTech, line 338: `requestContext.error` should be the original error from `handleOperation`.

router/core/graphql_prehandler.go
```diff
@@ -315,11 +317,29 @@ func (h *PreHandler) Handler(next http.Handler) http.Handler {
 			body:             body,
 		})
 		if err != nil {
-			requestContext.error = err
+			outputErr := err
+			var httpErr ApolloCompatibleHttpError
+			var reportErr ReportError
+			if h.executor.Resolver.Options().ApolloCompatibilityReplaceInvalidVariableError && errors.As(err, &httpErr) {
```
StarpTech, line 323: `httpErr` is returned for many reasons not just for invalid variables.

router/core/graphql_prehandler.go
```diff
@@ -315,11 +317,29 @@ func (h *PreHandler) Handler(next http.Handler) http.Handler {
 			body:             body,
 		})
 		if err != nil {
-			requestContext.error = err
+			outputErr := err
```
StarpTech, line 320: Please put this error handling code into a separate struct / function. We are having a hard time simplifying the pre-handler.

router/core/graphql_prehandler.go
```diff
@@ -315,11 +317,29 @@ func (h *PreHandler) Handler(next http.Handler) http.Handler {
...
+			var httpErr ApolloCompatibleHttpError
+			var reportErr ReportError
+			if h.executor.Resolver.Options().ApolloCompatibilityReplaceInvalidVariableError && errors.As(err, &httpErr) {
+				outputErr = &operationreport.ApolloCompatibilityError{
+					ExtensionCode: httpErr.ExtensionCode(),
+					Message:       httpErr.Error(),
+				}
+			} else if h.executor.Resolver.Options().ApolloCompatibilityReplaceUndefinedOperationFieldError && errors.As(err, &reportErr) {
```
StarpTech, line 328: I don't understand how you identified an `ReplaceUndefinedOperationField` error here. Also what happens when both modes are enabled? Does it mean `ApolloCompatibilityReplaceInvalidVariableError` has always precedence?

StarpTech (changes requested):

router/core/graphql_prehandler.go
Aenimus: The reason all the logic is here is because I thought the error should match what is actually provided to the user. So it's fine for the errors to be completely different?

router/core/graphql_prehandler.go
Aenimus: Ok but see other comment. Logic is only here due to to the span error. If error should be original i won't handle any of it here

router/core/graphql_prehandler.go
Aenimus: The opposite. If the operation itself is invalid validation of variables won't happen.

router/core/graphql_prehandler.go
Aenimus: This an apolloCompatibleHttpError. It has a different form to a typical http error. I think the variable name is confusing.

StarpTech: Oh yeah, we use the same naming for general http errors.

StarpTech (approved):
LGTM

PR #1326, opened 2024-10-31
fix: delete org and user workflows

thisisnithin:
## Checklist


thisisnithin:
Closing. We will manually delete namespaces to preserve integrity.

controlplane/src/core/repositories/OrganizationRepository.ts
```diff
@@ -849,9 +851,40 @@ export class OrganizationRepository {
...
     return this.db.transaction(async (tx) => {
       const oidcRepo = new OidcRepository(tx);
+      const fedGraphRepo = new FederatedGraphRepository(this.logger, tx, organizationId);
+      const subgraphRepo = new SubgraphRepository(this.logger, tx, organizationId);
+
+      const graphs = await fedGraphRepo.list({
+        limit: 0,
+        offset: 0,
+      });
+      const subgraphs = await subgraphRepo.list({
```
StarpTech, line 864: Is it really necessary? Can't we delete all targets of the org?

controlplane/src/bin/delete-user.ts
```diff
@@ -80,11 +90,14 @@ if (!isSafe) {
...
-  keycloakClient,
-  keycloakRealm: realm,
-});
+await userRepo.deleteUser(
+  {
+    id: user.id,
+    keycloakClient,
+    keycloakRealm: realm,
+  },
+  blobStorage,
```
StarpTech, line 99: Any reason why is this service dependency not part of he UserRepository constructor?

controlplane/src/bin/delete-user.ts
```diff
@@ -24,10 +27,17 @@ const {
   databaseTlsKey,
   webhookUrl,
   webhookSecret,
+  s3Storage,
 } = getConfig();
 
 const userId = process.env.USER_ID || '';
 
+const bucketName = extractS3BucketName(s3Storage);
```
StarpTech, line 35: Is it really necessary to put all the logic in here? Why don't we issue a job to the queue and the controlplane picks it up? In that way, we also leverage all benefits that the queue provides retry, history etc...

controlplane/src/core/repositories/OrganizationRepository.ts
```diff
@@ -849,9 +851,40 @@ export class OrganizationRepository {
     return result[0];
   }
 
-  public deleteOrganization(organizationId: string) {
+  public deleteOrganization(organizationId: string, blobStorage: BlobStorage) {
     return this.db.transaction(async (tx) => {
       const oidcRepo = new OidcRepository(tx);
+      const fedGraphRepo = new FederatedGraphRepository(this.logger, tx, organizationId);
```
StarpTech, line 857: If we go this road of deleting everything manuel, we need very good tests. Could you outline what is delete by constraints and what not?

StarpTech (changes requested):

controlplane/src/bin/delete-user.ts
thisisnithin: This would require refactoring way too many files since we have to touch other repos as well. Let's not block the PR

controlplane/src/core/repositories/OrganizationRepository.ts
thisisnithin: I have added a comment and a couple tests that cover what is deleted here

controlplane/src/bin/delete-user.ts
thisisnithin: This is run in GH actions which triggers a k8 job and it is transactional. It immediately gives us why a user can/cannot be deleted and what goes wrong. The benefits you listed is availbe with actions itself

controlplane/src/core/workers/DeleteUserQueue.ts
```diff
@@ -0,0 +1,141 @@
...
+
+      const user = await userRepo.byId(job.data.userId);
+      if (!user || !user.id || !user.email) {
+        throw new Error('User not found');
+      }
+
+      const { isSafe, soloOrganizations, unsafeOrganizations } = await orgRepo.canUserBeDeleted(job.data.userId);
+
+      if (!isSafe) {
+        this.input.logger.error(
```
StarpTech, line 89: I think this can be a info.

controlplane/src/core/workers/DeleteUserQueue.ts
```diff
@@ -0,0 +1,141 @@
...
+export class DeleteUserQueue {
+  private readonly queue: Queue<DeleteUserInput>;
+  private readonly logger: pino.Logger;
+
+  constructor(log: pino.Logger, conn: ConnectionOptions) {
+    this.logger = log.child({ queue: QueueName });
+    this.queue = new Queue<DeleteUserInput>(QueueName, {
+      connection: conn,
+      defaultJobOptions: {
+        removeOnComplete: true,
```
StarpTech, line 28: We should not delete it and keep a history for more than 90d.

controlplane/src/core/workers/ReactivateOrganizationWorker.ts
```diff
@@ -0,0 +1,114 @@
...
+
+export class ReactivateOrganizationQueue {
+  private readonly queue: Queue<ReactivateOrganizationInput>;
+  private readonly logger: pino.Logger;
+
+  constructor(log: pino.Logger, conn: ConnectionOptions) {
+    this.logger = log.child({ queue: QueueName });
+    this.queue = new Queue<ReactivateOrganizationInput>(QueueName, {
+      connection: conn,
+      defaultJobOptions: {
```
StarpTech, line 25: Same here

controlplane/src/core/workers/DeactivateOrganizationWorker.ts
```diff
@@ -0,0 +1,126 @@
...
+
+      await orgRepo.deactivateOrganization({
+        organizationId: job.data.organizationId,
+        reason: job.data.deactivationReason,
+        keycloakClient: this.input.keycloakClient,
+        keycloakRealm: this.input.keycloakRealm,
+        deleteOrganizationQueue: this.input.deleteOrganizationQueue,
+      });
+    } catch (err) {
+      this.input.logger.error(err, `Failed to deactivate organization with id ${job.data.organizationId}`);
```
StarpTech, line 96: Please use structured log fields for additional info like organization ID and job ID

controlplane/src/core/workers/ReactivateOrganizationWorker.ts
```diff
@@ -0,0 +1,114 @@
...
+  const worker = new Worker<ReactivateOrganizationInput>(
+    QueueName,
+    (job) => new ReactivateOrganizationWorker(input).handler(job),
+    {
+      connection: input.redisConnection,
+      concurrency: 10,
+    },
+  );
+  worker.on('stalled', (job) => {
+    log.warn(`Job ${job} stalled`);
```
StarpTech, line 108: Here as well. For ops, it could to log the job id, so we can filter for it.

controlplane/src/core/workers/DeactivateOrganizationWorker.ts
```diff
@@ -0,0 +1,126 @@
...
+const QueueName = 'organization.deactivate';
+const WorkerName = 'DeactivateOrganizationWorker';
+
+export interface DeactivateOrganizationInput {
+  organizationId: string;
+  organizationSlug: string;
+  deactivationReason?: string;
+}
+
+export class DeactivateOrganizationQueue {
```
StarpTech, line 18: I think it is time to ensure an interface for a worker implementation. Could you ensure that each worker implementation, implements an interface?

controlplane/src/core/workers/DeleteUserQueue.ts
```diff
@@ -0,0 +1,141 @@
...
+      await userRepo.deleteUser(
+        {
+          id: job.data.userId,
+          keycloakClient: this.input.keycloakClient,
+          keycloakRealm: this.input.keycloakRealm,
+        },
+        this.input.blobStorage,
+      );
+
+      this.input.platformWebhooks.send(PlatformEventName.USER_DELETE_SUCCESS, {
```
StarpTech, line 109: This is dangerous. This will send the mail without waiting for delivery correct? In case of an error, the job is not marked as failed and in case of a shutdown, the job might not be done in time.

StarpTech (changes requested):

controlplane/src/core/workers/DeleteUserQueue.ts
thisisnithin: Tha api is designed to not be blocking. It should be respected here as well, We log errors for these within the webhook service for which we can trigger manualy.

controlplane/src/core/workers/DeactivateOrganizationWorker.ts
thisisnithin: Done

controlplane/src/core/workers/ReactivateOrganizationWorker.ts
thisisnithin: Done

controlplane/src/core/workers/DeactivateOrganizationWorker.ts
thisisnithin: Done

controlplane/src/core/workers/ReactivateOrganizationWorker.ts
thisisnithin: Done

controlplane/src/core/workers/DeleteUserQueue.ts
thisisnithin: Done

controlplane/src/core/workers/DeleteUserQueue.ts
thisisnithin: Done

StarpTech (approved):
LGTM

PR #1328, opened 2024-11-03
feat(telemetry): improve memory consumption

StarpTech:
## Motivation and Context

Based on the included benchmark. The PR reduces memory pressure by many magnitudes. 

CleanShot 2024-11-04 at 00 01 05@2x

Other fixes:

- Host parsing issue with `localhost`
- Status code attribute wasn't attached in all cases

## Checklist


package.json
```diff
@@ -32,8 +32,8 @@
     "dist"
   ],
   "engines": {
-    "node": ">=20.0.0",
-    "pnpm": "8"
+    "node": ">=22.11.0",
```
df-wg, line 35: Given this change, can you also update https://github.com/wundergraph/cosmo/blob/af0ae3ba2d0afb4a637d101e93e06644cbafb3c4/CONTRIBUTING.md?plain=1#L13-L14 to reflect the new versions?

router/pkg/metric/metric_store.go
```diff
@@ -74,28 +74,34 @@ type (
...
+		MeasureRequestSize(ctx context.Context, contentLength int64, opts ...otelmetric.AddOption)
+		MeasureResponseSize(ctx context.Context, size int64, opts ...otelmetric.AddOption)
+		MeasureLatency(ctx context.Context, latency float64, opts ...otelmetric.RecordOption)
+		MeasureRequestError(ctx context.Context, opts ...otelmetric.AddOption)
+		MeasureOperationPlanningTime(ctx context.Context, planningTime float64, opts ...otelmetric.RecordOption)
 		Flush(ctx context.Context) error
 	}
 
 	Store interface {
+		MeasureInFlight(ctx context.Context, sliceAttr, attr []attribute.KeyValue) func()
```
df-wg, line 96: Can you add godocs here for how the interfaces are used, and how they are coupled?

df-wg (changes requested):
Overall LGTM, but two docs changes that I think should be added before it is merged

package.json
StarpTech: Good catch

router/pkg/metric/metric_store.go
StarpTech: Done

df-wg (approved):

PR #1329, opened 2024-11-03
fix(prometheus): reduce buckets to prevent cardinality issues

StarpTech:
## Motivation and Context

Users run into scraping performance issues due to high cardinality metrics. While metric labels can be filtered, there was no way to change the buckets of histograms. For our cloud integration, we went with 32 buckets, which is fine but we also used it for prometheus which produces a lot of data to export. This PR will introduce a different bucket configuration (10 buckets) for Prometheus histograms to reduce the overall export size.

## Checklist


df-wg (approved):
Can you add an explanation for the change and the motivation behind it in the PR body, so it'll be clearer if people look at the git history?

PR #1330, opened 2024-11-03
feat(router): implement automatic persistent queries (apq)

df-wg:
## Motivation and Context

A popularly requested feature is Automatic Persistent Queries, where the router would automatically persist queries if they are initially sent in the form: 
```
  curl --get http://localhost:4000/graphql \
  --header 'content-type: application/json' \
  --data-urlencode 'query={__typename}' \
  --data-urlencode 'extensions={"persistedQuery":{"version":1,"sha256Hash":"ecf4edb46db40b5132295c0291d62fb65d6759a9eedfa4d5d612dd5ec54a6b38"}}'
```

and then it can be used afterwards just by providing the `sha256Hash`, like: 
```
curl --get http://localhost:4000/graphql \
  --header 'content-type: application/json' \
  --data-urlencode 'extensions={"persistedQuery":{"version":1,"sha256Hash":"ecf4edb46db40b5132295c0291d62fb65d6759a9eedfa4d5d612dd5ec54a6b38"}}'
```

This feature can be enabled via the new top-level configuration:
```
# This is the existing configuration for the previous persisted operations: https://cosmo-docs.wundergraph.com/router/persisted-operations
persisted_operations:
  cache: 
    size: 100MB
  storage: 
    provider_id: s3
    object_prefix: wundergraph

# This is a new feature because persisted operations are related to clients which is not the way how APQ works
automatic_persisted_queries:
  enabled: true
  cache: # This object defines caching particulars
    size: 1024MB # Size must be in MB people can't think in operations and can vary heavily in size. This, at this point, is only relevant for an in-memory cache that we maintain
    ttl: 900 # set both for a local and a remote KV
  storage:
    provider_id: "my_redis"
    object_prefix: cosmo_apq

storage_providers:
  redis: 
    - id: "my_redis"
      url: "redis://localhost:6379"
``` 

## Checklist


router/core/operation_processor.go
```diff
@@ -610,6 +625,10 @@ func (o *OperationKit) normalizePersistedOperation() (cached bool, err error) {
 	o.parsedOperation.NormalizedRepresentation = o.kit.normalizedOperation.String()
 	o.parsedOperation.Request.Variables = o.kit.doc.Input.Variables
 
+	if o.operationProcessor.persistedOperationClient != nil && o.operationProcessor.persistedOperationClient.ApqEnabled() {
```
df-wg, line 628: This was meant to ensure that in cases where we're using APQ, the requests are cached in the proper location, rather than in the general persisted operation cache. However, this could also potentially mean that if a user has both persisted operations enabled and apq enabled, and they requested an operation from their file cached persisted operations, it would get cached in the local cache (because by this point in the operation, we don't know whether the request came from APQ or persisted operation). I welcome any suggestions for how we should address this, if it should be done at this point

router/core/operation_processor.go
```diff
@@ -372,10 +372,25 @@ func (o *OperationKit) FetchPersistedOperation(ctx context.Context, clientInfo *
...
 	// it's important to make a copy of the persisted operation data, because it's used in the cache
 	// we might modify it later, so we don't want to modify the cached data
-	o.parsedOperation.Request.Query = string(persistedOperationData)
+	if persistedOperationData != nil {
+		o.parsedOperation.Request.Query = string(persistedOperationData)
+	}
+
+	// If the operation was fetched with APQ, save it again to renew the TTL
+	if o.operationProcessor.persistedOperationClient.ApqEnabled() {
+		if err = o.operationProcessor.persistedOperationClient.SaveOperation(ctx, clientInfo.Name, o.parsedOperation.GraphQLRequestExtensions.PersistedQuery.Sha256Hash, o.parsedOperation.Request.Query); err != nil {
```
jensneuse, line 390: Will this mean that we're saving every request to the cache when APQ is enabled? This might result in excessive writes to the cache.

router/core/operation_processor.go
jensneuse: I think it's ok if you use both caches, the distributed (Redis) cache to load the APQ content, and the PO cache to cache the normalized version of the APQ. The naming might be slightly off, but I think it'll work fine. Can you remove this early return and validate if tests behave as expected?

I think it's important to cache this step also for APQ because otherwise APQ would never hit the normalization cache, right?

Again, I'm not 100% sure so please test and verify.

router/internal/persistedoperation/apq/client.go
```diff
@@ -0,0 +1,94 @@
...
+	cl.cacheLock = &sync.RWMutex{}
+
+	return cl, err
+}
+
+func (c client) Enabled() bool {
+	return c.enabled
+}
+
+func (c client) PersistedOperation(ctx context.Context, clientName string, sha256Hash string) ([]byte, error) {
```
jensneuse, line 71: I would use pointer receiver instead of value receiver (c *client) as a default. Otherwise each function call creates a client object with copies of all values, which might result in unintended behaviour, e.g. if you want to change something on the client object.

router/internal/persistedoperation/apq/client.go
```diff
@@ -0,0 +1,94 @@
...
+	if c.kvClient != nil {
+		return c.kvClient.Get(ctx, clientName, sha256Hash)
```
jensneuse, line 73: Would it make sense to use the kv client AND the local cache? Or just one?

router/internal/persistedoperation/apq/client.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	c.cacheLock.RLock()
+	defer c.cacheLock.RUnlock()
+	return c.cache.Get(clientName, sha256Hash), nil
+}
+
+func (c client) SaveOperation(ctx context.Context, clientName, sha256Hash string, operationBody []byte) error {
+	if c.kvClient != nil {
+		return c.kvClient.Set(ctx, sha256Hash, operationBody, c.ttl)
```
jensneuse, line 83: Same question as above. Use Cache AND kv client?

router/internal/persistedoperation/apq/redis.go
```diff
@@ -0,0 +1,67 @@
...
+		prefix: opts.Prefix,
+	}
+
+	return rclient, nil
+}
+
+func (r *redisClient) Get(ctx context.Context, clientName, operationHash string) ([]byte, error) {
+	cmd := r.client.Get(ctx, r.prefix+operationHash)
+	if errors.Is(cmd.Err(), redis.Nil) {
+		//return nil, &operationstorage.PersistentOperationNotFoundError{
```
jensneuse, line 50: Remove comment or clean up code?
Seems like this might not be 100% done yet?

router/pkg/config/config.schema.json
```diff
@@ -52,6 +52,24 @@
             }
           }
         },
+        "kv": {
+          "type": "array",
+          "items": {
+            "type": "object",
+            "required": ["url", "id"],
+            "additionalProperties": false,
+            "properties": {
```
jensneuse, line 61: Should we add kv type to distinguish kvs? E.g. we could use Redis or NATS as the kv.

router/pkg/config/config.schema.json
```diff
@@ -52,6 +52,24 @@
...
+              "id": {
+                "type": "string",
+                "description": "The ID of the storage provider. The ID is used to identify the storage provider in the configuration."
+              },
+              "url": {
```
jensneuse, line 66: Should we have additional fields for security or auth? Or is url enough to properly configure a Redis instance in production?

router/pkg/config/config.schema.json
```diff
@@ -130,6 +148,53 @@
...
     },
+    "automatic_persisted_queries": {
+      "type": "object",
+      "additionalProperties": false,
+      "description": "The configuration for the automatic persisted queries (APQ).",
+      "required": ["enabled"],
+      "properties": {
+        "enabled": {
+          "type": "boolean",
+          "description": "Enable fallback storage provider. The default value is false."
```
jensneuse, line 159: I don't understand the description. What is fallback here?

router/pkg/config/config.schema.json
```diff
@@ -130,6 +148,53 @@
...
+        },
+        "cache": {
```
jensneuse, line 161: Can you add a description on what the cache is used for vs the storage?
Maybe this description could also live one layer above, but it's also ok to explain it in the cache and storage sections.

router/pkg/config/config.schema.json
```diff
@@ -130,6 +148,53 @@
...
+                "minimum": "10MB"
+              },
+              "format": "bytes-string",
+              "description": "The size of the cache used. The default value is 100MB."
+            },
+            "ttl": {
+              "type": "integer",
+              "default": -1,
+              "minimum": -1,
+              "description": "The ttl of the cache. The ttl is used to specify the time-to-live of the cache. The default value is -1, which means that the cache never expires."
```
jensneuse, line 178: Is the ttl in seconds? Maybe make this explicit and clear.

jensneuse (changes requested):
Overall some minor requests.

There's one thing though I'd like you to add.

Can you create a markdown file that explains all possible flows and how they are implemented?

E.g. non APQ client with PO, APQ client without PO, APQ client without PO, etc...

I'd like to understand your mental model behind how each of the situations is handled, which is a bit hard to extract from code.

I think this will help understand if our general approach is right.

I think in general the situation is a bit complicated, especially when you have both APQ and PO enabled, which can make it hard to distinguish between the different cases.

router/core/operation_processor.go
df-wg: The one case that this causes issues with is `ttl expiration`, because we can't enforce that the KV will drop the key when the TTL is over. Does that matter?

router/internal/persistedoperation/apq/client.go
df-wg: I think that that would add redundant storage, what do you think? Apollo, for instance, only uses one at a time, it seems

router/pkg/config/config.schema.json
df-wg: I would assume just url to start with (with the auth string being included in the URL), but whatever you think makes sense

router/core/operation_processor.go
df-wg: It would. I can change the `Client` interface so that it will also return the source, which could be useful

router/pkg/config/config.schema.json
df-wg: I think users can configure an instance at this point with `redis://user:pass@localhost:6379`, and we can always add in additional fields for extra auth as customers request it

router/internal/persistedoperation/apq/client.go
StarpTech: I agree, how should that even work? @jensneuse are you suggesting to have as fallback cache or what is your motivation? I'd keep it simple for now.

router/internal/persistedoperation/README.md
```diff
@@ -0,0 +1,17 @@
...
+1. **No APQ, No Persisted Operations** &rarr; If a persisted operation is sent, the router will return an error, as there are no persisted operations stored. Even if a query is sent, the router will still error because APQ isn't enabled.
+1. **APQ and Persisted Operations** &rarr; In this scenario, the router will first check if the query is stored in the persisted operation files. If it is, the router will execute the query. If it is not, the router will check the APQ cache. If the query is found, the router will execute the query. If the query is not found, the router will check if the query was sent together with the persisted operation hash. If it was, the router will execute the query and save it in the APQ cache for the future. If it was not, the router will return an error.
```
StarpTech, line 17: This sounds doesn't right to me. We should check first if the query is in the APQ cache before looking into external sources like files.

router/internal/persistedoperation/operationstorage/cache.go
```diff
@@ -14,6 +16,25 @@ type OperationsCache struct {
...
 
+func NewOperationsCache(cacheSize int64) (*OperationsCache, error) {
+	if cacheSize <= 0 {
+		return &OperationsCache{}, nil
+	}
+
+	cache, err := ristretto.NewCache(&ristretto.Config[string, []byte]{
+		// assume an average of persistentAverageCacheEntrySize per operation, then
+		// multiply by 10 to obtain the recommended number of counters
+		NumCounters: (cacheSize * 10) / persistentAverageCacheEntrySize,
```
StarpTech, line 27: This calculation is very imprecise. We can't make any assumption on average size. I know this existed before and we should come up with something better.

router/core/operation_processor.go
```diff
@@ -369,13 +369,28 @@ func (o *OperationKit) FetchPersistedOperation(ctx context.Context, clientInfo *
 		return true, nil
 	}
 
-	persistedOperationData, err := o.operationProcessor.persistedOperationClient.PersistedOperation(ctx, clientInfo.Name, o.parsedOperation.GraphQLRequestExtensions.PersistedQuery.Sha256Hash)
+	persistedOperationData, isApq, err := o.operationProcessor.persistedOperationClient.PersistedOperation(ctx, clientInfo.Name, o.parsedOperation.GraphQLRequestExtensions.PersistedQuery.Sha256Hash)
```
StarpTech, line 372: APQ over GET must only work with operations of type Query. Mutations aren't allowed over GET and must be aborted with a proper err message.

https://www.apollographql.com/docs/apollo-server/performance/apq#using-get-requests-with-apq-on-a-cdn

router/internal/persistedoperation/operationstorage/cache.go
```diff
@@ -14,6 +16,25 @@ type OperationsCache struct {
 	Cache *ristretto.Cache[string, []byte]
 }
 
+func NewOperationsCache(cacheSize int64) (*OperationsCache, error) {
+	if cacheSize <= 0 {
+		return &OperationsCache{}, nil
+	}
+
+	cache, err := ristretto.NewCache(&ristretto.Config[string, []byte]{
```
StarpTech, line 24: Ristretto needs to be closed otherwise it leaks goroutines. The client needs to be shutdown on server shutdown.

router/pkg/config/config.schema.json
StarpTech: Sounds good to me. URL was sufficient in case of Redis.

StarpTech (changes requested):

router/internal/persistedoperation/operationstorage/cache.go
df-wg: Do you have a proposal for how we should create this?

StarpTech: You can have a custom cost function in ristretto and count the document bytes.

router/pkg/config/config.schema.json
df-wg: @StarpTech 
```
storage_providers:
  kv: 
    - id: "my_redis"
      url: "redis://localhost:6379"
      type: redis
    - id: "my_nats"
      url: "nats-url"
      type: nats
```

Sound ok?

router/core/operation_processor.go
df-wg: Added an explicit test to verify that this is rejected

router/internal/persistedoperation/operationstorage/cache.go
df-wg: Added to the close function

router/pkg/config/config.schema.json
StarpTech: Why do you think this necessary to specify? I think I'm lacking context here.

router/internal/persistedoperation/operationstorage/cache.go
df-wg: that cost function sounds pretty close to what I'd envision the default cost function to be, and it doesn't address what default num counters we should provide

router/pkg/config/config.schema.json
df-wg: @StarpTech So does that mean you don't think we need a new `kv` section in `storage_providers`? I'd argue that we should have one, but I do agree that i don't think we need to provide the `type` in the `kv`'s

router/internal/persistedoperation/operationstorage/cache.go
df-wg: Updated

router/internal/persistedoperation/apq/client.go
jensneuse: ok

df-wg: At this point, we decided not to do that, especially since we'll be reworking the L2 normalisation cache to include APQ. If that happens, we'll revisit this.

router/internal/persistedoperation/apq/client.go
```diff
@@ -0,0 +1,99 @@
...
+type Client interface {
+	Enabled() bool
+	PersistedOperation(ctx context.Context, clientName string, sha256Hash string) ([]byte, error)
+	SaveOperation(ctx context.Context, clientName, sha256Hash string, operationBody []byte) error
+	Close()
+}
+
+type KVClient interface {
+	Get(ctx context.Context, clientName, operationHash string) ([]byte, error)
+	Set(ctx context.Context, operationHash string, operationBody []byte, ttl int) error
```
jensneuse, line 26: Can you add a description that adds the ttl unit? (seconds?)

router/internal/persistedoperation/apq/client.go
```diff
@@ -0,0 +1,99 @@
...
+func (c *client) Enabled() bool {
+	return c.enabled
+}
+
+func (c *client) PersistedOperation(ctx context.Context, clientName string, sha256Hash string) ([]byte, error) {
+	if c.kvClient != nil {
+		return c.kvClient.Get(ctx, clientName, sha256Hash)
+	}
+
+	c.cacheLock.RLock()
```
jensneuse, line 76: As you're now using ristretto, which is thread safe, I don't think you still need a lock.

router/internal/persistedoperation/apq/redis.go
```diff
@@ -0,0 +1,63 @@
...
+func (r *redisClient) Get(ctx context.Context, clientName, operationHash string) ([]byte, error) {
+	cmd := r.client.Get(ctx, r.prefix+operationHash)
+	if errors.Is(cmd.Err(), redis.Nil) {
+		return nil, nil
+	}
+	return cmd.Bytes()
+}
+
+func (r *redisClient) Set(ctx context.Context, operationHash string, operationBody []byte, ttl int) error {
+	ttlD := time.Duration(float64(ttl) * float64(time.Second))
```
jensneuse, line 56: I think you can write this as:
```go
time.Duration(ttl) * time.Second
```

jensneuse (changes requested):

router/internal/persistedoperation/operationstorage/cache.go
```diff
@@ -14,6 +12,23 @@ type OperationsCache struct {
...
+		NumCounters: cacheSize * 10,
+		MaxCost:     cacheSize,
+		BufferItems: 64,
```
df-wg, line 23: No need to have a cost function, since every `set` operation passes the appropriate cost along (the length of the document)

router/internal/persistedoperation/apq/client.go
StarpTech: Seems to be still valid.

router/internal/persistedoperation/apq/redis.go
```diff
@@ -0,0 +1,63 @@
...
+		client: innerClient,
+		prefix: opts.Prefix,
+	}
+
+	return rclient, nil
+}
+
+func (r *redisClient) Get(ctx context.Context, clientName, operationHash string) ([]byte, error) {
+	cmd := r.client.Get(ctx, r.prefix+operationHash)
+	if errors.Is(cmd.Err(), redis.Nil) {
```
StarpTech, line 49: In what scenario does this happen?

df-wg: If it's a cache miss from redis, it returns an error of `redis.Nil`. In that scenario, we don't want to return an error because it could be that there is a top-level query that was attached to the request

router/core/operation_processor.go
```diff
@@ -994,11 +1042,22 @@ func NewOperationProcessor(opts OperationProcessorOptions) *OperationProcessor {
...
+			automaticPersistedOperationCacheTtl: float64(opts.AutomaticPersistedOperationCacheTtl),
 			persistedOperationVariableNames:     map[string][]string{},
 			persistedOperationVariableNamesLock: &sync.RWMutex{},
-			persistedOperationCache:             map[uint64]normalizedOperationCacheEntry{},
-			persistedOperationCacheLock:         &sync.RWMutex{},
 		}
+
+		cacheSize := opts.PersistedOperationCacheSize
+		if cacheSize <= 0 {
+			cacheSize = 1024 * 1024
```
StarpTech, line 1052: 1MB as a default seems to be very low. I'd choose the default as in the config.

router/internal/persistedoperation/apq/redis.go
StarpTech: thanks

router/core/operation_processor.go
```diff
@@ -729,6 +756,10 @@ type normalizedOperationCacheEntry struct {
 	operationType            string
 }
 
+func (e normalizedOperationCacheEntry) Len() int {
+	return len(e.normalizedRepresentation) + len(e.operationType) + len(e.operationType)
```
StarpTech, line 760: You added operationType twice.

StarpTech (changes requested):

PR #1332, opened 2024-11-04
chore: sync go work dependencies

Noroth:
## Motivation and Context

Apply missing dependency updates after router update.


Aenimus (approved):
LGTM

PR #1344, opened 2024-11-07
feat(graphqlmetrics): implement more efficient batching of clickhouse entries

Noroth:
- **fix: implement batching behavior for the metrics service**
- **chore: add environment config for batch processor**
- **chore: use worker pool for processing**
- **chore: improve batch processor logging**

## Motivation and Context

Clickhouse works better with bigger batches. Currently for each request that is sent to the metrics endpoints, the provided data would be sent to Clickhouse directly. 

Instead of doing that, the requests should be collected for a certain amount of time and then sent as a bigger chunk.


graphqlmetrics/config/config.go
```diff
@@ -35,6 +35,10 @@ type Config struct {
 	IsPrometheusEnabled  bool   `default:"false" envconfig:"PROMETHEUS_ENABLED"`
 	PrometheusListenAddr string `default:"127.0.0.1:8088" envconfig:"PROMETHEUS_LISTEN_ADDR"`
 	PrometheusPath       string `default:"/metrics" envconfig:"PROMETHEUS_PATH"`
+
+	MaxCostThreshold        int           `default:"10000" envconfig:"MAX_COST_THRESHOLD"`
```
StarpTech, line 39: Please prefix the environment variables to understand that they belong together.

graphqlmetrics/cmd/main.go
```diff
@@ -149,7 +156,7 @@ func main() {
 	wg.Add(1)
 	go func() {
 		defer wg.Done()
-		ms.Shutdown(cfg.ShutdownDelay)
```
StarpTech, line 152: Why did you remove it?

graphqlmetrics/pkg/batch/processor.go
```diff
@@ -0,0 +1,180 @@
...
+// receiveElement receives a batch of buffer adds them to the buffer.
+// If the buffer is full, the batch processing will be invoked.
+func (p *Processor[T]) receiveElement(element T) {
+	p.buffer = append(p.buffer, element)
+
+	if p.costFunc != nil {
+		p.currentBufferCost += p.costFunc(element)
+	}
+
+	p.logger.Debug("current buffer cost", zap.Int("currentCost", p.currentBufferCost))
```
StarpTech, line 171: Please use snake_case everywhere.

graphqlmetrics/core/metrics_service.go
```diff
@@ -33,355 +43,377 @@ type MetricsService struct {
...
+		)
+	})
+
+	return nil
 }
 
-func (s *MetricsService) Shutdown(deadline time.Duration) {
-	s.pool.StopAndWaitFor(deadline)
+// calculateRequestCost the total number of entries of metrics batch.
+func calculateRequestCost(item SchemaUsageRequestItem) int {
```
StarpTech, line 411: Please add a unit test for this.

graphqlmetrics/pkg/batch/processor.go
```diff
@@ -0,0 +1,180 @@
...
+		p.receiveElement(batch)
+	}
+
+	p.logger.Info("processing remaining items in the buffer", zap.Int("buffer_size", len(p.buffer)))
+	if len(p.buffer) > 0 {
+		p.process()
+	}
+}
+
+func (p *Processor[T]) canAcceptBatch() bool {
```
StarpTech, line 140: How does it determine that the processor is ready to accept when it always return true?

graphqlmetrics/pkg/batch/processor.go
```diff
@@ -0,0 +1,180 @@
...
+		case <-ticker.C:
+			p.logger.Debug("interval completed - processing batch")
+			p.process()
+		case element := <-p.queue:
+			p.logger.Debug("receiving element from queue",
+				zap.Any("element", element),
+				zap.Int("queue_size", len(p.queue)),
+			)
+
+			p.receiveElement(element)
```
StarpTech, line 109: You did not reset the timer after we have sent a batch.

StarpTech (changes requested):

graphqlmetrics/pkg/batch/processor.go
Noroth: the metrics_server wasn't using `snake_case` that's why I followed the existing naming. I will update everything

graphqlmetrics/cmd/main.go
Noroth: Initially I removed the pool and wanted to ensure to process everything in the batch processor. It does make sense to have a timeout here. I will re-add it.

graphqlmetrics/pkg/batch/processor.go
StarpTech: Thank you

graphqlmetrics/pkg/batch/processor.go
StarpTech: I got it. It used to block new requests on shutdown.

StarpTech (approved):
LGTM

PR #1345, opened 2024-11-07
chore: improve buffer usage

jensneuse:
- reduce inuse_space by re-using less buffers that can grow very big, e.g. in transport
- reduce object sizes in default structs in hot paths of planner
- re-use less buffers in loader that can grow big


StarpTech (approved):
LGTM

PR #1349, opened 2024-11-08
fix(router): don't flush all redis keys while shutdown

git-hulk:
Currently, shutdown router will flush all keys in Redis instance. This might cause unexpected behavior if users are using Redis instance for other purposes besides the rate limit. And from my understanding, it also makes no sense to delete all rate limiter keys once one of router instances was shutdowned.

## Checklist


git-hulk:
@StarpTech We're studying cosmo since we would like to introduce a gateway for GraphQL, and found this while going through the codebase. Would you mind having a look at this? I think it's a serious issue from my side.

StarpTech:
Hi @git-hulk, this was an oversight from us, great catch! The goal here was to drain the connection. But this is not necessary.

StarpTech:
The term flush has different meaning across the board and providers. 😌

git-hulk:
@StarpTech Thanks for your reply and confirmation.

StarpTech (approved):
LGTM

PR #1352, opened 2024-11-08
fix(graphqlmetrics): improve batch preparation

Noroth:
## Motivation and Context

We should only prepare batches if we really have data to write. it seems like aborting the batch causes Clickhouse to track an error. 

In addition the Clickhouse dependency version was updated


StarpTech (approved):
LGTM

PR #1354, opened 2024-11-08
feat: enables the full cosmo docker example to be built upon the latest rel…

gpazevedo:
## Motivation and Context

The full cosmo docker example currently builds some images using the latest code, which may introduce instability due to untested changes. This PR modifies the example to use the last released code version for building images, ensuring greater stability and reliability.

To maintain flexibility, the --last-release flag has been added. When this flag is provided, the example will explicitly use the last released version. Otherwise, the current behavior is preserved.

## Checklist

### DIFF
--- a/examples/full-cosmo-docker/README.md
+++ b/examples/full-cosmo-docker/README.md
@@ -16,7 +16,7 @@ This example demonstrates how to run the entire Cosmo platform locally with Dock
 1. Start the platform:
 
 ```shell
-./start.sh
+./start.sh --last-release
 ```
 
 2. Navigate to the [Studio Playground](http://localhost:3000/wundergraph/default/graph/mygraph/playground) and login before with the default credentials:
@@ -43,4 +43,4 @@ query MyEmployees {
 }
 ```
 
-After you are done, you can clean up the demo by running `./destroy.sh`.
+After you are done, you can clean up the demo by running `./destroy.sh --last-release`.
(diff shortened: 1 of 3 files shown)


StarpTech:
Hi @gpazevedo thanks for the PR. Since this a monorepo, composed of multiple services. What commit are you referring to latest? IMO we would need a way to pass a set of ENV variables to the docker compose file to lock to specific version of the stack.

gpazevedo:
> Hi @gpazevedo thanks for the PR. Since this a monorepo, composed of multiple services. What commit are you referring to latest? IMO we would need a way to pass a set of ENV variables to the docker compose file to lock to specific version of the stack.

Hi @StarpTech thanks for the review.
The "full-cosmo-docker" demo has two main steps: 
- Deploy locally the containers specified on the docker-compose.full.yml, that deliver all cosmo components and the demo subgraphs. One of them (cdn-server) is built using the current repository code, not necessarily the latest released version.
- Create and publish a demo federated graph based on the demo subgraphs using the latest released "wgc".

I see two options to have a demo based on released versions:
- Use the latest versions of each of the Cosmo components, and to do so, build the cdn-server image based on its latest release code;
- Use the latest release of all Cosmo components when the latest "wgc" was released.

I chose to use the second option.
The command "git tag --sort=-v:refname"  gets exactly this:
the latest version of the Cosmo component whose name is the last alphabetically (wgc)

I agree that we should be explicit about the chosen release. 
I suggest filtering the releases to just the component chosen: wgc or cdn. 
"git tag --sort=-v:refname | grep wgc | head -n 1"
 instead of 
"git tag --sort=-v:refname | head -n 1"

Did I understand it right?
What is the most adequate solution?

PR #1357, opened 2024-11-09
chore: remove go workspace, handle them personal to each developer

StarpTech:
### Introduction
Although go workspace has been built to simplify managing multiple modules in a single codebase, we run into couple of very serious issues that we were not able to solve deterministically.

One example: We upgraded a dependency that is not used in any other Go module. This upgrade had the effect that another module was upgraded. Due to this behaviour it was impossible to understand the dependency tree because the change was not reflected in the mod or sum file of the module. Luckily, our CI has catched this before it become a serious issue.

### Conclusion
We used workspaces the wrong way. Meanwhile, community has built an opinion around the usage of Go workspaces. Go workspaces should be handled as personal to each developer. 

https://github.com/SeldonIO/seldon-core/pull/4676

This makes sense, if you see them as a more manageable approach to replace directives.

More information https://earthly.dev/blog/go-workspaces/

## Checklist


df-wg (approved):
In general, lgtm. One additional change I'd like to see would be explicit recommendations in Contributing.md, perhaps, about how it's best to configure a personal workspace to make things smooth, if there is any difference from our current workflow

thisisnithin (approved):

StarpTech:
@df-wg addressed, thanks.

PR #1358, opened 2024-11-10
feat(router): allow users to have multiple wildcards cores allow_origins

### LINKED ISSUE #879: Cors AllowWildcard never set - Wildcard origins not taken into account

df-wg:
## Motivation and Context

A previous effort, #880, gave the initial steps to enabling multiple wildcards in `cors.allow_origins`. This PR takes it further, ensuring that we can handle a config with the configuration:
```
name: "CORS_ALLOW_ORIGINS"
value: "https://*.example.com,https://example.com,https://*.example.io:*,https://*.example.org,https://*.mytest.amplifyapp.com"
```

Closes #879 

## Checklist


router-tests/cors_test.go
```diff
@@ -0,0 +1,205 @@
+package integration
```
Aenimus, line 1: Perhaps a test for max recursion depth? Change the test instance to 2 or so and check that it fails?

df-wg: good point! Added

router/pkg/cors/config.go
```diff
@@ -17,7 +17,8 @@ type cors struct {
 }
 
 var (
-	DefaultSchemas = []string{
+	maxRecursionDepth = 10 // Safeguard against deep recursion
```
Aenimus, line 20: I think 10 probably makes much more sense!

df-wg: If necessary, we can make it configurable, but there doesn't seem to me to be a need for that at this point

Aenimus: agreed

router/pkg/cors/cors.go
```diff
@@ -123,21 +123,27 @@ func (c *Config) parseWildcardRules() [][]string {
 			continue
 		}
 
-		if c := strings.Count(o, "*"); c > 1 {
```
StarpTech, line 126: Please provide a bench for the regex so we can validate before and after impact.

StarpTech (changes requested):

router/pkg/cors/cors.go
df-wg: added two benches

df-wg:
Bench with wildcard: 
Screenshot 2024-11-11 at 7 42 56 PM

Bench without wildcard:
Screenshot 2024-11-11 at 7 43 44 PM

StarpTech (approved):
LGTM

StarpTech (approved):
LGTM

Issue #1179, opened 2024-09-16
Incorrect ClientInfo Handling When Loading Persisted Operations from Cache


janeli1:
### Component(s)

router

### Component version

N.A.

### wgc version

N.A.

### controlplane version

N.A.

### router version

0.110.1

### What happened?

If possible, please create a PR with a failing test to illustrate the issue clearly.
Otherwise, please attach a minimum reproduction through a GitHub repository that includes
essential information such as the relevant subgraph SDLs.
Please also make sure that the instructions for the reproduction are clear, tested, and fully accurate.

## Description
There is an issue with the caching mechanism for Persisted Operations. When a request with a valid Persisted Operation ID (POID) is made, the corresponding query body is correctly loaded from the cache. However, if a subsequent request is made with the same POID but an incorrect or different clientInfo, the cached query body is still passed through, regardless of the mismatched clientInfo. This behavior leads to inaccurate operations being processed based on stale or incorrect client data, potentially causing unintended results in the application's response.

This reported issue is caused by code here:
https://github.com/wundergraph/cosmo/blob/router%400.110.1/router/core/operation_processor.go#L317
The load PO from cache is checking only POID.
This is different from how the router check on POclient cache here with ClientInfo and the POID: https://github.com/wundergraph/cosmo/blob/router%400.110.1/router/internal/persistedoperation/client.go#L73

## Steps to Reproduce
1. Send a HTTP request with a valid POID. Got a valid response.
2. Send a HTTP request with the same POID but will different clientInfo. Got a valid response. 

## Expected Result
1. Send a HTTP request with a valid POID. Got a valid response.
2. Send a HTTP request with the same POID but will different clientInfo. When clientInfo is not supported, expect an error.

## Environment
OS: (e.g., "Ubuntu 20.04")
Package Manager: pnpm, npm, yarn, etc
Compiler(if manually compiled): (e.g., "go 14.2")

StarpTech:
Hi @janeli1, thank you for reporting. We will take a look and come back to you soon.

PR #1361, opened 2024-11-12
fix: enable redis url injection via env var, allow apq from diff clients

df-wg:
## Motivation and Context

This solves two issues pointed out: 
* We want to enable customers to inject a secure redis url, with auth, rather than having it in a config file that can be read. 
* While Persisted Operations need to be bound by client (see #1179), APQ should not be, because it's meant to share operations across clients. This PR ensures that that works, even with a local cache

## Checklist


StarpTech (approved):
LGTM

PR #1381, opened 2024-11-17
feat: better epoll detection, allow to disable epoll

StarpTech:
## Motivation and Context

This PR enables users to disable EPOLL support. In addition, we improved EPOLL support detection by a real integration test. In practice, users won't need to disable it manually anymore. I tested it manually on docker with limited host capabilities.

I also get rid of the terms EPOLL and KQUEUE in the configuration because they depends on the operating system. We abstract them away.

### Breaking changes

The following internal engine flags (including their ENV equivalent) have been renamed for clarity.

```
EnableWebSocketEpollKqueue => EnableNetPoll
EpollKqueuePollTimeout => WebSocketClientPollTimeout
EpollKqueueConnBufferSize => WebSocketClientConnBufferSize
WebSocketReadTimeout => WebSocketClientReadTimeout
```

Requires: https://github.com/wundergraph/graphql-go-tools/pull/984

## Checklist


router-tests/go.mod
```diff
@@ -176,5 +176,5 @@ require (
 replace (
 	github.com/wundergraph/cosmo/demo => ../demo
 	github.com/wundergraph/cosmo/router => ../router
-// github.com/wundergraph/graphql-go-tools/v2 => ../../graphql-go-tools/v2
+//  github.com/wundergraph/graphql-go-tools/v2 => ../../graphql-go-tools/v2
```
Aenimus, line 179: I don't think you meant to commit this

router-tests/testenv/testenv.go
```diff
@@ -628,9 +628,9 @@ func configureRouter(listenerAddr string, testConfig *Config, routerConfig *node
 			EnablePersistedOperationsCacheResponseHeader: true,
 			EnableNormalizationCacheResponseHeader:       true,
 		},
-		EpollKqueuePollTimeout:         300 * time.Millisecond,
-		EpollKqueueConnBufferSize:      1,
-		WebSocketReadTimeout:           time.Millisecond * 100,
+		WebsocketClientPollTimeout:     300 * time.Millisecond,
+		WebsocketClientConnBufferSize:  1,
+		WebsocketClientReadTimeout:     time.Millisecond * 100,
```
Aenimus, line 633: Why `time * constant` and then `constant * time`? I think it's best to pick a single style and stick to it. WDYT?

router/core/websocket.go
```diff
@@ -121,19 +121,31 @@ func NewWebsocketMiddleware(ctx context.Context, opts WebsocketMiddlewareOptions
...
-	if opts.EnableWebSocketEpollKqueue {
-		poller, err := epoller.NewPoller(opts.EpollKqueueConnBufferSize, opts.EpollKqueuePollTimeout)
-		if err == nil {
-			opts.Logger.Debug("Epoll is available")
-
-			handler.epoll = poller
-			handler.connections = make(map[int]*WebSocketConnectionHandler)
-			go handler.runPoller()
+	if opts.EnableNetPoll {
+		if err := netpoll.Supported(); err != nil {
```
Aenimus, line 125: Is there a way we can simplify this nested if else? It's not so easy to read IMO

router/core/factoryresolver.go
```diff
@@ -84,11 +85,18 @@ func NewDefaultFactoryResolver(
 		factoryLogger = abstractlogger.NewZapLogger(log, abstractlogger.DebugLevel)
 	}
 
+	var netPollConfig graphql_datasource.NetPollConfiguration
+
+	netPollConfig.ApplyDefaults()
+
+	netPollConfig.Disable = !enableNetPoll
```
Aenimus, line 92: Would it have made more sense for this field to be the same as the option? i.e., `netPollConfig.Enable`? Was the logic already written for Disable?

Also, IMO, I prefer boolean state representations to be an adjective, i.e., `DisableD` (or `EnableD`).

Aenimus:
The tag release you reference doesn't have a package that is used somewhere in Cosmo according to the failed tests.

router-tests/go.mod
StarpTech: Why not? I corrected the identation.

Aenimus: I see. But the other lines are a tab character, and this is a second space. It still doesn't align. At least not on my end.

router-tests/testenv/testenv.go
StarpTech: I didn't touch this. Those are style issue and should be ensured by a linter. We don't have such rules and therefore personal preference.

Aenimus: It doesn't matter which is picked, but it's very odd to have two contrasting styles almost directly juxtaposed.

StarpTech:
> The tag release you reference doesn't have a package that is used somewhere in Cosmo according to the failed CI.

This is fine because we haven't released the engine yet.

router/core/factoryresolver.go
StarpTech: This code existed already. I like you thoughts here but it would be more efficient to review only the code I touched. We can address your points in a refactor. Nonetheless, I addressed the points because it was easy to do.

StarpTech:
> Ah, you were using replace with your local branch? But then should this PR be a draft?

Yes, could be, but it was ready to review. I was only waiting for the engine release.

router/core/factoryresolver.go
Aenimus: I just wanted to write all my thoughts here. I didn't request any changes, so making no changes would also be understandable.

router/core/websocket.go
StarpTech: Probably, but with more efforts. A single nested if should be fine.

router-tests/testenv/testenv.go
StarpTech: Addressed, but IMO we should not focus on this in a PR.

Aenimus (approved):
LGTM; I believe CI will be fine upon engine tag update.

PR #1390, opened 2024-11-20
feat(router): add ability to set auth scopes in custom module

df-wg:
## Motivation and Context

Users occasionally want to be able to set authorization scopes in a custom module, based on an external/legacy auth system. This enables that functionality, and creates a test module demonstrating it. 

## Checklist


router/pkg/authentication/authentication.go
```diff
@@ -53,6 +55,20 @@ func (a *authentication) Claims() Claims {
...
 
+func (a *authentication) SetScopes(scopes []string) {
+	if a == nil {
+		return
+	}
+	if a.claims == nil {
+		a.claims = make(Claims)
+	}
+	existingScopes := a.claims["scope"]
+	if existingScopes != nil && existingScopes.(string) != "" {
```
df-wg, line 66: Question for reviewers: should this function overwrite the existing scopes, or set them in addition? I had assumed that it would set them in addition

StarpTech: Good question. I'd leave it up to the user, if he wants to copy the current values before setting the new values. So it should not append.

router/pkg/authentication/authentication.go
```diff
@@ -53,6 +55,20 @@ func (a *authentication) Claims() Claims {
...
+		scopes = append(scopes, existingScopes.(string))
+	}
+	a.claims["scope"] = strings.Join(scopes, " ")
```
StarpTech, line 69: Could you reference https://datatracker.ietf.org/doc/html/rfc8693#name-request

router-tests/modules/set_scopes_test.go
```diff
@@ -0,0 +1,76 @@
+package module_test
+
+import (
+	"github.com/stretchr/testify/require"
+	"github.com/wundergraph/cosmo/router-tests/jwks"
+	"github.com/wundergraph/cosmo/router-tests/testenv"
+	setScopesModule "github.com/wundergraph/cosmo/router/cmd/custom-set-scopes/module"
```
StarpTech, line 7: Router tests should not reference to an example. Please create on in the test-suite

StarpTech (changes requested):

router/pkg/authentication/authentication.go
df-wg: added a comment with it. I couldn't find an HTTP constant for "scope"

router-tests/modules/set_scopes_test.go
df-wg: That's what we currently do with the existing modules test, we use the test module. Even so, I should move this test module to a separate directory for the test suite?

StarpTech: Then by accident. Modules in router/cmd/ are demonstration exampls, not intended for testing.

StarpTech (changes requested):

router-tests/modules/set_scopes_test.go
```diff
@@ -0,0 +1,120 @@
...
+		Name:         jwksName,
+		URL:          authServer.JWKSURL(),
+		TokenDecoder: tokenDecoder,
+	}
+	authenticator, err := authentication.NewHttpHeaderAuthenticator(authOptions)
+	require.NoError(t, err)
+	return []authentication.Authenticator{authenticator}, authServer
+}
+
+func TestCustomModuleSetScopes(t *testing.T) {
```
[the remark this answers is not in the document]
df-wg: the tests use a query which queries `startDate`, which uses the `@requireScopes` directive
https://github.com/wundergraph/cosmo/blob/25652db26aea984830c8f451d6ba2c265f38fbf5/demo/pkg/subgraphs/employees/subgraph/schema.graphqls#L137

StarpTech (approved):
LGTM

PR #1392, opened 2024-11-21
feat: add validation for V2 @external directive

Aenimus:
## Checklist


.github/workflows/codeql.yml
```diff
@@ -18,6 +18,8 @@ on:
   pull_request:
     branches:
       - main
+    paths-ignore:
```
StarpTech, line 23: This will exclude it only for PR not on main. You can do https://stackoverflow.com/a/78835420

composition/src/errors/errors.ts
```diff
@@ -709,7 +709,7 @@ export function undefinedFieldInFieldSetErrorMessage(
 ): string {
   return (
     ` The following field set is invalid:\n  "${fieldSet}"\n` +
-    ` This is because of the selection set corresponding to the field coordinate "${parentTypeName}.${fieldName}".\n` +
+    ` This is because of the selection set corresponding to the Field coordinates "${parentTypeName}.${fieldName}".\n` +
```
StarpTech, line 712: Why is field capitalized?

composition/src/errors/errors.ts
```diff
@@ -709,7 +709,7 @@ export function undefinedFieldInFieldSetErrorMessage(
 ): string {
   return (
     ` The following field set is invalid:\n  "${fieldSet}"\n` +
```
StarpTech, line 711: Here lowercased

StarpTech (changes requested):

composition/src/errors/errors.ts
Aenimus: GraphQL jargon like Object, Interface, Enum, etc. are capitalised in our messages now. If they're not, it's because I missed them.

composition/src/errors/errors.ts
Aenimus: Never mind—looks like I chose to keep field set lower case.

composition/tests/fieldset-directives.test.ts
```diff
@@ -15,6 +15,7 @@ import {
   invalidSelectionOnUnionErrorMessage,
   invalidSelectionSetDefinitionErrorMessage,
   invalidSelectionSetErrorMessage,
+  N_A,
```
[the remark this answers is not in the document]
Aenimus: Is it necessary? The tests call the normalizer directly, and the API doesn't demand a subgraph name (it does from the federation flow). So the name is set to default value N_A. Soon, even the normalizer will always require a subgraph name. There's a ticket for it.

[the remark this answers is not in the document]
Aenimus: I will change the constant name to be clearer.

StarpTech (approved):
LGTM

PR #1397, opened 2024-11-24
fix: provider should be specified in the config.yaml

alepane21:
## Checklist

### DIFF
--- a/router-tests/testenv/testenv.go
+++ b/router-tests/testenv/testenv.go
@@ -88,6 +88,17 @@ func Run(t *testing.T, cfg *Config, f func(t *testing.T, xEnv *Environment)) {
 	f(t, env)
 }
 
+func RunWithError(t *testing.T, cfg *Config, f func(t *testing.T, xEnv *Environment)) error {
+	t.Helper()
+	env, err := createTestEnv(t, cfg)
+	t.Cleanup(env.Shutdown)
+	if err == nil {
+		f(t, env)
+	}
+
+	return err
+}
+
 func Bench(b *testing.B, cfg *Config, f func(b *testing.B, xEnv *Environment)) {
 	b.Helper()
 	b.StopTimer()
@@ -146,6 +157,7 @@ type Config struct {
 	Logger                             *zap.Logger
 	AccessLogger                       *zap.Logger
 	AccessLogFields                    []config.CustomAttribute
+	ModifyEventsConfiguration          func(cfg *config.EventsConfiguration)
 }
 
 type SubgraphsConfig struct {
@@ -492,10 +504,9 @@ func createTestEnv(t testing.TB, cfg *Config) (*Environment, error) {
 		}
 	}
 
+	var routerStartErr = make(chan error, 1)
 	go func() {
-		if err := rr.Start(ctx); err != nil {
-			t.Fatal("Could not start router", zap.Error(err))
-		}
+		routerStartErr <- rr.Start(ctx)
 	}()
 
 	graphQLPath := "/graphql"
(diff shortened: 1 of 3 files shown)


router-tests/testenv/testenv.go
```diff
@@ -423,6 +436,21 @@ func createTestEnv(t testing.TB, cfg *Config) (*Environment, error) {
...
+	if !cfg.ForcePubSubChecks {
+		for _, datasource := range routerConfig.EngineConfig.DatasourceConfigurations {
+			if customEvents := datasource.CustomEvents; customEvents != nil {
+				customEvents.DontVerify = true
+			}
+		}
+		for _, ffConfig := range routerConfig.FeatureFlagConfigs.GetConfigByFeatureFlagName() {
+			for _, datasource := range ffConfig.EngineConfig.DatasourceConfigurations {
+				if customEvents := datasource.CustomEvents; customEvents != nil {
+					customEvents.DontVerify = true
```
StarpTech, line 448: I'm not sure why we need this flag. It is also not the right way to extend the node_pb definition when this information is set at the router.

StarpTech (changes requested):

router-tests/testenv/testenv.go
devsergiy: +1 why do we need to disable this validation?

The only case is when you don't have edfs configuration at all, but then buildPubSub configuration won't be called

If we really need such an option I would suggest to name it differently - smth like

```
DisableEventProviderIdExistenceValidation
```

Just something more describing an intent of that option

StarpTech: Create a separate graph configuration without the EDFS graph. In that case, you don't have to deal with cross-concern issues.

alepane21: I made the changes that we discussed!

router-tests/testenv/testenv.go
```diff
@@ -605,9 +616,10 @@ func createTestEnv(t testing.TB, cfg *Config) (*Environment, error) {
 		}
 	}
 
-	if err := rr.Start(ctx); err != nil {
-		return nil, err
-	}
+	var routerStartErr = make(chan error, 1)
+	go func() {
```
StarpTech, line 620: Why is this needed? The server is already listening on a seperate go routine. Start return an error only when an error happened at initialization.

router/core/graph_server.go
```diff
@@ -1131,6 +1131,11 @@ func (s *graphServer) buildPubSubConfiguration(ctx context.Context, engineConfig
 					break
 				}
 			}
+
+			_, ok = s.pubSubProviders.nats[providerID]
+			if !ok {
+				return fmt.Errorf("failed to find Nats provider with ID \"%s\"", providerID)
```
StarpTech, line 1137:
```suggestion
				return fmt.Errorf("failed to find Nats provider with ID \"%s\". Ensure the provider definition is part of the config", providerID)
```

StarpTech (changes requested):

router-tests/testenv/testenv.go
alepane21: The problem is that the initialization process can take some time, so I need the channel to be sure that when the error is returned, it is caught.

router/core/graph_server.go
alepane21: Fixed for NATS and Kafka error message, thanks!

StarpTech (approved):
LGTM

router-tests/testenv/testenv.go
StarpTech: Discussed and updated.

PR #1410, opened 2024-12-02
fix: race on client request stop

jensneuse:
Execute synchronous subscription updates, e.g. when using SSE or Multipart, on the goroutine of the request. This resolves races between ctx cancelled and net/http server finishing request while an update is being written and flushed into the conn.


StarpTech (approved):
LGTM

PR #1413, opened 2024-12-03
fix: ignore internal cost for ristretto caches

Noroth:
## Motivation and Context

Ristretto caches contain an internal cost, which will be added upon each item that is supposed to be written to a cache. As an example with a cache size of `1024` and a cost per item of `1` in `Set()`, we would anticipate, that we could have up to 1024 items in the cache.

If `IgnoreInternalCost` is not set to `true`, for each `Set` operation, the provided cost will be added to an internal cost which defaults to around `56`. Therefore we would have a total cost of 57 for adding a single item to the cache.

Instead of now being able to store `1024` items, we are only able to store `17` items before keys are starting to get evicted.

**This PR changes the config to Ignore the internal cost per cache**

> [!WARNING]
> This change will improve the cache utilization but also have an impact on memory. 
> Example storing pointers to a struct with the size of `80` byte in a cache with `1024` max cost.
> In theory we would then have `ptr_size * struct_size * cache_size * size_of_internal_struct`
> `8 * 80 * 1024 * 56 = 36700160` up to 36 MB of memory per cache
> (deviations possible based on OS)

Before:
On the bottom graphs we can see that the `y` value for the cost on the left is higher than the `y` value of the keys on the right. 
image

After:
With the change, the cache cost will now reflect the actual number of keys.
image

## Checklist


StarpTech (approved):
LGTM

jensneuse (approved):

PR #1415, opened 2024-12-03
feat: add error message on invalid JSON in edfs message

alepane21:
## Motivation and Context

When a malformed message is received on the subscription, a general ‘Internal Server Error’ is currently returned. Instead, we should return a more specific error to provide the client with additional information.

## Checklist


router/core/errors.go
```diff
@@ -74,10 +76,12 @@ func getErrorType(err error) errorType {
 		return errorTypeEDFS
 	}
 	var invalidWsSubprotocolErr graphql_datasource.InvalidWsSubprotocolError
-
 	if errors.As(err, &invalidWsSubprotocolErr) {
 		return errorTypeInvalidWsSubprotocol
 	}
+	if strings.HasPrefix(err.Error(), "cannot parse JSON:") {
```
jensneuse, line 82: Is there a way to define this error in a typesafe way, which we can evaluate with `errors.Is` instead of making a string prefix check? If the error message changes at some point, this will fail. Wrapping errors can easily cause this.

jensneuse (changes requested):

router/core/errors.go
alepane21: The issue lies in how the error is handled: it’s currently returned as a general error [here in graphql-go-tools](https://github.com/wundergraph/graphql-go-tools/blob/v2.0.0-rc.130/v2/pkg/engine/resolve/resolvable.go#L128), and it originates [here in astjson](https://github.com/wundergraph/astjson/blob/44485579ffa51f8a2870469f2746faf7179543de/parser.go#L57).

This setup leaves me without a clear way to differentiate or handle the error specifically within this repository.

I see two possible solutions:
	1.	Modify the error returned in graphql-go-tools.
	2.	Adjust the error generation in astjson.

Which approach do you think would be more appropriate? Or is there a better way to handle this that I might be missing?

jensneuse (approved):

StarpTech (approved):
LGTM

PR #1416, opened 2024-12-03
chore: add rfc for resolve directive

jensneuse:


rfc/resolve-directive.md
```diff
@@ -0,0 +1,189 @@
...
+
+## Solution
+
+I propose to add "internal" fields which we mark as `@inaccessible`.
+These can be used by the Router to resolve information that is useful to other Subgraphs, but not directly accessible as keys.
+
+In addition, I propose a new directive `@resolve` which allows us to define a query on an input field.
+This input field can be market as `@inaccessible` to prevent exposing it to the public API.
+As such, clients are not able to define the input field in their queries.
+Instead, the Router will execute the query defined in the `@resolve` directive and pass the result to the input.
```
Aenimus, line 34: `@resolve` only makes sense as a name if we intend this functionality for several use cases.

In this RFC, what is actually being conducted is a prerequisite or a prevalidation.

rfc/resolve-directive.md
```diff
@@ -0,0 +1,189 @@
...
+```graphql
+# Product Admin Subgraph
+
+input RemoveReviewInput {
+    productID: ID!
+    reviewID: ID!
+    # This field is used to validate the review exists
+    # if the Router resolves it to null, the mutation will fail because the field is not nullable
+    # if it's resolved to a value, the mutation will have access to the content as part of the input
+    review: RemoveReviewReviewInput! @inaccessible @resolve(query: "query Review($id: ID!) { reviewByID(id: $id) { id }}", variables: "{ \"id\": {{ .entity.reviewID }} }")
```
Aenimus, line 86: I don't really understand this, honestly.

Why is this input necessary at all? Why isn't `@resolve` added to `@reviewID`? And whatever ID is passed by the user to `reviewID` must be found within the validation query, i.e., `reviewByID`?

Aenimus: I also wonder whether this syntax is overly verbose.

rfc/resolve-directive.md
```diff
@@ -0,0 +1,189 @@
...
+
+As an alternative, it's also possible to use the `@resolve` directive on a field to resolve the value of the field.
+
+```graphql
+# User Subgraph
+type User @key(fields: "id") {
+    id: ID!
+    name: String! @requires(fields: "info")
+    email: String! @requires(fields: "info")
+    info: UserInfo @inaccessible @resolve(query: "query { info: userInfo { id name email }}")
```
Aenimus, line 165: So here, `User.info` is delegating to `@resolve`, which is calling `Query.userInfo`? Although I don't see any queries defined in this subgraph—does that mean the query can be defined anywhere?

Aenimus: And if you can use any query defined anywhere, how do you make sure the mapping/relationship is correct?

rfc/resolve-directive.md
jensneuse: There needs to be a way to pass a header or claim, e.g. "sub", as an argument. How would you solve this?

rfc/resolve-directive.md
jensneuse: Yes, you can query the whole graph.
Composition needs to validate this query.

PR #1417, opened 2024-12-03
fix: disable query plan for subscriptions

thisisnithin:


Aenimus (approved):
LGTM

StarpTech:
@thisisnithin Do we ignore this on the router already or how is it handled when the header is set?

PR #1421, opened 2024-12-05
feat(router): add request_error boolean field to access logs

df-wg:
## Motivation and Context

With our work in adding access logs, we want to add a boolean field that users can watch to give them a simple indication if there was an error found in the request, whether a panic or a graphql error. It is also present in subgraph access logs, so it will exhibit which subgraph had the error

## Checklist

### DIFF
--- a/router-tests/structured_logging_test.go
+++ b/router-tests/structured_logging_test.go
@@ -1645,7 +1645,6 @@ func TestAccessLogs(t *testing.T) {
 					"path":                   "/graphql",
 					"query":                  "", // http query is empty
 					"ip":                     "[REDACTED]",
-					"error":                  "Failed to fetch from Subgraph 'products' at Path: 'employees'.",   // Automatically added from error
 					"service_name":           "service-name",                                                     // From request header
 					"response_error_message": "Failed to fetch from Subgraph 'products' at Path: 'employees'.",   // From context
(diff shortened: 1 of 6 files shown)


router-tests/structured_logging_test.go
```diff
@@ -1567,20 +1637,22 @@ func TestAccessLogs(t *testing.T) {
...
-					"operation_type":   "query",                                                            // From context
+					"log_type":               "client/subgraph",
+					"subgraph_name":          "products",
+					"subgraph_id":            "3",
+					"status":                 int64(403),
+					"method":                 "POST",
+					"path":                   "/graphql",
+					"query":                  "", // http query is empty
+					"ip":                     "[REDACTED]",
+					"error":                  "Failed to fetch from Subgraph 'products' at Path: 'employees'.",   // Automatically added from error
```
df-wg, line 1648: Looking at the normal access logs, it seems like we only automatically add `error` when we encounter a panic issue, but not a normal graphql error. Should I not be adding `error` automatically for subgraphs, and keep it restricted to the `response_error_message` and `request_error` context fields they can add?

StarpTech: That's a great observation. For now, I recommend to follow your proposal and reserve `error` for a different use cases.

StarpTech (approved):
LGTM

PR #1423, opened 2024-12-06
feat: implement otel cardinality limit

Noroth:
## Motivation and Context

To ensure that possible growing metric cardinality will not affect the router stability we should define a limit. The recommended default limit for cardinality is at `2000` 

https://opentelemetry.io/docs/specs/otel/metrics/sdk/

> [!NOTE]
> The OTEL Go sdk does not yet expose a way to configure a custom limiter and needs to be set via an environment variable. 
> This behavior might change in the future.

## Checklist


router/bench-random.js
```diff
@@ -0,0 +1,179 @@
...
+
+export const options = {
+    stages: [
+        { duration: '15s', target: 20 },
+        { duration: '15s', target: 50 },
+        { duration: '20s', target: 100 },
+    ],
+};
+
+function generateOperationName(length) {
```
StarpTech, line 17: You can use 
```ts
import { randomString } from 'https://jslib.k6.io/k6-utils/1.2.0/index.js';
```

router/pkg/metric/metric_store.go
```diff
@@ -141,6 +156,19 @@ func NewStore(opts ...Option) (Store, error) {
 	return h, nil
 }
 
+// setCardinalityLimit sets the cardinality limit for open telemetry.
+// This feature is experimental in otel-go and may be exposed in a different way in the future.
+// In order to avoid creating a large number of metric streams, we set a hard limit that can be collected for a single instrument.
+func setCardinalityLimit(limit int) error {
+	if limit < 0 {
+		// setting a limit of 0 disables  the cardinality limit
```
StarpTech, line 164: space

router/pkg/metric/metric_store.go
```diff
@@ -141,6 +156,19 @@ func NewStore(opts ...Option) (Store, error) {
...
+		limit = 0
+	}
+
+	val := strconv.Itoa(limit)
+	return os.Setenv("OTEL_GO_X_CARDINALITY_LIMIT", val)
```
StarpTech, line 169: This is fine but it is a potential race because setting an env within a test will effects all other tests. We have to write those tests in a separate file and not in parallel.

router/cmd/instance.go
```diff
@@ -2,22 +2,24 @@ package cmd
...
 	"github.com/wundergraph/cosmo/router/pkg/controlplane/selfregister"
 	"github.com/wundergraph/cosmo/router/pkg/cors"
+	"github.com/wundergraph/cosmo/router/pkg/logging"
 	"go.uber.org/automaxprocs/maxprocs"
-
-	"github.com/wundergraph/cosmo/router/core"
 	"go.uber.org/zap"
 )
 
+const cardinalityLimit = 2000
```
StarpTech, line 21: I think this belongs to the metric package.

StarpTech (changes requested):

router/cmd/instance.go
Noroth: Yeah that might be more suited. I honestly just put it there because I was only using it to create the router and otherwise it doesn't have any other use.

I will move it

router/pkg/metric/metric_store.go
Noroth: Good catch

router/pkg/metric/metric_store.go
```diff
@@ -141,6 +159,19 @@ func NewStore(opts ...Option) (Store, error) {
 	return h, nil
 }
 
+// setCardinalityLimit sets the cardinality limit for open telemetry.
+// This feature is experimental in otel-go and may be exposed in a different way in the future.
+// In order to avoid creating a large number of metric streams, we set a hard limit that can be collected for a single instrument.
+func setCardinalityLimit(limit int) error {
+	if limit < 0 {
+		// setting a limit of 0 disables the cardinality limit
+		limit = 0
```
StarpTech, line 168: Why should prevent someone from disabling it at all? I think we should just support the full API and only enforce a default limit.

router/pkg/metric/metric_store.go
```diff
@@ -114,12 +126,18 @@ type (
...
+	h := &Metrics{
+		logger: zap.NewNop(),
+	}
 
 	for _, opt := range opts {
 		opt(h)
 	}
 
+	if err := setCardinalityLimit(h.cardinalityLimit); err != nil {
+		h.logger.Warn("failed to set cardinality limit", zap.Error(err))
```
StarpTech, line 138: Always uppercase in the beginning

StarpTech (changes requested):

StarpTech (approved):
LGTM

PR #1429, opened 2024-12-10
chore: rfc foundation for @openfed__prerequisite

Aenimus:
## Checklist

### DIFF
--- a/rfc/openfed__prerequisite.md
+++ b/rfc/openfed__prerequisite.md
@@ -19,7 +19,7 @@ This is the proposed definition of the directive (functionality is discussed in 
 ```graphql
 directive @openfed__prerequisite(
   resolveEntity: openfed__ResolveEntityInput
-  resolveQuery: openfed__QueryInput
+  resolveQuery: openfed__ResolveQueryInput
 ) on ARGUMENT_DEFINITION
 
 input openfed__ResolveEntityInput {
--- a/rfc/openfed__prerequisite.md
+++ b/rfc/openfed__prerequisite.md
@@ -1,0 +1,408 @@
+---
+title: "@openfed__prerequisite directive to conduct behaviour before another operation."
+author: David Stutt
+---
+
+# Overview
+The `@openfed__prerequisite` directive (henceforth simply "prerequisite") intends to be a flexible and powerful addition
+to GraphQL Federation.
+When included in a schema, the directive will always perform its defined behaviour before the rest of the operation.
+This would allow the user to:
+- Validate an input that references an entity.
+- Validate an input against another operation.
+
+Potential idea and discussion point:
+- Inject input through the response of an exterior query.
+
+# Definition
+This is the proposed definition of the directive (functionality is discussed in the appropriate section):
+```graphql
+directive @openfed__prerequisite(
+  resolveEntity: openfed__ResolveEntityInput
+  resolveQuery: openfed__QueryInput
+) on ARGUMENT_DEFINITION
+
+input openfed__ResolveEntityInput {
+  typeName: String!
+  fields: openfed__InputSet!
+}
+
+input openfed__ResolveQueryInput {
+  query: String!
+}
+
+scalar openfed__InputSet
+```
+
+# Functionality
+
+The `prerequisite` directive (currently) defines two optional arguments.
+The arguments are mutually exclusive; and although they are both optional (limitation of GraphQL), one (and only one) 
+argument _must_ be defined.
+Failure to define an argument, or defining both, will result in a composition error.
+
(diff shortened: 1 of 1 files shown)


rfc/openfed__prerequisite.md
```diff
@@ -0,0 +1,408 @@
...
+
+### The typeName input
+The `typeName` input corresponds to the `__typename` of an entity that defines an entity resolver in the same subgraph
+that defines the directive.
+
+Composition will fail if the type referenced by the `typeName` input is not an Object that defines a valid `@key`
+directive with the `resolvable` argument either omitted or explicitly set to true.
+
+### The fields input
+The `fields` input is type `openfed__InputSet!` (henceforth simply `InputSet`).
```
Aenimus, line 57: I am not convinced the name of this input is best. Some other potential ideas:

- `MappedFieldSet`
- `MapSet`
- `InputToFieldSet`

df-wg: I think that `InputToFieldSet` is too verbose. I personally like `MappedFieldSet`

df-wg: Or you can make it clear that it's specific for `openfed__ResolveEntityInput` by calling it `ResolveFieldSet`

rfc/openfed__prerequisite.md
```diff
@@ -0,0 +1,408 @@
...
+- Naming is irrelevant. If the value of the argument is valid, it will be propagated to the key field.
+
+A composition error will occur if any of these rules are broken.
+
+##### Examples:
+1. Valid: Argument name `id` and key field name `id` match.
+```graphql
+type Mutation {
+    updateEmail(
+      id: ID! @openfed__prerequisite(resolveEntity: { typeName: "User", fields: "id" })
```
clayne11, line 80: If I'm understanding correctly, you have to provide the key fields, and then the rest of the fields get fetched and provided for you by the Router? Do the non-ID fields get stripped out of the composed schema? Otherwise I imagine this would be confusing and people would try passing in these fields.

I'm not totally following the use case — if I want to update the email, why would I fetch the email from the User subgraph? Don't I want to overwrite that value with a new email?

rfc/openfed__prerequisite.md
```diff
@@ -0,0 +1,408 @@
...
+If the entity representation call is successful, data will be returned.
+This means that entity reference is valid and the operation cannot proceed.
+
+### Entity Representation Call returns null
+If the entity representation call is unsuccessful, null will be returned.
+This means that entity reference is invalid and operation cannot proceed.
+Cosmo will produce an error and the operation will not be committed.
+
+### Potential discussion points
+- Multiple entities (batching)
```
clayne11, line 403: This seems very important. Is this API intended to only work for mutations or is this usable for queries as well?

rfc/openfed__prerequisite.md
```diff
@@ -0,0 +1,408 @@
...
+  reviews: [Review!]!
+}
+```
+
+### Entity Representation Call returns data
+If the entity representation call is successful, data will be returned.
+This means that entity reference is valid and the operation cannot proceed.
+
+### Entity Representation Call returns null
+If the entity representation call is unsuccessful, null will be returned.
```
clayne11, line 398: Null will be returned by what / to what? In general, I think a better strategy is to pass the error to the subgraph and allow it to read the error and determine how to proceed. Some errors on some fields will be fail open, and others will be fail closed. IMO we should allow the "client" to decide how to handle failures.

rfc/openfed__prerequisite.md
Aenimus: You want to update an email on a user. How do you know that user you're trying to update with the input is a valid user?

rfc/openfed__prerequisite.md
clayne11: What if it matches, but some of the fields return errors?

Aenimus: I'm not sure what you mean. This is an entity representation call like any other. We just provide indirect access to it through the same subgraph.

Aenimus: I think there is some confusion here. This directive validates an entity. It is not perform across-subgraph calls. You send the primary keys and check whether it matches an entity.

The functionality that I think you're describing is something else. This RFC is currently on the sidebench while thinking about what and how we'd like to address injection/inter-subgraph resolutions.

clayne11: Got it. If you do intend to work on this more it would be helpful to add a couple examples showing the actual requests the Router would make to each subgraph including the request body to better understand the call patterns and intent.

PR #1437, opened 2024-12-12
feat: implement cache warmup using filesystem

jensneuse:


router/core/cache_warmup.go
```diff
@@ -0,0 +1,278 @@
...
+		ClientInfo: item.Client,
+		TraceOptions: resolve.TraceOptions{
+			Enable: false,
+		},
+		ExecutionOptions: resolve.ExecutionOptions{
+			SkipLoader:                 true,
+			IncludeQueryPlanInResponse: false,
+			SendHeartbeat:              false,
+		},
+		TrackSchemaUsageInfo: true,
```
StarpTech, line 255: This should be coupled to `s.graphqlMetricsConfig.Enabled`

router/core/cache_warmup.go
```diff
@@ -0,0 +1,278 @@
...
+	if err != nil {
+		return err
+	}
+
+	if w.complexityLimits != nil {
+		_, _, _ = k.ValidateQueryComplexity(w.complexityLimits, k.kit.doc, w.routerSchema, k.parsedOperation.IsPersistedOperation)
+	}
+
+	planOptions := PlanOptions{
+		Protocol:   OperationProtocolHTTP,
```
StarpTech, line 245: This won't work with subscriptions.

router/core/router.go
```diff
@@ -1707,6 +1709,15 @@ func WithClientHeader(cfg config.ClientHeader) Option {
 	}
 }
 
+func WithCacheWarmupConfig(cfg *config.CacheWarmupConfiguration) Option {
+	return func(r *Router) {
+		if cfg.Workers <= 0 {
```
StarpTech, line 1714: By design, we never do config validation or default handling in `With` methods. Please do it centrally in `NewRouter`

router/pkg/config/config.schema.json
```diff
@@ -1647,6 +1647,57 @@
...
+        "properties": {
+          "enabled": {
+              "type": "boolean",
+              "description": "Enable the cache warmup.",
+              "default": false
+          },
+          "source": {
+            "type": "string",
+            "description": "The source of the cache warmup items can be filesystem, cdn (Cosmo), or s3.",
+            "enum": ["filesystem", "cdn","s3"]
```
StarpTech, line 1663: If we only support filesystem, we can omit cdn and s3 to fail fast.

StarpTech (changes requested):

router/core/cache_warmup.go
jensneuse: We didn't use this at all so I've removed it.

router/core/router.go
jensneuse: We must validate this here because the central validation would not work as we're unable to use it in testenv.

router/core/cache_warmup.go
```diff
@@ -0,0 +1,280 @@
...
+	cfg.Log.Info("Cache warmup - start",
+		zap.Int("workers", cfg.Workers),
+		zap.Int("items_per_second", cfg.ItemsPerSecond),
+		zap.Duration("timeout", cfg.Timeout),
+	)
+	start := time.Now()
+	completed, err := w.run(ctx)
+	if err != nil {
+		if errors.Is(err, context.DeadlineExceeded) {
+			cfg.Log.Error("Cache warmup - timeout",
```
jensneuse, line 69: Add comment: consider increasing timeout

StarpTech (approved):
LGTM

PR #1448, opened 2024-12-16
feat(router): remove default high cardinality attributes

Noroth:
## Motivation and Context

Certain metric attributes can cause an excessive amount of metric cardinality. In order to add more flexibility, those attributes should be removed from the default set and be configurable.

These changes are not applied when using the default cloud exporter.

> [!WARNING]
> This is a breaking change as by default certain attributes are now omitted from the default set of attributes. 
> If metric backends make use of those attributes, they need to be configured or the graphs need to be updated accordingly

## Checklist


router/core/graphql_handler.go
```diff
@@ -205,7 +204,6 @@ func (h *GraphQLHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 		err := h.executor.Resolver.ResolveGraphQLSubscription(ctx, p.Response, writer)
 		requestContext.dataSourceNames = getSubgraphNames(p.Response.Response.DataSources)
-		requestContext.telemetry.AddCustomMetricStringSliceAttr(ContextFieldOperationServices, requestContext.dataSourceNames)
```
StarpTech, line 208: Where are those fields added if not here?

router/core/context.go
```diff
@@ -137,6 +138,96 @@ var metricAttrsPool = sync.Pool{
...
+
+	return attributeMapper{
+		enabled:   enabled,
+		attr:      attr,
+		lookupMap: set,
+	}
+}
+
+func (r *attributeMapper) mapAttributes(attributes []attribute.KeyValue) []attribute.KeyValue {
+	if !r.enabled {
```
StarpTech, line 183: I tested it locally and `enabled` is always false. Shouldn't it be called whenever I want to add a custom metric attribute?

```yaml
telemetry:
  metrics:
    attributes:
      - key: a
        value_from:
          context_field: operation_name
```

StarpTech (changes requested):

router/core/graphql_handler.go
Noroth: They were already added here. I didn't see that they were used in the `graphql_handler.go`:

https://github.com/wundergraph/cosmo/blob/0c6bd041e88bc6026f837fcb5391b663211a829d/router/core/graphql_prehandler.go#L200-L203

router-tests/testenv/testenv.go
```diff
@@ -223,6 +223,7 @@ type Config struct {
 	SubgraphAccessLogsEnabled          bool
 	SubgraphAccessLogFields            []config.CustomAttribute
 	AssertCacheMetrics                 *CacheMetricsAssertion
+	DisableSimulateCloudExporter       bool
```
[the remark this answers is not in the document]
Noroth: I need a bit more clarity on this, because we don't have exporters in tests and basically all of our tests kind of "simulate" to have a cloud exporter.

router/core/context.go
Noroth: I understood that this is the correct behavior? If the default cloud exporter is set, we do not want to filter out or remap any attributes and keep the attributes in.

StarpTech (approved):
LGTM

PR #1449, opened 2024-12-16
fix(router): ensure subgraph timeouts/nil responses are handled correctly

df-wg:
## Motivation and Context

In the event of an unsuccessful call to a subgraph (for instance, an unresponsive server or a timeout), the `http.Response` will be `nil`. In general, our code needs to be resilient to handle that eventually, and we should be verifying that the `http.Request\Response` isn't `nil` before we access subfields, to avoid segfaults. 

In this case, customers reported seeing issues with header propagation when encountering unresponsive calls. This PR adds a nil check for the response to ensure that it doesn't happen, and adds testing to verify it for the future. 

## Checklist


router/core/header_rule_engine.go
```diff
@@ -263,6 +263,11 @@ func (h *HeaderPropagation) OnOriginRequest(request *http.Request, ctx RequestCo
 }
 
 func (h *HeaderPropagation) OnOriginResponse(resp *http.Response, ctx RequestContext) *http.Response {
+	// In the case of an error response, it is possible that the response is nil
+	if resp == nil {
```
jensneuse, line 267: Wouldn't it be the right thing to skip calling ANY OnOriginResponse hook when the response is nil instead of checking it in the handlers?
Is there any benefit in calling the OnOriginResponse hook when there's no origin response due to an error, e.g. timeout?
https://github.com/wundergraph/cosmo/blob/246bcff2029868acf2384569a4ae871dbf7b74ef/router/core/transport.go#L149

jensneuse (changes requested):

router/core/header_rule_engine.go
StarpTech: Unfortunately, no because a user could do something after the failed response e.g. measure, trace etc... 
We could also make the error accessible to the user in the `ctx RequestContext`. It is currently [inaccessible](https://github.com/wundergraph/cosmo/blob/fbd6bac9522b663468acc2a95ea24fc32b142d47/router/core/modules.go#L36) for no reasons.

jensneuse: I understand. How about passing a func to OnOriginResponse?

```go
type OnOriginResponseContext interface {
	RequestContext() RequestContext
	OriginResponse() (*http.Response, error)
}

func (h *HeaderPropagation) OnOriginResponse(ctx OnOriginResponseContext) *http.Response {
	
	res,err := ctx.OriginResponse()
	if err != nil {
		return nil
	}
```

Such an interface would prevent a lot of bugs when implementers of this hook are unaware that resp can be nil.

StarpTech: I like 👍 Let's bring those learnings into the new module system.

df-wg: Should I change that now? Or does the change to the interface require more discussion? I don't know how consumers are using it, and I don't want to spring a breaking change with no warning

StarpTech: We don't want to break it especially not in a fix PR. Let's focus on fixing the issue and have the discussion afterwards.

df-wg: Amazing. So in that case, can you review the PR?

StarpTech: Yes

StarpTech (approved):
LGTM

PR #1452, opened 2024-12-17
feat(router): expose engine statistic metrics

Noroth:
## Motivation and Context

We would like to expose metrics for engine statistics related to subscriptions.

* number of active connections
* number of active subscriptions
* number of active triggers
* count of sent messages from the engine

## Checklist


alepane21 (commented):
Just a test for Ludwig

alepane21 (dismissed):
Just another test for Ludwig

router/pkg/config/config.schema.json
```diff
@@ -851,6 +851,11 @@
                   "default": false,
                   "description": "Enable the collection of metrics for the GraphQL operation router caches. The default value is false."
                 },
+                "engine_stats" : {
```
StarpTech, line 854: We might want to add more in the future. I'd try to be more specific e.g.

```yaml
engine_stats
  connections: true
  websockets: true
```

This also allow the user to opt-in to specific metrics without buying into all metrics which is always a cost factor for them.

StarpTech (changes requested):
Only a few nuts but overall great approach and reuse of WebSocket stats.

router/pkg/config/config.schema.json
```diff
@@ -935,6 +940,11 @@
...
+                  "type": "boolean",
+                  "default": false,
+                  "description": "Enable the collection of metrics for the Engine statistics. The default value is false."
```
StarpTech, line 946: Maybe be a bit more verbose in what metrics we will enable and why they are useful.

router/core/graph_server.go
```diff
@@ -834,7 +874,7 @@ func (s *graphServer) buildGraphMux(ctx context.Context,
 			Subgraphs:                configSubgraphs,
 			RouterEngineConfig:       routerEngineConfig,
 			PubSubProviders:          s.pubSubProviders,
-			Reporter:                 s.websocketStats,
+			Reporter:                 s.engineStats,
```
StarpTech, line 877: How does it behave with multiple mux? Let's meet about this. In theory each mux has different base attributes and different stats.

StarpTech (changes requested):

router/core/graph_server.go
StarpTech: Decided to cover all with the base graph config

StarpTech (approved):
LGTM

PR #1476, opened 2024-12-23
feat: improve rate limiting with better customization

jensneuse:
This PR adds the following 2 features:

1. Being able to add a suffix to the rate liming key from a client request header or claim: This allows to rate limit different users. Currently, there's only a global rate limit
2. Being able to return a custom status code when we reject a request, e.g. 429 (can be a company security policy)


router/core/ratelimiter.go
```diff
@@ -60,11 +82,35 @@ func (c *CosmoRateLimiter) RateLimitPreFetch(ctx *resolve.Context, info *resolve
 	return &resolve.RateLimitDeny{}, nil
 }
 
+func (c *CosmoRateLimiter) generateKey(ctx *resolve.Context) string {
+	if c.keySuffixFromRequestHeader {
+		v := ctx.Request.Header.Get(c.requestHeaderName)
+		if v != "" {
+			return fmt.Sprintf("%s:%s", ctx.RateLimitOptions.RateLimitKey, v)
```
StarpTech, line 89: We should validate the value. Enforcing a max length and strip leading whitespaces from it.

router/core/ratelimiter.go
```diff
@@ -60,11 +82,35 @@ func (c *CosmoRateLimiter) RateLimitPreFetch(ctx *resolve.Context, info *resolve
...
+		}
+	}
+	if c.keySuffixFromClaim {
+		auth := authentication.FromContext(ctx.Context())
+		if auth != nil {
+			claims := auth.Claims()
+			if v, ok := claims[c.claimName]; ok {
+				return fmt.Sprintf("%s:%s", ctx.RateLimitOptions.RateLimitKey, v)
```
StarpTech, line 97: `v` might not be a string. We should validate it.

router/pkg/config/config.go
```diff
@@ -428,6 +433,7 @@ type RateLimitSimpleStrategy struct {
 	Burst                   int           `yaml:"burst" envDefault:"10" env:"RATE_LIMIT_SIMPLE_BURST"`
 	Period                  time.Duration `yaml:"period" envDefault:"1s" env:"RATE_LIMIT_SIMPLE_PERIOD"`
 	RejectExceedingRequests bool          `yaml:"reject_exceeding_requests" envDefault:"false" env:"RATE_LIMIT_SIMPLE_REJECT_EXCEEDING_REQUESTS"`
+	RejectStatusCode        int           `yaml:"reject_status_code" envDefault:"200" env:"RATE_LIMIT_SIMPLE_REJECT_STATUS_CODE"`
```
StarpTech, line 436: Wouldn't this be incompatible with GraphQL over http spec?

StarpTech (changes requested):

StarpTech (approved):
LGTM

router/core/ratelimiter.go
```diff
@@ -60,11 +83,39 @@ func (c *CosmoRateLimiter) RateLimitPreFetch(ctx *resolve.Context, info *resolve
 	return &resolve.RateLimitDeny{}, nil
 }
 
+func (c *CosmoRateLimiter) generateKey(ctx *resolve.Context) string {
+	if c.keySuffixFromRequestHeader {
```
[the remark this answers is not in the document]
chensunny: police like this 

https://stellate.co/docs/graphql-security/rate-limiting/api-reference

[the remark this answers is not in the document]
jensneuse: Implemented using expr lang. Thank you for the feedback.

router/pkg/config/config.schema.json
```diff
@@ -1523,6 +1528,10 @@
         "debug": {
           "type": "boolean",
           "description": "Enable the debug mode for the rate limit."
+        },
+        "key_suffix_expression": {
+          "type": "string",
+          "description": "Use Expr-Lang to define a key suffix for the rate limit, e.g. by using request headers, claims, or a combination of both with a fallback strategy."
```
StarpTech, line 1534:
```suggestion
          "description": "The expression to define a key suffix for the rate limit, e.g. by using request headers, claims, or a combination of both with a fallback strategy. The expression is specified as a string and needs to evaluate to a string. Please see https://expr-lang.org/ for more information."
```

PR #1478, opened 2024-12-23
fix(router): ensure persisted operation hits both with and without op name

df-wg:
## Motivation and Context

We previously included operation name in the persisted operation cache key. As a result, though, when a persisted operation was stored with a operation name, and someone called it without that operation name, it wouldn't return a cache hit, despite the fact that it was a match. This PR removes the operation name from the cache key, given that it is irrelevant to how specialized we want it to be

## Checklist


StarpTech (approved):
LGTM

StarpTech (approved):
LGTM

PR #1480, opened 2024-12-25
feat: allow to conditionally block mutation via expressions

StarpTech:
## Motivation and Context

This pull request introduces the capability to conditionally block mutations, subscriptions, and non-persisted operations. Instead of relying on YAML configuration alone, we have implemented support for dynamic expressions that are evaluated at runtime. Users can leverage request headers, URL parameters, query parameters, and authentication context (such as claims and scopes) to define the condition when operations are blocked. If no condition is provided all operations are blocked.

### Example Configuration

Block by a header condition

```yaml
security:
  block_mutations:
    enabled: true
    condition: |
      // Case insensitive
      request.header.Get('x-block-mutation') == 'yes'
```

Block by a scope condition

```yaml
security:
  block_mutations:
    enabled: true
    condition: |
      'write' in request.auth.scopes
```

The expression parser and runtime are powered by [expr-lang](https://expr-lang.org/), a widely used library in the Go ecosystem. We chose `expr` for its elegant DSL and robust error reporting.  

During router bootstrap, the system will validate that the expression are valid and returns the expected type (a boolean for conditions). If the expression evaluates to an incorrect type, the process will fail with a detailed error message. 

This step forward, will also lay the foundation for a more dynamic approach in applying configuration or extracting values at runtime.

> [!WARNING]  
> As part of the preparations for Cosmo V1, targeted for release in Q1 2025, this pull request introduces essential changes to enhance long-term stability and maintainability. While we strive to minimize breaking changes, they are sometimes necessary to lay the foundation for a more robust and scalable system. This update includes non-backward compatible modifications, specifically impacting the configuration of operation blocking mechanisms.

## Checklist


router-tests/block_operations_test.go
```diff
@@ -0,0 +1,202 @@
...
+	"io"
+	"net/http"
+	"strings"
+	"testing"
+)
+
+func TestBlockOperations(t *testing.T) {
+	t.Parallel()
+
+	t.Run("block mutations", func(t *testing.T) {
```
Noroth, line 17: Test naming could be improved because you are not only testing blocking behavior

```suggestion
	t.Run("test blocking rules for mutations", func(t *testing.T) {
```

router/core/context.go
```diff
@@ -2,6 +2,7 @@ package core
 
 import (
 	"context"
+	"github.com/wundergraph/cosmo/router/internal/expr"
```
Noroth, line 5: nit: sort imports

router/core/context.go
```diff
@@ -245,6 +246,8 @@ type requestContext struct {
 	graphQLErrorCodes []string
 	// telemetry are the base telemetry information of the request
 	telemetry *requestTelemetryAttributes
+	// expressionContext is the context for expressions parser when evaluating dynamic expressions
```
Noroth, line 249: Maybe?
```suggestion
	// expressionContext is the context that will be provided to a compiled expression in order to retrieve data via dynamic expressions
```

router/core/operation_blocker.go
```diff
@@ -2,6 +2,10 @@ package core
 
 import (
 	"errors"
+	"fmt"
+	"github.com/expr-lang/expr/vm"
+	"github.com/wundergraph/cosmo/router/internal/expr"
+	"go.uber.org/zap"
```
Noroth, line 8: nit: sort imports

router/core/websocket.go
```diff
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/wundergraph/cosmo/router/internal/expr"
```
Noroth, line 9: nit: sort imports

router/internal/expr/expr.go
```diff
@@ -0,0 +1,166 @@
...
+	"context"
+	"errors"
+	"fmt"
+	"github.com/expr-lang/expr"
+	"github.com/expr-lang/expr/file"
+	"github.com/expr-lang/expr/vm"
+	"github.com/wundergraph/cosmo/router/pkg/authentication"
+	"net/http"
+	"net/url"
+	"reflect"
```
Noroth, line 13: nit: sort imports

router/internal/expr/expr.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+// Get returns the first value associated with the given key (Exported).
+// For convenience, we export a function to make the work with headers case-insensitive.
+func (r RequestHeaders) Get(key string) string {
+	return r.Header.Get(key)
+}
+
+// LoadRequest loads the request object into the context.
+func (r *RequestRootContext) LoadRequest(req *http.Request) {
```
Noroth, line 58: This function is basically always called directly after creating a new context with 
`expr.RequestRootContext{}`

Maybe this could be a 
```go
NewFromRequest(req *http.Request) RequestRootContext {}
```

router/pkg/config/config.go
```diff
@@ -334,13 +334,28 @@ type EngineExecutionConfiguration struct {
...
+type BlockSubscriptionConfiguration struct {
+	Enabled   bool   `yaml:"enabled" envDefault:"false" env:"BLOCK_SUBSCRIPTION_ENABLED"`
+	Condition string `yaml:"condition" env:"BLOCK_SUBSCRIPTION_CONDITION"`
+}
+
+type BlockNonPersistedConfiguration struct {
+	Enabled   bool   `yaml:"enabled" envDefault:"false" env:"BLOCK_NON_PERSISTED_ENABLED"`
+	Condition string `yaml:"condition" env:"BLOCK_NON_PERSISTED_CONDITION"`
+}
+
```
Noroth, line 351: These types basically are the same and could be simplified into a single type

```suggestion
type BlockConfiguration struct {
	Enabled   bool   `yaml:"enabled" envDefault:"false" env:"ENABLED"`
	Condition string `yaml:"condition" env:"CONDITION"`
}
```

The SecurityConfiguration type could then be replaced with `envPrefix`

Also I think in general the 3 fields cannot have an `env` config anyways as they are complex types?

```go
type SecurityConfiguration struct {
	BlockMutations              BlockMutationConfiguration     `yaml:"block_mutations" envPrefix:"SECURITY_BLOCK_MUTATIONS_"`
	BlockSubscriptions          BlockSubscriptionConfiguration `yaml:"block_subscriptions" envPrefix:"SECURITY_BLOCK_SUBSCRIPTIONS_"`
	BlockNonPersistedOperations BlockNonPersistedConfiguration `yaml:"block_non_persisted_operations" envPrefix:"SECURITY_BLOCK_NON_PERSISTED_OPERATIONS_"`
```

router/internal/expr/expr.go
```diff
@@ -0,0 +1,166 @@
...
+// See https://github.com/expr-lang/expr/issues/734
+type RequestRootContext struct {
+	Request RequestContext
+}
+
+// RequestContext is the context for the request object in expressions. Be aware, that only value receiver methods
+// are exported in the expr environment. This is because the expressions are evaluated in a read-only context.
+type RequestContext struct {
+	Auth   RequestAuth
+	Header RequestHeaders
```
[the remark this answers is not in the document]
StarpTech: Yes, could be. Good point.

router/internal/expr/expr.go
```diff
@@ -0,0 +1,166 @@
...
+type RequestHeaders struct {
+	// Header is the http headers. We expose the header object to work with header case-insensitive
+	Header http.Header
+}
+
+// Get returns the first value associated with the given key (Exported).
+// For convenience, we export a function to make the work with headers case-insensitive.
+func (r RequestHeaders) Get(key string) string {
+	return r.Header.Get(key)
+}
```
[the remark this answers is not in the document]
StarpTech: There is no need to suffix it with header if the user still reference to `request.header` in the expression

router-tests/block_operations_test.go
```diff
@@ -0,0 +1,202 @@
+package integration_test
+
+import (
+	"github.com/stretchr/testify/require"
+	"github.com/wundergraph/cosmo/router-tests/testenv"
+	"github.com/wundergraph/cosmo/router/core"
+	"github.com/wundergraph/cosmo/router/pkg/config"
```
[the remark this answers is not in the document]
StarpTech: Either we automate it or ignore. It is just noise otherwise on a pr review.

router-tests/block_operations_test.go
```diff
@@ -0,0 +1,202 @@
...
+				defer res.Body.Close()
+				require.Equal(t, http.StatusOK, res.StatusCode)
+				data, err = io.ReadAll(res.Body)
+				require.NoError(t, err)
+				require.Equal(t, `{"data":{"addFact":{"description":"description"}}}`, string(data))
+			})
+		})
+	})
+
+	t.Run("block non-persisted operations", func(t *testing.T) {
```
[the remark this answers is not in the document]
StarpTech: 👍

router-tests/block_operations_test.go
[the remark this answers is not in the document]
Noroth: gci is a nice formatter for go imports.

router/pkg/config/config.go
StarpTech: I like. I use envPrefix.

router/internal/expr/expr.go
StarpTech: We would still need to conditionally run `LoadAuth`. I'd then rather keep it consistent with `Load`.

router-tests/block_operations_test.go
StarpTech: Improved

router/core/context.go
StarpTech: Updated it in a separate commit, thx

router/internal/expr/expr.go
```diff
@@ -17,41 +17,39 @@ import (
...
 // are exported in the expr environment. This is because the expressions are evaluated in a read-only context.
-type RequestContext struct {
-	Auth   RequestAuth
-	Header RequestHeaders
-	URL    RequestURL
+type Request struct {
+	Auth   RequestAuth    `expr:"auth"`
+	URL    RequestURL     `expr:"url"`
+	Header RequestHeaders `expr:"header"`
 }
```
Noroth, line 29: I'm fine with the current solution, this would only be another suggestion to avoid the type casting

```suggestion
type accessor interface{
        Get(string) string
}

type Request struct {
	Auth   RequestAuth    `expr:"auth"`
	URL    RequestURL     `expr:"url"`
	Header accessor `expr:"header"`
}
```

Noroth (commented):
Let me know about my last comment otherwise LGTM

router/core/websocket.go
```diff
@@ -792,7 +793,10 @@ func (h *WebSocketConnectionHandler) parseAndPlan(payload []byte) (*ParsedOperat
 	opContext.name = operationKit.parsedOperation.Request.OperationName
 	opContext.opType = operationKit.parsedOperation.Type
 
-	if blocked := h.operationBlocker.OperationIsBlocked(operationKit.parsedOperation); blocked != nil {
+	exprCtx := expr.RequestRootContext{}
+	exprCtx.LoadRequest(registration.clientRequest)
```
jensneuse, line 797: It looks to me like at this point, we've not loaded Auth from the request into the exprCtx.

router/internal/expr/expr.go
```diff
@@ -0,0 +1,169 @@
...
+	"github.com/wundergraph/cosmo/router/pkg/authentication"
+	"net/http"
+	"net/url"
+	"reflect"
+)
+
+// RequestRootContext is the context for expressions parser when evaluating dynamic expressions
+// If possible function calls should be avoided in the expressions as they are much more expensive
+// See https://github.com/expr-lang/expr/issues/734
+type RequestRootContext struct {
```
jensneuse, line 19: As this is the expression context and we're in the expr package, I'd just call this Context.
As we've discussed in #1481 this context will also contain Response, Field, and other values, so I'd keep it generic.

router/internal/expr/expr.go
```diff
@@ -0,0 +1,169 @@
...
+	Host string `expr:"host"`
+	// Path is the path of the URL
+	Path string `expr:"path"`
+	// Query is the parsed query parameters
+	Query map[string]string `expr:"query"`
+}
+
+// RequestHeaders is the context for the headers object in expressions. A user can access the headers directly by
+// key and get the array representation of the header values or use the Get method to get the first value.
+type RequestHeaders map[string][]string
```
jensneuse, line 47: have you considered using http.Header?

router/internal/expr/expr.go
```diff
@@ -0,0 +1,169 @@
...
+type RequestAuth struct {
+	IsAuthenticated bool           `expr:"isAuthenticated"`
+	Type            string         `expr:"type"`
+	Claims          map[string]any `expr:"claims"`
+	Scopes          []string       `expr:"scopes"`
+}
+
+// LoadAuth loads the authentication context into the request object.
+// Must only be called when the authentication was successful.
+func (r *RequestRootContext) LoadAuth(ctx context.Context) {
```
jensneuse, line 88: Do you think it's good to split this into a separate public func?
There's a high likelihood that someone forgets to call it.
How about creating a single func to init an expr.Context from a client request?

router/internal/expr/expr.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+// ResolveBoolExpression evaluates the expression and returns the result as a bool. The exprContext is used to
+// provide the context for the expression evaluation.
+func ResolveBoolExpression(vm *vm.Program, ctx RequestRootContext) (bool, error) {
+	if vm == nil {
+		return false, nil
+	}
+
+	r, err := expr.Run(vm, ctx)
```
jensneuse, line 145: It appears to me as if expr.Run is not thread safe.
I think we need to have a mutex around it, and potentially compile one expr per CPU and then have a slice of expressions (one per CPU) where we can pick one vm from a semaphore.

router/internal/expr/expr.go
```diff
@@ -0,0 +1,169 @@
...
+		return nil, handleExpressionError(err)
+	}
+
+	return v, nil
+}
+
+// ResolveStringExpression evaluates the expression and returns the result as a string. The exprContext is used to
+// provide the context for the expression evaluation.
+func ResolveStringExpression(vm *vm.Program, ctx RequestRootContext) (string, error) {
+	r, err := expr.Run(vm, ctx)
```
jensneuse, line 125: Thread safe?

jensneuse (changes requested):

router/internal/expr/expr.go
StarpTech: That what @Noroth suggested. but`LoadAuth` needs to be called conditionally anyway. For this reason, I wanted to follow the `Load*` pattern rather than going with 2 variants.

router/internal/expr/expr.go
StarpTech: Why do you think it needs to be thread safe? You suggestions sounds like a big pre optimization that isn't required at this point.

router/internal/expr/expr.go
StarpTech: Not required by the implementation.

router/core/websocket.go
StarpTech: Good point, a test is missing.

router/internal/expr/expr.go
StarpTech: Sounds good to me

router/internal/expr/expr.go
StarpTech: In your proposal, the user won't be able to anymore to access the headers by map. I'm also fine with your direction. We don't have the need yet to do it and an interface is a good way to ensure the right contract.

Noroth (approved):
LGTM

router/internal/expr/expr.go
jensneuse: Is this exposed by default or do we have to expose it explicitly?

router/internal/expr/expr.go
jensneuse: I think it needs to be thread safe because you have one single operation blocker which gets called by many requests in different goroutines.

router/internal/expr/expr.go
jensneuse: Why does it need to be called conditionally? Maybe we can aim to create an env context in a central place for each request?

router/internal/expr/expr.go
StarpTech: It is thread safe https://github.com/expr-lang/expr/issues/683#issuecomment-2213958821

router/internal/expr/expr.go
StarpTech: I revisited the implementation and you're right it is required but luckily `vm.program` is already thread safe.

router/internal/expr/expr.go
StarpTech: Everything is exposed by default but only exported methods are accessible within the VM.

router/internal/expr/expr.go
StarpTech: Clarified

router/core/graphql_prehandler.go
```diff
@@ -344,6 +327,26 @@ func (h *PreHandler) Handler(next http.Handler) http.Handler {
 			authenticateSpan.End()
 
 			r = validatedReq
+
+			requestContext.expressionContext.Request.Auth = expr.LoadAuth(r.Context())
+		}
+
+		err = h.handleOperation(r, variablesParser, &httpOperation{
```
StarpTech, line 334: I moved auth before we handle the operation. Make sense to me.

router/pkg/authentication/authentication.go
```diff
@@ -93,12 +93,13 @@ func Authenticate(ctx context.Context, authenticators []Authenticator, p Provide
...
 			continue
 		}
-		if claims != nil {
-			return &authentication{
-				authenticator: auth.Name(),
-				claims:        claims,
-			}, nil
-		}
+
+		// If authentication succeeds, we return the authentication
```
StarpTech, line 97: This was a flaw. As long as the authenticator return no error, we can assume the authentication was successful even when no claims have been configured.

jensneuse (approved):

PR #1481, opened 2025-01-02
chore: add draft for expr context

jensneuse:


rfc/expr-context.md
```diff
@@ -0,0 +1,115 @@
...
+```yaml
+context:
+  agent: "request.auth.claims.sub || request.client.ip || request.header.X-Forwarded-For"
+```
+
+At runtime, the following expression can be used to set additional fields in the context object or to override or unset existing fields:
+
+```
+# set agent to undefined
+unset("request.context.agent")
```
StarpTech: Expr is side effect free, this won't be possible.

rfc/expr-context.md
```diff
@@ -0,0 +1,134 @@
...
+
+The operation object provides information about the operation being executed. This information is parsed and validated by the router.
+
+```
+request.operation.name // e.g. MyQuery
+request.operation.type // query, mutation, subscription
+request.operation.hash // hash of the operation
+```
+
+# fetch.*
```
StarpTech, line 81: I'm not a fan of the fetch terminology. It's not used anywhere on the server world. It always reminds me to the browser.

rfc/expr-context.md
```diff
@@ -0,0 +1,134 @@
...
+fetch.request.method // GET, POST, PUT, DELETE 
+fetch.request.body.query // GraphQL query
+fetch.request.body.operation_name // GraphQL operation name
+fetch.request.body.variables // GraphQL variables
+fetch.request.body.extensions // GraphQL extensions
+fetch.request.uri.host // subgraph host
+fetch.request.uri.path // e.g. /graphql
+fetch.request.uri.port // e.g. 443
+
+fetch.subgraph.name // name of the subgraph being fetched from
```
StarpTech, line 97: How do you reference to the subgraph request?

rfc/expr-context.md
```diff
@@ -0,0 +1,134 @@
...
+request.body.variables
+request.body.extensions
+request.uri.host
+request.uri.path
+request.uri.port
+
+request.client.name
+request.client.version
+request.client.ip
+request.routerConfigVersion
```
StarpTech, line 24: This information is not request specific.

rfc/expr-context.md
```diff
@@ -0,0 +1,134 @@
...
+```
+
+# request.context
+
+The context object provides information about the context of the request. A generic object that can be used to store any information that is relevant to the request.
+
+```
```
StarpTech, line 31: I'd interpret context as an intermediate datastructure to make handling easier. Agent is a good use case if is more than the raw header value.

StarpTech (changes requested):

rfc/expr-context.md
jensneuse: Everything from Router to Subgraphs in the Federation space is called fetch

rfc/expr-context.md
jensneuse: You can use a function:
https://expr-lang.org/docs/functions

StarpTech: It is a principle. We should avoid that. Expr should only be used to get / compute a value.

rfc/expr-context.md
StarpTech: I understand but because federation is protocol agnostic. Here we clearly reference to http primitives like headers.

rfc/expr-context.md
```diff
@@ -0,0 +1,142 @@
...
+## request
+
+The client request object is a read-only object that provides information about the incoming request. It is available at all times during the request's lifecycle.
+
+```
+request.id
+request.header
+request.method
+request.body.query
+request.body.operation_name
```
StarpTech, line 14: We should stick with camelCase as we do everywhere else.

rfc/expr-context.md
```diff
@@ -0,0 +1,142 @@
...
+
+Cosmo Router provides a context object to an expr-lang VM that can be used to access information about the request, response, and other relevant information. The context object is available at all times during the request's lifecycle.
+
+## request
+
+The client request object is a read-only object that provides information about the incoming request. It is available at all times during the request's lifecycle.
+
+```
+request.id
+request.header
```
StarpTech, line 11: I'm good with it to be lowercase for all fields "camelCase"

rfc/expr-context.md
StarpTech: For simplicity, I suggest to be `CamelCase` (first letter uppercase) because it works that way with Go exported methods. Otherwise, we would need to construct an any map.

rfc/expr-context.md
```diff
@@ -0,0 +1,142 @@
...
+request.method
+request.body.query
+request.body.operation_name
+request.body.variables
+request.body.extensions
+request.uri.host
```
StarpTech, line 17: Let's go with `url`

PR #1486, opened 2025-01-04
fix(expressions): avoid exposing more methods as desired

StarpTech:
## Motivation and Context

https://github.com/expr-lang/expr/issues/744

## Checklist

### DIFF
--- a/router/internal/expr/expr.go
+++ b/router/internal/expr/expr.go
@@ -17,7 +17,8 @@ import (
 * Naming conventions:
 * - Fields are named using camelCase
 * - Methods are named using PascalCase (Required to be exported)
-* - Methods should be exported through a custom type to avoid accidental methods that can mutate the context
+* - Methods should be exported through a custom type to avoid exposing accidental methods that can mutate the context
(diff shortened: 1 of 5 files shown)


router/internal/expr/expr.go
```diff
@@ -62,10 +55,22 @@ type RequestURL struct {
...
 
+type RequestHeaders struct {
+	Header http.Header `expr:"-"` // Do not expose the full header
+}
+
+// Get returns the value of the header with the given key. If the header is not present, an empty string is returned.
+// The key is case-insensitive and transformed to the canonical format.
+func (r RequestHeaders) Get(key string) string {
+	return r.Header.Get(key)
+}
```
Noroth, line 66: I genuinely think that having an interface here would be the cleanest and most Go style solution. I understand that we have to wait for the expr team to implement the fix but let's flag this with a TODO and the link to the Github issue so that we remember to update the code in the future.

What do you think?

StarpTech: Sounds good to me.

Noroth (approved):
LGTM

PR #1488, opened 2025-01-05
feat(jwk): upgrade JWK library, ensure tokens are validated, retry on network issues

StarpTech:
## Motivation and Context

This PR comes with a bunch of important improvements regarding JWK handling:

- Check `.Valid` after parsing tokens
- Upgrade to the latest JWK library to enforce HTTP retry on network hiccups or JWK server outage. This allows us to get rid of the behaviour that we allow the server to be operational even when the JWK hasn't been initialized.
- Use `jwkset` library functions to create a JWK server implementation in tests. Our own implementation had an error that wasn't catched by the previous JWK library. It came to the surface after upgrade.

I tested the implementation for RSA and ed25519 but went with RSA because it was the previous way.

## Checklist


router/internal/httpclient/retryableclient.go
```diff
@@ -7,12 +7,16 @@ import (
 	"time"
 )
 
-func NewRetryableHTTPClient(logger *zap.Logger) *http.Client {
+func NewRetryableHTTPClient(logger *zap.Logger, maxRetry int) *http.Client {
```
df-wg, line 10: nit: If we use 5 in all cases, why are we making this configurable? Or can we have 5 as a dedicated const here for our default value?

df-wg (approved):
overall, LGTM!

router/internal/httpclient/retryableclient.go
StarpTech: I wanted to make it a bit more configurable. Might not be the best solution.

StarpTech: Let me revert it.

router-tests/utils.go
```diff
@@ -35,7 +36,7 @@ func configureAuth(t *testing.T) ([]authentication.Authenticator, *jwks.Server)
 	authServer, err := jwks.NewServer(t)
 	require.NoError(t, err)
 	t.Cleanup(authServer.Close)
-	tokenDecoder, _ := authentication.NewJwksTokenDecoder(zap.NewNop(), authServer.JWKSURL(), time.Second*5)
+	tokenDecoder, _ := authentication.NewJwksTokenDecoder(context.Background(), zap.NewNop(), authServer.JWKSURL(), time.Second*5)
```
jensneuse, line 39: Would it be better to create a helper func that takes *testing.T and returns a context that gets cancelled when t.Cleanup is called? I'm not sure about the token decoder implementation, so it might or might not have any meaningful effect.

jensneuse: https://github.com/wundergraph/cosmo/pull/1488/files#diff-f705492dfd8e245ece0ac151b18e19be3818b88ebf3857da8b9fcd07f0ca3890R54
I think it could be beneficial

jensneuse (changes requested):

router-tests/utils.go
StarpTech: Good point, I'm gonna move it into the test func and use the global environment context.

StarpTech: Done.

jensneuse (approved):

PR #1499, opened 2025-01-08
feat(router): enable using redis clusters for rate limiting and apq

df-wg:
## Motivation and Context

Previously, users were only able to provide a single redis instance to Cosmo for both APQ/Rate Limiting. This didn't take advantage of Redis' built-in cluster mode, which takes care of horizontal scaling for the users. 

This PR enables that. In order to do it, users can provide a list of their cluster URLs instead of the singular URL they used to provide, in the now renamed `urls` field. In order to opt in to cluster mode, users have to set `cluster_enabled: true` in their configuration for both APQ and Rate Limiting

> [!WARNING]  
> As part of the preparations for Cosmo V1, targeted for release in Q1 2025, this pull request introduces essential changes to enhance long-term stability and maintainability. While we strive to minimize breaking changes, they are sometimes necessary to lay the foundation for a more robust and scalable system.

## Before: 
```yaml
rate_limit:
  enabled: true
  strategy: "simple"
  storage:
    url: "testuser:testpass@localhost:8000"
    key_prefix: "cosmo_rate_limit"  

storage_providers:
  redis:
    - id: "my_redis"
      url: "test:testpass@localhost:8000"
```

## After
```yaml
rate_limit:
  enabled: true
  strategy: "simple"
  storage:
    cluster_enabled: true
    urls:
      - "testuser:testpass@localhost:8000"
      - "test2:testpass@localhost:8001"
    key_prefix: "cosmo_rate_limit"  

storage_providers:
  redis:
    - id: "my_redis"
      cluster_enabled: true
      urls:
        - "test:testpass@localhost:8000"
        - "test2:testpass@localhost:8001"
```

## Migration Path:
[ ] Rename `storage_providers.redis.url` to `storage_providers.redis.urls`, and `rate_limit.storage.url` to `rate_limit.storage.urls`, as the first value of a list

## Checklist


docker-compose.yml
```diff
@@ -242,6 +242,93 @@ services:
     profiles:
       - dev
 
+  redis-cluster-node-0:
```
df-wg, line 245: Cluster mode requires a minimum of 6 nodes in the cluster - 3 masters and 3 slaves

StarpTech: Never heard about that it is a requirement to make it work? As far I remember 3/3 is a production requirement for failover but due testing sir can go lower.

router/internal/persistedoperation/operationstorage/redis/rdcloser.go
```diff
@@ -0,0 +1,72 @@
...
+	Close() error
+}
+
+type RedisCloserOptions struct {
+	Logger   *zap.Logger
+	URL      string
+	Password string
+}
+
+func NewRedisCloser(opts *RedisCloserOptions) (RDCloser, error) {
```
StarpTech, line 22: Could you add a brief description about what a Redis Closer is?

router-tests/ratelimit_test.go
```diff
@@ -669,6 +669,143 @@ func TestRateLimit(t *testing.T) {
...
+						Enabled:  true,
+						Strategy: "simple",
+						SimpleStrategy: config.RateLimitSimpleStrategy{
+							Rate:                    1,
+							Burst:                   1,
+							Period:                  time.Second * 2,
+							RejectExceedingRequests: false,
+						},
+						Storage: config.RedisConfiguration{
+							Url:       clusterUrls,
```
StarpTech, line 700: According to the router configuration and json schema, only single URL is supported. Could you please provide a full example and update schema description?

router/internal/persistedoperation/operationstorage/redis/rdcloser.go
```diff
@@ -0,0 +1,72 @@
...
+	var rdb RDCloser
+	if !strings.Contains(opts.URL, ",") {
+		options, err := redis.ParseURL(opts.URL)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse the redis connection url %s: %w", opts.URL, err)
```
StarpTech, line 27: The URL can contains authentication information and should not be logged.

router/internal/persistedoperation/operationstorage/redis/rdcloser.go
```diff
@@ -0,0 +1,72 @@
...
+			options.Password = opts.Password
+		}
+		rdb = redis.NewClient(options)
+	}
+
+	if !isFunctioningClient(rdb) || isClusterClient(rdb) {
+		opts.Logger.Info("Detected that redis is running in cluster mode.")
+		stripped := strings.ReplaceAll(opts.URL, "redis://", "")
+		rdb = redis.NewClusterClient(&redis.ClusterOptions{
+			Addrs:    strings.Split(stripped, ","),
```
StarpTech, line 40: This can be done at the config level. We can split the value by comma through a struct tag.

router/internal/persistedoperation/operationstorage/redis/rdcloser.go
```diff
@@ -0,0 +1,72 @@
...
+			Password: opts.Password,
+		})
+		if !isFunctioningClient(rdb) {
```
StarpTech, line 43: If a single url is not supported how does a user use single master node as before?

StarpTech (changes requested):

router/internal/persistedoperation/operationstorage/redis/rdcloser.go
StarpTech, line 43: We should make it also part of the readiness check of the router but only if it result in downtime. If we can swallow errors it is fine.

StarpTech (changes requested):

docker-compose.yml
df-wg: Good point. On an closer read of the the docs they say that there's a minimum of 3 master nodes, with a recommendation of at least 1 replica, but I dropped it down to 3

router/internal/persistedoperation/operationstorage/redis/rdcloser.go
df-wg: I don't think it needs to be part of the readiness check, from what I can see we swallow redis errors as they're just for caching

router/internal/persistedoperation/apq/redis.go
```diff
@@ -28,20 +28,19 @@ func NewRedisClient(opts *RedisOptions) (KVClient, error) {
 		return nil, errors.New("storage config is nil")
 	}
 
-	options, err := redis.ParseURL(opts.StorageConfig.URL)
-	if err != nil {
-		return nil, fmt.Errorf("failed to parse the redis connection url: %w", err)
-	}
-
-	innerClient := redis.NewClient(options)
+	rdb, err := rd.NewRedisCloser(&rd.RedisCloserOptions{
```
chensunny, line 31: here is no password transmit to RedisCloserOptions

router/pkg/config/config.go
```diff
@@ -438,8 +438,9 @@ type RateLimitErrorExtensionCode struct {
 }
 
 type RedisConfiguration struct {
-	Url       string `yaml:"url,omitempty" envDefault:"redis://localhost:6379" env:"RATE_LIMIT_REDIS_URL"`
-	KeyPrefix string `yaml:"key_prefix,omitempty" envDefault:"cosmo_rate_limit" env:"RATE_LIMIT_REDIS_KEY_PREFIX"`
+	URL         string   `yaml:"url,omitempty" envDefault:"redis://localhost:6379" env:"RATE_LIMIT_REDIS_URL"`
```
chensunny, line 441: 1. Here is No Password  definition ；
2. Is not easy to external  Redis sentinel  client with the config of master name and the password for
sentinel and redis instance

chensunny: 3. How to choice the DB

router/internal/persistedoperation/apq/redis.go
df-wg: Good catch! In this case, we haven't settled on an ideal method to allow consumers to inject passwords, and from what we've heard, they're satisfied with using a URL connection string which has a password in it. If you have a particular use case or reason to have a password in the configuration, please let me know. Otherwise, I'll resolve this.

router/pkg/config/config.go
df-wg: I didn't fully understand comments #2 and #3 here, would you mind clarifying more? I don't have major experience with Redis sentinel mode, so I appreciate your insights here

Regardless, I think that we may choose to close the loop on this PR and merge it in first, before adding password/DB choice configurations

router/internal/persistedoperation/operationstorage/redis/rdcloser.go
```diff
@@ -0,0 +1,84 @@
...
+	} else {
+		options, err := redis.ParseURL(opts.URLs[0])
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse the redis connection url: %w", err)
+		}
+		options.Password = opts.Password
+		rdb = redis.NewClient(options)
+
+		if isClusterClient(rdb) {
+			opts.Logger.Info("Detected that redis is running in cluster mode. You may encounter issues as a result")
```
StarpTech, line 52: Should this be of log level warn?

StarpTech (approved):
LGTM

router/internal/persistedoperation/operationstorage/redis/rdcloser.go
df-wg: good point, addressed

PR #1504, opened 2025-01-09
chore: add RFC for @openfed__description

Aenimus:
## Checklist

### DIFF
--- a/rfc/openfed__setDescription.md
+++ b/rfc/openfed__setDescription.md
@@ -1,0 +1,331 @@
+---
+title: "@openfed__setDescription: Propagate specific descriptions to the federated graph"
+author: David Stutt
+---
+
+This RFC proposes a new directive, `@openfed__setDescription`, which aims to deliver the following:
+1. Propagate a specific description to the federated graph
+2. Propagate a description to the federated graph that is different to the description on the subgraph 
+(internal/private vs. external/public).
+3. Choose not to propagate any descriptions to the federated graph at all.
+
+# Motivations
+
+GraphQL type definitions can (and often do, e.g., entity Objects) exist in multiple subgraphs at any one time—a "shared
+definition".
+Each of these shared definitions allows a description to be defined.
+However, the federated graph can only define a single description.
+In the event that multiple shared definitions define a description, the rules that WunderGraph Cosmo follows are quite 
+simple:
+1. Store the first description encountered.
+2. If a longer description (by character length) is encountered, update the record to this longer description.
+3. If a another description of the same length is encountered, keep the current description record (which will be the 
+same length).
+4. Propagate the description to the federated graph.
+
+Customers have requested the ability to have more granular control over which description should be propagated to the
+federated graph, if any.
+
+# Proposal
+
+A new directive, `@openfed__setDescription`, aims to facilitate all desired granularity for how and when descriptions
+are propagated to the federated graph.
+
+## Definition
+
+The potential definition of the directive:
+```graphql
+  directive @openfed__setDescription(public: Boolean! = true, content: String) on 
+    ARGUMENT_DEFINITION | ENUM | ENUM_VALUE | FIELD_DEFINITION | INPUT_FIELD_DEFINITION 
+    | INPUT_OBJECT | INTERFACE | OBJECT | SCALAR | SCHEMA | UNION
+```
+
+The directive can be defined on any type system directive location (all of which allow a description to be defined).
+
+## Arguments
+
+The directive definition proposes two arguments.
+
+### public
+- Required but defaults to `true`.
+- Declares whether the description is to be propagated in the federated graph.
+- On any given instance of a shared definition, only one `@openfed__setDescription` instance may set `public` to
+`true` or there will be a composition error.
+- If `public` is set to `false`, the description will _never_ be propagated to the federated graph.
+This means the description instance will effectively be completely ignored by federation, 
+_e.g._, even when following "normal" WunderGraph Cosmo description rules.
+
+### content
+- Optional.
+- Does nothing if the `public` argument is set to `false`.
+- Allows the definition of a new description, _e.g._, one that is different to the actual description in the subgraph
+(which may be intended to be private/internal), or for a definition (such as an extension) that cannot define 
+descriptions.
+
+# Considerations/discussion points
+
+The following is a list of considerations and/or potential discussion points:
+
+## Single source of precedence/single public description
+The federated graph can only define a single description.
+This description is "normally" the longest description (the first of that length in a tie).
+If the `@openfed__setDescription` defines the argument `public` as `true` (the default value), that description will be 
+propagated to the federated graph.
+If the `@openfed__setDescription` defines the argument `content`, the value of the content argument will "override"
+what the "original" description.
+If more than one shared definition instance defines the `@openfed__setDescription` `public` argument as `true`, a
+composition error will be returned.
+Only one shared definition instance may declare a public description.
+
+## Parent level vs. child level
+If `@openfed__setDescription` is defined on a parent level, _e.g._, an `Object type`, should all its children (_i.e._,
+fields) "inherit" the public declaration?
+For example:
+```graphql
+# subgraph A
+
+  """
+  A.Query description.
+  """
+  type Query @openfed__setDescription {
+    """
+    A.Query.dummy description.
+    """
+    dummy: String! @shareable
+  }
+```
+
+```graphql
+# subgraph B
+
+  """
+  B.Query description, which is longer than A.
+  """
+  type Query {
+    """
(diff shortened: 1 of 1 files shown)


rfc/openfed__setDescription.md
```diff
@@ -0,0 +1,307 @@
+---
+title: "@openfed__setDescription: Propagate specific descriptions to the federated graph"
```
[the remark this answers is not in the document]
devsergiy: could it be just `openfed__description`?

rfc/openfed__setDescription.md
```diff
@@ -0,0 +1,307 @@
...
+### public
+- Required but defaults to `true`.
+- Declares whether the description is to be propagated in the federated graph.
+- On any given instance of a shared definition, only one `@openfed__setDescription` instance may set `public` to
+`true` or there will be a composition error.
+- If `public` is set to `false`, the description will _never_ be propagated to the federated graph.
+This means the description instance will effectively be completely ignored by federation, 
+_e.g._, even when following "normal" WunderGraph Cosmo description rules.
+
+### content
```
devsergiy, line 58: should it be named `overrideContent` or just `override`?

it will make more verbose what it stands for, without looking into directive definition

Aenimus: Great suggestion!

rfc/openfed__setDescription.md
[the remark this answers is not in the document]
Aenimus: I like it!

rfc/openfed__setDescription.md
```diff
@@ -0,0 +1,307 @@
...
+    A.Query.dummy description.
+    """
+    dummy: String! @shareable
+  }
+```
+
+It's also possible a new argument, _e.g._, `inheritance: Boolean! = false`, could dictate whether the behaviour 
+defined on the parent is "inherited" by children, _e.g._, all `public: true` or all `public: false`.
+The child should also be able to override this behaviour by explicitly defining its own directive.
+I'm mostly of the opinion that descriptions should be explicit and granular, and inherited/coerced behaviour might be
```
devsergiy, line 146: I believe explicit behavior is more predictable

The only question is about the situation when we set a description for a type from subgraph `A` but fields are still using descriptions from subgraph `C`

Aenimus: It may also be cumbersome if, for example, an Object defines 50 Fields, but you just want that subgraph to set _all_ public descriptions , _i.e._, for the parent and its children.

rfc/openfed__setDescription.md
```diff
@@ -0,0 +1,308 @@
...
+If `@openfed__description` is defined on a parent level, _e.g._, an `Object type`, should all its children (_i.e._,
+fields) "inherit" the public declaration?
+For example:
+```graphql
+# subgraph A
+
+  """
+  A.Query description.
+  """
+  type Query @openfed__description {
```
datascientist2, line 91: I prefer the explicit nature of this option, but when we discussed a similar approach internally many in the team felt that it would lead to a lot of "noise" in the schema.

rfc/openfed__setDescription.md
```diff
@@ -0,0 +1,308 @@
...
+descriptions.
+
+# Considerations/discussion points
+
+The following is a list of considerations and/or potential discussion points:
+
+## Single source of precedence/single public description
+The federated graph can only define a single description.
+This description is "normally" the longest description (the first of that length in a tie).
+If the `@openfed__description` defines the argument `public` as `true` (the default value), that description will be
```
datascientist2, line 73: FWIW, I had previously considered something along the lines of
```
@openfed__supergraphDescription
```
The idea being "this is the description that does into the Supergraph".  Folks on the team felt it was redundant, but I always come down on the side of verbosity in APIs.

Aenimus: I think this might not be accurate in this case because the directive can be used to keep a description private/internal. In such a case, the federated graph is not involved.

rfc/openfed__setDescription.md
```diff
@@ -0,0 +1,308 @@
...
+
+# Motivations
+
+GraphQL type definitions can (and often do, e.g., entity Objects) exist in multiple subgraphs at any one time—a "shared
+definition".
+Each of these shared definitions allows a description to be defined.
+However, the federated graph can only define a single description.
+In the event that multiple shared definitions define a description, the rules that WunderGraph Cosmo follows are quite 
+simple:
+1. Store the first description encountered.
```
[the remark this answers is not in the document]
Aenimus: Currently, it is the order that is provided by the user.

rfc/openfed__setDescription.md
Aenimus: In the other comment thread I said:
> It may also be cumbersome if, for example, an Object defines 50 Fields, but you just want that subgraph to set _all_ public descriptions , _i.e._, for the parent and its children.

And I can certainly see this becoming unwieldy if you just wanted to paint everything with the same brush.

datascientist2: I think in our case, we're only going to see a smattering of usages.

rfc/openfed__setDescription.md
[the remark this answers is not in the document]
datascientist2: That is stable enough, however you guys will probably need to commit to not changing that in the future without explicit opt-in by clients.

rfc/openfed__setDescription.md
datascientist2: True.  We didn't have this private/internal functionality.

I would love to have something about this directive that provides just a bit more intuitive context.  `description` is such an overloaded term that it will be meaningless to a lot of our newbies and they'll need to pause whenever they come across it.

I wonder if we could change the arguments a little:
```
directive @openfed__description(useInSupergraph: Boolean! = false, override: String)
```
Which would force usages like this
```
type MyType @openfed__description(useInSupergraph=true)
```

I realize this is mainly targeted at newbies reading the raw schema, which may or may not be a significant problem.

rfc/openfed__setDescription.md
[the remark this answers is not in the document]
Aenimus: It's possible we should do something more deterministic like sort the subgraph by name alphabetically.

rfc/openfed__setDescription.md
Aenimus: I'm open to any directive names you feel are better for the directive. Originally, the directive was named `@openfed__setDescription`. I suggested also `@openfed__configureDescription`. Upon Sergiy's feedback, I opted for `@openfed__description` because we're doing a bit more than just setting.

The reason it is `public: Boolean! = true` is because I envisioned the main use case being control which description is propagated to the federated graph. The secondary  use case is where one would need to be explicit, _i.e._, to keep a description internal by setting `public` to `false`.

rfc/openfed__setDescription.md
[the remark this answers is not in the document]
df-wg: Sorting alphabetically definitely makes sense. But what happens if a new subgraph with an alphabetically earlier name is added to the supergraph, with a description of the same length?

rfc/openfed__setDescription.md
```diff
@@ -0,0 +1,308 @@
...
+`true` or there will be a composition error.
+- If `public` is set to `false`, that description instance will _never_ be propagated to the federated graph.
+This means the description instance will effectively be completely ignored by federation, 
+_e.g._, even when following "normal" WunderGraph Cosmo description rules.
+Description instances defined in other subgraphs can still be propagated.
+
+### override
+- Optional.
+- Does nothing if the `public` argument is set to `false`.
+- Allows the definition of a new description, _e.g._, one that is different to the actual description in the subgraph
```
jensneuse, line 62: This doesn't seem clear to me. What use case does this solve? Is there the issue that one subgraph can easily steal the ownership of a description from another one? Why is this needed?

rfc/openfed__setDescription.md
```diff
@@ -0,0 +1,308 @@
...
+propagated to the federated graph.
+If the `@openfed__description` defines the argument `override`, the value of the argument will "override" the "original" 
+description.
+If more than one shared definition instance defines the `@openfed__description` `public` argument as `true`, a
+composition error will be returned.
+Only one shared definition instance may declare a public description.
+
+## Parent level vs. child level
+If `@openfed__description` is defined on a parent level, _e.g._, an `Object type`, should all its children (_i.e._,
+fields) "inherit" the public declaration?
```
jensneuse, line 83: I like this approach

Aenimus:
The directive has been implemented:
https://cosmo-docs.wundergraph.com/federation/directives/openfed__configuredescription

A second directive will be implemented at a later date to provide "cascade functionality" for descriptions from parent to child.

PR #1507, opened 2025-01-10
feat: add wgc request ID generation & propagation

turnabout:
## Motivation and Context

The `wgc` CLI currently does not generate a unique identifier for each request made to the control plane. The goal is to generate one when executing a command, propagate it to the control plane (on all requests sent to it), and attach it to control plane logs for easier debugging.

In the case of a control plane error during the execution, the request ID is printed to the user.

## Checklist

---

* `wgc`
  * Generated the UUID using `node:crypto` and attaching to the global config once. This results in the same id being reused on all requests sent to the control panel by a command
  * Added support for the printing of the request ID on errors with a Client interceptor (`requestIdInterceptor`), so the solution is reused in all subcommands
* Control panel
  * Modified `enrichLogger` to extract the request ID header (if it exists) and include it as additional log fields


controlplane/src/core/util.ts
```diff
@@ -79,9 +79,12 @@ export const enrichLogger = (
   logger: FastifyBaseLogger,
   authContext: Partial<AuthContext & GraphKeyAuthContext>,
 ) => {
+  const requestId = ctx.requestHeader.get('x-request-id') as string ?? undefined;
```
StarpTech, line 82: Would it make sense to populate a new ID here when we don't have one?

cli/src/core/config.ts
```diff
@@ -38,12 +39,14 @@ export const config = {
   checkAuthor: process.env.COSMO_VCS_AUTHOR || '',
   checkCommitSha: process.env.COSMO_VCS_COMMIT || '',
   checkBranch: process.env.COSMO_VCS_BRANCH || '',
+  requestId: randomUUID(),
```
StarpTech, line 42: Why would we want to have a single static request ID?

cli/src/core/client/client.ts
```diff
@@ -15,6 +19,22 @@ export interface Client {
...
+/**
+ * Interceptor printing the request ID to help with debugging, when a request results in an error.
+ */
+export const requestIdInterceptor: Interceptor = (next) => async (req) => {
+  const res = await next(req);
+
+  if ((res.message?.response as Response)?.code === EnumStatusCode.ERR) {
+    console.log(pc.yellow('---'));
+    console.log(pc.yellow('Something went wrong while processing a request for this command.'));
+    console.log(pc.yellow(`Request ID: ${config.requestId}`));
```
StarpTech, line 31: This should print the request ID not a static one. I also think we should return the requestID as a response header from the controlplane and print it here. That actually ensures that we work with the right ID. This behaviour would also ensure that clients that don't generate an ID get one.

cli/src/core/client/client.ts
```diff
@@ -15,6 +19,22 @@ export interface Client {
...
 
+/**
+ * Interceptor printing the request ID to help with debugging, when a request results in an error.
+ */
+export const requestIdInterceptor: Interceptor = (next) => async (req) => {
+  const res = await next(req);
+
+  if ((res.message?.response as Response)?.code === EnumStatusCode.ERR) {
+    console.log(pc.yellow('---'));
+    console.log(pc.yellow('Something went wrong while processing a request for this command.'));
```
StarpTech, line 30: I like the interceptor approach because we have it centrally. Can we add more context here? RPC method? Time of error? etc.. everything that might be helpful when customers sharing it.

cli/test/get-base-headers.test.ts
```diff
@@ -0,0 +1,15 @@
+import { describe, test, expect } from 'vitest';
```
StarpTech, line 1: Please add an integration test that validates that an RPC call gets the request header.

StarpTech (changes requested):

cli/test/get-base-headers.test.ts
turnabout: I added a test for this (`request-header.test.ts`). Since the interceptor adding the header does its work from inside the transport created in the `CreateClient` function, I needed a way to intercept requests before they actually go out. So I added an `interceptors` option to the `ClientOptions` passed to `CreateClient`, and using it to test the presence of a valid `x-request-id`

cli/src/core/client/client.ts
turnabout: Added more contextual data

controlplane/src/core/util.ts
turnabout: I had that thought as well. It could only be useful, so I replaced the `undefined` by a new UUID

cli/src/core/config.ts
turnabout: Like mentioned by email, this stemmed from a misunderstanding.

cli/src/core/client/client.ts
turnabout: I opted to generate it inside the same interceptor. So I'm now hooking before the call (gen the ID) and after (log ID in case of error). This should accomplish the same and felt simpler to manage

cli/test/request-id-interceptor.test.ts
```diff
@@ -0,0 +1,89 @@
...
+
+vi.mock('node:crypto', () => ({
+  randomUUID: () => mockedId,
+}));
+
+const ERR_RESPONSE = { message: { response: { code: EnumStatusCode.ERR, details: 'An error occurred' } }, method: { name: 'WhoAmI' } };
+const SUCCESS_RESPONSE = { message: { response: { code: EnumStatusCode.OK, details: 'No error here' } } };
+
+
+describe('requestIdInterceptor()', () => {
```
StarpTech, line 17: Instead of mocking it you could instantiate a buf client in the tests and provide a dummy service implementation that validates end to end if the request ID arrives. In the same way, you could also test that in case of an error the informations are logged.

cli/test/request-id-interceptor.test.ts
```diff
@@ -0,0 +1,89 @@
...
+
+let mockedId = 'mocked-id';
+
+const consoleLogSpy = vi.spyOn(console, 'log').mockImplementation(() => {});
+
+vi.mock('node:crypto', () => ({
+  randomUUID: () => mockedId,
+}));
+
+const ERR_RESPONSE = { message: { response: { code: EnumStatusCode.ERR, details: 'An error occurred' } }, method: { name: 'WhoAmI' } };
```
StarpTech, line 13: Instead of creating this here statically, we can just test a client end to end. We have examples for it.

StarpTech (changes requested):

cli/test/request-id-interceptor.test.ts
turnabout: Isn't that what I did with `request-header.test.ts`? This test is testing the interceptor in isolation, and the other one is creating a client instance, injecting it into a command, executing it, then checking the outgoing requests have the header.

StarpTech: You're right. I didn't get the latest updates.

cli/test/request-header.test.ts
```diff
@@ -0,0 +1,38 @@
+import { describe, test, expect } from 'vitest';
+import { Command } from 'commander';
+import { Response, WhoAmIResponse } from '@wundergraph/cosmo-connect/dist/platform/v1/platform_pb';
+import { CreateClient } from '../src/core/client/client.js';
+import WhoAmI from '../src/commands/auth/commands/whoami.js';
+import { config } from '../src/core/config.js';
+
+
+const interceptor = (req) => {
+  expect(req.header.get('x-request-id'), "must be a valid UUID").toMatch(/^[\da-f]{8}-[\da-f]{4}-[0-5][\da-f]{3}-[089ab][\da-f]{3}-[\da-f]{12}$/i);
```
StarpTech, line 10: This is done multiple times. Can we make a util function for reuse?

StarpTech (changes requested):

cli/test/request-header.test.ts
turnabout: Added test util function `expectUuid()`

controlplane/src/core/util.ts
```diff
@@ -79,9 +79,12 @@ export const enrichLogger = (
   logger: FastifyBaseLogger,
   authContext: Partial<AuthContext & GraphKeyAuthContext>,
 ) => {
+  const requestId = (ctx.requestHeader.get('x-request-id') as string) ?? randomUUID();
```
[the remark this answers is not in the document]
turnabout: Removed

StarpTech (changes requested):

cli/test/request-id-interceptor.test.ts
```diff
@@ -0,0 +1,89 @@
...
+  });
+
+  test('Should log the request ID when the response errored', async () => {
+    const req = {
+      header: {
+        set: (key, val) => {}
+      }
+    };
+
+    await requestIdInterceptor(() => ERR_RESPONSE)(req);
```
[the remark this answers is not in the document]
turnabout: I understand your point. The way I see it, this test is testing the interceptor in isolation (as much as possible) while the other is testing the complete feature. I usually like having both of these kinds of tests together, to give me the peace of mind an implementation works as expected by itself, but also in the context of a complete feature.

But I guess there is significant overlap here. If you think they're not needed or not aligned with the rest of the project, I can certainly remove them.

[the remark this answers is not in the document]
StarpTech: Thanks for clarifying. I do a see a benefit of combining them. Please do 😊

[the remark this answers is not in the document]
turnabout: I removed the test file `request-id-interceptor.test.ts` and ported the two test cases that didn't overlap between them to `request-header.test.ts` (I assumed that's what you meant by "combining them"). The test cases are:

* 'Should log the request ID when the response errored'
* 'Should NOT log the request ID when the response was successful'

cli/src/core/client/client.ts
```diff
@@ -1,20 +1,51 @@
...
+export const requestIdInterceptor: Interceptor = (next) => async (req) => {
+  const requestId = randomUUID();
+
+  req.header.set('x-request-id', requestId);
+
+  const timeStamp = new Date().toISOString();
+  const res = await next(req);
+
+  if ((res.message?.response as Response)?.code === EnumStatusCode.ERR) {
+    console.error(res);
```
turnabout, line 36: To remove - was probably left by accident while debugging locally

PR #1509, opened 2025-01-10
fix(router): remove wildcard from router graphql path

Noroth:
## Motivation and Context

`Mount()` appends a wildcard `*` to the graphql endpoint. We want to make sure we only hit the endpoint on the configured path. Wildcard url paths should not be taken into account and any other path than the configured paths should return 404.

## Checklist

### DIFF
--- a/router-tests/graphql_over_get_test.go
+++ b/router-tests/graphql_over_get_test.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/stretchr/testify/require"
 	"github.com/wundergraph/cosmo/router-tests/testenv"
+	"github.com/wundergraph/cosmo/router/core"
 )
 
 func TestOperationsOverGET(t *testing.T) {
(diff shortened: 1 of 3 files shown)


devsergiy (approved):

router-tests/graphql_over_get_test.go
```diff
@@ -54,6 +55,97 @@ func TestOperationsOverGET(t *testing.T) {
...
+				OperationName: []byte(`Employees`),
+				Query:         `query Employees { employees { id } }`,
+			})
+
+			require.NoError(t, err)
+			require.Equal(t, http.StatusNotFound, res.Response.StatusCode)
+		})
+	})
+
+	t.Run("Should not create wildcard for root path", func(t *testing.T) {
```
StarpTech, line 110: Please also add a tests (I thought we would have that already) that the playground and the endpoint can be on the same path.

StarpTech (changes requested):

router-tests/graphql_over_get_test.go
Noroth: Will do

StarpTech: ty

StarpTech (approved):
LGTM

PR #1512, opened 2025-01-13
feat(cli): add support for reading schema from stdin in subgraph check command

L42y:
## Motivation and Context
I want to introspect the schema then pipe to wgc for subgraph check without saving it to actual file first, as demonstrated in Apollo's Rover: https://www.apollographql.com/docs/rover/commands/subgraphs#subgraph-check

## Checklist

### DIFF
--- a/cli/src/commands/subgraph/commands/check.ts
+++ b/cli/src/commands/subgraph/commands/check.ts
@@ -9,6 +9,15 @@ import { BaseCommandOptions } from '../../../core/types/types.js';
 import { verifyGitHubIntegration } from '../../../github.js';
 import { handleCheckResult } from '../../../handle-check-result.js';
 
+// Helper function to read from stdin
+async function readStdin(): Promise<Buffer> {
+  const chunks: Buffer[] = [];
+  for await (const chunk of process.stdin) {
+    chunks.push(Buffer.from(chunk));
+  }
+  return Buffer.concat(chunks);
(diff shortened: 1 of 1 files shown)


cli/src/commands/subgraph/commands/check.ts
```diff
@@ -50,8 +47,27 @@ export default (opts: BaseCommandOptions) => {
       });
     }
 
-    // submit an empty schema in case of a delete check
-    const schema = schemaFile ? await readFile(schemaFile) : Buffer.from('');
+    // Handle schema input
```
StarpTech, line 50: @L42y the apollo approach in accepting the schema from stdin is unusual. Why should we use the `--schema` flag when piping data is already implicit? Additionally, `--schema` is declared as a way to specify a file path.

What do you think about this approach?

```
data | wgc subgraph check my-graph
```

This would follow the behaviour of common unix tools. An example to implement this is demonstrated [here](https://github.com/tj/commander.js/issues/137#issuecomment-50937953). This can be then also reused across all other commands.

StarpTech (changes requested):

cli/src/commands/subgraph/commands/check.ts
StarpTech: Let me know what you think, I think both approaches has it's benefits.

PR #1515, opened 2025-01-19
fix(cache operation): swallow cache errors and other improvements

StarpTech:
## Motivation and Context

- Don't abort the router start or config update when the cache errors for any reasons e.g. after max retry of PO download
- Don't include client version and name to the operations manifest when unknown values like `missing`, `unknown` are used.
- PersistentId was not displayed in the UI due to a typo.
- Don't pretty print cache operations because a user wants to understand what's the exact operation content in the cache
- Catch nil-pointer exception when operation content is empty.

## Checklist

### DIFF
--- a/controlplane/src/core/repositories/CacheWarmerRepository.ts
+++ b/controlplane/src/core/repositories/CacheWarmerRepository.ts
@@ -59,8 +59,8 @@ export class CacheWarmerRepository {
         OperationHash as operationHash,
         OperationName as operationName,
         OperationPersistedID as operationPersistedID,
-        ClientName as clientName,
-        ClientVersion as clientVersion,
+        if(ClientName = 'unknown', '', ClientName) as clientName,
+        if(ClientVersion = 'missing', '', ClientVersion) as clientVersion,
         func_rank(${quantile}, BucketCounts) as rank,
         func_rank_bucket_lower_index(rank, BucketCounts) as b,
         round(func_histogram_v2(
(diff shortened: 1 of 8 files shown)


controlplane/src/core/repositories/CacheWarmerRepository.ts
```diff
@@ -49,6 +49,7 @@ export class CacheWarmerRepository {
     const parsedDateRange = isoDateRangeToTimestamps(dateRange, rangeInHours);
     const [start, end] = getDateRange(parsedDateRange);
     const quantile = 0.9;
+    const minPlanningTimeInMs = 0;
```
StarpTech, line 52: We can make this configurable in the future.

controlplane/src/core/repositories/CacheWarmerRepository.ts
```diff
@@ -58,8 +59,8 @@ export class CacheWarmerRepository {
         OperationHash as operationHash,
         OperationName as operationName,
         OperationPersistedID as operationPersistedID,
-        ClientName as clientName,
-        ClientVersion as clientVersion,
+        if(ClientName = 'unknown', '', ClientName) as clientName,
+        if(ClientVersion = 'missing', '', ClientVersion) as clientVersion,
```
df-wg, line 63: would you rather '' as opposed to a clear "missing" value?

df-wg (approved):

controlplane/src/core/repositories/CacheWarmerRepository.ts
StarpTech: Yes, but currently the router emits those values to express empty values. In the future, we have to change that.

PR #1519, opened 2025-01-20
feat(router): optimize playground delivery, add concurrency_limit to config

df-wg:
## Motivation and Context

Previously, we were recomputing this template every time and loading it into memory every time you accessed the endpoint. This should fix that, and ensure we only compute it once. 

## Checklist

### DIFF
--- a/router/internal/graphiql/playgroundhandler.go
+++ b/router/internal/graphiql/playgroundhandler.go
@@ -12,17 +12,25 @@ type PlaygroundOptions struct {
 }
 
 type Playground struct {
-	next http.Handler
-	opts *PlaygroundOptions
+	next          http.Handler
+	opts          *PlaygroundOptions
+	templateBytes []byte
 }
 
 func NewPlayground(opts *PlaygroundOptions) func(http.Handler) http.Handler {
 	return func(next http.Handler) http.Handler {
-		return &Playground{
+		p := &Playground{
 			next: next,
 			opts: opts,
 		}
+		p.initPlayground()
+		return p
 	}
+}
+
+func (p *Playground) initPlayground() {
+	tpl := strings.Replace(p.opts.Html, "{{graphqlURL}}", p.opts.GraphqlURL, -1)
+	p.templateBytes = []byte(tpl)
 }
 
 func (p *Playground) ServeHTTP(w http.ResponseWriter, r *http.Request) {
@@ -35,12 +43,9 @@ func (p *Playground) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	tpl := strings.Replace(p.opts.Html, "{{graphqlURL}}", p.opts.GraphqlURL, -1)
-	resp := []byte(tpl)
-
 	w.Header().Set("Content-Type", "text/html; charset=utf-8")
-	w.Header().Set("Content-Length", strconv.Itoa(len(resp)))
+	w.Header().Set("Content-Length", strconv.Itoa(len(p.templateBytes)))
 
 	w.WriteHeader(http.StatusOK)
-	_, _ = w.Write(resp)
+	_, _ = w.Write(p.templateBytes)
 }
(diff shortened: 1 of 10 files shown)


router/internal/graphiql/playgroundhandler.go
```diff
@@ -12,19 +12,28 @@ type PlaygroundOptions struct {
 }
 
 type Playground struct {
-	next http.Handler
-	opts *PlaygroundOptions
+	next          http.Handler
+	opts          *PlaygroundOptions
+	templateBytes *[]byte
```
df-wg, line 17: I assumed that we'd prefer to keep the pointer in memory rather than the full bytes slice, but I can change this back

jensneuse: a slice internally already is a pointer (to the underlying array), so it's more typical to just keep the slice instead of a pointer to the slice

router/internal/graphiql/playgroundhandler.go
```diff
@@ -35,12 +44,9 @@ func (p *Playground) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-	tpl := strings.Replace(p.opts.Html, "{{graphqlURL}}", p.opts.GraphqlURL, -1)
-	resp := []byte(tpl)
-
 	w.Header().Set("Content-Type", "text/html; charset=utf-8")
-	w.Header().Set("Content-Length", strconv.Itoa(len(resp)))
+	w.Header().Set("Content-Length", strconv.Itoa(len(*p.templateBytes)))
 
 	w.WriteHeader(http.StatusOK)
-	_, _ = w.Write(resp)
+	_, _ = w.Write(*p.templateBytes)
```
jensneuse, line 51: I think I'd do this instead:
```go
_, _ = bytes.NewBuffer(resp).WriteTo(w)
```

Explanation:

Internally, the net/http package has a 4kb buffer. The response size of the HTML is 3mb. As such, the internal buffer will have to expand (grow -> copy) multiple times to reach 3mb. Initializing the buffer with 3mb will eliminate the necessary grow calls.

In addition, this could become a security concern as an attacker could overwhelm the router just by sending a lot of traffic to this endpoint, as it will create a lot of buffers.

To mitigate this, we could add a simple semaphore e.g. to limit the concurrency in this handler to 10, which will limit the memory usage to 30mb.

jensneuse (changes requested):

router/pkg/config/config.schema.json
```diff
@@ -1075,10 +1075,36 @@
...
+          "description": "Enable the GraphQL Playground. The GraphQL Playground is a web-based GraphQL IDE that allows you to interact with the GraphQL API. The default value is true. If the value is false, the GraphQL Playground is disabled.",
+          "default": true
+        },
+        "path": {
+          "type": "string",
+          "format": "x-uri",
+          "default": "/",
+          "description": "The path of the GraphQL Playground. The GraphQL Playground is a web-based GraphQL IDE that allows you to interact with the GraphQL API. The default value is '/'."
+        },
+        "limit": {
```
jensneuse, line 1094: rename to "concurrency_limit" to make it explicit what we're limiting?

jensneuse (changes requested):
One minor.
Last thing, can you add a simple e2e test to the router-tests to ensure that the playground is returning HTML when configured with defaults?
I think we're currently not testing this.

df-wg:
> One minor. Last thing, can you add a simple e2e test to the router-tests to ensure that the playground is returning HTML when configured with defaults? I think we're currently not testing this.

@jensneuse isn't that present here? https://github.com/wundergraph/cosmo/blob/e21577a5a73537482f714302acf337af10260152/router/internal/graphiql/playgroundhandler_test.go#L11-L30

Noroth:
@jensneuse I recently added a test for the playground

https://github.com/wundergraph/cosmo/blob/e21577a5a73537482f714302acf337af10260152/router-tests/graphql_over_get_test.go#L151-L182

jensneuse (approved):

PR #1521, opened 2025-01-21
feat: add ApolloCompatibilitySkipEnumValueValidation

flymedllva:
## Motivation and Context

Need to support ApolloCompatibilitySkipEnumValueValidation opts in graphql go tools https://github.com/wundergraph/graphql-go-tools/pull/1031


Aenimus:
Hi @flymedllva,

I'm afraid we still don't understand the use case. Why do you wish to use an Enum and then ignore the contract? Why not use a `String` at that point?

flymedllva:
Hey

Yes, I agree that it is better to use a`String`, but we moved from Apollo and there the enums were passed through as strings and everything is written according to this concept, it is a big task to change, and specifically here we would be satisfied if we had the opportunity to leave the behavior to which we are accustomed

Aenimus:
Hi @flymedllva 

We think this flag could be more harmful than helpful, so we won't be considering it at this time.

PR #1529, opened 2025-01-23
fix(router): enable health checks during startup

Noroth:
## Motivation and Context

Health checks are currently only available after the router has been initialized. The expected intended behavior is, that the health checks are available during setup and report liveness and correct readiness information.

`ready`
* during initialization -> false
* after initialization -> true

## Checklist

### DIFF
--- a/router-tests/testenv/testenv.go
+++ b/router-tests/testenv/testenv.go
@@ -6,11 +6,9 @@ import (
 	"context"
 	"crypto/tls"
 	"crypto/x509"
-	_ "embed"
 	"encoding/json"
 	"errors"
 	"fmt"
-	"github.com/hashicorp/consul/sdk/freeport"
 	"io"
 	"log"
 	"math/rand"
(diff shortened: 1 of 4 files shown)


router-tests/testenv/testenv.go
```diff
@@ -30,42 +28,38 @@ import (
...
 	"github.com/wundergraph/cosmo/router/core"
 	nodev1 "github.com/wundergraph/cosmo/router/gen/proto/wg/cosmo/node/v1"
 	"github.com/wundergraph/cosmo/router/pkg/config"
+	"github.com/wundergraph/cosmo/router/pkg/controlplane/configpoller"
+	"github.com/wundergraph/cosmo/router/pkg/logging"
+	rmetric "github.com/wundergraph/cosmo/router/pkg/metric"
+	pubsubNats "github.com/wundergraph/cosmo/router/pkg/pubsub/nats"
+	"github.com/wundergraph/graphql-go-tools/v2/pkg/engine/datasource/pubsub_datasource"
+
+	_ "embed"
```
jensneuse, line 62: what's this?

router-tests/testenv/testenv.go
```diff
@@ -717,7 +711,7 @@ func createTestEnv(t testing.TB, cfg *Config) (*Environment, error) {
 		}
 	}
 
-	waitErr := e.WaitForServer(ctx, e.RouterURL+"/health/live", 100, 10)
+	waitErr := e.WaitForServer(ctx, e.RouterURL+"/health/ready", 100, 10)
```
jensneuse, line 714: Can the route be configured? If yes, should we read the config or hardcode this here?

jensneuse (changes requested):
Can you add a test where you mock the cache warmup loader interface?
1. start server
2. check live true ready false
3. set mock to "finish" warmup phase
4. check live true ready true

router-tests/testenv/testenv.go
Noroth: We can configure it, the router just has a fallback to a default value and I suppose that's why we never configured it in the testenv

```go
	// Default values for health check paths

	if r.healthCheckPath == "" {
		r.healthCheckPath = "/health"
	}
	if r.readinessCheckPath == "" {
		r.readinessCheckPath = "/health/ready"
	}
	if r.livenessCheckPath == "" {
		r.livenessCheckPath = "/health/live"
	}

```

router/core/http_server.go
```diff
@@ -44,13 +50,20 @@ func newServer(opts *httpServerOptions) *server {
 		MaxHeaderBytes: opts.maxHeaderBytes,
 	}
 
+	// Create default handler for liveness and readiness
+	httpRouter := chi.NewMux()
```
StarpTech, line 54: When we add them here, don't we need to remove them from `newGraphServer`?

StarpTech (changes requested):

router-tests/testenv/testenv.go
StarpTech: @jensneuse did you mix it up? This is testenv.

router/core/http_server.go
Noroth: No because we are creating a completely new mux and replace the existing one in the Swap function.

router-tests/testenv/testenv.go
jensneuse: My point was, if we ever write a test to validate the path of ready, live and health can be changed, this WaitFor will fail.

Noroth:
> Can you add a test where you mock the cache warmup loader interface?
> 
> 1. start server
> 2. check live true ready false
> 3. set mock to "finish" warmup phase
> 4. check live true ready true

This is a good question. The `StartRouter` is a blocking operation, also in Testenv we start the router blocking... So basically all logic that we implement in the callback function will be called once the router has finished setting up.
Also we need the xEnv to know the endpoint to call the router so we can't do it in parallel within the test function.

Currently I tested it manually with a delay timer and running curl requests. 

I think we can do that, but it might take a bit of time to have a safe implementation in the testenv.

StarpTech (approved):
LGTM

PR #1533, opened 2025-01-24
feat(router): enable starting the router without subgraphs

df-wg:
## Checklist


router/pkg/routerconfig/s3/client.go
```diff
@@ -89,13 +89,21 @@ func (c Client) RouterConfig(ctx context.Context, version string, modifiedSince
...
+	return io.ReadAll(reader)
+}
+
+func (c Client) RouterConfig(ctx context.Context, version string, modifiedSince time.Time) (*routerconfig.Response, error) {
+	body, err := c.getConfigFile(ctx, version, modifiedSince)
 	if err != nil {
 		var minioErr minio.ErrorResponse
 		if errors.As(err, &minioErr) && minioErr.StatusCode == http.StatusNotModified {
 			return nil, configpoller.ErrConfigNotModified
+		} else if !strings.Contains(err.Error(), "NoSuchKey") {
```
df-wg, line 101: I couldn't test it locally, nor find the actual error to match it on. This is based on the error code listed here: https://github.com/minio/minio-go/blob/8b220d8e58cc4e1bbb667b732a130c7b647d8db9/s3-error.go#L39

router/pkg/routerconfig/client.go
```diff
@@ -21,3 +21,43 @@ type ConfigNotFoundError interface {
 	error
 	FederatedGraphId() string
 }
+
+func GetDefaultConfig() []byte {
```
df-wg, line 25: I would assume this is better than using a //go:embed to store it

router-tests/testenv/testdata/configWithStatic.json
```diff
@@ -0,0 +1,35 @@
...
+    ],
+    "graphqlSchema": "schema {\n  query: Query\n}\ntype Query {\n  hello: String\n}",
+    "fieldConfigurations": [
+      {
+        "typeName": "Query",
+        "fieldName": "hello"
+      }
+    ]
+  },
+  "version": "1a7c0b1a-839c-4b6f-9d05-7cb728168f57",
```
df-wg, line 32: Should I make this version v0.0.1 or some clearer version?

router/pkg/routerconfig/cdn/client.go
```diff
@@ -228,6 +216,29 @@ func (cdn *Client) RouterConfig(ctx context.Context, version string, modifiedSin
...
+func (cdn *Client) RouterConfig(ctx context.Context, version string, modifiedSince time.Time) (*routerconfig.Response, error) {
+	body, err := cdn.getRouterConfig(ctx, version, modifiedSince)
+	if err != nil && errors.Is(err, ErrConfigNotFound) {
+		body = routerconfig.GetDefaultConfig()
+	} else if err != nil {
+		return nil, err
+	}
+
+	/*
+	* Serialize the response body to a RouterConfig object
```
jensneuse, line 231: Deserialize/Unmarshal?

router-tests/integration_test.go
```diff
@@ -49,6 +49,19 @@ func TestSimpleQuery(t *testing.T) {
...
+func TestNoSubgraphConfig(t *testing.T) {
+	t.Parallel()
+
+	testenv.Run(t, &testenv.Config{
+		RouterConfigJSONTemplate: testenv.ConfigWithStaticJSONTemplate,
+	}, func(t *testing.T, xEnv *testenv.Environment) {
+		res := xEnv.MakeGraphQLRequestOK(testenv.GraphQLRequest{
+			Query: `query { hello }`,
+		})
+		require.JSONEq(t, `{"data":{"hello":"world"}}`, res.Body)
```
df-wg, line 61: This doesn't truly replicate the test, because we are starting the router with this config, instead of validating that this config is provided automatically

router/pkg/routerconfig/s3/client.go
jensneuse, line 101: Can this be more precise instead of matching string contains? e.g. matching a specific error type or an error ID?

router/pkg/routerconfig/client.go
```diff
@@ -21,3 +21,43 @@ type ConfigNotFoundError interface {
...
+            "typeName": "Query",
+            "fieldNames": [
+              "hello"
+            ]
+          }
+        ],
+        "customStatic": {
+          "data": {
+            "kind": "STATIC_CONFIGURATION_VARIABLE",
+            "staticVariableContent": "{\"hello\": \"world\"}"
```
jensneuse, line 44: Can you update the "world" string to the following template:

```
Cosmo Router is ready!

Follow this guide to deploy your first Supergraph:
https://cosmo-docs.wundergraph.com/tutorial/from-zero-to-federation-in-5-steps-using-cosmo

Build Info:

Version {version}
Commit {commit}
Date {date}
```

the following publicly exposed vars can be used

```go
core.Version
core.Commit
core.Date
```

jensneuse (dismissed):

router/pkg/routerconfig/client.go
```diff
@@ -21,3 +21,43 @@ type ConfigNotFoundError interface {
 	error
 	FederatedGraphId() string
 }
+
+func GetDefaultConfig() []byte {
+	return []byte(`
```
jensneuse, line 26: Can you define the config in code instead of JSON bytes?
Using JSON bytes can obfuscate breaking changes in the future as we wouldn't see if the bytes are incompatbile.

router/pkg/routerconfig/s3/client.go
df-wg: I couldn't test it out to verify what the actual error type would be, but it didn't seem like minio defines errors, from what I could see

router/pkg/routerconfig/client.go
df-wg: I wasn't able to have any new lines in the return string, plus introducing `core.*` made an import cycle that would have been difficult to fix

router/pkg/routerconfig/s3/client.go
StarpTech: Why do make the call at all when we know that we don't expect anything to load?

router-tests/testenv/testdata/configWithStatic.json
StarpTech: Why is this config needed, when we can build it through code?

router/pkg/routerconfig/s3/client.go
df-wg: We can't know, before making the call, whether or not we'll have a new version of the config to load. It's only upon the response that we can see if there is a config or not, and even if we don't have a valid response, we would only specifically want to initialize the config with the default config if it's clear there's no existent config yet. Otherwise, we could end up causing the clients an interruption, by overwriting their formerly valid config with our default static config

router-tests/integration_test.go
StarpTech: We should start to write proper integration tests where we don't build a router binary that is never used in production. You can run the router as it is and pass environment variables.

router/pkg/routerconfig/s3/client.go
df-wg: Good point. I updated it

router-tests/testenv/testenv.go
```diff
@@ -315,6 +341,76 @@ type LogObservationConfig struct {
...
+	res.Config.Telemetry.Metrics.OTLP.Enabled = false
+	testCdn := SetupCDNServer(t)
+
+	ctx, cancel := context.WithCancelCause(context.Background())
+	cdnConfig := config.CDNConfiguration{
+		URL:       testCdn.URL,
+		CacheSize: 1024 * 1024,
+	}
+
+	router, err := routercmd.NewRouter(context.Background(), routercmd.Params{
```
StarpTech, line 374: Instead of doing a hybrid approach, it would be great if we could purely start a router by spawning a process and configure it in the same way as a user would do. By environment variable or router config. We could maintain different presets of .env and router config for maximum flexibility. We should avoid the programmatic api completely.

StarpTech (changes requested):

router-tests/testenv/testenv.go
df-wg: Finally worked out all the issues with it, feel free to take a look when you have time 😄

router-tests/testenv/testexec.go
```diff
@@ -0,0 +1,136 @@
...
+	env := runRouterBin(t, ctx, cfg, routerDir)
+	t.Cleanup(env.Shutdown)
+
+	// Execute the test case with the environment
+	f(t, env)
+}
+
+// BuildRouter runs `make build` inside the router directory and fails the test on error.
+func buildRouterBin(t *testing.T, ctx context.Context) {
+	t.Helper()
```
StarpTech, line 39: How do we avoid building the binary multiple times as soon as we add more tests? I recommend to build the binary once even when you call it multiple times. Additionally, we should build it with the right build flags including race.

StarpTech (changes requested):

router-tests/testenv/testexec.go
```diff
@@ -0,0 +1,159 @@
...
+	cmd := exec.Command(fullBinPath, "--override-env", envFile)
+	cmd.Dir = t.TempDir()
+	newCtx, cancel := context.WithCancelCause(context.Background())
+	runCmdWithLogs(t, ctx, cmd, false)
+
+	// Graceful shutdown on context cancel
+	go func() {
+		<-newCtx.Done()
+		_ = cmd.Process.Signal(os.Interrupt)
+		time.Sleep(2 * time.Second)
```
StarpTech, line 106: Why sleep? The router is configured with a graceful shutdown timeout. You can pass it as an ENV or use the default. In all scenarios, the router should be able to graceful shutdown.

StarpTech (changes requested):

router-tests/testenv/testexec.go
df-wg: Addressed

StarpTech (approved):
LGTM

PR #1534, opened 2025-01-24
feat: add compatibility handshake between router and execution config

Aenimus:
## Motivation and Context

1. Propagate a compatibility version in the router execution configuration through the "compatibility_version" property. This is composed thus: <router execution configuration version><composition package version>.
2. Add a handshake in the router:
- If the router execution config version is higher than the threshold defined in the router, reject the execution configuration.
- If the router execution config version is lower than the threshold defined in the router, warn the user that some functionality may not work or may be different.

## Checklist


router/core/router.go
```diff
@@ -1334,6 +1348,36 @@ func (r *Router) Shutdown(ctx context.Context) (err error) {
 	return err
 }
 
+func (r *Router) isRouterCompatibleWithExecutionConfig(compatibilityVersion string) bool {
```
StarpTech, line 1351: Please add a test for this.

StarpTech (changes requested):

StarpTech:
Add a PR description.

composition/scripts/get-composition-version.mjs
```diff
@@ -0,0 +1,20 @@
+import fs from 'fs';
+import { fileURLToPath } from 'node:url';
+import path from 'node:path';
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+const compositionVersion = '{{$COMPOSITION__VERSION}}';
+
+// From pnpm v10+, hooks will not be supported.
```
StarpTech, line 9: It is supported but we would need to whitelist modules that are allowed to run it.

StarpTech (approved):
LGTM

composition/scripts/get-composition-version.mjs
Aenimus: Updated the comment.

PR #1537, opened 2025-01-27
fix(router): handle repeated headers in response

cmtm:
When multiple headers with the same name were sent in the response from
a subgraph, only one would be included in the response to the client
when forwarding headers. Fix it so that they're all forwarded.


StarpTech:
Hi @cmtm, thanks for the PR. Please add an integration test to verify the correct behaviour. Example: https://github.com/wundergraph/cosmo/blob/main/router-tests/header_set_test.go

cmtm:
I've added a unit test. I've been unable to run any integration tests locally.
https://discord.com/channels/738739428314316823/1333891899551121438

Is the unit test sufficient?

router-tests/header_propagation_test.go
```diff
@@ -97,20 +99,20 @@ func TestHeaderPropagation(t *testing.T) {
...
 	}
 
-	setSubgraphPropagateHeader := func(header, valA, valB string) testenv.SubgraphsConfig {
+	setSubgraphPropagateHeader := func(header string, valA, valB []string) testenv.SubgraphsConfig {
 		return testenv.SubgraphsConfig{
 			Employees: testenv.SubgraphConfig{
 				Middleware: func(handler http.Handler) http.Handler {
 					return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-						w.Header().Set(header, valA)
+						w.Header()[header] = valA
```
devsergiy, line 107: Hi @cmtm 

Manually modifying the underlying map is not fully equal to [header.Set](https://pkg.go.dev/net/http#Header.Set), because the wrapper always uses canonical form of the header name

The simplest fix will be
```
w.Header()[textproto.CanonicalMIMEHeaderKey(header)] = valA
```

devsergiy (changes requested):

router-tests/header_propagation_test.go
StarpTech: We should write a test that verify it.

cmtm: Done. I wrote a test for it, but I'm not sure how the http.Header type handles canonicalization internally. If it canonicalizes when deserializing an http request, we might not be able to properly test the canonicalization with it. We'd need to use a lower level construct.

cmtm: Note that I didn't change the behavior of this handler, since it's okay if it doesn't do canonicalization.

devsergiy: why it is okay? we should use canonical header names whenever it is possible

router/core/header_rule_engine_test.go
```diff
@@ -54,6 +54,42 @@ func TestPropagateHeaderRule(t *testing.T) {
...
+				},
+			},
+		})
+		assert.Nil(t, err)
+
+		rr := httptest.NewRecorder()
+
+		clientReq, err := http.NewRequest("POST", "http://localhost", nil)
+		require.NoError(t, err)
+		clientReq.Header.Add("X-Test-1", "test1")
```
devsergiy, line 75: manually modify the header here with not canonical names

router/core/header_rule_engine_test.go
```diff
@@ -54,6 +54,42 @@ func TestPropagateHeaderRule(t *testing.T) {
...
+		updatedClientReq, _ := ht.OnOriginRequest(originReq, &requestContext{
+			logger:           zap.NewNop(),
+			responseWriter:   rr,
+			request:          clientReq,
+			operation:        &operationContext{},
+			subgraphResolver: NewSubgraphResolver(nil),
+		})
+
+		assert.Len(t, updatedClientReq.Header, 1)
+		assert.Equal(t, []string{"test1", "test2"}, updatedClientReq.Header.Values("X-Test-1"))
```
devsergiy, line 90: this line should read properly canonical form

the important thing that we should send headers to the subgraph server in canonical form

devsergiy (changes requested):

router-tests/header_propagation_test.go
cmtm: If I understand this correctly, this is the test code which mocks a subgraph and returns the headers requested by the test. The goal of this test is to verify the behavior of the cosmo router. Ensuring it properly handles non-canonicalized headers send by the subgraph are part of the test.

router/core/header_rule_engine_test.go
cmtm: discussing above

router/core/header_rule_engine_test.go
cmtm: I don't think this is where it belongs. It doesn't seem like it's the rules engines job to canonicalize the incoming headers. At least, it's not doing it before this change. This change isn't making any change to any of the code that's handling incoming headers. It's only changing it for headers in the response.

router/core/header_rule_engine.go
```diff
@@ -324,31 +324,34 @@ func (h *HeaderPropagation) applyResponseRule(propagation *responseHeaderPropaga
...
-func (h *HeaderPropagation) applyResponseRuleKeyValue(res *http.Response, propagation *responseHeaderPropagation, rule *config.ResponseHeaderRule, key, value string) {
+func (h *HeaderPropagation) applyResponseRuleKeyValue(res *http.Response, propagation *responseHeaderPropagation, rule *config.ResponseHeaderRule, key string, values []string) {
+	// Since we'll be setting the header map directly, we need to canonicalize the key
+	key = http.CanonicalHeaderKey(key)
 	switch rule.Algorithm {
 	case config.ResponseHeaderRuleAlgorithmFirstWrite:
 		propagation.m.Lock()
 		if val := propagation.header.Get(key); val == "" {
-			propagation.header.Set(key, value)
+			propagation.header[key] = values
```
devsergiy, line 345: @cmtm apologize, maybe I confused you by commenting code in tests

the main problem here is that this change has different behavior than it was before
this was applying the canonical form of a header name
```
propagation.header.Set(key, value)
```
this one
```
propagation.header[key] = values
```

is not doing that, because you manually modify the header map

cmtm: @devsergiy, does this make sense? We should still be canonicalizing it.

devsergiy: yep it is fine now
sorry if I missed it earlier, hard to track changes with the force pushes

devsergiy (approved):

PR #1545, opened 2025-01-29
feat(router): template expressions for access logs

SkArchon:
## Motivation and Context

The goal is to extend the router to allow defining custom access log attributes by [template expressions](https://cosmo-docs.wundergraph.com/router/configuration/template-expressions).

In order to do this the PR adds an "expression" field which can be set to process any log messages based on these exprlang expressions

However on top of this, the PR also goes ahead and adds the request `error` to the `expressionContext`.

This means that the user could now use expressions for access logs
```
access_logs:
  enabled: true
  router:
    fields:
    - key: "testthere"
      value_from:
        expression: "request.error ?? request.header.Get('X-WG-TRACE')"
  subgraphs:
    enabled: true
    fields:
      - key: "my-header"
        value_from:
          expression: "request.header.Get('X-WG-TRACE') == 'true' ? 'aa1' : 'aa2'"
```

**In Progress / Remaining Work atm**
Adding more of an "integration test" (right now I've written a unit test) for testing the expressions
docs

## Checklist


SkArchon:
The initial example given to me was

`request.error ?? request.body`

However, it seems like the body is not part of the expression context, when I was looking at adding it to the context I noticed it was of type `io.Reader` (readerCloser exact IIRC), which means that if the body was already read it is likely to be empty (of we will empty it by reading the io.Reader) + we would need to write code (I am assuming here) to output this to the log also.

As such, because of this + because the core specification was about adding the "error" to the context and allowing template expressions, I did not focus on this, and would like to discuss this more before continuing on deciding to add `body`.


[comment is WIP]

router/core/request_context_fields.go
```diff
@@ -32,6 +34,23 @@ const (
 
 // Helper functions to create zap fields for custom attributes.
 
+func NewExpressionLogField(val any, key string, defaultValue any) zap.Field {
+	// Depending on the condition exprlang will dereference reportError sometimes
+	// and as the method receivers are all pointers (*reportError)
+	// the Error() function wont get called unless its of type *reportError
+	switch result := val.(type) {
+	case reportError:
```
SkArchon, line 42: As the comment mentions, depending on the expression `exprlang` will return `reportError` instead of the pointer version `*reportError`, which means that Error() wont end up being called

router/internal/requestlogger/access_log_expression.go
```diff
@@ -0,0 +1,32 @@
+package requestlogger
```
SkArchon, line 1: The other function that tests the logger lies in this package `requestlogger`, but creating of the requestContext which is needed for evaluating expressions is in the `core` package, and *is not* exported.

Which means if I was to write my test cases the same way here in the `subgraph_logger_test.go` file, I will need to export the `buildRequestContext` function (and related struct), which seemed like a bad idea for this one test.

But it also made me wonder if that code should be in a separate (maybe "inner-core" package, or this package merged with core since its quite small?). I ended up writing a unit test directly `access_log_field_handler_test` in the core package.

SkArchon:
In Progress / Remaining Work atm:
* Adding more of an "integration test" (right now I've written a unit test) for testing the expressions
* There is a bug with the key being duplicated with null and the default value

router/pkg/config/config.schema.json
```diff
@@ -2554,6 +2558,10 @@
                 "type": "string",
                 "description": "The name of the response header from which to extract the value. The value is only extracted for subgraph access logs"
               },
+              "expression": {
+                "type": "string",
+                "description": "The field name of the context from which to extract the value. The value is only extracted when a context is available otherwise the default value is used."
```
StarpTech: That doesn't sound right.

router/pkg/config/config.schema.json
```diff
@@ -821,6 +821,10 @@
                         "type": "string",
                         "description": "The name of the request header from which to extract the value. The value is only extracted when a request context is available otherwise the default value is used."
                       },
+                      "expression": {
+                        "type": "string",
+                        "description": "The field name of the context from which to extract the value. The value is only extracted when a context is available otherwise the default value is used."
```
StarpTech: Here as well.

router/core/request_context_fields.go
StarpTech: I can't follow. `reportError` is an internal error from our router and has nothing to do with `expr`. Could you eloborate?

router/core/access_log_field_handler_test.go
```diff
@@ -0,0 +1,117 @@
+package core
+
+import (
```
StarpTech, line 3: I like that your wrote unit tests but we prefer integration tests. Please look in `router-tests`. We already have examples how to assert log statements.

The following cases needs to be covered:

- Add a custom router log field that reacts to an error
- Add a custom subgraph log field that reacts to an error
- Ensure that invalid expression are catched.

router/core/graphql_prehandler.go
```diff
@@ -217,7 +217,7 @@ func (h *PreHandler) Handler(next http.Handler) http.Handler {
 
 		executionOptions, traceOptions, err := h.parseRequestOptions(r, clientInfo, requestLogger)
 		if err != nil {
-			requestContext.error = err
+			AssignErrors(requestContext, err)
```
StarpTech, line 220: Instead of writing a utility function to assign the error, you could implement `ResolveAnyExpression` similar as with`ResolveStringExpression` on the `requestContext`. This would follow the pattern and the error can be set right before we pass the expressionContext to the template engine.

router/internal/requestlogger/access_log_expression.go
```diff
@@ -0,0 +1,32 @@
...
+	Key     string
+	Default string
+	Expr    *vm.Program
+}
+
+func GetAccessLogConfigExpressions(attributes []config.CustomAttribute) ([]ExpressionAttribute, error) {
+	exprSlice := make([]ExpressionAttribute, 0)
+	for _, sAttribute := range attributes {
+		if expr := sAttribute.ValueFrom.Expression; expr != "" {
+			expression, err := exprlocal.CompileAnyExpression(expr)
```
StarpTech: We should evaluate if the expression returns a loggable value. Is it possible to return no value like undefined? There is a checker module `github.com/expr-lang/expr@v1.16.9/checker` that can be used to check the return type of an expression.

StarpTech (changes requested):

StarpTech:
```
 note on adding request.body The initial example given to me was

request.error ?? request.body
```

That's correct. I'd suggest to finish it up to the point we discussed it. We can add this later. The customer wanted to log the GraphQL operation in case of an error.

router/core/graphql_prehandler.go
SkArchon: Agree on the `ResolveAnyExpression`, however I want to clarify about assigning errors in the `ResolveAnyExpression` function.

My assumption was that whenever an error occurred that the error should be available as part of the expressionContext. So if anyone did 
`reqContext.expressionContext.errror` at any time they wont get null (which would be different from `requestContext.error`).

With this change they would get null, but from what you are saying is that they should not access the expressionContext this way so **its ok**, instead they should access it via any `Resolve*Expression` method, for which we will set the error to the expressionContext before the resolving step?

StarpTech: Correct. Thanks for the explanation.

router/core/request_context_fields.go
StarpTech: Now I understand. You were referring to the returned error of the expression. Why do you believe it is important toi align this. The value should not matter because we will pass it straight to the logger which accepts any.

SkArchon: Even though the logger accepts any, the final log message ends up becoming a memory location. This becomes a problem when a customer expects the error message to be printed.

router/core/access_log_field_handler_test.go
SkArchon: Tests added

router/core/context.go
```diff
@@ -253,11 +253,43 @@ type requestContext struct {
...
+			},
+		}
+		return expr.ResolveAnyExpression(expression, copyContext)
+	}
+
+	return c.ResolveAnyExpression(expression)
+}
+
+func (c *requestContext) ResolveAnyExpression(expression *vm.Program) (any, error) {
+	c.setErrorContext()
```
StarpTech, line 282: Rather than doing it in every method, should we introduce a setter on the request context and do it there? I think you misunderstood my previous feedback.

router/core/context.go
```diff
@@ -253,11 +253,43 @@ type requestContext struct {
...
+		return
+	}
+	c.expressionContext.Request.Error = &ExprWrapError{c.error}
+}
+
+func (c *requestContext) ResolveAnyExpressionWithErrorOverride(expression *vm.Program, err error) (any, error) {
+	currContextRequest := c.expressionContext.Request
+	if err != nil {
+		// We don't really want to do a deep clone since we only know that error only can get modified
+		copyContext := expr.Context{
```
StarpTech, line 267: Why do we need to create a new `expr.Context` ?

StarpTech (changes requested):

router/core/context.go
SkArchon, line 267: I ran into a problem where when an error occurs in the subgraph, it's not assigned to the `requestContext.error` at the point of being called.

Thus any expressions like `request.error?? 'no-error'` returned `'no-error'`. Observing the same pattern that was used in the predefined context fields for `request_error` which passes an error directly, I used a similar approach / workaround. Creating a copy of the expressionContext (non deep) and assigning it to resolve the expression.

router/core/errors.go
```diff
@@ -316,3 +316,14 @@ func writeOperationError(r *http.Request, w http.ResponseWriter, requestLogger *
 		writeRequestErrors(r, w, http.StatusInternalServerError, graphqlerrors.RequestErrorsFromError(errInternalServer), requestLogger)
 	}
 }
+
+type ExprWrapError struct {
```
SkArchon, line 320: This is to circumvent this problem described here:
https://github.com/wundergraph/cosmo/pull/1545#discussion_r1936100140

While testing further I noticed that there can be other error types, so it was safer to wrap the error in a type we can control, so we have a deterministic error type outputted from the `exprlang` expression.

(I didn't come across anything atm, but if there is a better existing type to use, we can reuse that)

router/core/context.go
```diff
@@ -253,11 +253,43 @@ type requestContext struct {
 	expressionContext expr.Context
 }
 
+func (c *requestContext) setErrorContext() {
+	if c.error == nil {
+		return
+	}
+	c.expressionContext.Request.Error = &ExprWrapError{c.error}
```
SkArchon, line 260: Refer https://github.com/wundergraph/cosmo/pull/1545/files#r1936122019 and https://github.com/wundergraph/cosmo/pull/1545#discussion_r1936100140 for why we wrap

router/core/request_context_fields.go
StarpTech: I'd expect that whatever the user returns in an expression that zap is able to log it and that the pointer is dereferenced automatically before printing.

router/core/request_context_fields.go
```diff
@@ -68,7 +86,15 @@ func NewDurationLogField(val time.Duration, attribute config.CustomAttribute) za
 	return zap.Skip()
 }
 
-func AccessLogsFieldHandler(attributes []config.CustomAttribute, err any, request *http.Request, responseHeader *http.Header) []zapcore.Field {
+func AccessLogsFieldHandler(
+	logger *zap.Logger,
+	attributes []config.CustomAttribute,
+	exprAttributes []requestlogger.ExpressionAttribute,
+	passedErr error,
```
SkArchon, line 93: The previous type was "any" because it took inputs from panics and errors, I went ahead and split it into two as we want to maintain the error type in the `ExpressionContext`, but if we are ok to make it even more generic just for the panic, we can merge the fields again.

router/core/request_context_fields.go
SkArchon: From my testing so far no, but let me double check a bit more and get back to you.

router/internal/requestlogger/access_log_expression.go
```diff
@@ -0,0 +1,53 @@
...
+
+func GetAccessLogConfigExpressions(attributes []config.CustomAttribute) ([]ExpressionAttribute, error) {
+	exprSlice := make([]ExpressionAttribute, 0)
+	for _, sAttribute := range attributes {
+		exprString := sAttribute.ValueFrom.Expression
+		if sAttribute.ValueFrom == nil || exprString == "" {
+			continue
+		}
+
+		err := exprlocal.ValidateAnyExpression(exprString)
```
SkArchon, line 24: Validate against nil and a few other types, we can add / remove types later as necessary

SkArchon:
@StarpTech done the changes we discussed, which were mainly

1. WrapError and assigning a pointer for the error for logging -> 
What we discussed: WrapErrors in a separate method (so whoever needs it can call it)

2. Reassigning the context expression temporarily for subgraphs ->
What we discussed: Remove the handling of expressions in the subgraphs

However, two questions regarding tests

1. I ran the following to verify tests `make test-fresh` inside cosmo/router, and `go test -race -v ./...` inside cosmo/router-tests, I roughly ran through the `router-ci`, github actions, just want to check if there is anything I might have missed I would need to run?

2. It seems that the `router-tests` fail on some tests, this is even on the main branch without my changes. I am looking into it atm, but is it possible to get CI or someone to check my branch on their machine, would be helpful in case I am missing some things in my setup still

router/pkg/config/config.schema.json
```diff
@@ -2595,6 +2595,10 @@
                 "type": "string",
                 "description": "The name of the response header from which to extract the value. The value is only extracted for subgraph access logs"
               },
+              "expression": {
```
SkArchon, line 2598: @StarpTech This config is still shared between subgraphs and router (i.e. :- there wont be a schema validation error if you specify "expression" in a subgraph config, but it will be a no-op) , I was trying out using using allOf to get merging two subschemas to work, but didn't make too much progress yet (I didn't want to duplicate the schema config for router and subgraph of course for just one field)

If its good with you, I would prefer to open this as a separate PR?

PR #1549, opened 2025-01-30
fix(router): parse accept header per rfc 9110

df-wg:
## Motivation and Context

Users noted that we weren't parsing and recognizing Accept headers appropriately (per the rules [here](https://www.rfc-editor.org/rfc/rfc9110.html#name-accept)). Eventually, a solution for this will be rolled out as part of the stdlib (https://github.com/golang/go/issues/19307), at which point we can use that to remove this complexity and standardize our approach. 
## Checklist


StarpTech (approved):
LGTM

PR #1563, opened 2025-02-04
fix(router): enable redis to work with auth

### LINKED ISSUE #1557: Password-protected Redis storage provider connection errors
### Component(s)

router

### router version

v0.170.0

## Description

The recent PR https://github.com/wundergraph/cosmo/pull/1499 has introduced the ability to connect to Redis, when running in cluster-mode, which was previously unsupported. As a result of this change, it is my understanding that connecting to Redis clusters or indeed standalone instances with a password is no longer functioning correctly.

https://github.com/wundergraph/cosmo/pull/1499/files#diff-3173118cd9e8f3849a4d398d232609e32cac4991fd962e78280d9c68c14b67abR26 - `NewRedisCloser` is passed an options bundle, which never has `Password` populated (hence will be `""`). 
- For `cluster_mode: false`, the password parsed from the URL using the `go-redis` library is replaced by the value provided in this options struct. 
- For `cluster_mode: true`, the password is mapped into the `go-redis` library's `ClusterOptions` struct

In both cases, having an `""` value will cause issues for connections requiring authentication, and result in a `NOAUTH` response from the Redis server.

Additionally TLS connections using the `rediss://` URL scheme appear to be unsupported when using `cluster_enabled: true`. The manual stripping of the scheme from each URL strips only `redis://` (non-TLS) URLs. Doing this outside of the `go-redis` library also neglects to configure the `ClusterOptions.TLSConfig` option, which results in a non-TLS connection being attempted to a TLS Redis instance, also failing.

fwiw: a symmetrical function to `redis.ParseURL` exists in the backing `go-redis` library, specifically for parsing cluster-mode URLs into the `ClusterOptions` struct, called `redis.ParseClusterURL`. If used, this would provide greater flexibility for consumers, allowing timeouts, connection pools, etc to be configured.

## Steps to Reproduce
* Scenario 1: Standalone Redis connections *
1. Start a local Redis instance: `docker run -p 6379:6379 redis`
2. Set a password on the instance, by connecting with `redis-cli` and running:
```
CONFIG SET requirepass "mypass"
```
3. Configure Cosmo Router with a Redis storage provider: 
```toml
#... 
 storage:
    cluster_enabled: false
    urls:
      - "redis://default:mypass@localhost:6379"
# ...
```
4. Start Cosmo Router
5. Notice error in logs:
```
failed to bootstrap router: failed to create a functioning redis client 
```

* Scenario 2: Cluster mode Redis connections *
1. Start a local Redis cluster-mode cluster, with a password of "mypass"
2. Configure Cosmo Router with a Redis storage provider: 
```toml
#... 
 storage:
    cluster_enabled: true
    urls:
      - "redis://default:mypass@localhost:6379"
      - "redis://default:mypass@localhost:6380"
      - "redis://default:mypass@localhost:6381"
# ...
```
3. Start Cosmo Router
4. Notice similar error in logs

(For the TLS issue, supply `rediss://` as the scheme, and notice this is not stripped in the URLs supplied to the underlying library)

## Expected Result
Connection to Redis cluster-mode and/or standalone instances over non-TLS and/or TLS connections, with the ability to supply a valid password.

## Actual Result
- `failed to bootstrap router: failed to create a functioning redis client`

### Log output

```shell
failed to bootstrap router: failed to create a functioning redis client 
```

df-wg:
## Motivation and Context

Previous work at refactoring the redis connection added a regression, where we weren't handling redis authentication appropriately. This PR introduces a number of fixes:
* Correctly parse redis cluster URLs and parse options using the designated `ParseClusterOptions`
* Ensure that we only provide auth if added
* Add integration tests with ADL protected redis cluster

Fixes #1557 

## Checklist

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -243,19 +243,7 @@ services:
     profiles:
       - dev
 
-#  3 node minimum for a cluster, per redis documentation
-  redis-cluster-node-0:
-    image: redis:${DC_REDIS_VERSION:-7.2.4}-alpine
-    command: redis-server /usr/local/etc/redis/redis.conf
-    networks:
-      - primary
-    ports:
-      - "7000:6379"
-    volumes:
-      - ./docker/redis/redis-cluster.conf:/usr/local/etc/redis/redis.conf
-    profiles:
-      - dev
-
+  #  3 node minimum for a cluster, per redis documentation
   redis-cluster-node-1:
     image: redis:${DC_REDIS_VERSION:-7.2.4}-alpine
     command: redis-server /usr/local/etc/redis/redis.conf
@@ -263,6 +251,7 @@ services:
       - primary
     ports:
       - "7001:6379"
+      - "16371:16379"
     volumes:
       - ./docker/redis/redis-cluster.conf:/usr/local/etc/redis/redis.conf
     profiles:
@@ -275,6 +264,20 @@ services:
       - primary
     ports:
       - "7002:6379"
+      - "16372:16379"
+    volumes:
+      - ./docker/redis/redis-cluster.conf:/usr/local/etc/redis/redis.conf
+    profiles:
+      - dev
+
+  redis-cluster-node-3:
+    image: redis:${DC_REDIS_VERSION:-7.2.4}-alpine
+    command: redis-server /usr/local/etc/redis/redis.conf
+    networks:
+      - primary
+    ports:
+      - "7003:6379"
+      - "16373:16379"
     volumes:
       - ./docker/redis/redis-cluster.conf:/usr/local/etc/redis/redis.conf
     profiles:
@@ -286,9 +289,9 @@ services:
     networks:
       - primary
     depends_on:
-      - redis-cluster-node-0
       - redis-cluster-node-1
       - redis-cluster-node-2
+      - redis-cluster-node-3
     volumes:
       - ./docker/redis/:/usr/local/etc/redis/
     profiles:
@@ -331,6 +334,6 @@ volumes:
   redis:
   redis-slave:
   redis-cluster-configure:
-  redis-cluster-node-0:
   redis-cluster-node-1:
   redis-cluster-node-2:
+  redis-cluster-node-3:
--- a/router-tests/ratelimit_test.go
+++ b/router-tests/ratelimit_test.go
@@ -682,7 +682,7 @@ func TestRateLimit(t *testing.T) {
 		)
 
 		t.Run("cluster url ordering", func(t *testing.T) {
-			t.Run("diff auth in different urls", func(t *testing.T) {
+			t.Run("uses auth from first url", func(t *testing.T) {
 				t.Parallel()
 				clusterUrlSlice = []string{"redis://cosmo:test@localhost:7000", "redis://cosmo1:test1@localhost:7001", "redis://cosmo2:test2@localhost:7002"}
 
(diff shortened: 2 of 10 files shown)


df-wg:
I verified, running the steps in https://github.com/wundergraph/cosmo/issues/1557, that it solves the attached issues

router-tests/ratelimit_test.go
```diff
@@ -675,16 +675,133 @@ func TestRateLimit(t *testing.T) {
...
 		var (
-			clusterUrlSlice = []string{"localhost:7000", "localhost:7001", "localhost:7002"}
-			password        = "test"
+			clusterUrlSlice     = []string{"redis://cosmo:test@localhost:7000", "redis://cosmo:test@localhost:7001", "redis://cosmo:test@localhost:7002"}
+			noSchemeClusterUrls = []string{"localhost:7000", "localhost:7001", "localhost:7002"}
+			user                = "cosmo"
+			password            = "test"
 		)
 
+		t.Run("cluster url ordering", func(t *testing.T) {
```
endigma, line 684: I could be wrong, but it looks like this could be a table test instead of 3x the main test code, if not a tdt, perhaps a "test runner" function?

router/internal/persistedoperation/operationstorage/redis/rdcloser.go
```diff
@@ -32,20 +33,30 @@ func NewRedisCloser(opts *RedisCloserOptions) (RDCloser, error) {
...
+		rdb = redis.NewClusterClient(clusterOps)
 	} else {
 		options, err := redis.ParseURL(opts.URLs[0])
 		if err != nil {
 			return nil, fmt.Errorf("failed to parse the redis connection url: %w", err)
 		}
-		options.Password = opts.Password
+		if opts.Password != "" {
+			options.Password = opts.Password
+		}
```
endigma, line 59: `opts` + `options` used so close together is confusing, especially to skim

router/internal/persistedoperation/operationstorage/redis/rdcloser.go
```diff
@@ -60,6 +71,33 @@ func NewRedisCloser(opts *RedisCloserOptions) (RDCloser, error) {
...
+		secondaryURL, parseErr := url.Parse(rawURL)
+		if parseErr != nil {
+			opts.Logger.Warn(fmt.Sprintf("Skipping invalid Redis URL %q: %v", rawURL, parseErr))
+			continue
+		}
+
+		// Strip schema, username, and password
+		addr := secondaryURL.Host
+		if secondaryURL.User != nil && parsedUrl.User != nil && parsedUrl.User.String() != "" && secondaryURL.User.Username() != parsedUrl.User.Username() {
+			opts.Logger.Warn(fmt.Sprintf("Stripping credentials from secondary Redis address: %q", addr))
```
endigma, line 90: nitpick, zap is a structured logging library, you can write something like this instead of using fmt.Sprintf which essentially turns Zap.Logger into Zap.SugaredLogger performance wise.
```go
opts.Logger.Warn("Stripping credentials from secondary Redis address", zap.String("addr", addr))
```

If you'd really like to use Printf semantics, Zap supports this via turning the logger into a sugaredlogger and vise versa at any point, like this:
```go
sl := l.Sugar()

sl.Warnf(...)

l = sl.Desugar()

File: router/internal/persistedoperation/operationstorage/redis/rdcloser.go (hunk 2 above)
Comment (df-wg) [reply router/internal/persistedoperation/operationstorage/redis/rdcloser.go]: Updated with a more clear name, good point

File: router-tests/testenv/testenv.go [hunk 4]
```diff
@@ -106,6 +106,15 @@ func Run(t *testing.T, cfg *Config, f func(t *testing.T, xEnv *Environment)) {
 	}
 }
 
+// Run runs the test and fails the test if an error occurs
```
Noroth, line 109: copy 🍝

alepane21 (changes requested):
I would avoid the port 7000, because is used by AirPlay and can cause issues, like it was on my PC!

docker-compose.yml
```diff
@@ -243,38 +243,41 @@ services:
...
+  #  3 node minimum for a cluster, per redis documentation
+  redis-cluster-node-1:
     image: redis:${DC_REDIS_VERSION:-7.2.4}-alpine
     command: redis-server /usr/local/etc/redis/redis.conf
     networks:
       - primary
     ports:
-      - "7000:6379"
+      - "7001:6379"
+      - "16371:16379"
```
df-wg, line 254: Expose an additional port to use, per their docs: https://redis.io/docs/latest/operate/oss_and_stack/management/config-file/
```
# In certain deployments, Redis Cluster nodes address discovery fails, because
# addresses are NAT-ted or because ports are forwarded (the typical case is
# Docker and other containers).
#
# In order to make Redis Cluster working in such environments, a static
# configuration where each node knows its public address is needed. The
# following four options are used for this scope, and are:
#
# * cluster-announce-ip
# * cluster-announce-port
# * cluster-announce-tls-port
# * cluster-announce-bus-port
#
# Each instructs the node about its address, client ports (for connections
# without and with TLS) and cluster message bus port. The information is then
# published in the header of the bus packets so that other nodes will be able to
# correctly map the address of the node publishing the information.
#
# If tls-cluster is set to yes and cluster-announce-tls-port is omitted or set
# to zero, then cluster-announce-port refers to the TLS port. Note also that
# cluster-announce-tls-port has no effect if tls-cluster is set to no.
#
# If the above options are not used, the normal Redis Cluster auto-detection
# will be used instead.
#
# Note that when remapped, the bus port may not be at the fixed offset of
# clients port + 10000, so you can specify any port and bus-port depending
# on how they get remapped. If the bus-port is not set, a fixed offset of
# 10000 will be used as usual.
#
# Example:
#
# cluster-announce-ip 10.1.1.5
# cluster-announce-tls-port 6379
# cluster-announce-port 0
# cluster-announce-bus-port 6380
```

alepane21 (approved):

PR #1567, opened 2025-02-04
fix: inject a per-subgraph http.Client to fix timeout race issue

endigma:
## Checklist


router/core/factoryresolver.go
```diff
@@ -61,19 +63,22 @@ type DefaultFactoryResolver struct {
...
 	kafkaPubSubBySourceID map[string]pubsub_datasource.KafkaPubSub,
 ) *DefaultFactoryResolver {
+	transportFactory := NewTransport(transportOptions)
 
 	defaultHttpClient := &http.Client{
+		Timeout:   transportOptions.SubgraphTransportOptions.RequestTimeout,
 		Transport: transportFactory.RoundTripper(enableSingleFlight, baseTransport),
 	}
 	streamingClient := &http.Client{
+		Timeout:   transportOptions.SubgraphTransportOptions.RequestTimeout,
```
devsergiy, line 81: please remove it, streaming client could not have a timeout

endigma: 🤦

router/core/subgraph_transport_test.go
devsergiy, line 1: as we removed unit tests

could you please add integration test for timeouts

based on this examples

this query hits one subgraph
https://github.com/wundergraph/cosmo/blob/e6f4b9bf9d642d39933528dd1654a9c9d969e565/router-tests/graphql_over_get_test.go#L20

this query hits 2 subgraphs

basically it was the same in the removed tests

devsergiy (approved):

router/core/factoryresolver.go
```diff
@@ -369,6 +379,15 @@ func (l *Loader) Load(engineConfig *nodev1.EngineConfiguration, subgraphs []*nod
 			}
 
 			dataSourceName := l.subgraphName(subgraphs, in.Id)
+
+			factory, err := l.resolver.ResolveGraphqlFactory(dataSourceName)
+			if err != nil {
+				return nil, err
+			}
+			if factory == nil {
```
StarpTech, line 387: When can this happen?

router/core/subgraph_transport.go
```diff
@@ -43,22 +42,8 @@ func (tt *TimeoutTransport) RoundTrip(req *http.Request) (*http.Response, error)
 	subgraph := rq.ActiveSubgraph(req)
```
StarpTech, line 42: This will also fix the case when timeouts has been defined for all sugraphs right? Please add a test that covers it.

StarpTech (changes requested):

router/core/factoryresolver.go
endigma: I can look into this, but I just moved this code to make the subgraph name available, the error handling here is as it was.

router/core/subgraph_transport.go
endigma: Added

router/core/factoryresolver.go
endigma: I checked, and this appears to be logically impossible, but it is pre-existing in all the factory consumers in this file. I'll ping the git blame before removing them.

endigma: Removed

demo/pkg/subgraphs/test1/subgraph/schema.resolvers.go
```diff
@@ -73,6 +73,21 @@
 	return big, nil
 }
 
+// LongResponse is the resolver for the longResponse field.
+func (r *queryResolver) LongResponse(ctx context.Context, artificalDelay int, bytes int) (*string, error) {
+	if artificalDelay > 0 {
+		time.Sleep(time.Duration(artificalDelay) * time.Millisecond)
+	}
+
+	b := make([]byte, bytes)
```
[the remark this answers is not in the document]
endigma: Yes mr AI, that is the point

router/core/factoryresolver.go
Aenimus: My fault. I wrote it like this to be extra safe but we should just trust the contract (and that the contract will never change).

demo/pkg/subgraphs/test1/subgraph/schema.resolvers.go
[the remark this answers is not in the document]
Aenimus: This will blow up if a negative int is provided, no?

devsergiy (approved):

StarpTech (approved):
LGTM

PR #1577, opened 2025-02-10
fix: makes default config options work properly within maps

endigma:
## Motivation and Context

We noticed that default config options are not correctly merged into parts of the config that use maps. This adds support for correctly merging in these defaults at config load time.

## Checklist

## Todo

- [x] Check that this wouldn't be a better fit inside of the router initialization
- [x] Double check test coverage


router/pkg/config/config.go
```diff
@@ -146,7 +146,8 @@ type TrafficShapingRules struct {
 	// Apply to requests from clients to the router
 	Router RouterTrafficConfiguration `yaml:"router"`
 	// Subgraphs is a set of rules that apply to requests from the router to subgraphs. The key is the subgraph name.
-	Subgraphs map[string]*GlobalSubgraphRequestRule `yaml:"subgraphs,omitempty"`
+	Subgraphs         map[string]*GlobalSubgraphRequestRule `yaml:"-"`
+	SubgraphsNillable map[string]*NillableRequestRule       `yaml:"subgraphs,omitempty"`
```
StarpTech, line 150: Why do you want to express nillable in the type name?

router/pkg/config/config.go
```diff
@@ -946,6 +965,40 @@ func LoadConfig(configFilePath string, envOverride string) (*LoadResult, error)
 		}
 	}
 
+	// Merge in defaults for subgraph TrafficShapingRules
+	defaultRequestRule := GlobalSubgraphRequestRule{
```
StarpTech, line 969: This would be the second place where we defined these defaults. I'd suggest fallback to the default in `NewTransportRequestOptions` and reuse it. In that case, it works in the tests and code is more coherent.

Example

```go
func NewTransportRequestOptions(cfg config.GlobalSubgraphRequestRule) TransportRequestOptions {
	return TransportRequestOptions{
		RequestTimeout:        or(cfg.RequestTimeout, 60* time.Second),
        // ...
	}
}
```

StarpTech (changes requested):

StarpTech (approved):
LGTM

PR #1584, opened 2025-02-11
feat(router): make header matching rules case insensitive

df-wg:
## Motivation and Context

Many users have found themselves in confusing places when they've tried to apply our [header propagation rules](https://cosmo-docs.wundergraph.com/router/proxy-capabilities/response-header-operations#enabling-header-propagation), because they neglected to use the case insensitive prefix to the regex. 

HTTP Header names are by [RFC](https://www.w3.org/Protocols/rfc2616/rfc2616.html) case insensitive, so there should be no need to explicitly declare them as case insensitive. 

Any case where we were using the `<header>.Get(HEADER)` was by its nature case insensitive, but this specifically needed to fix the header matching rules 

## Checklist


alepane21 (approved):
I was initially concerned that prepending the prefix to the regexp might break some existing patterns. However, I couldn’t identify any issues during testing. In the worst case, a previously non-functional rule might start working unexpectedly, but this risk seems minimal and non-blocking.

PR #1602, opened 2025-02-18
feat(engine): apollo router compatible non-2XX response errors

endigma:
## Motivation and Context

- Adds config option `/apollo_router_compatibility_flags/subrequest_http_error/enabled` which enables Apollo Router like behaviour when subgraphs return non-2XX http status.
- Bumps `graphql-go-tools` to (v2.0.0-rc.156)[https://github.com/wundergraph/graphql-go-tools/releases/tag/v2.0.0-rc.156] to support this functionality

## Checklist

### Features

* apollo-router-like non-ok http status errors ([#1072](https://github.com/wundergraph/graphql-go-tools/issues/1072)) ([e685c29](https://github.com/wundergraph/graphql-go-tools/commit/e685c29331c0d1879ff8e099d4441047fbddf054))

### DIFF
--- a/router/core/executor.go
+++ b/router/core/executor.go
@@ -48,13 +48,14 @@ type Executor struct {
 }
 
 type ExecutorBuildOptions struct {
-	EngineConfig             *nodev1.EngineConfiguration
-	Subgraphs                []*nodev1.Subgraph
-	RouterEngineConfig       *RouterEngineConfiguration
-	PubSubProviders          *EnginePubSubProviders
-	Reporter                 resolve.Reporter
-	ApolloCompatibilityFlags config.ApolloCompatibilityFlags
-	HeartbeatInterval        time.Duration
+	EngineConfig                   *nodev1.EngineConfiguration
+	Subgraphs                      []*nodev1.Subgraph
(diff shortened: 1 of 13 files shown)


router/core/executor.go
```diff
@@ -96,6 +97,10 @@ func (b *ExecutorConfigurationBuilder) Build(ctx context.Context, opts *Executor
 		options.ResolvableOptions.ApolloCompatibilityReplaceInvalidVarError = true
 	}
 
+	if opts.ApolloRouterCompatibilityFlags.SubrequestHTTPError.Enabled {
```
Aenimus, line 100: Now we have two, should we add an enable all like we have for the other original compatibility flags?

devsergiy (approved):

router/core/executor.go
endigma: For the two current flags and the use case I don't think an enable all is necessary yet, you'd only want to enable these if you depend on them, to aid migration to more standards compliant code. There's a good chance you could migrate from Apollo Router to Cosmo without hitting either of these issues I think.

PR #1609, opened 2025-02-20
fix: set responseWriter in requestContext to the wrapped one

endigma:
correctly sets `requestContext.responseWriter` to the wrapped one used for capturing status, bytes written, etc


StarpTech (approved):
LGTM

PR #1623, opened 2025-02-24
fix(cli): resolve broken console outputs

SkArchon:
## Motivation and Context

This PR fixes two issues

**Writing to stderr**
Writing to stderr, we have changed console.log to console.warn. `console.warn` in nodejs is just an alias for `console.error` which writes to stderr. When we write to stderr anytime data gets piped to a different source, only the stdout is piped. Write now messages like deprecation messages get piped out because they are printed to stdout, this will change it to stderr then.

**Cleaning up process.exit**
When we use process.exit it means that any async processes wont get to run on the event loop, because console.log can be async this means that the everything might not get written to stdout. The workaround is either keep process.exit and use process.stdout.write, and set a callback, which we wait until is run before exiting the process. However we opted to simply do an early return. The nodejs docs itself suggests this approach.
We can look at removing process.exit separately if needed
See More Here: https://nodejs.org/api/process.html#processexitcode

I cleaned up all the process.exit functions, and I manually tested each file that was modified for the code path it was modified (i.e. verifying the early returns), this was mostly for the error paths since it was mainly `exit(1)`s, but tested the happy path for the `exit(0)`s.

On a seperate note, I saw that program.error of the commander library also internally calls process.exit, but we can look at refactoring this out later (if it even makes sense)

## Checklist

### DIFF
--- a/cli/src/commands/graph/monograph/commands/list.ts
+++ b/cli/src/commands/graph/monograph/commands/list.ts
@@ -85,7 +85,7 @@ export default (opts: BaseCommandOptions) => {
     }
 
     if (options.raw || options.json) {
-      console.log(filteredGraphs);
+      console.log(JSON.stringify(filteredGraphs));
       return;
     }
 
--- a/cli/src/commands/graph/monograph/commands/list.ts
+++ b/cli/src/commands/graph/monograph/commands/list.ts
@@ -81,12 +81,12 @@ export default (opts: BaseCommandOptions) => {
     }
 
     if (options.raw) {
-      console.log(pc.yellow('Please use the --json option. The --raw option is deprecated.'));
+      console.warn(pc.yellow('Please use the --json option. The --raw option is deprecated.'));
     }
 
     if (options.raw || options.json) {
       console.log(filteredGraphs);
-      process.exit(0);
+      return;
     }
 
     const graphsTable = new Table({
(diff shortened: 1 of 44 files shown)


JivusAyrus (changes requested):
This issue is tagged with happenning, and their problem was with `wgc subgraph check`. Should we fix that command also in this PR?

JivusAyrus:
- subgraph commands like publish, check, update and delete.
- federated commands like create, update and delete

cli/src/commands/graph/monograph/commands/list.ts
```diff
@@ -77,16 +77,16 @@ export default (opts: BaseCommandOptions) => {
...
 
     if (options.raw) {
-      console.log(pc.yellow('Please use the --json option. The --raw option is deprecated.'));
+      console.warn(pc.yellow('Please use the --json option. The --raw option is deprecated.'));
     }
 
     if (options.raw || options.json) {
-      console.log(filteredGraphs);
-      process.exit(0);
+      console.log(JSON.stringify(filteredGraphs));
```
SkArchon, line 88: The output seemed a bit off and I couldnt parse it with jq without using JSON.stringify, also follows the same pattern of the other list functions

SkArchon:
@JivusAyrus I ended up cleaning all the process.exits since there was only about half left after replacing the federated-graph, monogrpah and subgraph functions. I have manually tested these changes for the code paths that were modified (i.e. verifying the early returns), this was mostly for the error paths since it was mainly `exit(1)`s, but tested the happy path for the `exit(0)`s

JivusAyrus (approved):

PR #1625, opened 2025-02-25
ci: optimize bun binaries and put experimental in the name

endigma:
- Gzips the built binaries for artifact size optimization
- Added `experimental` to released filenames


endigma:
fwiw, the bun flags here do nearly nothing
image

endigma:
`build_test` wasn't reporting to the required check correctly, so I've split the PR workflow into `build_test` and `build_upload_matrix`, the latter depending on the former's success

StarpTech (approved):
LGTM

PR #1628, opened 2025-02-25
fix: correct and deterministic dist folder structure

endigma:
- Changes `tsconfig.json` to set `rootDir` which should make `dist` folder structure constant in the future
- Adds import assertion and doesn't do esmodule import from json

### DIFF
--- a/cli/src/core/config.ts
+++ b/cli/src/core/config.ts
@@ -3,7 +3,7 @@ import { join } from 'pathe';
 import yaml from 'js-yaml';
 import envPaths from 'env-paths';
 
-import { version } from '../../package.json';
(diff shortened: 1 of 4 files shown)


cli/src/core/config.ts
```diff
@@ -3,7 +3,7 @@ import { join } from 'pathe';
 import yaml from 'js-yaml';
 import envPaths from 'env-paths';
 
-import { version } from '../../package.json';
+import info from '../../package.json' with { type: "json" };
```
Aenimus, line 6: Why are we not deconstructing? `{ version }`

Aenimus (approved):
LGTM

PR #1634, opened 2025-02-26
test: run kafka as github CI service instead of container

endigma:
- patch for seemingly load related integration test failures perhaps as a result of all the new CLI tests
- runs kafka as github CI service instead of container for the time being
- kafka service configuration is transplanted from the `docker-compose.yml` file

I intend to find a nicer solution soon.

### DIFF
--- a/router-tests/testenv/testenv.go
+++ b/router-tests/testenv/testenv.go
@@ -366,6 +366,16 @@ func createTestEnv(t testing.TB, cfg *Config) (*Environment, error) {
 	if cfg.EnableKafka {
 		if os.Getenv("CI") == "true" {
 			cfg.KafkaSeeds = []string{"localhost:9092"}
+
+			client, err := kgo.NewClient(
+				kgo.SeedBrokers(cfg.KafkaSeeds...),
+			)
+			if err != nil {
+				return nil, err
+			}
+
+			kafkaClient = client
+			kafkaAdminClient = kadm.NewClient(kafkaClient)
 		} else {
 			kafkaStarted.Add(1)
 			go func() {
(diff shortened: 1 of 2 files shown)


router-tests/testenv/testenv.go
```diff
@@ -364,22 +364,36 @@ func createTestEnv(t testing.TB, cfg *Config) (*Environment, error) {
 	}
 
 	if cfg.EnableKafka {
-		kafkaStarted.Add(1)
-		go func() {
-			defer kafkaStarted.Done()
+		if os.Getenv("CI") == "true" {
```
StarpTech, line 367: This makes it impossible or hard to test locally. Why do you think running kafka in CI won't result in flaky tests? Is it all about how we instantiate Kafka?

StarpTech (changes requested):

router-tests/testenv/testenv.go
endigma: Kafka still runs locally, it just doesn't start a container if it's in CI. It's something to do with load from what we could tell during debugging, not due to container startup or cleanup but just connection timeouts. 

I believe containers started by `services` are a bit different in GHA than ones started by docker within the job, not sure exactly how but the tests run faster and fail less when it's a service.

endigma: FWIW the GHA service container is run with the same environment as the one it uses locally

StarpTech: In summary it improves but doesn't eliminate the cause?

StarpTech (approved):
LGTM

PR #1635, opened 2025-02-26
ci: dry-run versioning before release build

endigma:
## Motivation and Context

Versioning in CLI releases is wrong currently due to the build happening before `package.json` is updated in `lerna publish`. I looked pretty deep into lerna docs to see if I could do this more "correctly" by separating the version and publish commands, but nothing seems to work except running "version" and then using your own script to handle publishing.

I'd like not to fix what isn't broken, so I do the simple thing and dry-run the release (which adds changelog and `package.json` and reset before and after building.

## Checklist


StarpTech (approved):
LGTM

PR #1639, opened 2025-02-26
feat: option to remove scopeinfo from prometheus metrics

endigma:
- **test: improve config tests with parallel, fix testify parameter ordering**
- **feat: config to disable otel scope info in prometheus metrics**

# Todo

- [x] Unit tests, if applicable
- [x] Investigate defaults for this portion of the config
- [x] E2E test

### DIFF
--- a/router-tests/prometheus_test.go
+++ b/router-tests/prometheus_test.go
@@ -2986,25 +2986,12 @@ func TestPrometheus(t *testing.T) {
 
 			require.Len(t, requestsInFlightMetricsFiltered, 2)
 
-			require.NotContains(t, requestsInFlightMetricsFiltered[0].Label, &io_prometheus_client.LabelPair{
-				Name:  PointerOf("otel_scope_name"),
-				Value: PointerOf("cosmo.router.prometheus"),
-			})
-
-			require.NotContains(t, requestsInFlightMetricsFiltered[0].Label, &io_prometheus_client.LabelPair{
-				Name:  PointerOf("otel_scope_version"),
-				Value: PointerOf("0.0.1"),
-			})
-
-			require.NotContains(t, requestsInFlightMetricsFiltered[1].Label, &io_prometheus_client.LabelPair{
-				Name:  PointerOf("otel_scope_name"),
-				Value: PointerOf("cosmo.router.prometheus"),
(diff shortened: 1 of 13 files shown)


router-tests/prometheus_test.go
```diff
@@ -50,8 +50,6 @@ func TestPrometheus(t *testing.T) {
 			requestTotalMetrics := requestTotal.GetMetric()
 
 			require.Len(t, requestTotalMetrics, 2)
-			require.Len(t, requestTotalMetrics[0].Label, 12)
```
StarpTech, line 53: Why did you remove it?

router-tests/prometheus_test.go
```diff
@@ -2998,6 +2952,120 @@ func TestPrometheus(t *testing.T) {
...
+
+			mfFiltered, err = promRegistryFiltered.Gather()
+			require.NoError(t, err)
+
+			requestsInFlightFiltered := findMetricFamilyByName(mfFiltered, "router_http_requests_in_flight")
+			requestsInFlightMetricsFiltered := requestsInFlightFiltered.GetMetric()
+
+			require.Len(t, requestsInFlightMetricsFiltered, 2)
+
+			require.Equal(t, []*io_prometheus_client.LabelPair{
```
StarpTech, line 2989: Wouldn't it be easier to read if we would only test if those labels aren't included?

StarpTech (changes requested):

router-tests/prometheus_test.go
endigma: We check the entire contents immediately after, so this just produces unhelpful test failures

router-tests/prometheus_test.go
endigma: I'm not sure what you mean, this is how every single test in this entire package functions. I could test for presence/absence of the otel labels by looping but it where the entire object is not too complex I don't see a big issue in comparing the full metrics captured

router-tests/prometheus_test.go
StarpTech: Alright, thanks

StarpTech (approved):
LGTM

PR #1642, opened 2025-02-27
feat(router): add RouterOnRequest hook

SkArchon:
## Motivation and Context

The user needs to be able to modify request headers BEFORE anything, this means adding a new hook that users can write using custom modules.

The name of the hook is `RouterOnRequest`. The hook runs at the start of the request before certain internal middleware, notably `authentication` and `tracing`. The purpose of the hook is for the user to do any pre manipulation on the request start.

An example case would be if the user wants to change the authentication header based on some custom logic, this would be the hook for the user to do it with a custom module, because it runs BEFORE the router handles authentication.

## Checklist


router/cmd/custom/module/module.go
```diff
@@ -104,6 +111,7 @@ func (m *MyModule) Module() core.ModuleInfo {
 // Interface guard
 var (
 	_ core.RouterMiddlewareHandler = (*MyModule)(nil)
+	_ core.PreRequestMiddleware    = (*MyModule)(nil)
```
StarpTech, line 114: I'd like to suggest a different name. I think `RouterOnRequest` is more appropriate.

StarpTech (changes requested):

router/cmd/custom/module/module.go
SkArchon: Are you suggesting `RouterOnRequestMiddlewareHandler` or `RouterOnRequest`?

Separate note, actually missed adding "Handler" originally

SkArchon: I have a worry that people might get confused between `RouterMiddleware` and `RouterOnRequest`, do you think `RouterOnRequestStart` might make more sense?

StarpTech: I like your idea but I don't believe it becomes an issue. Also if you add a `Start` people are looking for `End`. I think this creates inconsistency. Look at https://fastify.dev/docs/v2.15.x/Documentation/Lifecycle/ it reflects the proposed structure.

SkArchon: I don't have a problem with the name itself, more about a worry of a distinction with the "RouterMiddleware" hook (Fastify does not seem to have that problem since they don't have 'RouterMiddleware"). but agree on folks looking for the `End` if we do `Start`.

However I checked some other frameworks as well and a few seem to also use `OnRequest` also, so will stick to that.

SkArchon: Modified

SkArchon:
The Hook is added after building the request context and after the recovery handler, but before everything else. The sinle star after the number (*) is used to indicate `httpRouter.Use` calls inside if conditions
 
```
1  : httpRouter.Use(recoveryhandler)     : Recover from panics
2  : RequestSize                         : Limit max body size if not multipart
3* : HandleCompression                   : HandleCompression if specified, this is on by default
4  : RequestID                           : If RequestID is not already set generated and set in header
5  : RealIP                              : Set the real ip of the service that sent the request
6* : Cors                                : If Cors is enabled

buildGraphMux function
============
7  : buildRequestContext                 : The first function where the request context is built and set
8  : recoveryHandler                     : Set Recovery Handler

**PRE REQUEST MIDDLEWARE NEW HOOK ADDED HERE**
9  : addCommonTraceAttribute             : Add Telemetery
10*: traceHandler.Handler                : Add Tracing if it is enabled
11*: requestLogger                       : Add requestLogger if it is enabled (request is logged after subsequent middlewares are executed)

     (Init prehandlerStruct: Uses graphql Prehandler struct)
12: httpRouter.Use(s.playgroundHandler, wsMiddleware) : if s.playgroundConfig.Enabled && s.graphqlPath == s.playgroundConfig.Path
   : OR httpRouter.Use(wsMiddleware)                 : else

13 : graphqlPrehandler                   : Prehandler
14 : routerMiddlewares                   : routerMiddlewares from custom modules if specified
============

7 to 14 is repeated for every featureFlagMux

15 : Some gzip wrapper?                  : gzhttp.NewWrapper
16*: CookieWhitelist middleware          : if headerRules are not null this is initialized (but the headerRules values can be empty though)
```

StarpTech (approved):
LGTM

PR #1646, opened 2025-03-03
feat(router): add support for jwks discovery via openid-configuration endpoint

Noroth:
## Motivation and Context

This PR adds support for JWKs discovery via the `.well-known/openid-configuration` endpoint 
https://openid.net/specs/openid-connect-discovery-1_0.html

Manually tested against **Auth0**

## Checklist


router-tests/authentication_test.go
```diff
@@ -942,6 +943,85 @@ func TestAlgorithmMismatch(t *testing.T) {
...
+			token, err := authServer.TokenForKID(c.KID(), nil)
+			require.NoError(t, err)
+
+			tokens[c.KID()] = token
+		}
+
+		return tokens, authenticators
+	}
+
+	t.Run("Should discover JWKs from OIDC discovery endpoint", func(t *testing.T) {
```
alepane21, line 1005: I would add some more tests:
- endpoint not responding correctly (404 or response non as expected)
- endpoint taking a long time to respond

alepane21 (changes requested):

router-tests/authentication_test.go
Noroth: I can add a test for the 404 scenario.

The tests do not really add a lot of value. For a successful router startup we need to be able to retrieve keys from the endpoint. Otherwise the router will fail during startup, which is intended behavior. 

The test would simply show that the function to create a token decoder returns an error after the timeout.

Therefore we would never achieve to test requests against the router because we couldn't even start it.

alepane21 (approved):

PR #1650, opened 2025-03-04
[WIP] fix: test fix

SkArchon:
## Checklist


router/core/errors.go
```diff
@@ -257,12 +263,18 @@ func writeMultipartError(w http.ResponseWriter, requestErrors graphqlerrors.Requ
...
 
-	resp, err := wrapMultipartMessage(responseBytes)
+	isSubscription := operationType == "subscription"
+	resp, err := wrapMultipartMessage(responseBytes, isSubscription)
 	if err != nil {
 		return err
 	}
 
-	resp = append(resp, '\n')
+	if isSubscription {
```
SkArchon, line 272: FYI: I am not sure if we actually need this, as from what I gathered we call this from the router (and not from the gotools subscription code), but I need to look into subscriptions a bit more, and see how errors are emitted for subscriptions via gotools.

Subscriptions are also broken for the apollo kotlin client,  so I will take a look at both of these together.

SkArchon:
Manual scenarios that I tested
```
URQL:
 - Non Subscriptions
    - With Payload (Errors): does not work
    - With Payload (Data): does not work
    - Without Payload (Errors): Works, shows errors
    - Without Payload (Data): Works, shows data 
 - Subscriptions
    - Could not get to work with fetch multipart yet via the proxy, since the poll doesn’t really work via the test proxy I am using, need to figure out how to setup and use graphqlgotools and remove payload to test this

Apollo JS Client
 - Non Subscriptions
   - With Payload (Errors): Works
   - With Payload (Data): Works
   - Without Payload (Errors): Works
   - Without Payload (Data): Works
 - Subscriptions
    - Works with both payload and non payload elements

Apollo Kotlin Client (3.8.2)
 - Non Subscriptions
   - With Payload (Errors): Empty / Swallows error (after \r fix)
   - With Payload (Data):
   - Without Payload (Errors): Shows Error
   - Without Payload (Data):
 - Subscriptions
    - Entirely broken without \r fix
    - Partially broken with \r fix, after the stream is done it errors out
       - Works as long as wrapped in payload
       - Errors out when not wrapped in payload
     
Relay / Hasura go client
 - I took a look at these clients, but they do not have support for mulltipart (GitHub search on these repos before testing for multipart returned nothing). In fact for relay a separate github issue pointed to using apollo client internally for multipart support.
```

router/core/errors.go
StarpTech: This is called from `writeRequestErrors` which can be called when using subscription over SSE or multipart.

endigma:
You may also want to try with HoudiniGraphql client

router/core/errors.go
SkArchon: So what I am worried about is duplicate end boundaries, so in flushwriter on complete (which is called through graphqlgotools) we write
```
_, _ = f.writer.Write([]byte("--" + multipartBoundary + "--\n"))
```
since we have this, if I remove the if condition I want to make sure that we cant have duplicate end boundaries.
That is
```
---graphql---
---graphql---
```
 Thus the if condition to be on the safe side, will verify and remove in a pr.

devsergiy (approved):

StarpTech (approved):
LGTM

PR #1656, opened 2025-03-05
fix(router): wrong description of the expression field in config.schema.json

git-hulk:
Just a simple wrong description in config.schema.json.


git-hulk:
The failure of the integration test should be fixed in https://github.com/wundergraph/cosmo/pull/1657

PR #1657, opened 2025-03-05
fix(ci): no need to use secrets when running the integration test

git-hulk:
## Motivation and Context

The external PRs have no permission to read the secrets, which will cause failure when running the integration test. It's unnecessary to use the secrets since there's no place where the credentials was used.

## Checklist

### DIFF
--- a/.github/workflows/router-ci.yaml
+++ b/.github/workflows/router-ci.yaml
@@ -145,9 +145,6 @@ jobs:
           --health-interval 10s
           --health-timeout 5s
           --health-retries 5
-        credentials:
-          username: ${{secrets.DOCKER_USERNAME}}
-          password: ${{secrets.DOCKER_PASSWORD}}
         ports:
           - 6379:6379
       kafka:


.github/workflows/router-ci.yaml
```diff
@@ -145,9 +145,6 @@ jobs:
           --health-interval 10s
           --health-timeout 5s
           --health-retries 5
-        credentials:
```
StarpTech, line 148: Hi @git-hulk, the docker credentials are used to avoid rate limit issues.

git-hulk: I see, does it sound good to change the DockerHub to other mirror sources instead of failing the integration, which is quite unfriendly for external contributors? please refer to https://github.com/actions/runner-images/issues/1445.

StarpTech: This is a good question and we also thought about it but it's not a priority for us right now. Pointing to a docker mirror can be a security risk. We need to evaluate different solutions.

git-hulk:
1.

PR #1661, opened 2025-03-06
feat: add execution of check_query_planner

alepane21:
With this PR we introduced a new check on the cosmo repo: `query-planner-ci.yaml`.
This check will run everytime the wgc cli or the router or some of their dependencies has been changed in a PR.
It will call a workflow in the cosmo-celestial repo and add a message in this PR.

Also part of this PR the query-planner has been changed to validates query before query planning: if a query is not valid, we don't want to trigger an error, because it could be just a wrong query (like a frontend doing an outdated query or to a wrong federated graph).

I also remove the `make test` from the `Dockerfile` after discussing it with Dustin.

## Motivation and Context

Catch unwanted changes to query plans after changes to cosmo.

## Checklist

- [X] Documentation has been updated on [https://github.com/wundergraph/cosmo-docs](https://github.com/wundergraph/cosmo-docs).


StarpTech (approved):
LGTM

PR #1664, opened 2025-03-07
feat: add request body to expression context

SkArchon:
## Motivation and Context

We want to add the request raw body to the expression context. This will allow users to use the body as the basis for expressions throughout the router.

## Checklist

### DIFF
--- a/router-tests/structured_logging_test.go
+++ b/router-tests/structured_logging_test.go
@@ -2237,6 +2237,262 @@ func TestFlakyAccessLogs(t *testing.T) {
 			})
 		})
 
+		t.Run("validate the evaluation of the body.raw expression for a query", func(t *testing.T) {
+			t.Parallel()
+
+			testenv.Run(t,
+				&testenv.Config{
+					AccessLogFields: []config.CustomAttribute{
+						{
+							Key: "service_name",
+							ValueFrom: &config.CustomDynamicAttribute{
+								RequestHeader: "service-name",
+							},
+						},
+						{
+							Key: "operation_hash",
+							ValueFrom: &config.CustomDynamicAttribute{
+								ContextField: core.ContextFieldOperationHash,
+							},
+						},
+						{
+							Key: "expression_body",
+							ValueFrom: &config.CustomDynamicAttribute{
+								Expression: "request.body.raw",
+							},
+						},
+					},
+					LogObservation: testenv.LogObservationConfig{
+						Enabled:  true,
+						LogLevel: zapcore.InfoLevel,
+					},
+				},
+				func(t *testing.T, xEnv *testenv.Environment) {
+					res := xEnv.MakeGraphQLRequestOK(testenv.GraphQLRequest{
+						Query:  `query employees { employees { id } }`,
+						Header: map[string][]string{"service-name": {"service-name"}},
+					})
+					require.JSONEq(t, employeesIDData, res.Body)
+					requestLog := xEnv.Observer().FilterMessage("/graphql")
+					requestLogAll := requestLog.All()
+					requestContext := requestLogAll[0].ContextMap()
+
+					expectedValues := map[string]interface{}{
+						"log_type":        "request",
+						"status":          int64(200),
+						"method":          "POST",
+						"path":            "/graphql",
+						"query":           "",
+						"ip":              "[REDACTED]",
+						"service_name":    "service-name",                                         // From request header
+						"operation_hash":  "1163600561566987607",                                  // From context
+						"expression_body": "{\"query\":\"query employees { employees { id } }\"}", // From expression
+					}
+					additionalExpectedKeys := []string{
+						"user_agent",
+						"latency",
+						"config_version",
+						"request_id",
+						"pid",
+						"hostname",
+					}
+					checkValues(t, requestContext, expectedValues, additionalExpectedKeys)
+				},
+			)
+		})
+
+		t.Run("validate the evaluation of the body.raw expression for a file upload", func(t *testing.T) {
+			testenv.Run(t, &testenv.Config{
+				AccessLogFields: []config.CustomAttribute{
+					{
+						Key: "operation_hash",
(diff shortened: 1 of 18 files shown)


router/pkg/config/config.schema.json
```diff
@@ -1998,6 +1998,11 @@
       "default": true,
       "description": "Enable the router registration. The router registration is used to register the router on the control-plane. The default value is true. This should not be modified unless you know what you are doing."
     },
+    "enable_body_in_expression_context": {
```
jensneuse, line 2001: How about this:

```json
{
  "include_request_body_in_expression_context": {
    "type": "object",
    "description": "Including the request body in the expression context can be useful, e.g. for logging or debugging purposes. However, as this can have a negative impact on performance and memory usage, it's disabled by default and can be enabled on a per-request basis using an expression.",
    "additionalProperties": false,
    "properties": {
      "enabled": {
        "type": "boolean",
        "default": false,
        "description": "Whether or not the request body should be included in the expression context."
      },
      "expression": {
        "type": "string",
        "default": "",
        "description": "An expression that will be evaluated to determine if the request body should be included in the expression context. If the expression is not provided, the request body will always be included if enabled is true."
      }
    }
  }
}
```

jensneuse: having such an expression to eval, we can look at a special header, or even the content size header to only enable this feature for very specific requests

router/core/graphql_prehandler.go
```diff
@@ -493,6 +496,15 @@ func (h *PreHandler) handleOperation(req *http.Request, variablesParser *astjson
 		}
 	}
 
+	// There is potential for the body to be quite large, for example because the query string has a few
+	// thousand lines, thus we allow users to enable this if they want to actually log the body
+	if h.enableBodyInExprContext {
```
alepane21, line 501: Being this an heavy operation, can we also check if the body is actually used by any expressions? So that if someone enables it but is not using it we still don't enable it. It shouldn't be too hard, I made something similar here: https://github.com/wundergraph/cosmo/blob/main/router/core/attribute_expressions.go#L58

router/internal/expr/expr.go
```diff
@@ -48,6 +48,7 @@ type Request struct {
 	Auth   RequestAuth    `expr:"auth"` // if changing the expr tag, the ExprRequestAuthKey should be updated
 	URL    RequestURL     `expr:"url"`
 	Header RequestHeaders `expr:"header"`
+	Body   RequestBody    `expr:"body"`
```
alepane21, line 51: I think you should also add a checker that if someone is using the body, but is not enabled, an error should be returned as if the expression is not valid. I think this is something you can do also using the Visitors like is done here: https://github.com/wundergraph/cosmo/blob/main/router/core/attribute_expressions.go#L58

router/core/graphql_prehandler.go
SkArchon: Good idea 👍

router/internal/expr/expr.go
SkArchon: I am not entirely sure about this part though, I can see cases where I have expressions where I log the query of the body, but I want to quickly toggle the parameter on and off in prod (to debug something). If there are lots of body queries it would be to easier disable it with a flag.

router/core/graphql_prehandler.go
SkArchon: Actually need to look a bit more, because this will involve adding this check to all expression uses.

router-tests/structured_logging_test.go
```diff
@@ -2237,6 +2237,189 @@ func TestFlakyAccessLogs(t *testing.T) {
 			})
 		})
 
+		t.Run("validate body is empty when enableBodyInExprContext is not set", func(t *testing.T) {
```
StarpTech, line 2240: Two points:

1. I think it is not intuitive when someone has to enable it. Did you consider using a function to make it lazy? `request.Body()`
2. The primary motivation of this feature was to provide a way to log the full body e.g. `string(request.Body())` I think the RFC is misleading here. We will provide the data already through `request.operation`. Exposing the same structure again does not have any benefit. We should handle  `Body` as an arbitrary bytes string.

My proposal is the following. We need to update the RFC.

1. A user can access the payload as follow `request.Body()` this is a getter to lazy load the full request buffer. We can have a limitReader that can be adjusted.
2. In order to get the string representation, a user can use it in a regular expression `request.Body()`
3. We will expose operation specific data as described [here](https://github.com/wundergraph/cosmo/pull/1481/files#diff-6e5f74edff90b87915462882ebd022cfd076afec3024dcd30dbd61549c16cc2bR90) in the future.

StarpTech (changes requested):

router-tests/structured_logging_test.go
devsergiy: What about multipart bodies? what should be available?

SkArchon: You will get this ```{\"query\":\"mutation($files: [Upload!]!) { multipleUpload(files: $files)}\",\"variables\":{\"files\":[null,null]}}```

router-tests/structured_logging_test.go
```diff
@@ -2237,6 +2237,172 @@ func TestFlakyAccessLogs(t *testing.T) {
 			})
 		})
 
+		t.Run("validate the evaluation of the body expression", func(t *testing.T) {
```
StarpTech, line 2240: Please add a test case that reflect that actual customer use case. We wanted to conditionally log the body in case of an error.

router/internal/expr/use_body.go
```diff
@@ -0,0 +1,43 @@
+package expr
```
StarpTech, line 1: Please add simple unit tests for it. Usage in function calls, arguments etc..

StarpTech (changes requested):

StarpTech (approved):
LGTM

PR #1666, opened 2025-03-07
feat: improvements to file watcher

endigma:
# Scope

- [x] Adds a new watcher implementation that does not rely on `inotify`, as it is tragically unreliable once users deploy the router in a remotely non-standard way (configured by kube configmaps, in podman machines on MacOS)
- [ ] Ports execution config hot reloading to the simpler implementation
- [ ] Implements hot reloading for the entire router config, including limited rollback/recover functionality if an invalid config is provided

### DIFF
--- a/router/pkg/watcher/simple_watcher.go
+++ b/router/pkg/watcher/simple_watcher.go
@@ -1,0 +1,49 @@
+package watcher
+
+import (
+	"context"
+	"os"
+	"time"
+
+	"go.uber.org/zap"
+)
+
+func SimpleWatch(ctx context.Context, logger *zap.Logger, interval time.Duration, path string, cb func()) error {
+	ticker := time.NewTicker(interval)
+	defer ticker.Stop()
+
+	var prevModTime time.Time
+
+	stat, err := os.Stat(path)
+	if err != nil {
+		logger.Debug("target file cannot be statted", zap.String("path", path), zap.Error(err))
+	} else {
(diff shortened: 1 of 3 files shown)


router/pkg/watcher/simple_watcher.go
```diff
@@ -0,0 +1,49 @@
...
+
+func SimpleWatch(ctx context.Context, logger *zap.Logger, interval time.Duration, path string, cb func()) error {
+	ticker := time.NewTicker(interval)
+	defer ticker.Stop()
+
+	var prevModTime time.Time
+
+	stat, err := os.Stat(path)
+	if err != nil {
+		logger.Debug("target file cannot be statted", zap.String("path", path), zap.Error(err))
```
StarpTech, line 19: Always uppercase

router/pkg/watcher/simple_watcher.go
```diff
@@ -0,0 +1,49 @@
...
+				// Reset the mod time so we catch any new file at the target path
+				prevModTime = time.Time{}
+
+				continue
+			}
+
+			logger.Debug("checking", zap.String("filename", path), zap.Time("prev", prevModTime), zap.Time("mod", stat.ModTime()))
+
+			if stat.ModTime().After(prevModTime) {
+				prevModTime = stat.ModTime()
```
StarpTech, line 42: No need to call it twice.

StarpTech (changes requested):

PR #1668, opened 2025-03-10
feat(controlplane): include base subgraph data in feature subgraph response

Noroth:
## Motivation and Context

The API does not yet return base subgraph information in the response when retrieving feature subgraphs. The protobuf definition already contains the fields but they were unused. 

The terraform provider for instance requires this information to make sure the state holds the correct information and matches the remote data.

## Checklist


controlplane/src/core/bufservices/subgraph/getSubgraphByName.ts
```diff
@@ -52,6 +52,8 @@ export function getSubgraphByName(
         namespace: subgraph.namespace,
         websocketSubprotocol: subgraph.websocketSubprotocol || '',
         isFeatureSubgraph: subgraph.isFeatureSubgraph,
+        baseSubgraphId: 'baseSubgraphId' in subgraph ? subgraph.baseSubgraphId : undefined,
+        baseSubgraphName: 'baseSubgraphName' in subgraph ? subgraph.baseSubgraphName : undefined,
```
Aenimus, line 56: Same as above but `subgraph.baseSubgraphName`

controlplane/src/core/bufservices/subgraph/getSubgraphById.ts
```diff
@@ -49,6 +49,8 @@ export function getSubgraphById(
         namespace: subgraph.namespace,
         websocketSubprotocol: subgraph.websocketSubprotocol || '',
         isFeatureSubgraph: subgraph.isFeatureSubgraph,
+        baseSubgraphId: 'baseSubgraphId' in subgraph ? subgraph.baseSubgraphId : undefined,
+        baseSubgraphName: 'baseSubgraphName' in subgraph ? subgraph.baseSubgraphName : undefined,
```
endigma, line 53: Not 100%, but I believe this is the equivalent of just `subgraph.baseSubgraphId`, if the field is not in `subgraph`, it is already undefined, no?

endigma:
You can rebase to fix controlplane CI

controlplane/src/core/bufservices/subgraph/getSubgraphById.ts
Noroth: We are expecting two different types for the subgraph. SubgraphDTO and FeatureSubgraphDTO. Both types contain the same fields except for the second type, which is a superset that also holds baseSubgraphId and baseSubgraphName. 

If you try to access those fields without casting or `in` you'll get an error

controlplane/src/core/repositories/SubgraphRepository.ts
```diff
@@ -733,7 +734,7 @@ export class SubgraphRepository {
     return subgraphs;
   }
 
-  private async getSubgraph(conditions: SQL<unknown>[]): Promise<SubgraphDTO | undefined> {
+  private async getSubgraph(conditions: SQL<unknown>[]): Promise<SubgraphDTO | FeatureSubgraphDTO | undefined> {
```
StarpTech, line 737: I'd suggest to split it into `getSubgraph` and `getFeatureSubgraph` repositories. This avoid fetching the feature subgraph for all calls and avoids that the logic becomes more complex.

StarpTech (changes requested):

StarpTech (approved):
LGTM

PR #1671, opened 2025-03-10
fix: polling based execution config watcher

endigma:
## Motivation and Context

- `inotify` and `fsnotify` are great, until various not-quite-standard filesystems and symlinks get thrown into the mix, then it doesn't seem so great anymore.
- The new system should be virtually edge-case free, and has a lot lower mental complexity, it just checks on a timer if the last modified date of the watched file has changed

## Theoretical limitations

- We trust the filesystem/OS to give us a useful and precise mod time value
- If you somehow circumvented the OS/fs and changed the file contents directly without changing mod time, we would not know


router/pkg/config/config.schema.json
```diff
@@ -264,6 +264,15 @@
                   "type": "boolean",
                   "default": false,
                   "description": "Enable the watch mode. The watch mode is used to watch the execution config file for changes. If the file changes, the router will reload the execution config without downtime."
+                },
+                "watch_interval": {
+                  "type": "string",
+                  "description": "The interval at which the file is checked for changes. The period is specified as a string with a number and a unit, e.g. 10ms, 1s, 1m, 1h. The supported units are 'ms', 's', 'm', 'h'.",
+                  "default": "1s",
+                  "duration": {
+                    "minimum": "100ms",
```
StarpTech, line 273: Because you said this is sensitive for the watcher to work, we should add a simple test.

StarpTech (changes requested):

router/pkg/config/config.schema.json
endigma: Added, and updated the schema to have no upper limit and a higher minimum (1s), and enforced `watch == true` if `watch_interval` is present at schema level, this should prevent mistakes where users set one and not the other.

router/pkg/watcher/simple_watcher.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+type SimpleWatcherOptions struct {
+	Interval time.Duration
+	Logger   *zap.Logger
+	Path     string
+	Callback func()
+}
+
+func LogSimpleWatch(ctx context.Context, options SimpleWatcherOptions) {
```
Noroth, line 18: as we basically require all options to be set, would it make sense to make them private and have a function `NewSimpleWatcherOptions()` with a bit of validation that returns an error instead of a panicking when calling Logger or Callback?

router/core/router.go
```diff
@@ -1116,45 +1117,36 @@ func (r *Router) Start(ctx context.Context) error {
 		}()
 
 		if r.executionConfig != nil && r.executionConfig.Watch {
+			go watcher.LogSimpleWatch(ctx, watcher.SimpleWatcherOptions{
```
Noroth, line 1120: The old watcher is unused now, correct? Any reason to keep unused code?

router/pkg/watcher/simple_watcher_test.go
```diff
@@ -0,0 +1,286 @@
...
+				Callback: func() {
+					wg.Done()
+				},
+			})
+		})
+
+		// Wait for the first cycle to complete to set baseline
+		time.Sleep(2 * watchInterval)
+
+		wg.Add(1)
```
Noroth, line 144: Just from a review perspective wg here looks a bit like a race condition. I personally would try to avoid time.Sleep. I suggest adding something to the waitgroup first and waiting 2 times or you could also work with channels and a select with something like `<-time.After`.

What do you think?

router/core/router.go
endigma: Posterity I guess, but its broken in other ways anyway. I'll remove it

router/pkg/watcher/simple_watcher.go
endigma: Maybe, I could also split it into like `NewSimpleWatcher()` and `func (*SimpleWatcher) Start(ctx)` or something.

router/pkg/watcher/simple_watcher_test.go
endigma: Can you clarify what you think looks like a race here? 

To clarify the control flow:

- We want to ensure the callback is run when the file changes
- We set up a waitgroup, and have the callback call `wg.Done()`
- We sleep a bit to let the first watcher cycle pass, otherwise it will not detect the change as the first pass result is "baseline"
- We delete the file, and wait for another cycle, deletions do not trigger reloads
- We write the file back, and then wait on the callback waitgroup to ensure it runs

WaitGroup is sound here because:

- If it runs more than once, `wg.Done()` will panic
- If it runs less than once, the test will time out and fail
- It it runs exactly once, the test will pass

Thoughts?

router/pkg/watcher/watcher.go
```diff
@@ -2,308 +2,63 @@ package watcher
...
 
-func (e Event) String() string {
-	return string(e.Op) + ":" + e.Path
-}
-
-func newEventSet() *eventSet {
-	return &eventSet{
-		events: map[string]Event{},
+func LogSimpleWatch(ctx context.Context, options SimpleWatcherOptions) {
+	if err := SimpleWatch(ctx, options); err != nil {
```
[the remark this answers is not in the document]
endigma: Like this? https://github.com/wundergraph/cosmo/blob/dabefaba23f0a6f7e14c9a0cf5381b9ce018a388/router/core/router.go#L1121-L1122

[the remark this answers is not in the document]
endigma: I could make it `("component", "execution config watcher")` ?

router/pkg/watcher/watcher.go
```diff
@@ -2,308 +2,63 @@ package watcher
...
-func (w *Watcher) Wait() error {
-	if err := w.errGroup.Wait(); err != nil {
-		if !errors.Is(err, ErrStop) {
-			return err
-		}
-		return nil
-	}
-	return nil
-}
+	ll.Debug("Watching file for changes", zap.Time("initialModTime", prevModTime))
```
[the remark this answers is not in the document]
endigma: fixed, another nice thing to add to a theoretical zap usage linter

router/pkg/watcher/watcher.go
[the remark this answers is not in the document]
StarpTech: I was thinking of `("component", "config_watcher", "watcher_name": "execution_config")`

router/pkg/watcher/simple_watcher.go
endigma: It has error handling and a sort of "init" phase now, with minimal bloat to the API

router/pkg/watcher/watcher.go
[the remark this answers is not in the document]
endigma: Done, what do you think of possibly passing `ctx/ll` to callbacks so they can inherit the attributes?

StarpTech (approved):
LGTM

PR #1680, opened 2025-03-11
fix: improve performance when updating member role

wilsonrivera:
## Motivation and Context

When updating member roles the operation could take anywhere from a few seconds to over a minute (eventually failing).

## Checklist

Both operations were executed with 7400 organizations.

Before:
before

After:
after

## Analysis

The `organizations` are created as groups inside `Keycloak` and the roles are subgroups of the organizations.

To find the identifier of the role in `Keycloak` we were searching by the role name, for example `admin`, this causes `Keycloak` to return all the organizations as the `admin` role should exist in all of them. This issue scales with the amount of organizations created in `Keycloak`.

Something that added to the issue is that we were performing this operation 3 times in a row to load the relevant organization roles.

## Solution

The solution is to use the organization identifier (`kcGroupId`) to list the subgroups (roles) of the organization and filter them by name.

Additionally, to avoid performing this operation multiple times in a row just with different role, a new method was introduced to load all organization roles at once and filter in memory.

### DIFF
--- a/.github/scripts/setup-keycloak.sh
+++ b/.github/scripts/setup-keycloak.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-export KC_VERSION=22.0.4
+export KC_VERSION=25.0.1
 curl -LO  https://github.com/keycloak/keycloak/releases/download/"${KC_VERSION}"/keycloak-"${KC_VERSION}".zip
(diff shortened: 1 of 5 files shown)


StarpTech (approved):
LGTM

PR #1686, opened 2025-03-13
fix(router): Example module missing registration. 

darren-west:
## Motivation and Context

The missing call to register means the module is never used and it confused me for a minute when trying to develop a custom module.


StarpTech:
Hi @darren-west thanks for the PR. However, this is not mergeable. We use the module in the test suite. This must have been an issue. You can duplicate the module and put it in `router-tests/modules` and remove the references from the module in the tests suite. After this we can merge it.

StarpTech:
@darren-west friendly ping. Would be great, if you could address it.

darren-west:
> @darren-west friendly ping. Would be great, if you could address it.

Hi @StarpTech sorry I missed your previous comment. I have now addressed this and pushed.

StarpTech:
Closed due to inactivtiy.

PR #1688, opened 2025-03-13
feat: implement checks configuration

wilsonrivera:
## Motivation and Context

As an organization, I would like to be able to configure the days considered when when performing subgraph checks by namespace.

## Checklist

https://github.com/user-attachments/assets/6b49a81e-ca6c-4797-bcdc-26fa05475591


controlplane/src/core/bufservices/subgraph/checkSubgraphSchema.ts
```diff
@@ -295,7 +301,8 @@ export function checkSubgraphSchema(
       featureId: 'breaking-change-retention',
     });
 
-    const limit = changeRetention?.limit ?? 7;
+    let limit = changeRetention?.limit ?? 7;
+    limit = clamp(namespace?.checksTimeframeInDays ?? limit, 1, limit);
```
JivusAyrus, line 305: If it's configured, use the configured value or else use the org limit

controlplane/migrations/0114_bright_carmella_unuscione.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE "namespaces" ADD COLUMN "checks_timeframe_in_days" integer;
```
JivusAyrus, line 1: Let's add a new table called "namespace_config" and you can add this column to the new table.

studio/src/components/checks/checks-config.tsx
```diff
@@ -0,0 +1,127 @@
...
+        className="flex flex-col gap-y-4"
+      >
+        <FormField
+          control={form.control}
+          name="timeframeInDays"
+          render={({ field }) => (
+            <FormItem>
+              <FormLabel>Days to consider</FormLabel>
+              <FormControl>
+                <Input
```
JivusAyrus, line 92: To maintain consistency should we go with a dropdown instead of a input?

JivusAyrus (changes requested):

controlplane/migrations/0114_bright_carmella_unuscione.sql
JivusAyrus: We can even migrate the boolean values in the namespaces table to the new table

JivusAyrus:
Should we add this config to the lint policy page itself and rename it to Check Configurations?

studio/src/pages/[organizationSlug]/checks-config.tsx
```diff
@@ -0,0 +1,58 @@
+import { NamespaceSelector } from "@/components/dashboard/NamespaceSelector";
```
Comment (JivusAyrus) [inline studio/src/pages/[organizationSlug]/checks-config.tsx:1]: Should we add this config to the lint policy page itself and rename it to Check Configurations?

JivusAyrus (changes requested):

studio/src/components/checks/checks-config.tsx
wilsonrivera: As in a set number of options? Lets say
- 1 day
- 3 days
- 1 week
- 2 weeks

and so on?

JivusAyrus: yep, we can do 
- 3 days
- 7 days
- 14 days
- 30 days ...

controlplane/src/core/bufservices/subgraph/checkSubgraphSchema.ts
wilsonrivera: Could there not be cases where the organization limit is changed?

controlplane/migrations/0114_bright_carmella_unuscione.sql
wilsonrivera: My concern with this approach is that users that have already modified these options will have to do it again (even if just a simple boolean).

controlplane/src/core/bufservices/subgraph/checkSubgraphSchema.ts
JivusAyrus: But that shouldn't cause a problem right?

controlplane/migrations/0114_bright_carmella_unuscione.sql
JivusAyrus: We can migrate those values to the new table

studio/src/pages/[organizationSlug]/checks-config.tsx
Comment (JivusAyrus) [reply studio/src/pages/[organizationSlug]/checks-config.tsx:1]: I have mixed feelings for this. Leaning more towards isolating it for checks. You can once discuss with dustin.

controlplane/src/core/bufservices/subgraph/checkSubgraphSchema.ts
wilsonrivera: It shouldn't, the only thing would be that if they downgrade, this configuration would override the new plan

controlplane/migrations/0114_puzzling_talos.sql
```diff
@@ -0,0 +1,14 @@
+CREATE TABLE IF NOT EXISTS "namespace_config" (
```
JivusAyrus, line 1: We should migrate the data from the old table right?

JivusAyrus (changes requested):

controlplane/migrations/0114_puzzling_talos.sql
wilsonrivera: We should, however, the way I've implemented this is, if the column for the new table doesn't have a value (is `null`), it will fallback to the value from the `namespaces` table, which we already have in memory.

Then when the customer updates any value, an `upsert` is performed with the `namespace id` as key.

JivusAyrus: I was thinking of deleting those values from the old table after the migration is done. What do you think?

wilsonrivera: I have reverted the migration to do copy the values to the new table and remove the old columns in a single migration instead of introducing a second one just to remove the columns

JivusAyrus (approved):

studio/src/components/checks/graph-pruning-issues-table.tsx
```diff
@@ -43,7 +43,7 @@
           <Button
             onClick={() => {
               router.push(
-                `/${router.query.organizationSlug}/lint-policy?namespace=${router.query.namespace}`,
+                `/${router.query.organizationSlug}/policies?namespace=${router.query.namespace}`,
```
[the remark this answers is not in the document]
JivusAyrus: This can be fixed by using user.currentOrganization.slug(useUser) and graphData.graph.namespace from GraphContext.

PR #1693, opened 2025-03-14
fix(websocket): improve write and read deadlines

StarpTech:
## Motivation and Context

- Use proper read and write deadlines when writing to the WebSocket connections to prevent staleness. Introducing `websocket_client_write_timeout`
- Stop using the WebSocket stats worker in production because it can slow down or even block event processing.
- Stop using test containers in tests. I often run into port and connectivity issues. We now always point to the local bootstrapped container or, in CI, to the Action Service.

Related Engine PR: https://github.com/wundergraph/graphql-go-tools/pull/1113

## Checklist


router/internal/wsproto/subscriptions_transport_ws.go
```diff
@@ -112,6 +129,9 @@ func (p *subscriptionsTransportWSProtocol) WriteGraphQLErrors(id string, errors
 	if err != nil {
 		return fmt.Errorf("encoding JSON: %w", err)
 	}
+	if err := p.conn.SetWriteDeadline(time.Now().Add(p.writeTimeout)); err != nil {
```
jensneuse, line 132: Does it make sense to move SetWriteDeadline & SetReadDeadline into conn.WriteJSON & conn.ReadJSON?

jensneuse (changes requested):

router/internal/wsproto/subscriptions_transport_ws.go
StarpTech: I think so yes. I'll give it a shot.

jensneuse (approved):

PR #1704, opened 2025-03-18
feat(router): schema usage lite export via prometheus

endigma:
## Motivation and Context

Some customers would like to use Prometheus for schema usage metrics, instead of the stack provided by Cosmo.

## Changes

- Added a new feature to export schema usage to prometheus
- New configuration section `/telemetry/metrics/prometheus/schema_usage`
- Option within the new object `include_operation_sha` controls whether or not the metrics include an attribute containing the sha256 of the operation, this can be disabled to reduce cardinality
    - This doesn't use the normal labels exclusion mechanism because computing the sha256 is not always required
    - Disabling here _and_ excluding the label has no extra effect, and disabling it via the label exclusion does work, but if this option is `true` the sha256 will unconditionally be computed
- `debug.config.yaml` now sets 
    - `log_level: debug`
    - `telemetry.prometheus.schema_usage.enabled: true`
    - `telemetry.prometheus.schema_usage.include_operation_sha: true`

## Config example

```yaml
telemetry:
  metrics:
    prometheus:
      schema_usage:
        enabled: true
        include_operation_sha: true

```


router-tests/prometheus_improved_test.go
```diff
@@ -0,0 +1,207 @@
...
+// asserting the values of unrelated metrics/labels.
+
+const (
+	SchemaFieldUsageMetricName = "router_graphql_schema_field_usage_total"
+
+	WgOperationSha256  = "wg_operation_sha256"
+	WgGraphQLFieldName = "wg_graphql_field_name"
+	WgGraphQLFieldType = "wg_graphql_field_type"
+	WgOperationName    = "wg_operation_name"
+	WgOperationType    = "wg_operation_type"
```
StarpTech, line 27: You can reference to the original one and use a helper function to make it snake case. In that case, we don't need to maintain two places.

StarpTech (changes requested):

router-tests/prometheus_improved_test.go
endigma: Done

StarpTech (approved):
LGTM

StarpTech (approved):

router/core/graph_server.go
```diff
@@ -460,11 +460,22 @@ func (s *graphMux) buildOperationCaches(srv *graphServer) (computeSha256 bool, e
 			}
 		}
 	} else if srv.persistedOperationsConfig.Safelist.Enabled || srv.persistedOperationsConfig.LogUnknown {
+		// THIS IS A BUG!! Tests in router-tests/safelist_test.go rely on the buggy functionality
```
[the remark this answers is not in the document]
endigma: Yes

PR #1713, opened 2025-03-20
fix: ensure to close any responses before retrying

SkArchon:
## Motivation and Context

Right now it seems that we do not close the response body before retrying, this causes the response to be nil for the next retry. However if we close it, the response from the next retry is not nil then. W
e now make sure to read the entire stream and close it before retrying. Golang code itself mentions `It is the caller's responsibility to close Body.`

## Checklist


router/internal/retrytransport/retry_transport.go
```diff
@@ -101,6 +104,20 @@ func (rt *RetryHTTPTransport) RoundTrip(req *http.Request) (*http.Response, erro
 	return resp, err
 }
 
+func (rt *RetryHTTPTransport) drainBody(resp *http.Response) {
+	if resp == nil || resp.Body == nil {
+		return
+	}
+	_, err := io.Copy(io.Discard, resp.Body)
```
Noroth, line 111: Do we need the extra step to discard the content? Shouldn't close be sufficient?

router/internal/retrytransport/retry_transport.go
```diff
@@ -101,6 +104,20 @@ func (rt *RetryHTTPTransport) RoundTrip(req *http.Request) (*http.Response, erro
...
+	if err != nil {
+		rt.Logger.Error("Failed draining when copying the body", zap.Error(err))
+	}
+	err = resp.Body.Close()
```
Noroth, line 115: We probably want to do something like defer body.Close() at the beginning of the function to make sure we always at least attempt to close the body, right?

SkArchon: I didn't use defer because we don't early return on errors. But nothing against moving to defer, will change.

router/internal/retrytransport/retry_transport.go
SkArchon: Yes, from my testing closing the body didn't make go mark `reusable=true`, but reading the entire body, even without closing did.

SkArchon:
When we close the body the following function in `http/transport` is called, so if we DON'T call drain and close the body `pc.reused` will be `false`

```
// markReused marks this connection as having been successfully used for a
// request and response.
func (pc *persistConn) markReused() {
	pc.mu.Lock()
	pc.reused = true
	pc.mu.Unlock()
}
```

Which means this affects the following on a retry RoundTrip call

```
if !pc.isReused() {
        // This was a fresh connection. There's no reason the server
	// should've hung up on us.
	//
	// Also, if we retried now, we could loop forever
	// creating new connections and retrying if the server
	// is just hanging up on us because it doesn't like
	// our request (as opposed to sending an error).
	return false
}
if _, ok := err.(nothingWrittenError); ok {
	// We never wrote anything, so it's safe to retry, if there's no body or we
	// can "rewind" the body with GetBody.
	return req.outgoingLength() == 0 || req.GetBody != nil
}
```

For retries the code after the `pc.isReused()` check is called, notably in the case of my testing `if _, ok := err.(nothingWrittenError); ok {` needs to be called to successfully retry, this is internally this error is thrown when we try to retry since the body is already read, then the retry *seems* to be retried when this error is present and the body is `rewindable`, which is the case since from my testing it used `bytes.Buffer` for the body.

router/internal/retrytransport/retry_transport.go
```diff
@@ -101,6 +104,26 @@ func (rt *RetryHTTPTransport) RoundTrip(req *http.Request) (*http.Response, erro
...
+		if err != nil {
+			rt.Logger.Error("Failed draining when closing the body", zap.Error(err))
+		}
+	}()
+
+	// When we close the body only will go internally marks the persisted connection as true
+	// which is important so that it can reuse the connection internally for retrying
+	_, err := io.Copy(io.Discard, resp.Body)
+	if err != nil {
+		rt.Logger.Error("Failed draining when copying the body", zap.Error(err))
```
Noroth, line 123: nit 
```suggestion
		rt.Logger.Error("Failed draining when discarding the body", zap.Error(err))
```

StarpTech (changes requested):
Please add additional tests for retries as discussed.

router/internal/retrytransport/retry_transport.go
```diff
@@ -101,6 +104,26 @@ func (rt *RetryHTTPTransport) RoundTrip(req *http.Request) (*http.Response, erro
...
+		if err != nil {
+			rt.Logger.Error("Failed draining when closing the body", zap.Error(err))
+		}
+	}()
+
+	// When we close the body only will go internally marks the persisted connection as true
+	// which is important so that it can reuse the connection internally for retrying
+	_, err := io.Copy(io.Discard, resp.Body)
+	if err != nil {
+		rt.Logger.Error("Failed draining when discarding the body", zap.Error(err))
```
StarpTech, line 123: This must be the request logger.

router/internal/retrytransport/retry_transport.go
```diff
@@ -101,6 +104,26 @@ func (rt *RetryHTTPTransport) RoundTrip(req *http.Request) (*http.Response, erro
...
 
+func (rt *RetryHTTPTransport) drainBody(resp *http.Response) {
+	if resp == nil || resp.Body == nil {
+		return
+	}
+
+	defer func() {
+		err := resp.Body.Close()
+		if err != nil {
+			rt.Logger.Error("Failed draining when closing the body", zap.Error(err))
```
StarpTech, line 115: Here too.

StarpTech (changes requested):

StarpTech (approved):
LGTM

StarpTech (approved):
LGTM

PR #1716, opened 2025-03-24
feat: allow to filter checks by subgraph name

wilsonrivera:
## Motivation and Context

As a user, I'd like to filter checks by one or more subgraphs on top of the date range and for it to be possible to share the link with other users.

## Checklist

The subgraph identifier gets added to a comma-separated query parameter which then is passed to ControlPlane. As validation, ControlPlane takes only UUID values (as preventive measures to the client passing invalid data) and ensures that the passed UUIDs belong to the organization and selected graph.

https://github.com/user-attachments/assets/d621a830-0cf3-4ed8-b35c-245ae28ae28f


proto/wg/cosmo/platform/v1/platform.proto
```diff
@@ -441,6 +441,7 @@ message GetChecksByFederatedGraphNameRequest {
   string startDate = 4;
   string endDate = 5;
   string namespace = 6;
+  repeated string subgraphs = 7;
```
StarpTech, line 444: It is not clear to me that we have applied a filter. Could you come up with a better name? e..g `include_subgraphs`

controlplane/src/core/repositories/SubgraphRepository.ts
```diff
@@ -700,14 +705,22 @@ export class SubgraphRepository {
       return [];
     }
 
+    const subgraphsToSelect =
+      Array.isArray(data.subgraphs) && data.subgraphs.length > 0 ? data.subgraphs.filter((v) => isValidUuid(v)) : [];
```
StarpTech, line 709: If the ids are valid should be done in the controller, so we can return early.

StarpTech (changes requested):

StarpTech (approved):
LGTM

StarpTech (approved):
LGTM

PR #1722, opened 2025-03-25
fix: improve url normalization

wilsonrivera:
## Motivation and Context

Improve the normalization of URLs. 

## Checklist

Previously, providing an URL without the protocol (e.g.: localhost:3000) would take the hostname as the protocol.

The fix consists of assuming the protocol is 'http' and having a flag to not include the protocol in the resulting normalized URL.


shared/test/utils.test.ts
```diff
@@ -19,6 +19,22 @@ test('Normalize urls', () => {
       input: 'https://example.com/a/b/c/d?query=123#fragment',
       expected: 'https://example.com/a/b/c/d',
     },
+    {
+      input: 'localhost:3000',
+      expected: 'localhost:3000',
+    },
+    {
+      input: '//localhost:3000',
```
StarpTech, line 27: Is this a valid url? The goal was not to make the url parsing resilient against any failure. We should still error when the url is not valid but handle localhost as a special case. I'd propose

```ts
/**
 * Normalize the URL by removing the trailing slash, fragments and query parameters.
 * Only the protocol, hostname, port and path are preserved.
 * If no protocol is provided, defaults to http.
 * @param url
 */
export function normalizeURL(url: string): string {
  // return empty
  if (!url) {
    return url;
  }

  // Add https:// protocol if not present
  const urlWithProtocol = url.includes('://') ? url : `https://${url}`;
  const parsedUrl = new URL(urlWithProtocol);
  let path = parsedUrl.pathname;

  // Remove the trailing slash if present
  if (path.endsWith('/')) {
    path = path.slice(0, -1);
  }

  const port = parsedUrl.port ? `:${parsedUrl.port}` : '';
  return `${parsedUrl.protocol}//${parsedUrl.hostname}${port}${path}`;
}
```

shared/test/utils.test.ts
```diff
@@ -19,6 +19,22 @@ test('Normalize urls', () => {
...
+      expected: 'localhost:3000',
+    },
+    {
+      input: 'http://example.com',
+      expected: 'http://example.com',
+    },
+    {
+      input: 'telnet://192.0.2.16:80/',
```
StarpTech, line 35: Path should no be discarded. Proxies can be sensitive to that. Not sure why we did this in the first place.

StarpTech (changes requested):

shared/test/utils.test.ts
wilsonrivera: With this approach, we'll always add a slash at the end of the path (parsing the url converts the empty path to `/`)

I propose something like this to remove the last slash if isn't needed
```
if (
    path.endsWith('/') &&
    !url.endsWith(`/${parsedUrl.search}${parsedUrl.hash}`) &&
    !url.endsWith(`/?${parsedUrl.hash}`)
  ) {
    path = path.slice(0, -1);
  }
```

StarpTech: Sounds good

shared/src/utils/util.ts
```diff
@@ -21,6 +21,8 @@ export function joinLabel({ key, value }: { key: string; value: string }) {
 /**
  * Normalize the URL by removing the trailing slash, fragments and query parameters.
```
StarpTech, line 22: This is no longer true.

shared/src/utils/util.ts
```diff
@@ -29,11 +31,15 @@ export function normalizeURL(url: string): string {
     return url;
   }
 
-  const parsedUrl = new URL(url);
-  let path = parsedUrl.pathname;
+  const urlToParse = url.includes('://') ? url : `https://${url}`;
```
StarpTech, line 34: I think for localhost, we should make an exception and fallback to http:// instead of https://. This is also how the browser behaves and more convenient when working locally.

https://github.com/nodejs/node/issues/13708#issuecomment-309039107

StarpTech (changes requested):
LGTM

shared/src/utils/util.ts
StarpTech: Basically, we only need to remove query and fragment information.

StarpTech: Example

```ts
function normalizeUrl(url: string): string {
  try {
    // Parse the URL using the built-in URL API.
    const parsedUrl = new URL(url);
    
    // Remove the query parameters and fragment.
    parsedUrl.search = "";
    parsedUrl.hash = "";
    
    // Return the normalized URL as a string.
    return parsedUrl.toString();
  } catch (error) {
    // Handle the case where the input URL is invalid.
    throw new Error("Invalid URL provided.");
  }
}

// Example usage:
const originalUrl = "localhost:3000/path?query=123#section";
const normalizedUrl = normalizeUrl(originalUrl);
console.log(normalizedUrl); // Output: "localhost:3000/path"

```

wilsonrivera: Unlike Go, JavaScript always adds a final slash when the path is empty (https://developer.mozilla.org/en-US/docs/Web/API/URL/pathname#pathname_with_invisible_segment). 

What we can do is have the same condition as before for the path before clearing the `search` and `hash` fields:
```javascript
...
  let removeFinalSlash =
      parsedUrl.pathname.endsWith('/') &&
      !url.endsWith(`/${parsedUrl.search}${parsedUrl.hash}`) &&
      !url.endsWith(`/?${parsedUrl.hash}`);

    parsedUrl.search = '';
    parsedUrl.hash = '';

    let result = parsedUrl.toString();
    if (removeFinalSlash) {
      result = result.slice(0, -1);
    }

    return result;
...
```

This way, when provided with `localhost:3000` we return the same value instead of `localhost:3000/`. Or would it be fine if the slash is added ?

PR #1723, opened 2025-03-25
fix: catch an error on provides with fragments

devsergiy:
### Bug Fixes

* catch an error on provides with fragments ([#1115](https://github.com/wundergraph/graphql-go-tools/issues/1115)) ([f4bb0af](https://github.com/wundergraph/graphql-go-tools/commit/f4bb0afa4b0124eff279635f7a0dc7d65d2e8554))


Aenimus (approved):
LGTM

Issue #1725, opened 2025-03-26
race condition in router shutdown function


yvz5:
### Component(s)

router

### Component version

code from master

### wgc version

--

### controlplane version

--

### router version

latest code

### What happened?

In the `func (r *Router) Shutdown(ctx context.Context) (err error)`  of the file `router\core\router.go` there are multiple goroutines trying to read / write to a shared variable `err` without being thread safe, this would cause a race condition. There are no tests for graceful shutdown too, `-race` flag would catch this.

### Router configuration

```yaml

```

### Router execution config

```json

```

### Log output

```shell

```

StarpTech:
This has been recently addressed. Please retest.

PR #1726, opened 2025-03-26
fix(router): shutdown race condition when writing to a shared err variable

yvz5:
## Motivation and Context

fixes https://github.com/wundergraph/cosmo/issues/1725


StarpTech (approved):
LGTM

StarpTech:
@yvz5 thanks for the fix!

Pavan-Chaithanya:
@StarpTech Tried to take a stab at this issue here: https://github.com/wundergraph/cosmo/pull/1736
Let me know your thoughts

endigma:
Hiya @yvz5, first of all, thanks for your contribution! As for this PR, we will be closing it as the issue has been fixed as part of #1773, which also improves many other parts of the shutdown process, including new tests to cover this functionality.

Thanks!

PR #1736, opened 2025-03-30
fix(router): race condition due to error in shutdown

Pavan-Chaithanya:
## Motivation and Context
Fixes https://github.com/wundergraph/cosmo/issues/1725
Fixed the race condition of Shutdown function. Refactored this function a bit so it is easier to work with.

## Checklist

More to discuss:
I haven't looked into the code enough. Is there any reason why the config poller and http server are not shut down in the goroutine as well?


StarpTech:
Hi @Pavan-Chaithanya, thanks for the PR. However, this issue has already been addressed in https://github.com/wundergraph/cosmo/pull/1726/files. I reviewed both solutions, and yours seems unnecessarily complex. There's no need for inline structs or functions in this case.´

PR #1738, opened 2025-03-31
fix: subgraph error code is not propagated AND panics on websocket request logging

SkArchon:
## Motivation and Context

The subgraph error codes are not propagated when we fail the validation, because they are only propagated right now when there are errors after a subgraph call. This will add any error codes from the validation step to the requestContext's errorCodes.

This fix will allow errorCodes from the validation step to be logged.

This PR also fixes nil pointer panics on the request logger for the start of a web socket request. See below for screenshot of websocket request.

image

## Checklist


StarpTech (approved):
LGTM

PR #1741, opened 2025-03-31
feat: improve delete organization workflow

wilsonrivera:
## Motivation and Context

Improve the organization deletion flow.

The previous flow allowed an organization to be deleted synchronously with an active subscription. The new flow prevents triggering an organization deletion when an active subscription is present, only after the organization subscription have been cancelled, the organization deletion is queued for 3 days, giving operators chance to "restore" the organization.

## Checklist

When the organization have an active subscription, the user won't be able to queue deletion for the organization until the subscription is cancelled.

active-subscription

The body for the deletion dialog have been updated:

delete-dialog

Once the deletion have been queued, all administrators (including the one who queued the deletion) will receive an email informing them about the operation and provide a way to revert if it not intended.
email-body

Once the organization is queued for deletion, non-admin users will be kicked out of the organization pages and admins will see the following banner on all the pages for the organization:

queued-banner

Finally, the `delete organization` card on the settings page is replaced with `restore organization`:

restore
restore-dialog

The PR also introduces a new way to write emails templates. Previously, the template was plain HTML, now we can use React to write the templates using `react-email` to transform to HTML. The way it works is, there is a new subpackage inside `controlplane`, called `emails`.

To modify or transform the templates to HTML, you would run the following command:

```bash
cd controlplane/emails
pnpm build
```

You can also preview the templates by running:

```bash
cd controlplane/emails
pnpm dev
```


controlplane/emails/package.json
```diff
@@ -0,0 +1,21 @@
...
+  "description": "",
+  "main": "index.js",
+  "type": "module",
+  "scripts": {
+    "dev": "email dev --dir src --port 3400",
+    "build": "email export --dir src --outDir ../src/templates/emails --pretty && pnpm run --filter ../ lint:fix"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
```
StarpTech, line 13: Should be apache as the repo.

controlplane/emails/package.json
```diff
@@ -0,0 +1,21 @@
...
+  "devDependencies": {
+    "@react-email/components": "0.0.34",
+    "@types/react": "18.3.3",
+    "@types/react-dom": "18.3.0",
+    "react": "18.3.1",
+    "react-email": "^4.0.3"
```
StarpTech, line 19: Stick version too?

controlplane/src/core/bufservices/organization/restoreOrganization.ts
```diff
@@ -0,0 +1,85 @@
...
+      return {
+        response: {
+          code: EnumStatusCode.ERR,
+          details: 'User is not a part of this organization.',
+        },
+      };
+    }
+
+    // Ensure that the user is an admin of the organization
+    if (!orgMember.roles.includes('admin')) {
```
StarpTech, line 52: This information is already provided on the authContext.

controlplane/src/core/repositories/OrganizationRepository.ts
```diff
@@ -852,6 +875,43 @@ export class OrganizationRepository {
     return result[0];
   }
 
+  public async queueOrganizationDeletion(input: {
+    organizationId: string;
+    queuedBy?: string;
+    deleteOrganizationQueue: DeleteOrganizationQueue;
+  }) {
+    const now = new Date();
```
StarpTech, line 883: We should make this whole thing a migration so in case of the job cant be scheduled that the tx is rollbacked.

StarpTech (changes requested):

StarpTech (approved):
LGTM

PR #1742, opened 2025-03-31
chore: add rfc for custom response rendering

jensneuse:


rfc/custom-response-rendering.md
```diff
@@ -0,0 +1,241 @@
...
+}
+
+// Register the renderer
+renderer := &CustomPrivacyRenderer{
+    config: map[string]interface{}{
+        "types": []string{"User"},
+        "fields": []string{"email"},
+    },
+}
+router.RegisterRenderer(renderer)
```
maxkomarychev, line 162: Can we have multiple renderers? Does order matter? Can we register "wildcard" or "fallback" rendered which will be called if no other rendered has matched?

rfc/custom-response-rendering.md
```diff
@@ -0,0 +1,241 @@
...
+
+## Example Usage
+
+```go
+type CustomPrivacyRenderer struct {
+    // Configuration
+    config map[string]interface{}
+}
+
+func (r *CustomPrivacyRenderer) Render(ctx *core.RequestContext, value ValueContext, writer io.Writer) (int, error) {
```
maxkomarychev, line 138: performance concern: can we enable/disable custom renders entirely per `RequestContext` to avoid evaluation per individual `ValueContext`? i.e. introduce another callback: which will control whether plugin should be called at all for the entire request.

PR #1744, opened 2025-04-01
feat: add support to set ClientInfo from initial payload when using WebSockets

jensneuse:
This PR adds support for setting ClientInfo from the initial payload when using WebSockets in the Cosmo Router. The main changes allow configuring how the router handles client information in WebSocket connections.

## Key Changes

- Added support for extracting ClientInfo from WebSocket initial payload
- Introduced new configuration options for WebSocket client info handling
- Added flexibility in how client information is processed

## Example Config

```yaml
websocket:
  client_info_from_initial_payload:
    # Enable extracting client info from initial payload
    enabled: true
    # Field name in payload for client name
    name_field: "graphql-client-name"
    # Field name in payload for client version
    version_field: "graphql-client-version"
    # Forward client info to request headers
    forward_to_request_headers:
      # Enable forwarding client info to request header
      # Allowing it to be forwarded to Subgraph requests
      enabled: true
      # target Header name for client name
      name_target_header: "graphql-client-name"
      # target Header name for client version
      version_target_header: "graphql-client-version"
```


router/pkg/config/config.go
```diff
@@ -577,6 +577,19 @@ type WebSocketConfiguration struct {
...
+	SetClientInfoFromInitialPayload WebSocketSetClientInfoFromInitialPayloadConfiguration `yaml:"set_client_info_from_initial_payload"`
+}
+
+type WebSocketSetClientInfoFromInitialPayloadConfiguration struct {
+	// Enabled true if the Router should set the client info from the initial payload of a Subscription Request to the Subgraph
+	Enabled bool `yaml:"enabled" envDefault:"true" env:"WEBSOCKETS_SET_CLIENT_INFO_FROM_INITIAL_PAYLOAD_ENABLED"`
+	// ExportToClientRequestHeaders true if the Router should export the client info to the client request headers
+	ExportToClientRequestHeaders bool `yaml:"export_to_client_request_headers" envDefault:"true" env:"WEBSOCKETS_SET_CLIENT_INFO_FROM_INITIAL_PAYLOAD_EXPORT_TO_CLIENT_REQUEST_HEADERS"`
+	// ClientNameHeaderName is the name of the header where the client name should be exported to
+	ClientNameHeaderName string `yaml:"client_name_header_name" envDefault:"graphql-client-name" env:"WEBSOCKETS_SET_CLIENT_INFO_FROM_INITIAL_PAYLOAD_CLIENT_NAME_HEADER_NAME"`
```
StarpTech, line 590: I suggest to make it an array because in the wild we already support apollo and `graphql-client-name` convention

router/pkg/config/config.go
```diff
@@ -577,6 +577,19 @@ type WebSocketConfiguration struct {
 	ForwardInitialPayload bool `yaml:"forward_initial_payload" envDefault:"true" env:"WEBSOCKETS_FORWARD_INITIAL_PAYLOAD"`
 	// Authentication configuration for the WebSocket Connection
 	Authentication WebSocketAuthenticationConfiguration `yaml:"authentication,omitempty"`
+	// SetClientInfoFromInitialPayload configuration for the WebSocket Connection
+	SetClientInfoFromInitialPayload WebSocketSetClientInfoFromInitialPayloadConfiguration `yaml:"set_client_info_from_initial_payload"`
+}
+
+type WebSocketSetClientInfoFromInitialPayloadConfiguration struct {
+	// Enabled true if the Router should set the client info from the initial payload of a Subscription Request to the Subgraph
+	Enabled bool `yaml:"enabled" envDefault:"true" env:"WEBSOCKETS_SET_CLIENT_INFO_FROM_INITIAL_PAYLOAD_ENABLED"`
```
StarpTech, line 586: I'd recommend `WEBSOCKETS_CLIENT_INFO_FROM_INITIAL_PAYLOAD_ENABLED` rather than using verbs in the naming.

router/pkg/config/config.go
```diff
@@ -577,6 +577,19 @@ type WebSocketConfiguration struct {
...
+	// ExportToClientRequestHeaders true if the Router should export the client info to the client request headers
+	ExportToClientRequestHeaders bool `yaml:"export_to_client_request_headers" envDefault:"true" env:"WEBSOCKETS_SET_CLIENT_INFO_FROM_INITIAL_PAYLOAD_EXPORT_TO_CLIENT_REQUEST_HEADERS"`
```
StarpTech, line 588: Forward might be more meaningful here. We are also already in the context. We can be less verbose with the naming.

```
ForwardHeaders
```

router/core/websocket.go
```diff
@@ -1074,6 +1078,36 @@ func (h *WebSocketConnectionHandler) Initialize() (err error) {
...
+	if h.setClientInfoFromInitialPayload.Enabled && h.initialPayload != nil {
+		var initialPayloadMap map[string]interface{}
+		err := json.Unmarshal(h.initialPayload, &initialPayloadMap)
+		if err != nil {
+			h.logger.Warn("Error parsing initial payload for client info", zap.Error(err))
+			return err
+		}
+
+		// Update client name if present
+		if clientName, ok := initialPayloadMap["graphql-client-name"].(string); ok {
```
StarpTech, line 1092: Why is this hard coded?

StarpTech (changes requested):

router/pkg/config/config.go
jensneuse: We export the value into a header, that makes no sense. 😅

router/core/websocket.go
jensneuse: We set a convention.

router/pkg/config/config.go
StarpTech: I took me a moment to understand you abstraction. Why do we want to make this fix? Let the user define which property will be used as value.

```yaml
enabled: true
source_field: "graphqlClientName"
target_header: "my-client-header"
```

`ExportToClientRequestHeaders` is not needed, you can check for existence of `target_header`

StarpTech (approved):
LGTM

PR #1746, opened 2025-04-01
feat(router): hot config reloading

endigma:
## Motivation and Context

As a user who deploys the router with a dynamic config from the filesystem, I'd like it such that when the file changes on disk, there is the option to have the router automatically reload itself when it detects this change.

## Changes

- Better responsibility boundary between `cmd/router/main.go` and `router/pkg/config/config.go`
    - `config.LoadConfig` no longer loads dotenv into the environment or decides the config path under any circumstance
    - These responsibilities fall to `main.Main` now
- The log level is now atomic and mutable at runtime if you hold a reference
- The router is now capable of executing a full reload when either the config file changes or SIGHUP signal is sent to the process
- Moved `randomDuration` utility out of `controlplane` package into `internal/timex` for reuse
- Added configuration for watching the config file and triggering a reload when a change is detected (example below)

## Config responsibility change

As a side effect of drawing a clearer boundary between the responsibilities of `main.go` and `config.go` for determining the config file to read, direct calls to `LoadConfig` will no longer check `CONFIG_FILE` environment variable. This is intentional, as this responsibility now falls to the caller.

`LoadConfig` continues to default to `config.DefaultConfigPath` when the provided path is `""`.

## Limitations

The config path loading and the file watcher are outside the reload loop, so:

- Configuration hot-reloading will not detect changes to the config path
- Configuration for the file watching behaviour itself will not be updated during reloads

## Configuration Example

```yaml
watch_config:
  enabled: true
  interval: "10s"

  startup_delay:
    enabled: false
    maximum: "10s"
```

## Checklist


router/cmd/main.go
```diff
@@ -71,59 +106,148 @@ func Main() {
...
+			time.Sleep(startupDelay)
+
+			if err := watchFunc(watcherCtx); err != nil {
+				if err != context.Canceled {
+					logger.Error("Error watching execution config", zap.Error(err))
+				}
+			}
+		}()
+
+		logger.Info("Watching router config file",
```
StarpTech, line 177: This needs to be split into multiple files. The reloader should be split from the actual router start logic.

docker/nats/build-push.sh
```diff
@@ -6,7 +6,7 @@ set -e
 # --- Config ---
 REPOSITORY="wundergraph/cosmo"
 IMAGE_NAME="nats"
-IMAGE_TAG="2.11.0"
+IMAGE_TAG="2.11.0-alpine"
```
StarpTech, line 9: What was the reason to change this?

docker-compose.yml
```diff
@@ -50,9 +50,9 @@ services:
   jaeger:
     image: jaegertracing/all-in-one:latest
     ports:
-      - '16686:16686'
-      - '14268:14268'
-      - '4319:4318'
+      - "16686:16686"
```
StarpTech, line 53: Please avoid unnecessary reformats.

StarpTech (changes requested):

docker/nats/build-push.sh
endigma: normal image does not have a linux distro, and thus cannot install `curl`, for the healthcheck

router/cmd/main.go
endigma: What do you mean? You'd like a general purpose command restarter? It needs to be in the same place as the router start logic in order to control it.

docker/nats/build-push.sh
endigma: as for the internal tag, i thought it best to keep them aligned? i can make it just `2.11.0`

router/cmd/main.go
endigma: Both the file watcher and SIGHUP just send a signal to a new channel `reloadChan`, which sits next to the old `ctx.Done()` as possible ways to trigger a router shutdown, with the former not setting a flag that causes process exit at the end of the graceful shutdown.

docker/nats/build-push.sh
StarpTech: Thanks for clarification

router/cmd/main.go
StarpTech: It would be great if we could abstract that a little bit better away without bloating up the main file. The file watcher code should not run in production.

router/pkg/supervisor/supervisor.go
```diff
@@ -0,0 +1,153 @@
...
+		return err
+	}
+
+	return nil
+}
+
+func (rs *RouterSupervisor) stopRouter() error {
+	// Enforce a maximum shutdown delay to avoid waiting forever
+	// Don't use the parent context that is canceled by the signal handler
+	shutdownCtx, cancel := context.WithTimeout(context.Background(), rs.resources.Config.ShutdownDelay)
```
StarpTech, line 92: Please stay with the previous log messages e.g.

```
logger.Info("Graceful shutdown of router initiated", zap.String("shutdown_delay", result.Config.ShutdownDelay.String()))
```

router/cmd/main.go
```diff
@@ -42,88 +49,176 @@ func Main() {
...
-	} else if result.DefaultLoaded {
-		logger.Info("Found default config file. Values in the config file have higher priority than environment variables",
-			zap.String("config_file", config.DefaultConfigPath),
+	sl := baseLogger.With(zap.String("component", "supervisor"))
+
+	rs := supervisor.NewRouterSupervisor(&supervisor.RouterSupervisorOpts{
+		Logger: sl,
+
+		LifecycleHooks: &supervisor.LifecycleHooks{
+			LoadResources: func(rr *supervisor.RouterResources) error {
```
StarpTech, line 108: What is the benefit of the callback if the supervisor is already coupled to the core / router ?

StarpTech (changes requested):

endigma:
Current state has a more minimal callback, and implemented other feedback

router/core/supervisor.go
```diff
@@ -0,0 +1,186 @@
...
+func (rs *RouterSupervisor) stopRouter() error {
+	// Enforce a maximum shutdown delay to avoid waiting forever
+	// Don't use the parent context that is canceled by the signal handler
+	shutdownCtx, cancel := context.WithTimeout(context.Background(), rs.resources.Config.ShutdownDelay)
+	defer cancel()
+
+	rs.logger.Info("Graceful shutdown of router initiated", zap.String("shutdown_delay", rs.resources.Config.ShutdownDelay.String()))
+
+	if err := rs.router.Shutdown(shutdownCtx); err != nil {
+		return fmt.Errorf("failed to shutdown router gracefully: %w", err)
```
StarpTech, line 102: Please verify with the old implementation. Logs are missing.

StarpTech (changes requested):

router/core/supervisor.go
endigma:
```
18:16:03 PM INFO core/supervisor.go:100 Graceful shutdown of router initiated {"hostname": "june.jlt.sh", "pid": 71083, "service": "@wundergraph/router", "service_version": "dev", "component": "supervisor", "shutdown_delay": "1m0s"}
```

What do you mean missing?

endigma: The location will have changed and it will have `component: supervisor`, is that an issue?

StarpTech: This

```go
	if err = router.Shutdown(shutdownCtx); err != nil {
		if errors.Is(err, context.DeadlineExceeded) {
			logger.Warn("Router shutdown deadline exceeded. Consider increasing the shutdown delay")
		}
		logger.Fatal("Could not shutdown router gracefully", zap.Error(err))
	} else {
		logger.Info("Router shutdown successfully")
	}

	logger.Debug("Router exiting")
```

StarpTech (approved):
LGTM

router/internal/timex/duration.go
```diff
@@ -0,0 +1,22 @@
+package timex
+
+import (
+	"math/rand"
```
Noroth, line 4: You probably rather want to use `math/rand/v2`
https://pkg.go.dev/math/rand/v2

Otherwise you need to seed

StarpTech (approved):
LGTM

router/internal/timex/duration.go
endigma: thanks

PR #1749, opened 2025-04-02
feat: query batching

SkArchon:
## Motivation and Context

This PR adds query batching, which includes the following
1. Ability to batch queries when enabled (not enabled by default)
2. Client tests (does not run automatically at the moment, includes js and kotlin clients)

## Checklist


router/internal/batch/batch.go
```diff
@@ -0,0 +1,107 @@
...
+	fn := func(w http.ResponseWriter, r *http.Request) {
+		// Read the request body for potential batching.
+		bodyBytes, err := io.ReadAll(r.Body)
+		if err != nil {
+			http.Error(w, "failed to read request", http.StatusBadRequest)
+			return
+		}
+
+		// Restore the body for further processing (if needed)
+		r.Body = io.NopCloser(bytes.NewBuffer(bodyBytes))
```
StarpTech, line 32: Can we avoid parsing the body twice or find a cheaper way to verify if it is a batch request?

router/internal/batch/batch.go
```diff
@@ -0,0 +1,107 @@
...
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	"go.opentelemetry.io/otel/trace"
+	"io"
+	"net/http"
+)
+
+const unlimitedBatchEntries = 0
+
+func Handler(maxEntriesPerBatch, maxRoutines int, handlerSent http.Handler, tracerProvider *sdktrace.TracerProvider) http.Handler {
+	tracer := tracerProvider.Tracer(
```
StarpTech, line 18: It's good practice to not create a tracer per request, although I believe it is cached. Could you create this before once?

router/internal/batch/batch.go
```diff
@@ -0,0 +1,107 @@
...
+		// Drain and close the semaphore.
+		for len(sem) > 0 {
+			<-sem
+		}
+		close(sem)
+
+		// Write out the batched response as a JSON array.
+		w.Header().Set("Content-Type", "application/json")
+		if err := json.NewEncoder(w).Encode(responses); err != nil {
+			http.Error(w, "failed to encode batched response", http.StatusInternalServerError)
```
StarpTech, line 102: Because we have a new root span (the batch) we need to ensure that all relevant span attributes are attached like IsError, graphID, operationHash etc.. We should also introduce another attribute to make it easier to identify batch requests.

In Studio, we should targeting a "working" version.

router/internal/batch/batch.go
```diff
@@ -0,0 +1,107 @@
...
+				rCopy = rCopy.WithContext(context.WithValue(rCopy.Context(), chi.RouteCtxKey, chi.NewRouteContext()))
+				rCopy.Body = io.NopCloser(bytes.NewBuffer(singleOp))
+
+				// Create a ResponseWriter that captures the output.
+				rw := newBufferingResponseWriter()
+				// Execute the mux handler for this operation.
+				handlerSent.ServeHTTP(rw, rCopy)
+
+				// Store the response (assuming the response is valid JSON).
+				responses[i] = rw.Body.Bytes()
```
StarpTech, line 83: This won't work with custom response header rules. I also don't think we should support this until really needed because it can complicate a lot of things. Batching is still not recommended and the motivation here is to support legacy clients.

StarpTech (changes requested):

router/internal/batch/batch.go
SkArchon: This is only created once, and not per request, it is outside of the handler func (I will make it a bit more explicit).

router/internal/batch/batch.go
SkArchon: I'm still looking into on tracing, will keep these in mind +1.

But can you clarify what you mean by "In Studio, we should targeting a "working" version."? (The spans already show up on studio)

router-tests/batch_test.go
```diff
@@ -0,0 +1,385 @@
...
+
+	t.Run("verify batched requests when max is unlimited", func(t *testing.T) {
+		t.Parallel()
+
+		testenv.Run(t,
+			&testenv.Config{
+				BatchingConfig: config.BatchingConfig{
+					Enabled:            true,
+					MaxConcurrent:      10,
+					MaxEntriesPerBatch: 0, // 0 means unlimited
```
SkArchon, line 121: @StarpTech do you think we actually need this? (this is not the default so someone needs to set this, its a big risk for them though, but its the same as setting 1000000 or something)

router/internal/batch/batch.go
StarpTech: Let's huddle and see if the user has enough context for investigating issues.

router-tests/batch_test.go
StarpTech: I'd always prefer defensive approach. In that case, a user can simply set it to a higher number if he is not interested in limiting but at least it was explicit.

router/internal/batch/batch.go
SkArchon: As discussed we will document this

router/internal/batch/batch_test.go
```diff
@@ -0,0 +1,147 @@
...
+		funcElapsed := time.Since(funcStart)
+		require.NoError(t, err)
+
+		readAllReader := bytes.NewBufferString(fullString)
+		readAllStart := time.Now()
+		_, err = io.ReadAll(readAllReader)
+		readAllElapsed := time.Since(readAllStart)
+		require.NoError(t, err)
+
+		require.Greater(t, readAllElapsed, funcElapsed)
```
SkArchon, line 29: Results of this test as follows
2025/04/08 15:34:04 readAllElapsed 647.25µs
2025/04/08 15:34:04 funcElapsed 4.25µs

It ofcourse only apply to large strings where the buffer is 2x or above the current buffer of 4kb

cc: @StarpTech

router/internal/batch/batch.go
SkArchon: I compared and added the missing attributes (which makes sense), and added tests for tracing.

router/core/graphql_prehandler.go
```diff
@@ -609,6 +610,15 @@ func (h *PreHandler) handleOperation(req *http.Request, variablesParser *astjson
 	}
 	requestContext.telemetry.addCommonAttribute(attributesAfterParse...)
 
+	if isBatchedRequest, _ := req.Context().Value(batch.IsBatchedRequestKey{}).(bool); isBatchedRequest {
+		if requestContext.operation.opType != "query" {
+			return &httpGraphqlError{
+				message:    "Batched requests can only contain queries",
```
StarpTech, line 616: Is this the same response as apollo? Additionally, will this error be tracked in metrics / tracing?

router/internal/batch/batch.go
```diff
@@ -0,0 +1,185 @@
...
+	return http.HandlerFunc(fn)
+}
+
+func addTracing(r *http.Request, bodyBytes []byte, clientHeader config.ClientHeader, batchOperationsLength int, baseOtelAttributes []attribute.KeyValue, version string) {
+	rootSpan := trace.SpanFromContext(r.Context())
+
+	stringBody := string(bodyBytes)
+
+	digest := xxhash.New()
+	digest.WriteString(stringBody)
```
StarpTech, line 171: There should be also Write method.

router/internal/batch/batch.go
```diff
@@ -0,0 +1,185 @@
...
+		w.Header().Set("Content-Type", "application/json")
+		if err := json.NewEncoder(w).Encode(responses); err != nil {
+			http.Error(w, "failed to encode batched response", http.StatusInternalServerError)
+		}
+	}
+
+	return http.HandlerFunc(fn)
+}
+
+func addTracing(r *http.Request, bodyBytes []byte, clientHeader config.ClientHeader, batchOperationsLength int, baseOtelAttributes []attribute.KeyValue, version string) {
```
StarpTech, line 165: Can we pull this out from this package to manage this logic centralized? This package shouldn't contain telemetry logic.

router/internal/batch/batch.go
```diff
@@ -0,0 +1,185 @@
...
+	rootSpan := trace.SpanFromContext(r.Context())
+
+	stringBody := string(bodyBytes)
+
+	digest := xxhash.New()
```
StarpTech, line 170: We do this already before. We should also not create a new hasher per request.

router/pkg/config/config.go
```diff
@@ -754,6 +754,12 @@ type AccessLogsConfig struct {
 	Subgraphs AccessLogsSubgraphsConfig `yaml:"subgraphs,omitempty" env:"ACCESS_LOGS_SUBGRAPH"`
 }
 
+type BatchingConfig struct {
+	Enabled            bool `yaml:"enabled" env:"BATCHING_ENABLED" envDefault:"false"`
+	MaxConcurrent      int  `yaml:"max_concurrent" env:"BATCHING_MAX_CONCURRENT" envDefault:"10"`
```
StarpTech, line 759: `MaxConcurrency` ?

router/core/graph_server.go
```diff
@@ -228,6 +230,27 @@ func newGraphServer(ctx context.Context, r *Router, routerConfig *nodev1.RouterC
...
+		})
+		httpRouter.Use(handler)
+	}
+
+	if s.batchingConfig.Enabled {
+		if s.batchingConfig.MaxConcurrentRoutines <= 0 {
+			return nil, errors.New("maxConcurrent must be greater than 0")
+		}
+		if s.batchingConfig.MaxEntriesPerBatch <= 0 {
+			return nil, errors.New("maxEntriesPerBatch must be greater than 0")
```
StarpTech, line 250: In addition to that we should also validate it via json schema

router/internal/batch/batch.go
```diff
@@ -0,0 +1,185 @@
...
+			return
+		}
+
+		batchOperationsLength := len(batchOperations)
+
+		addTracing(r, bodyBytes, handlerOpts.ClientHeader, batchOperationsLength, handlerOpts.BaseOtelAttributes, handlerOpts.RouterConfigVersion)
+
+		// When a max batch limit has been specified
+		if batchOperationsLength > handlerOpts.MaxEntriesPerBatch {
+			http.Error(w, "unable to process request", http.StatusBadRequest)
```
StarpTech, line 100: How does other implementations behave? Do they fail or do they also allow partial responses?

This error should be tracked via tracing.

router/internal/batch/batch.go
SkArchon: Can you elaborate a bit more, do you mean move it to tracing?

router/internal/batch/batch.go
SkArchon: Will need to check this.

router/core/graphql_prehandler.go
SkArchon: Apollo allows mutations, we don't; so this will be a new error. I would suggest we just allow queries for now and allow mutations if customers request it (as its harder to go from permissive -> restrictive after its out in the wild).
image

SkArchon: It allows mutations, we don't; so this will be a new error. I would suggest we just allow queries for now and allow mutations if customers request it (as its harder to go from permissive -> restrictive after its out in the wild).

router/internal/batch/batch.go
StarpTech: Yes, at a place that is tracing specific. I don't see an overlap here with batching.

router/core/graphql_prehandler.go
StarpTech: If it supports mutation by spec, we need to support it because we're building as a migration feature.

router/internal/batch/batch.go
SkArchon: Does not handle partials, just errors out. I made it similar.

router/core/graph_server.go
```diff
@@ -269,6 +310,60 @@ func newGraphServer(ctx context.Context, r *Router, routerConfig *nodev1.RouterC
...
+type TracingHandlerOpts struct {
+	traceConfig         *rtrace.Config
+	healthCheckPath     string
+	readinessCheckPath  string
+	livenessCheckPath   string
+	compositePropagator propagation.TextMapPropagator
+	tracerProvider      *sdktrace.TracerProvider
+}
+
+func NewTracingHandler(s TracingHandlerOpts) func(next http.Handler) http.Handler {
```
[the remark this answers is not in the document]
SkArchon: Note for self: done

router/internal/batch/batch.go
StarpTech: Alright

router/core/batch.go
```diff
@@ -0,0 +1,257 @@
...
+	}
+
+	// Drain and close the semaphore.
+	for len(sem) > 0 {
+		<-sem
+	}
+	close(sem)
+
+	// Write out the batched response as a JSON array.
+	w.Header().Set("Content-Type", "application/json")
```
StarpTech, line 172: Charset missing

StarpTech (changes requested):

StarpTech (approved):
LGTM

PR #1755, opened 2025-04-04
fix(websocket): handle ping/pong correctly

StarpTech:
## Motivation and Context

This PR implements proper Ping/Pong support for subgraphs to prevent connection drops after inactivity. Most of the changes were made in the engine: https://github.com/wundergraph/graphql-go-tools/pull/1122. We expose the `websocket_client_ping_interval` as a dedicated engine config to define the ping interval.

## Checklist


router-tests/websocket_test.go
```diff
@@ -2159,3 +2166,236 @@ func expectConnectAndReadCurrentTime(t *testing.T, xEnv *testenv.Environment) {
...
+							return
+						}
+
+					case "subscribe":
+						// Handle countEmp subscription
+						if payload, ok := msg["payload"].(map[string]interface{}); ok {
+							if query, ok := payload["query"].(string); ok && strings.Contains(query, "countEmp") {
+								go handleCountEmpSubscription(t, wsWriteCh, msg["id"], 200*time.Millisecond, totalUpdates)
+							}
+						}
```
endigma, line 2279: The `msg` type seems like it would be fairly easy to mostly if not fully unmarshal into a concrete struct, is that not possible? I believe it would make the test easier to read.

router-tests/websocket_test.go
```diff
@@ -2159,3 +2166,236 @@ func expectConnectAndReadCurrentTime(t *testing.T, xEnv *testenv.Environment) {
...
+				select {
+				case <-pingReceived:
+					pingCount++
+
+					if pingCount >= totalUpdates {
+						t.Logf("Received %d pings from router", pingCount)
+						return
+					}
+				case <-timeoutCh:
+					require.GreaterOrEqual(t, pingCount, 1, "Expected at least one ping from router")
```
endigma, line 2350: Maybe an upper bound on this check as well? Just to be thorough. I think there's a `require.WithinDelta` or something like that.

router-tests/websocket_test.go
```diff
@@ -2159,3 +2166,236 @@ func expectConnectAndReadCurrentTime(t *testing.T, xEnv *testenv.Environment) {
...
+	for i := 1; i <= totalUpdates; i++ {
+		response := map[string]interface{}{
+			"type": "next",
+			"id":   id,
+			"payload": map[string]interface{}{
+				"data": map[string]interface{}{
+					"countEmp": i,
+				},
+			},
+		}
```
endigma, line 2371: This is the same type as the `msg` in the other part of the test right? I think this would be a good candidate for a struct.

router/core/factoryresolver.go
```diff
@@ -4,12 +4,11 @@ import (
...
 	"encoding/json"
 	"fmt"
+	"github.com/wundergraph/graphql-go-tools/v2/pkg/engine/argument_templates"
 	"net/http"
 	"net/url"
 	"slices"
 
-	"github.com/wundergraph/graphql-go-tools/v2/pkg/engine/argument_templates"
-
 	"github.com/buger/jsonparser"
```
endigma, line 12: I know we don't have import sorting as a linter, but can this please go with the other graphql-go-tools/v2 imports 🥺

router/core/graph_server.go
```diff
@@ -5,6 +5,7 @@ import (
 	"crypto/ecdsa"
 	"errors"
 	"fmt"
+	"github.com/wundergraph/cosmo/router/internal/retrytransport"
```
endigma, line 8: Why the incorrect import reorder?

router/go.mod
```diff
@@ -31,7 +31,7 @@ require (
 	github.com/tidwall/gjson v1.18.0
 	github.com/tidwall/sjson v1.2.5
 	github.com/twmb/franz-go v1.16.1
-	github.com/wundergraph/graphql-go-tools/v2 v2.0.0-rc.168
+	github.com/wundergraph/graphql-go-tools/v2 v2.0.0-rc.168.0.20250404210002-57dea1f1c1a8
```
endigma, line 34: This doesn't look right

endigma (changes requested):

router-tests/websocket_test.go
```diff
@@ -2159,3 +2166,236 @@ func expectConnectAndReadCurrentTime(t *testing.T, xEnv *testenv.Environment) {
...
+	t.Parallel()
+
+	t.Run("epoll", func(t *testing.T) {
+
+		t.Parallel()
+
+		totalUpdates := 5
+
+		// Channel to collect pings from the router
+		pingReceived := make(chan bool, totalUpdates)
```
endigma, line 2180: What's the motivation of buffering this channel if we just skip over the blocking send later?

router/pkg/config/config.schema.json
```diff
@@ -2408,7 +2408,7 @@
           "type": "string",
           "format": "go-duration",
           "default": "1s",
-          "description": "The timeout for the poll loop of the WebSocket client implementation. The period is specified as a string with a number and a unit, e.g. 10ms, 1s, 1m, 1h. The supported units are 'ms', 's', 'm', 'h'."
```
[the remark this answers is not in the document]
StarpTech: I find it very verbose and repetitive. I'll re-add it and we can make a decision later.

router/go.mod
StarpTech: It is right. We first need to release the engine. While this allows me to test it.

router/pkg/config/config.schema.json
```diff
@@ -2419,7 +2419,21 @@
           "type": "string",
           "format": "go-duration",
           "default": "5s",
-          "description": "The timeout for the websocket read of the WebSocket client implementation. The period is specified as a string with a number and a unit, e.g. 10ms, 1s, 1m, 1h. The supported units are 'ms', 's', 'm', 'h'."
+          "description": "The timeout for the websocket read of the WebSocket client implementation."
+        },
+        "websocket_client_write_timeout": {
+            "type": "string",
+            "format": "go-duration",
```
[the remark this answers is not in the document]
StarpTech: Legacy. It will be validated by the JSON schema library. Back then or even today, the official spec had no duration format. However, I think your assumption is right we don't have to do it anymore.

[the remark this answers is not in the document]
StarpTech: This needs to be addressed in a separate PR.

router-tests/websocket_test.go
StarpTech: I didn't find a utility function. I add another assert.

router/pkg/config/config.schema.json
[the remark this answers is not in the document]
StarpTech: That's not correct. We have implemented it this way to validate min/max with the JSON schema approach. Doing this with pattern would be tedious or not even possible.

router/core/factoryresolver.go
StarpTech: I see your point but it is not very productive to enforce a rule that we didn't agree on. If we want to enforce it we should align on it as a team and automate it. Once option is to use `goimports` which formats imports and does go fmt afterwards. There is IDE support for Goland and VsCode.

router/core/graph_server.go
StarpTech: Reordered with goimports. Please see discussion above.

router-tests/websocket_test.go
StarpTech: It is possible. Although it doesn't add much value and I see your point and fixed it.

router/core/flushwriter.go
```diff
@@ -55,6 +55,9 @@ func (f *HttpFlushWriter) Complete() {
 			_, _ = f.writer.Write([]byte("--" + multipartBoundary + "--\r\n"))
 		}
 	}
+
+	f.flusher.Flush()
```
jensneuse, line 59: Can you add a comment explaining this?

jensneuse (approved):
one minor comment

router/core/flushwriter.go
StarpTech: Done

router-tests/testenv/testenv.go
```diff
@@ -1910,68 +1914,176 @@ func (e *Environment) WaitForTriggerCount(desiredCount uint64, timeout time.Dura
...
 	}
 
-	return messageType, p, err
+	// This should not be reached
+	return 0, nil, fmt.Errorf("failed to read from WebSocket: %w", err)
 }
 
-func DeflakeWSReadJSON(t testing.TB, conn *websocket.Conn, v interface{}) (err error) {
-	for i := 0; i < 5; i++ {
+func WSReadJSON(t testing.TB, conn *websocket.Conn, v interface{}) (err error) {
```
endigma, line 1960: If these don't necessarily need exponential backoff, perhaps [`require.EventuallyWithT`](https://pkg.go.dev/github.com/stretchr/testify/require#EventuallyWithT) would do and make the helper function less complex?

router-tests/testenv/testenv.go
```diff
@@ -1910,68 +1914,176 @@ func (e *Environment) WaitForTriggerCount(desiredCount uint64, timeout time.Dura
...
+		duration := b.Duration()
+		t.Logf("WebSocket write JSON error (attempt %d/%d): %v - retrying in %v...",
+			attempts, maxAttempts, err, duration)
+
+		time.Sleep(duration)
 	}
 
-	return err
+	// This should not be reached
+	return fmt.Errorf("failed to write JSON to WebSocket: %w", err)
```
endigma, line 2086: Just a nit, given you're already changing every call-site with the rename, if these functions get `t testing.TB`, and the `err` return is always just handled by `require.NoError`, why not just `t.Fail/t.Failnow` instead of returning errors? Did I miss a `require.Error/ErrorContains` somewhere?

endigma (approved):
Two nits

router-tests/testenv/testenv.go
StarpTech: Good thoughts. The idea here was to make it more reliable. I think your idea could also be enough.

StarpTech: Eventually works with a callback that makes writing it harder.

router-tests/testenv/testenv.go
StarpTech: Because it is a low level function. I don't want to assume something. Maybe people want to check for the error.

PR #1762, opened 2025-04-07
feat: organization audit logs should retain for 90d after org

wilsonrivera:
## Motivation and Context

Audit logs should be retained for 90 days after an organization have been deleted for visibility.

The way we aim to achieve this is by removing the `foreign key` constraint from the organization id on the audits table to prevent the automatic deletion of logs while still allowing deletion of organization. We have also introduced a new column on the audits log table to keep the `organization slug` have a human value to identify the organization (a log entry is created when the organization slug is updated).

A new queue and worker have been introduces to  asynchronously delete the audit logs after 90 days only after the organization deletion worker is executed.

## Checklist


controlplane/src/db/schema.ts
```diff
@@ -1621,11 +1621,7 @@ export const auditLogs = pgTable(
   'audit_logs', // auditlogs
   {
     id: uuid('id').notNull().primaryKey().defaultRandom(),
-    organizationId: uuid('organization_id')
-      .notNull()
-      .references(() => organizations.id, {
-        onDelete: 'cascade',
-      }),
+    organizationId: uuid('organization_id').notNull(), // we don't want the audit log to be dropped when the organization is deleted
```
StarpTech, line 1624: After deleting the org, how we will be able to identify an org in the logs?

StarpTech (changes requested):

controlplane/src/db/schema.ts
wilsonrivera: I have updated the audit logs table to include the `organization slug`,  that way we can at least identify the organization after being deleted. I opted for the slug as we already have it and is supposed to be unique.

wilsonrivera: The migration also populates the field for the existing logs

router/demo.config.yaml
```diff
@@ -5,6 +5,31 @@
 
 version: "1"
 
+dev_mode: true
```
StarpTech, line 8: Please revert.

StarpTech (changes requested):

router/demo.config.yaml
wilsonrivera: Done, sorry about that

StarpTech (approved):
LGTM

StarpTech:
Please add a few details to the PR how we will retain it.

PR #1766, opened 2025-04-09
fix: solve issue caused when receiving a date range where the start is greater than the end

wilsonrivera:
## Motivation and Context

When the client sends a date range where the start is greater than the end (for example, selecting the same date on Studio but making the start time greater than the end time) would result in a `500` error.

This PR aims to solve this issue by inverting the values when this condition occurs.

This PR also solves a visual issue where the days were not aligned correctly on the calendar.

## Checklist

preview


controlplane/src/core/repositories/analytics/SubgraphMetricsRepository.ts
```diff
@@ -491,6 +508,12 @@ export class SubgraphMetricsRepository {
     whereSql,
     queryParams,
   }: GetSubgraphMetricsProps) {
+    if (dateRange.start > dateRange.end) {
```
StarpTech, line 511: Should we put this into a helper fn?

StarpTech (changes requested):

StarpTech (approved):
LGTM

PR #1769, opened 2025-04-09
🐞fix(router-tests): remove freeport logger init method

robmorgan:
## Motivation and Context

This PR removes the init method from `testenv`. It is calling a `freeport.SetLogLevel(freeport.ERROR)` method which I believe never existed and is likely an LLM hallucination. Note: this is not a problem if you fork the Cosmo repo, but in our case we are importing many of the Cosmo packages into our own custom router and we are unable to use `testenv`.

## Checklist

### DIFF
--- a/router-tests/testenv/testenv.go
+++ b/router-tests/testenv/testenv.go
@@ -85,10 +85,6 @@ var (
 	demoKafkaProviders             = []string{myKafkaProviderID}
 )
 
-func init() {
-	freeport.SetLogLevel(freeport.ERROR)
-}
-
 // Run runs the test and fails the test if an error occurs
 func Run(t *testing.T, cfg *Config, f func(t *testing.T, xEnv *Environment)) {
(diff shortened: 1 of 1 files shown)


Noroth:
Hi @robmorgan. 
This function does not exist in the original consul sdk. 

The freeport package has a very verbose logging, so we requested to have a way to change the logging behavior. Sadly we never received a response from Hashicorp and created a fork for now. 

https://github.com/hashicorp/consul/pull/22103#issuecomment-2781166929

See the replace directive here.

https://github.com/wundergraph/cosmo/blob/ddfa355588bb0f87446d2160161428b94e63ea7b/router-tests/go.mod#L178C2-L178C112

robmorgan:
Hi @Noroth, thanks for the response! I understand what you are trying to do now. I've tested that replace directive and its a viable workaround for us, so I'm going to close this PR out for now.

PR #1771, opened 2025-04-10
fix(router): graceful shutdown blocked by in-flight requests due to lock

Noroth:
## Motivation and Context

This PR fixes an issue with the graceful shutdown. 

The write lock actually caused the existing read locks to be blocked. During the server shutdown we attempt to close idle connections but the connections have never been served and were waiting for the handler function. The handler function is blocked by read locks, which were blocked by the write lock.

As we set the write lock before shutting down the server, we prevented any inflight requests from completing. The only reason for the write lock is to set the `graphServer` and `handler` to nil.

By moving the write lock further down, we can ensure that any inflight requests are being processed before shutting down the http server without blocking the handler.

Also unrelated improvement:
`operationHashCache` was not closed yet

## Checklist

### DIFF
--- a/router-tests/events/kafka_events_test.go
+++ b/router-tests/events/kafka_events_test.go
@@ -323,7 +323,7 @@ func TestKafkaEvents(t *testing.T) {
 
 			require.Eventually(t, func() bool {
 				return counter.Load() == 4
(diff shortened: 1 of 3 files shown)


StarpTech (approved):
LGTM

PR #1793, opened 2025-04-15
fix: solve an issue where fields were marked as deprecated incorrectly for `input` and `enum` types

wilsonrivera:
## Motivation and Context

This solves an issue where fields were incorrectly being flagged as deprecated for `input` and `enum` types when viewing them directly.

## Checklist

### DIFF
--- a/studio/src/lib/schema-helpers.ts
+++ b/studio/src/lib/schema-helpers.ts
@@ -12,6 +12,7 @@ import {
   GraphQLScalarType,
   GraphQLSchema,
   GraphQLUnionType,
+  InputValueDefinitionNode,
   Kind,
   Location,
   FieldDefinitionNode,
@@ -57,7 +58,7 @@ export type GraphQLTypeCategory =
 export type ParsedGraphQLField = {
   name: string;
   description?: string;
-  deprecationReason?: string;
+  deprecationReason?: string | null;
   authenticated?: boolean;
   requiresScopes?: string[][];
   defaultValue?: any;
@@ -67,7 +68,7 @@ export type ParsedGraphQLField = {
     description: string;
     defaultValue: any;
     type: string;
-    deprecationReason?: string;
+    deprecationReason: string | null;
     loc?: Location;
   }>;
   loc?: Location;
@@ -120,14 +121,7 @@ export const mapGraphQLType = (
     return {
       ...common,
       category: "inputs",
-      fields: Object.values(graphqlType.getFields()).map((field) => ({
-        name: field.name,
-        description: field.description || "",
-        deprecationReason: field.deprecationReason || "",
-        defaultValue: field.defaultValue,
-        type: field.type.toString(),
-        loc: field.astNode?.loc,
-      })),
+      fields: Object.values(graphqlType.getFields()).map(field => maybeParseField(field.astNode)!).filter(Boolean),
     };
   }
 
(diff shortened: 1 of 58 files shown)


studio/src/lib/schema-helpers.ts
```diff
@@ -57,7 +58,7 @@ export type GraphQLTypeCategory =
 export type ParsedGraphQLField = {
   name: string;
   description?: string;
-  deprecationReason?: string;
+  deprecationReason?: string | null;
```
Aenimus, line 61: no `?`, right?

studio/src/lib/schema-helpers.ts
```diff
@@ -120,14 +121,7 @@ export const mapGraphQLType = (
...
       category: "inputs",
-      fields: Object.values(graphqlType.getFields()).map((field) => ({
-        name: field.name,
-        description: field.description || "",
-        deprecationReason: field.deprecationReason || "",
-        defaultValue: field.defaultValue,
-        type: field.type.toString(),
-        loc: field.astNode?.loc,
-      })),
+      fields: Object.values(graphqlType.getFields()).map(field => maybeParseField(field.astNode)!).filter(Boolean),
```
Aenimus, line 124: Can you write this without `!`? that looks like a code smell imo

studio/src/lib/schema-helpers.ts
wilsonrivera: I'm leaving it optional as the fields for the `union` type doesn't seems to care about this

studio/src/lib/schema-helpers.ts
wilsonrivera: The thing is that it would be needed either after the `maybeParseField` or after the `field.astNode`;  `astNode` is nullable

endigma: Why does `maybeParseField` return a nullable result at all if we're going to panic-level non-null assert it here?

studio/src/lib/schema-helpers.ts
```diff
@@ -632,6 +631,11 @@ export const extractDirectives = (node: ASTNode | undefined | null): ExtractedDi
       case "deprecated":
         const deprecatedDirValues = getArgumentValues(GraphQLDeprecatedDirective, directive);
         result.deprecationReason = (deprecatedDirValues.reason ?? "") as string;
```
endigma, line 633: What would the type of this be if you didn't `as` it? This looks like it should be a string anyways? If not, is there a good reason why it isn't?

endigma (changes requested):

studio/src/lib/schema-helpers.ts
wilsonrivera: Originally, it didn't include the `InputValueDefinition`, but this has changed since, I've updated the code to remove `maybeParseField` as is redundant

studio/src/lib/schema-helpers.ts
endigma: Ok, but you're already doing (x ?? "whatever") so shouldn't typescript pick that up as a type assertion?

wilsonrivera: It doesn't, it assumes the union `{} | string` sadly

endigma (approved):

PR #1811, opened 2025-04-23
fix: improve the queries of the check page

JivusAyrus:
## Motivation and Context
This PR intends to improve the queries used to fetch the checks of a federated graph.
- Fetching the total check count conditionally reduces one query in most cases.
- Add the start and end date to the query which fetches the total checks.

## Checklist


controlplane/src/core/bufservices/check/getChecksByFederatedGraphName.ts
```diff
@@ -91,15 +93,27 @@ export function getChecksByFederatedGraphName(
       includeSubgraphs,
     });
 
-    const totalChecksCount = await subgraphRepo.getChecksCount({ federatedGraphTargetId: federatedGraph.targetId });
+    let totalChecksCount = 0;
+    // fetch the total checks count, only if the checks count based on the date range is 0
```
StarpTech, line 97: Sounds to me like a pre-opt. Do you think it is really worth to do i, plus adding the field to the proto?

StarpTech (changes requested):

controlplane/src/core/bufservices/check/getChecksByFederatedGraphName.ts
JivusAyrus: I feel this should solve the issue.
We were seeing queries in gcp without filters on dates and thinking it was getting clipped, but we had a query without those filters to fetch the checks count of the fed graph. Without those filters, the join was happening between the entire tables.
That particular count was only responsible to show the users commands to run a check when the count was 0.

Now, I pass the dates, which are the max number of days the users can view their checks based on their plan and that is done optionally, as if the checks query with date range return checks then we dont run it at all, as we already know that the checks exist.

JivusAyrus: Regarding the proto, i removed a field called totalChecksCount and added hasChecks

StarpTech (approved):
LGTM

PR #1816, opened 2025-04-28
feat: add specific request tracing hooks

SkArchon:
## Motivation and Context

Sometimes subgraph requests can fail for users, and it might not be obvious why, by providing a way to log and add more span details, this can help users to figure out what is actually happening at the request level.

In order to do this the PR introduces the following
* Add a traceclient wrapper: We create a custom roundtripper that uses the http.ClientTrace wrapper. This wrapper stores information from its hook into the request's context, which can be accessed outside the wrapper. (This is not tied to expressions and uses its own separate structs, but as an optimization we only wrap the roundTripper if `subgraph.request.clientTrace` used in any expression.)

* Add subgraph to the expression context: This holds information about the currently executing subgraph. If `clientTrace` is used in an expression, we convert the struct used by the `traceclient` to the expression struct.

* Enable expressions for subgraphs: At the moment we only allow defining expressions on access logs at the router level, this PR adds the capability to use expressions for access logs at the subgraph level.

* Add `tracingAttributes`: Right now we have `telemetryAttributes` and `metricAttributes`, if a user defines `telemetryAttributes` we will attach all values to every span in the request AS WELL as metrics. We add `tracingAttributes`, which does the same but only attaches attributes to spans.

* Evaluation of expressions that use `subgraph.*`, any expression that uses `subgraph` is evaluated in the `engine_hooks_loader`, because `subgraph.*` is empty outside of this. In order to do this we separate out expressions during the compilation step for both `telemetryAttributes` and `tracingAttributes` if they only should run in the `subgraph`. This prevents expressions focused on subgraphs being attached to unrelated spans.

**Caveats**

* All expected output values for expressions are strings, which means if the output value is time, int, bool, the user will need to cast it to string. This can be done using the `string` function in exprlang. This follows the current behaviour of `telemetryAttributes`, which only allows strings.

* Users should be aware of nils, if a hook wasnt executed for some reason the equivalent expr struct field would be nil. 

**Figuring out issues using this PR**
Since the goal of this PR is to help users find out subgraph timeout, not connecting, etc issues, they can use the following to figure out (note that non string expressions need to be wrapped with string, but we will be ignoring using string to make it easy to read)

* Assume that they want all values from the `traceclient` logged. For dialStart and dialDone, users will need to just log the slice
```
    Key: "connCreateTime", expr: `subgraph.request.clientTrace.connCreate?.time`
    Key: "connCreateHost", expr: `subgraph.request.clientTrace.connCreate?.hostPort`   
    Key: "connAcquired", expr: `subgraph.request.clientTrace.connAcquired?.reused`
    ...   
```

* This is going to litter the subgraph `Engine - Fetch` span with information, if users are interested in Spans which error out, users can add the following
   `subgraph.request.error != nil ? subgraph.request.clientTrace.connCreate?.time : nil`

* Users can get more specific information, if they are interested in duration used to aquire a connection, users could
   `subgraph.request.clientTrace.connAcquired.time - subgraph.request.clientTrace.connCreate.time`

* Use the `GetGroupedDials()` helper, to get Start and Done dials and process durations
```
let groupedDials = subgraph.request.clientTrace.GetGroupedDials();
let dialDurations = map(groupedDials, #.doneTime - #.startTime);
dialDurations
```

More examples can be found at: https://github.com/wundergraph/cosmo/pull/1816/files#diff-a76256094262e1d7bf2857b5e8be79e45bcfcd298f918a42fb8db6e267e4c984R178-R179
 
**ExprStructure**
```
Subgraph:
  id: string
  name: string
  request:
    error: error
    clientTrace:
      ...
      # clientTraceHooks
```

Example Configuration for access logs (we only added expressions)
```
access_logs:
  subgraphs:
    enabled: true
    fields:
      - key: "dnsHost"
        value_from:
          expression: "subgraph.error ?? subgraph.operation.trace.dnsStart.host"
```

Example Configuration for tracing (We added tracing.attributes)
```
telemetry:
  tracing:
    attributes:
      - key: "wg.tracing.dnsHost"
        value_from:
           expression: "subgraph.error ?? subgraph.operation.trace.dnsStart.host"
```

## Checklist


router/internal/httpclient/traceclient.go
```diff
@@ -0,0 +1,211 @@
...
+			return cs.Name
+		}
+	}
+	return fmt.Sprintf("%d", id)
+}
+
+// We only use this for Got1xxResponse response at the moment
+// it can also be used for WroteHeaderField if we enable it later
+func (t *TraceInjectingRoundTripper) addAndGetAttrVal(value map[string]any, operationString string) []any {
+	t.mu.Lock()
```
SkArchon, line 168: We create a client trace transport per request, but even within the same request some operations can get called multiple times. From the code

> Any particular hook may be nil. Functions may be
> called concurrently from different goroutines and some may be called
> after the request has completed or failed.

Notably `Got1xxResponse` and `WroteHeaderField`, of which the latter we are skipping since we have `WroteHeaders`.

router/internal/expr/expr.go
```diff
@@ -93,6 +96,18 @@ type RequestAuth struct {
 	Scopes          []string       `expr:"scopes"`
 }
 
+// Subgraph Related
+type Subgraph struct {
```
SkArchon, line 100: Following the spec (except for the "Trace" attribute) from https://github.com/wundergraph/cosmo/pull/1481

router/pkg/config/config.go
```diff
@@ -66,17 +65,32 @@ type ResponseTraceHeader struct {
...
 	ParentBasedSampler  bool                `yaml:"parent_based_sampler" envDefault:"true" env:"TRACING_PARENT_BASED_SAMPLER"`
 	Exporters           []TracingExporter   `yaml:"exporters"`
 	Propagation         PropagationConfig   `yaml:"propagation"`
 	ResponseTraceHeader ResponseTraceHeader `yaml:"response_trace_id"`
+	Subgraph            SubgraphTracing     `yaml:"subgraph"`
 
 	TracingGlobalFeatures `yaml:",inline"`
 }
 
+type RequestTransportTrace struct {
```
SkArchon, line 89: We don't need this anymore, however we would still want to look at either adding a "Enabling" flag, or if no one has referenced "subgrpah.trace" we can auto disable the tracing. Favouring the latter.

router/pkg/config/config.go
```diff
@@ -1056,7 +1071,7 @@ func LoadConfig(configFilePath string, envOverride string) (*LoadResult, error)
 		cfg.Config.SubgraphErrorPropagation.Enabled = true
 		cfg.Config.SubgraphErrorPropagation.PropagateStatusCodes = true
 		cfg.Config.SubgraphErrorPropagation.OmitLocations = false
-		cfg.Config.SubgraphErrorPropagation.AllowedExtensionFields = unique.SliceElements(append(cfg.Config.SubgraphErrorPropagation.AllowedExtensionFields, "code", "stacktrace"))
+		//cfg.Config.SubgraphErrorPropagation.AllowedExtensionFields = unique.SliceElements(append(cfg.Config.SubgraphErrorPropagation.AllowedExtensionFields, "code", "stacktrace"))
```
SkArchon, line 1074: To revert

router/core/request_context_fields.go
```diff
@@ -44,7 +45,7 @@ func NewExpressionLogField(val any, key string, defaultValue any) zap.Field {
 		val = &assertVal
 	}
 
-	if v := val; v != "" {
+	if v := val; v != "" && v != nil {
```
SkArchon, line 48: While we validate no explicit nil can be returned, on case we use the optional chaining operator we can still get nil, which will result in no default value being set.
(This could be considered a bit of a breaking change, if nil is returned somehow we don't set the default value right now.)

SkArchon:
Because we also introduce custom telemetry attributes, we don't have a way right now to allow custom attributes in studio. We will need to (initial thoughts at least) add a UI in studio which can allow users to specify what custom telemetry attributes they want to view in studio traces.

router/core/transport.go
```diff
@@ -59,16 +60,23 @@ type sfCacheItem struct {
...
 	enableSingleFlight bool,
+	visitorManager *expr.VisitorGroup,
 ) *CustomTransport {
 	ct := &CustomTransport{
 		metricStore: metricStore,
 	}
 
+	baseRoundTripper := roundTripper
+
+	// TODO: Fix this bug where the round tripper is not set to the base round tripper
```
SkArchon, line 75: Fixing

StarpTech (changes requested):

router/internal/httpclient/traceclient.go
```diff
@@ -0,0 +1,147 @@
...
+		ConnectStart: func(network, addr string) {
+			start := SubgraphDialStart{
+				Time:    time.Now(),
+				Network: network,
+				Address: addr,
+			}
+
+			dialStartMu.Lock()
+			defer dialStartMu.Unlock()
+			eC.DialStart = append(eC.DialStart, start)
```
SkArchon, line 102: From my testing I verified that multiple connection starts and done's can occur, I believe its important to record them all. Since we don't want a data race I use a lock.

router/core/attribute_expressions.go
```diff
@@ -15,48 +14,32 @@ import (
 type attributeExpressions struct {
 	// expressions is a map of expressions that can be used to resolve dynamic attributes
 	expressions map[string]*vm.Program
+
 	// expressionsWithAuth is a map of expressions that can be used to resolve dynamic attributes and acces the auth
 	// argument
 	expressionsWithAuth map[string]*vm.Program
-}
-
-type VisitorCheckForRequestAuthAccess struct {
```
SkArchon, line 23: Just moved this to the expr package where other visitors are stored.

PR #1818, opened 2025-04-28
chore: replaced linter action with golangci-lint

alepane21:
Replaced custom linter action with golangci-lint action in CI workflow

## Motivation and Context

We want to be sure that tests are as parallel as possible. To add the tests I've converted the ci linter custom action to use golangci-lint and ported old checks.

After executing the checks some errors come up and I've already fixed them:
- added missing t.Parallel and added ignore rules when needed
- removed unused code
- changed multiple if to switch

## Checklist

- [X] Documentation has been updated on [https://github.com/wundergraph/cosmo-docs](https://github.com/wundergraph/cosmo-docs).

### DIFF
--- a/router-tests/.golangci.yml
+++ b/router-tests/.golangci.yml
@@ -1,12 +1,7 @@
 version: "2"
 linters:
   enable:
-    - errcheck
     - paralleltest
-    - govet
-    - staticcheck
-    - ineffassign
-    - unused
   settings:
     errcheck:
       exclude-functions:
--- a/router-tests/.golangci.yml
+++ b/router-tests/.golangci.yml
@@ -1,0 +1,27 @@
+version: "2"
+linters:
+  enable:
+    - errcheck
+    - paralleltest
+    - govet
+    - staticcheck
+    - ineffassign
+    - unused
+  settings:
+    errcheck:
+      exclude-functions:
+        - (io.ReadCloser).Close
+        - (*net/http.Server).Close
+        - (net/http.ResponseWriter).Write
+        - (*github.com/gorilla/websocket.Conn).Close
+        - (*compress/gzip.Reader).Close
+        - (io.Closer).Close
+        - (*mime/multipart.Writer).Close
+    paralleltest:
+      ignore-missing: false
+      ignore-missing-subtests: false
+  exclusions:
+    rules:
+      - path: "lifecycle/shutdown_test.go"
+        linters:
+          - paralleltest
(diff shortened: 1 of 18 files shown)


router-tests/.golangci.yml
```diff
@@ -0,0 +1,27 @@
+version: "2"
+linters:
+  enable:
+    - errcheck
+    - paralleltest
+    - govet
+    - staticcheck
+    - ineffassign
+    - unused
```
endigma, line 9: You don't need to re-enable linters that are in the default set, if you want to explicitly control which linters are enabled (including the defaults) you should set `default: none` above `enable`.

router-tests/.golangci.yml
```diff
@@ -0,0 +1,27 @@
...
+        - (io.ReadCloser).Close
+        - (*net/http.Server).Close
+        - (net/http.ResponseWriter).Write
+        - (*github.com/gorilla/websocket.Conn).Close
+        - (*compress/gzip.Reader).Close
+        - (io.Closer).Close
+        - (*mime/multipart.Writer).Close
+    paralleltest:
+      ignore-missing: false
+      ignore-missing-subtests: false
```
endigma, line 22: I do not recommend this linter at the moment, it just makes a lot of yellow and there is precedent discussion for not wanting parallelization for the sake of parallelization.

router-tests/.golangci.yml
alepane21: Thanks for the tip! I fixed it by only adding the one no in the default list.

router-tests/.golangci.yml
StarpTech: I agree with @endigma here. Adding parallelization should be reasoned. [In the worst case](https://threedots.tech/post/go-test-parallelism/), it could even make them slower. We have already run into a resource exhaustion scenario.

Regarding MCP @alepane21 I did it on purpose to not further stress out port allocation. Making them parallel didn't speed them up much.

StarpTech (changes requested):
I like the initiative but this needs more thoughts. We should not add parallelization for everything and using switch case rather than if/else should not be enforced.

alepane21:
> I like the initiative but this needs more thoughts. We should not add parallelization for everything and using switch case rather than if/else should not be enforced. I suggest to postpone this discussion.

I'll cancel the Pull Request then, thank you. It was an issue that I had on the pipeline, probably created too many months ago!

PR #1827, opened 2025-05-04
fix: refactor telemetry tests

SkArchon:
## Motivation and Context

Right now the telemetry tests use array index for tests. Thus if the array element order changes due to a change in the code the tests will fail. This tests fixes this by making the telemetry tests get the required array (scope metric) element by name, instead of relying on the array index then.

This is opened as a separate PR, before the router gauge PR.

## Checklist

### DIFF
--- a/router-tests/telemetry/telemetry_test.go
+++ b/router-tests/telemetry/telemetry_test.go
@@ -36,7 +36,7 @@ import (
 )
 
 const (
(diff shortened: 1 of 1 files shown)


alepane21 (approved):

StarpTech (approved):
LGTM

PR #1833, opened 2025-05-06
feat(playground): add support for state sharing via URL with user control over what to include

akshaygarg576:
## Motivation and Context
Wanted to make it easier for users to share their GraphQL Playground setup with others—especially during debugging or collaboration. This feature lets users generate a shareable URL that captures exactly what they’re working on: the query, variables, headers — with full control over what to include. It’s something customers specifically asked for, and it helps teams quickly reproduce each other's environments without any backend setup or manual copy-pasting. As part of intial iteration, I'm starting with sharing only variables, operations and headers. In future, it can be expanded to pre-flight, pre-operation and post-operation scripts too.

## 🎥 Demo Recording
https://www.loom.com/share/3137040f21a14806853a22b6c41eef0c

## 🛠️ Implementation Overview
The task consists of 3 major parts: 
- Creating a share playground modal with options to include. Different options include:
  - [x] Operation (always included)
  - [ ] Variables
  - [ ] Headers
- Encoding the selected state and generating a shareable URL
- Decoding the shared state and hydrating the playground accordingly

## 🧩 Architecture & Design Decisions
- Injected the new `SharePlaygroundModal` into the GraphiQL environment using the `#share-button` container id
- `SharePlaygroundModal` lives in its own file and is composed of smaller components for clarity, maintainability, and scalability. It uses the `useSharePlaygroundModal` hook for local UI state, while keeping business logic (like `buildStateToShare` and getting localstorage values) extracted to utility files to ensure separation of concerns and testability.
- Modal includes state options, warnings, and a "Copy Link" button. The final shareable URL is generated using `lz-string` and embedded as a `?playgroundUrlState=...` query param. A new link is generated everytime the user clicks on an option.
- `useHydratePlaygroundStateFromUrl` handles decoding and hydration logic, calling utility functions like `extractStateFromUrl`. 
- The chronology of events in `useHydratePlaygroundStateFromUrl` hook are as follows:
  - First stores the decompressed state from URL in a state variable
  - Cleans up the url so that GraphiQL's creation of new tabs doesn't interfere with hydrating
  - Allow GraphiQL to perform all its necessary rendering cycle (including calling onTabChange)
  - Once GraphiQL has done completing its duties, then we create a new tab and hydrate the state in this new tab (including setting to localstorage wherever required)
  - Once hydration is completed, we set `isHydrated` to true in the `PlaygroundContext`. This is done so that custom-scripts.tsx who also updates the script:tabsState lcoalstorage first waits for the hydration to complete and then update the localstorage to avoid any race conditions
  - If during decompresion and extraction, there's an error we show error tooltip to customers, and avoid creating a new tab
- Structured the feature to keep UI, state, and business logic separate. React components and hooks consume utility functions (buildStateToShare, decompressState, etc.), but utilities never depend on components — ensuring a clean unidirectional flow that avoids cyclic dependencies and makes shared logic reusable across contexts

### 🧠 Compression Strategy
I went with `lz-string` (compressToEncodedURIComponent) to encode the playground state into a URL-safe format. It’s lightweight, works reliably across all major browsers (unlike `CompressionStream`, which isn't supported in older browser versions -- before Feb 2023), and is widely used in tools like `CodeSandbox` and `JSFiddle`.
I didn’t add a fallback for lz-string since it’s stable, browser-compatible, and doesn’t rely on external APIs. Errors are already gracefully handled with user-friendly messaging and fallbacks to default state.
I looked into other options like `CompressionStream`, `pako`, and `lzma`, but most either lacked full browser support or felt like overkill for what we needed.

## 🚧 Edge Case Handling
- When the URL exceeds 2000 characters, a warning is shown and the "Copy Link" button is disabled. The warning disappears only when the URL length is shortened by deselecting some options.
- If a malformed or corrupted URL is loaded:
  - We show a toast with a user-friendly error message
  - Clear the invalid state from the URL
  - Log the issue in dev environments
- When the user copies the link, a user-friendly toast message is shown to user
- Reset modal state each time it is opened to avoid confusion
- Show success/failure toasts on link copy or generation errors

### 🚀 Performance
- Since the entire feature is client-side and on-demand, it has zero impact on runtime performance unless the modal is explicitly used
- Wrapped the URL generation logic in useCallback to avoid unnecessary repetitions during re-renders
- State hydration happens only on page load and is gated by a presence check for the query param, minimizing unnecessary work

### 🔒 Security
- No sensitive information like auth token is included in the URL compression
- Upon decompression, the scripts are visible to the user in UI, but they aren't immediately executed unless the user runs the request manually.

## ❗Limitations/Assumptions
- The sharing functionality applies only to the current active tab. This is aligned with the UX since the share button appears within the toolbar of the selected tab.

## 🧪 Test Coverage
Covered unit tests for business logic
- Encoding/decoding logic (`createCompressedStateUrl`, `decompressState`, `extractStateFromUrl`)
- Option filtering (`buildStateToShare`)

## 🔮 Future improvements

- Add Playwright tests for Modal interaction, URL sharing and state restoration, and Error handling for corrupted states
- Adding support for sharing the pre-flight, pre-operation and post-operation state. Consider removing GraphiQL as it is introducing a lot of race conditions

## 🧾 Additional Updates
- Updated local setup documentation for better clarity and reduced onboarding friction

## Checklist


StarpTech (approved):
LGTM

PR #1842, opened 2025-05-08
fix: split flaky and non flaky integration jobs

SkArchon:
## Motivation and Context

Right now if flaky integration tests fail we need to rerun the entire stack, which means that integration tests can fail on the second run. Creating an endless cycle, we separate out the jobs for now so we can restart them independently.

image

## Checklist


alepane21 (approved):
This looks good. It should at least help reduce the time we lose when dealing with flaky test failures.

PR #1846, opened 2025-05-10
fix: cleanup router config version from metrics

SkArchon:
## Motivation and Context

Since we now have router info as an attribute, we would like to cleanup the router_config_version for non cosmo cloud exporters, and only append back the router_config_version if its specified in metric.attributes.

We use the existing `attribute_mapper` which already filters out "router_config_version" if it is not specified in metric.attributes. However we introduce this to engine and runtime metrics.

We also add test to test the following scenarios
1. User is using the default cosmo cloud exporter
2. User is using a custom exporter
3. User is using a custom exporter but has specific "router_config_version" in metric attributes.

## Checklist


router/core/websocket.go
```diff
@@ -61,7 +60,6 @@ type WebsocketMiddlewareOptions struct {
 
 	WebSocketConfiguration *config.WebSocketConfiguration
 	ClientHeader           config.ClientHeader
-	Attributes             []attribute.KeyValue
```
SkArchon, line 64: Cleanup unused attributes

SkArchon:
TODO: Need to add telemetry attributes also to the `getRouterConfigVersionOverride` function.

router-tests/telemetry/telemetry_test.go
```diff
@@ -9454,3 +9454,1889 @@ func TestFlakyTelemetry(t *testing.T) {
...
+		if useCustomExporter == UseCloudExporter && addOperationInfo {
+			mid = append(mid,
+				otel.WgOperationHash.String("1163600561566987607"),
+				otel.WgOperationName.String(""),
+			)
+		}
+		return append(before, mid...)
+	}
+
+	t.Run("Verify metrics when there is a router config version metric attribute", func(t *testing.T) {
```
StarpTech, line 9479: Can we simplify this? These are few K lines of code just to validate if we have an attribute set. It's super hard for me to understand what these tests really tests.

router/core/router.go
```diff
@@ -2198,3 +2199,19 @@ func or[T any](maybe *T, or T) T {
 	}
 	return or
 }
+
+func isNotDefaultCloudExporter(metricConfig *rmetric.Config) bool {
```
StarpTech, line 2203: I think we can put this into the metric package. I'd avoid negation. How about `HasCloudExporterConfigured`

router/core/graph_server.go
```diff
@@ -675,26 +673,21 @@ func (s *graphServer) buildGraphMux(ctx context.Context,
...
-	}
+	exprManager := expr.CreateNewExprManager()
 
 	metricsEnabled := s.metricConfig.IsEnabled()
 
 	// we only enable the attribute mapper if we are not using the default cloud exporter
-	enableAttributeMapper := !(s.metricConfig.IsUsingCloudExporter || rmetric.IsDefaultCloudExporterConfigured(s.metricConfig.OpenTelemetry.Exporters))
-
-	exprManager := expr.CreateNewExprManager()
+	baseMuxAttributes := getBaseMuxAttributes(routerConfigVersion, s.baseOtelAttributes, featureFlagName)
```
StarpTech, line 681: Do we need to do this twice? Can't we collect the attributes only once before (for NewRuntimeMetrics) and add here only the feature flag attribute? I find it hard to read when we introduce smaller functions mapping functions that essentially nothing doing.

StarpTech (changes requested):

router-tests/telemetry/telemetry_test.go
SkArchon: I removed all attributes, since base otels are set to every attribute. We validate one.

router/pkg/metric/meter.go
```diff
@@ -294,13 +294,18 @@ func NewOtlpMeterProvider(ctx context.Context, log *zap.Logger, c *Config, servi
 	return mp, nil
 }
 
-// IsDefaultCloudExporterConfigured checks if the default cloud exporter is configured in the provided exporters.
-func IsDefaultCloudExporterConfigured(c []*OpenTelemetryExporter) bool {
-	for _, exp := range c {
+// IsUsingDefaultCloudExporter checks if the provided metricConfig is using the default cloud exporter.
+func IsUsingDefaultCloudExporter(metricConfig *Config) bool {
```
SkArchon, line 298: Merged the function

StarpTech (approved):
LGTM

SkArchon:
Users must ensure that oltp metrics is disabled or the default exporter is not present.

PR #1848, opened 2025-05-12
feat: edfs refactor kafka and nats with a datasource per field

alepane21:
## Motivation and Context

Today changing or adding an EDFS provider requires changes to cosmo and the engine, making it harder than it should.
The main point of this PR is to make it easier to add future EDFS providers.

So I moved all the pubsub code from the engine to the pubsub package in the router and refactored it to make more sense with the new structure.

There is a README that will help implementing new EDFS providers (manually or using AI).

### Important points
- I added a test to verify that kafka publish is working, there was nothing testing it before;
- connections operations (connection, closing connection) should happen as before;
- switching from a previous version to the new one should just works, without any change to the configuration or schema.

## Design

|Interface | Responsibility|
|-- | --|
|ProviderBuilder | Allows the router to build a Provider starting from DataSourceConfiguration and router event config.|
|Provider | Owns provider‑wide resources for a single providerId (e.g. NATS connection pool).|
|EngineDataSourceFactory | Implements ResolveDataSource + SubscribeDataSource for one GraphQL field.|
|Adapter | Thin, protocol‑specific driver that actually publishes / subscribes.|

## Execution flow

- When the router’s Loader (router/core/factoryresolver.go) encounters a PubSub nodes, it collects them and build the data sources and PubSubProviders by passing all the collected datasources to the PubSubProviderBuilder
- The PubSubProviderBuilder will create a single PubSubDataSource for each root node with its adapter and event data already assigned
  - this will generate more data sources, but will reduce the operations to do after the schema has been loaded
- The adapter will hands off the actual publish / subscribe work to the appropriate external driver (Kafka, NATS, etc.).

## Checklist


router/pkg/pubsub/README.md
```diff
@@ -35,45 +48,96 @@ type AdapterInterface interface {
...
-- [`*nodev1.DataSourceConfiguration`](../../gen/proto/wg/cosmo/node/v1/node.pb.go#DataSourceConfiguration), that contains everything you need about the provider data parsed from the schema.
-- `*plan.DataSourceMetadata`, usually not needed
-- [`config.EventsConfiguration`](../config/config.go#EventsConfiguration) that contains the config needed to setup the provider connection
-- `*zap.Logger`
-- `hostName string`, useful if you need to identify the connection based on the local host name
-- `routerListenAddr string`, useful if you need to identify the connection based on different router instances in the same host
+You can use the [kafka implementation](./kafka/engine_datasource_test.go) as a reference.
 
-The responsability of the factory is to initialize the PubSubProvider, like in this implementation for an `ExampleProvider`:
+#### Add pubsub_datasource_test.go
```
devsergiy: It could be a table like necessary tests

| Implementation file | test file | reference tests |
| -------- | ------- | ------- |
| pubsub_datasource.go | pubsub_datasource_test.go | [kafka implementation](./kafka/pubsub_datasource_test.go) |

router/pkg/pubsub/datasource/provider.go
```diff
@@ -2,19 +2,35 @@ package datasource
...
-	"github.com/wundergraph/cosmo/router/pkg/config"
-	"github.com/wundergraph/graphql-go-tools/v2/pkg/engine/plan"
-	"go.uber.org/zap"
 )
 
-type ProviderFactory func(ctx context.Context, in *nodev1.DataSourceConfiguration, dsMeta *plan.DataSourceMetadata, config config.EventsConfiguration, logger *zap.Logger, hostName string, routerListenAddr string) (PubSubProvider, error)
-
 type ArgumentTemplateCallback func(tpl string) (string, error)
 
+type Lifecycle interface {
```
devsergiy: maybe ProviderLifecycle?

router/pkg/pubsub/datasource/provider.go
```diff
@@ -2,19 +2,35 @@ package datasource
...
+	Startup(ctx context.Context) error
+	Shutdown(ctx context.Context) error
+}
+
+// PubSubProvider is the interface that the PubSub provider must implement
 type PubSubProvider interface {
+	// ID Get the provider ID as specified in the configuration
```
devsergiy, line 19: you could include lifecycle here like this

```go
type PubSubProvider interface {
  ProviderLifecycle
  ID() string
  TypeID() string
}
```

router/pkg/pubsub/datasource/provider_impl.go
```diff
@@ -0,0 +1,45 @@
+package datasource
+
+import (
+	"context"
+
+	"go.uber.org/zap"
+)
+
+type PubSubProviderImpl struct {
```
devsergiy: what do you think about naming interface `Provider` and this implementation `PubSubProvider`?

router/pkg/pubsub/kafka/adapter.go
```diff
@@ -239,3 +221,18 @@ func (p *Adapter) Shutdown(ctx context.Context) error {
 
 	return nil
 }
+
+func NewAdapter(ctx context.Context, logger *zap.Logger, opts []kgo.Opt) (*Adapter, error) {
```
devsergiy: here it could be named `Adapter` for the interface and `ProviderAdapter` for the implementation

router/pkg/pubsub/datasource/provider.go
devsergiy: or AdapterLifecycle

router/pkg/pubsub/kafka/engine_datasource.go
```diff
@@ -12,6 +13,21 @@ import (
...
+}
+
+type PublishEventConfiguration struct {
+	ProviderID string          `json:"providerId"`
+	Topic      string          `json:"topic"`
+	Data       json.RawMessage `json:"data"`
+}
+
+func (s *PublishEventConfiguration) MarshalJSONTemplate() string {
+	return fmt.Sprintf(`{"topic":"%s", "data": %s, "providerId":"%s"}`, s.Topic, s.Data, s.ProviderID)
```
devsergiy, line 30: should we marshal struct itself with json.Marshal?

router/pkg/pubsub/datasource/factory.go
```diff
@@ -4,35 +4,32 @@ import (
...
+		pubSubDataSourceFactory: pubSubDataSourceFactory,
+		executionContext:        executionContext,
 	}
 }
 
-type Factory struct {
-	providers        []PubSubProvider
-	executionContext context.Context
-	config           config.EventsConfiguration
+type Factory[P, E any] struct {
```
devsergiy: let's name it planner factory

router/pkg/pubsub/datasource/provider.go
```diff
@@ -2,19 +2,35 @@ package datasource
...
+}
+
+// PubSubProviderBuilder is the interface that the provider builder must implement.
+type PubSubProviderBuilder[P, E any] interface {
+	// TypeID Get the provider type id (e.g. "kafka", "nats")
+	TypeID() string
+	// BuildProvider Build the provider and the adapter
+	BuildProvider(options P) (PubSubProvider, error)
+	// BuildDataSource Build the data source for the given provider and event configuration
+	BuildDataSource(data E) (PubSubDataSource, error)
```
devsergiy: Currently, we have this intermediate factory `PubSubDataSourceFactory` which calls provider `BuildDataSource` internally

`PubSubDataSource` - it is also de facto a factory for the EngineDataSource

Is it possible to eliminate the intermediate factory?

If it is not feasible or will make it more complicated
What do you think about renaming `PubSubDataSource` to smth like `PubSubEngineDataSourceFactory` to have more clear info about its intent

router/pkg/pubsub/kafka/pubsub_datasource.go
```diff
@@ -4,50 +4,55 @@ import (
...
+const (
+	EventTypePublish EventType = iota
+	EventTypeSubscribe
+)
+
 type PubSubDataSource struct {
-	EventConfiguration *nodev1.KafkaEventConfiguration
-	KafkaAdapter       AdapterInterface
+	fieldName  string
+	typeName   string
```
devsergiy: typeName is unused

router/pkg/pubsub/README.md
alepane21: Good idea, thanks!

router/pkg/pubsub/kafka/engine_datasource.go
alepane21: I think it is the easier way to allow raw content in the `data` field, so I kept it like this, but I'm open to better ways!

devsergiy: is json.Marshal handles json.RawBytes in the incorrect way?

router/pkg/pubsub/kafka/adapter.go
alepane21: make sense, thanks!

router/pkg/pubsub/datasource/provider.go
alepane21: The only usage of the intermediate factory was as databag for having a builder and an event passed down to the planner. So I created a PlannerConfig struct with the two fields and remove the useless builder!

I really like the PubSubEngineDataSourceFactory, but I think that the PubSub prefix is now almost useless. Initially I introduced it to make a clear distinction between engine DataSource and PubSubDataSource ... but now is not useful anymore, so I just called it EngineDataSourceFactory and remove the PubSub prefix where it made no sense.

router/pkg/pubsub/kafka/engine_datasource.go
alepane21: No, the problem is that the content of data could be not valid JSON, like
```json
{"id":$$0$$,"update":$$1$$}
```

devsergiy: please add a comment about that

router/pkg/pubsub/pubsub.go
```diff
@@ -0,0 +1,187 @@
...
+type dsConfAndEvents[E GetEngineEventConfiguration] struct {
+	dsConf *DataSourceConfigurationWithMetadata
+	events []E
+}
+
+func (e *ProviderNotDefinedError) Error() string {
+	return fmt.Sprintf("%s provider with ID %s is not defined", e.ProviderTypeID, e.ProviderID)
+}
+
+// BuildProvidersAndDataSources is a generic function that builds providers and data sources for the given provider
```
devsergiy: this comment is not up to date I believe

alepane21: Fixed, thanks!

devsergiy (approved):
Great Job!

PR #1862, opened 2025-05-14
feat: add new metric store with trace connection metrics

SkArchon:
## Motivation and Context

This PR adds a set of new metric store called "ConnectionMetricStore", which is used to collect measurements related to information when making a subgraph request. We introduce the following metric attributes as part of the `router connection metric` store.

All metrics are slicecable (dimensions) by `subgraph`, `server.address` and `server.port` (the last two are semconv standard attributes, we went with these names because other attributes like net.peer.name have been deprecated in favour of it)

**router.http.client.max_connections**: Static gauge which keeps values for any maxConnsPerHost configuration
**router.http.client.active_connections**: A gauge which keeps track of active connections
**router.http.client.acquire_connection_duration**: A histogram of the total time taken to acquire a connection

In order to do this we introduce a traceclient, as well as a tracedialer. The first has hooks for operations such as get_conn, etc. The last is a wrapper on the `http.transport` `WrapDialer` method. `WrapDialer` is ONLY called when a new connection is created, thus we can use the function call as a fact that a connection was created.

## Checklist


router/core/trace_dialer.go
```diff
@@ -0,0 +1,53 @@
...
+func NewTraceDialer() *TraceDialer {
+	return &TraceDialer{
+		connectionPoolStats: metric.NewConnectionPoolStats(),
+	}
+}
+
+type dialerFunc func(ctx context.Context, network, address string) (net.Conn, error)
+
+func (t *TraceDialer) WrapDial(base dialerFunc) dialerFunc {
+	return func(ctx context.Context, network, address string) (net.Conn, error) {
```
SkArchon, line 22: Used JMeter to load test this, as otherwise counter doesnt go over one, counter seems to resets back to 0 from my testing

SkArchon:
TODO: Didn't push the prom tests, need to fix the prometheus tests asserts, as they are failing when the values seem to be equal.

router/internal/retrytransport/retry_transport.go
```diff
@@ -99,6 +100,8 @@ func (rt *RetryHTTPTransport) RoundTrip(req *http.Request) (*http.Response, erro
 		// drain the previous response before retrying
 		rt.drainBody(resp, requestLogger)
 
+		httpclient.IncrementRetryCount(req.Context())
```
StarpTech, line 103: Is this called anywhere else? A user could have retry disabled.

router/internal/httpclient/traceclient.go
```diff
@@ -0,0 +1,147 @@
...
+	req = req.WithContext(httptrace.WithClientTrace(req.Context(), trace))
+	trip, err := t.base.RoundTrip(req)
+
+	return trip, err
+}
+
+func IncrementRetryCount(ctx context.Context) {
+	fromContext := GetClientTraceFromContext(ctx)
+	fromContext.RetryCountLatestIndex++
+	fromContext.ClientTraces = append(fromContext.ClientTraces, &ClientTrace{})
```
StarpTech, line 45: Why do we need to collect them?

router/internal/httpclient/types.go
```diff
@@ -0,0 +1,144 @@
...
+	Error         error
+	Network       string
+	Address       string
+}
+
+// GetGroupedDials returns a slice of DialCombined structs
+// Which are combined based on the network (tcp/udp/...) and address
+// We sort by the first dial that was successful, so that the most likely
+// used dial is first in the list
+func (r ClientTrace) GetGroupedDials() []DialCombined {
```
StarpTech, line 99: This can be simplified. We can take the first DialStart event and ConnectStart to calculate the duration.

```go
var (
        mu            sync.Mutex
        startAll      time.Time
    )

    trace := &httptrace.ClientTrace{
        ConnectStart: func(_, _ string) {
            mu.Lock()
            if startAll.IsZero() {
                startAll = time.Now()
            }
            mu.Unlock()
        },
        GotConn: func(info httptrace.GotConnInfo) {
              mu.Lock()
              total := time.Since(startAll)
              mu.Unlock()
              fmt.Printf("Full connect (TCP+TLS) took %v\n", total)
        },
    }
```

StarpTech (changes requested):

router/internal/retrytransport/retry_transport.go
SkArchon: Its only called if retry is enabled.

router/internal/httpclient/types.go
SkArchon: While we cannot use this, because we need to get the first **successful** "ConnectDone" that finished first. We could simplify it a bit by only keeping a map for the ConnectStart, and within this function itself. Which I will do.

router/internal/httpclient/traceclient.go
SkArchon: We right now record all the information on the retries into the metric store because there could be information (latency, etc) spikes there that might be useful for the user, esp if its a common thing.

router/internal/httpclient/types.go
StarpTech: Also when `GotConn` is called after we successfully connected?

router/internal/httpclient/traceclient.go
StarpTech: Why do we have special handling for retries? We are really only interested in durations. A retry would emit another data point. We don't need to distinguish them yet.

router/core/graph_server.go
```diff
@@ -117,13 +119,25 @@ func newGraphServer(ctx context.Context, r *Router, routerConfig *nodev1.RouterC
 		return nil, fmt.Errorf(`the compatibility version "%s" is not compatible with this router version`, routerConfig.CompatibilityVersion)
 	}
 
+	isConnStoreEnabled := r.Config.metricConfig.OpenTelemetry.ConnectionStats || r.Config.metricConfig.Prometheus.ConnectionStats
+	var traceDialer *TraceDialer
+	if isConnStoreEnabled {
+		// An important caveat is that since we create http transports per base and subgraph
```
SkArchon, line 125: to remove comment

router/pkg/otel/attributes.go
```diff
@@ -46,10 +46,12 @@ const (
 	WgIsBatchingOperation              = attribute.Key("wg.operation.batching.is_batched")
 	WgBatchingOperationsCount          = attribute.Key("wg.operation.batching.operations_count")
 	WgBatchingOperationIndex           = attribute.Key("wg.operation.batching.operation_index")
-
 	// HTTPRequestUploadFileCount is the number of files uploaded in a request (Not specified in the OpenTelemetry specification)
 	HTTPRequestUploadFileCount = attribute.Key("http.request.upload.file_count")
 
+	WgHost       = attribute.Key("wg.conn.host")
```
StarpTech, line 52: Should exist in semconv

router/pkg/metric/oltp_connection_metric_store.go
```diff
@@ -0,0 +1,106 @@
...
+	"github.com/wundergraph/cosmo/router/pkg/otel"
+	"go.opentelemetry.io/otel/attribute"
+
+	otelmetric "go.opentelemetry.io/otel/metric"
+	"go.opentelemetry.io/otel/sdk/metric"
+	"go.uber.org/zap"
+)
+
+const (
+	cosmoRouterConnectionMeterName    = "cosmo.router.connection"
```
StarpTech, line 17: `cosmo.router.connections`

router/pkg/metric/connection_measurements.go
```diff
@@ -0,0 +1,66 @@
+package metric
+
+import (
+	"fmt"
+	otelmetric "go.opentelemetry.io/otel/metric"
+)
+
+// Connection metric constants
+const (
+	maxConnections            = "router.http.client.connection.max"
```
StarpTech, line 10: `router.http.client.max_connections`
`router.http.client.active_connections`
`router.http.client.connections.aquire_duration`

router-tests/prometheus_test.go
```diff
@@ -4227,6 +4229,19 @@ func TestPrometheus(t *testing.T) {
 
 }
 
+func getHost(connectionTotal *io_prometheus_client.Metric) string {
+	host := ""
+	for _, label := range connectionTotal.Label {
+		if label.Name == nil || label.Value == nil {
+			continue
+		}
+		if *label.Name == "wg_conn_host" {
```
StarpTech, line 4238: Reference to a const

StarpTech (changes requested):

router/pkg/otel/attributes.go
```diff
@@ -78,3 +78,25 @@ var (
 	RouterServerAttribute    = WgComponentName.String("router-server")
 	EngineTransportAttribute = WgComponentName.String("engine-transport")
 )
+
+const (
+	ServerAddress = attribute.Key("server.address")
+	ServerPort    = attribute.Key("server.port")
+)
+
+func GetServerAttributes(host string) []attribute.KeyValue {
```
SkArchon, line 87: According to semconv, there were a bunch of similar attributes that we could have used but all of them seem to have been deprecated in favour of `server.address` and `server.port`, thus we have to split the port and host component.

In case a port is not provided / bad host, we stick to returning only the host component.

cc: @StarpTech

router/pkg/otel/attributes.go
```diff
@@ -46,10 +47,11 @@ const (
 	WgIsBatchingOperation              = attribute.Key("wg.operation.batching.is_batched")
 	WgBatchingOperationsCount          = attribute.Key("wg.operation.batching.operations_count")
 	WgBatchingOperationIndex           = attribute.Key("wg.operation.batching.operation_index")
-
 	// HTTPRequestUploadFileCount is the number of files uploaded in a request (Not specified in the OpenTelemetry specification)
 	HTTPRequestUploadFileCount = attribute.Key("http.request.upload.file_count")
 
+	WgConnReused = attribute.Key("wg.conn.reused")
```
StarpTech, line 53: We should avoid shortcuts in attribute names. Proposal `wg.http.client.reused_connection`

Similar metric: https://opentelemetry.io/docs/specs/semconv/http/http-metrics/#metric-httpclientopen_connections

router/pkg/metric/connection_measurements.go
```diff
@@ -0,0 +1,66 @@
...
+import (
+	"fmt"
+	otelmetric "go.opentelemetry.io/otel/metric"
+)
+
+// Connection metric constants
+const (
+	maxConnections            = "router.http.client.max_connections"
+	connectionsActive         = "router.http.client.active_connections"
+	connectionAcquireDuration = "router.http.client.connections.acquire_duration"
```
StarpTech, line 12: To follow the naming schema we could also use `router.http.client.acquire_connection_duration`

https://opentelemetry.io/docs/specs/semconv/http/http-metrics/#metric-httpclientopen_connections

StarpTech (changes requested):

router-tests/prometheus_test.go
```diff
@@ -4229,17 +4230,22 @@ func TestPrometheus(t *testing.T) {
 
 }
 
-func getHost(connectionTotal *io_prometheus_client.Metric) string {
-	host := ""
+func getPort(connectionTotal *io_prometheus_client.Metric) string {
+	// We can get the key from the otel attribute, but we need to make an initialization
+	// also convert the key to prometheus which uses _ instead of . like otel
+	serverPortKey := strings.ReplaceAll(string(otel.ServerPort.String("").Key), ".", "_")
```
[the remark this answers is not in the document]
SkArchon: I referenced another place which `repalced` the string in prometheus test. Will update both

router/pkg/metric/connection_measurements.go
SkArchon: I thought we went aligned on prefixing it with `connections` because for future metrics like `router.http.client.connections.dns_duration` etc everything will be prefixed by connections are you suggesting that for future metrics we also don't prefix, i.e. :- router.http.client.dns_connection_duration ?

router/pkg/metric/connection_metric_store.go
```diff
@@ -0,0 +1,144 @@
...
+	otelProvider, promProvider *metric.MeterProvider,
+	metricsConfig *Config,
+	connectionPoolStats *ConnectionPoolStats,
+) (*ConnectionMetrics, error) {
+	connMetrics := &ConnectionMetrics{
+		baseAttributes: baseAttributes,
+		logger:         logger,
+	}
+
+	if metricsConfig.OpenTelemetry.ConnectionStats {
```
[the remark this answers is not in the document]
SkArchon: Also for the sdkmetric.NewMeterProvider() this is a meter-provider no-op, not a metric-provider no-op, you mean to have a no-op interface for the meter providers right, so we can remove the nil checks?

PR #1870, opened 2025-05-19
feat: add subgraph expressions and tracing attributes

SkArchon:
## Motivation and Context

This PR adds subgraph expressions and tracing attributes. 

**Subgraph ExpressionContext**
The following is introduced. We only have a minimal expression context for subgraph as it stands. We introduce the following

```
subgraph:
  id # subgraph id
  name #subgraph name
  request
    error #subgraph error if any occurred
    clientTrace # we can add any future trace related params here without polluting the request interface
      connAcquireDuration
```

**Add Tracing Attributes**
We already have `metric` and `telemetry` attributes, however we do not have `tracing` attributes, which can only be used to add values to `tracing` or `spans`. This is introduced as customers might be interested in how long a `connectionAcquire` took for a subgraph.

**Process subgraph expressions for telemetry**
Like we have `attributesWithAuth` we introduce `attributesWithSubgraph` which is essentially attributes that use `subgraph` in the expression context. We separate these values out for all three of telemetry / tracing and metric attributes and process them for every subgraph call.

**Process Subgraph Expressions In Logs And Tracing/Telemetry Attributes**
We have expressions for router access logs, however we do not use expressions for the subgraph access logger right now, we add this feature and allow users to define `subgraph` access log expressions.

## Checklist

### DIFF
--- a/router/core/context.go
+++ b/router/core/context.go
@@ -9,7 +9,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/expr-lang/expr/vm"
 	"go.opentelemetry.io/otel/attribute"
 	"go.uber.org/zap"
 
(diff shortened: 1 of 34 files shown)


router/core/context.go
```diff
@@ -269,13 +270,7 @@ func (c *requestContext) SetError(err error) {
 }
 
 func (c *requestContext) ResolveAnyExpressionWithWrappedError(expression *vm.Program) (any, error) {
```
SkArchon, line 272: to cleanup

PR #1880, opened 2025-05-22
feat: add proper enabled configuration flag for plugins

Noroth:
## Motivation and Context

We recently removed the `enabled` flag for the plugins as it didn't make sense the way it was implemented. Instead of enabling only the configuration options this PR adds an implementation to handle enabling of plugins in general. 

Enabled now will make sure that no plugins are loaded when `enabled` is set to false.

## Checklist


demo/.gitignore
```diff
@@ -1 +1,4 @@
 config.json
+
+bin
+generated
```
endigma, line 4: We should be committing generated code to the demo, no? This is generally considered best practice, and we do it with all the gqlgen subgraphs.

demo/Makefile
```diff
@@ -2,4 +2,19 @@ generate:
...
+
+plugin-generate:
+	pnpx wgc@latest router plugin build ./pkg/subgraphs/projects --generate-only --go-module-path github.com/wundergraph/cosmo/demo/pkg/subgraphs/projects
+
+plugin-build-binary:
+	cd pkg/subgraphs/projects/ && go build -o ./bin/$(shell go env GOOS)_$(shell go env GOARCH) ./src/
+
+plugin-build-integration: plugin-build
+	rm -rf ../router/plugins/*
+	cp -R pkg/subgraphs/projects ../router/plugins/
```
endigma, line 20: This is a very dangerous operation to run inside a Makefile I think, perhaps the router should just be pointed to a plugins directory within demo in a config file?

router-tests/router_plugin_test.go
```diff
@@ -0,0 +1,36 @@
...
+)
+
+func TestRouterPlugin(t *testing.T) {
+	t.Parallel()
+
+	testenv.Run(t, &testenv.Config{
+		RouterConfigJSONTemplate: testenv.ConfigWithPluginsJSONTemplate,
+		RouterOptions: []core.Option{
+			core.WithPlugins(config.PluginsConfiguration{
+				Path: "../router/plugins",
```
endigma, line 19: Perhaps point this at something in `../demo` or testdata or something, I don't like this being dependent on the current state of the router dir.

endigma (changes requested):

demo/.gitignore
endigma: If you want to ensure it is identical to committed version in CI, I recommend something like this:

```bash
#!/bin/bash

set -euo pipefail

task alyx:gen -f

status=$(git status --porcelain)
statusv=$(git diff)

if [ -n "$status" ]; then
	echo "you need to generate and commit the changes"
	echo "$statusv"
	exit 1
fi
```

PR #1888, opened 2025-05-23
feat: communicate downstream errors more effectively

endigma:
## Motivation and Context

When a subgraph closes unexpectedly, the router previously sent a normal "complete" event to clients, causing them to believe this was a natural closure. This PR changes it such that:

- In the default case, the router will mime the subgraph and close the connection abruptly, as though the client was connected directly
- In a configured case, the router will send an error ahead of the above behaviour for extra information.

## Todo

- [ ] Configuration for the error message needs to be plumbed through
- [ ] Update to a release version of go-tools, currently based on https://github.com/wundergraph/graphql-go-tools/pull/1158

## Checklist


router-tests/websocket_test.go
```diff
@@ -2360,34 +2265,185 @@ func TestWebSocketPingIntervalForGraphQLTransportWS(t *testing.T) {
...
-			timeoutCh := time.After(5 * time.Second)
-			pingCount := 0
+			pingCount := int(pingsReceived.Load())
+			require.GreaterOrEqual(t, pingCount, 1, "Expected at least one ping from router")
+			require.LessOrEqual(t, pingCount, totalUpdates, "Expected no more than %d pings from router", totalUpdates)
+		})
+	})
+}
+
+func TestWebsocketDownstreamError(t *testing.T) {
```
StarpTech: Please make it much clearer what this test is about.

router-tests/websocket_test.go
```diff
@@ -2360,34 +2265,185 @@ func TestWebSocketPingIntervalForGraphQLTransportWS(t *testing.T) {
...
 		})
+	}
 
-	})
+	return wsMiddleware, pingsReceived
 }
 
 // Helper function to handle countEmp subscription
-func handleCountEmpSubscription(t *testing.T, wsWriteCh chan<- wsMessage, id string, updateInterval time.Duration, totalUpdates int) {
+func handleCountEmpSubscription(t *testing.T, wsWriteCh chan<- wsJSONMessage, wsCloseCh chan<- wsCloseMessage, id string, updateInterval time.Duration, totalUpdates int, complete bool) {
```
StarpTech, line 2500: Does it mean, when "complete" is false it will close it?

router-tests/websocket_test.go
StarpTech: Please add a test where the subgraph closes the connection normally.

ws-countemp-start
```diff
@@ -0,0 +1,2 @@
+{"type":"connection_init","payload":{}}
```
StarpTech: What's the use of this?

StarpTech (changes requested):

ws-countemp-start
endigma: shouldn't have been committed, it's stdin for websocat, will fix

router-tests/websocket_test.go
endigma: Technically duplicative of the test directly above this one, but done so the single test output makes sense

router-tests/websocket_test.go
endigma: sort of, added a comment to clarify

router-tests/websocket_test.go
endigma: done

router/internal/wsproto/proto.go
```diff
@@ -15,13 +19,19 @@ type Proto interface {
...
+	Close(id string) error
 }
 
-type JSONConn interface {
-	ReadJSON(v interface{}) error
-	WriteJSON(v interface{}) error
+type ProtoConn interface {
+	ReadJSON(v any) error
+	WriteJSON(v any) error
+	WriteCloser(code ws.StatusCode, reason string) error
```
jensneuse: Rename to "WriteClose" or "WriteCloseFrame" or "WriteCloseMessage"?

router/internal/wsproto/proto.go
```diff
@@ -15,13 +19,19 @@ type Proto interface {
 	Pong(*Message) error
 	WriteGraphQLData(id string, data json.RawMessage, extensions json.RawMessage) error
 	WriteGraphQLErrors(id string, errors json.RawMessage, extensions json.RawMessage) error
+
 	// Done is sent to indicate the requested operation is done and no more results will come in
 	Done(id string) error
+
+	// Close uncerimoniously closes the connection, with an optional error message ahead of it
+	Close(id string) error
```
jensneuse: Should error message be named id?

router/internal/wsproto/subscriptions_transport_ws.go
```diff
@@ -119,6 +121,18 @@ func (p *subscriptionsTransportWSProtocol) WriteGraphQLErrors(id string, errors
 	})
 }
 
+func (p *subscriptionsTransportWSProtocol) Close(id string) error {
+	if err := p.conn.WriteCloser(ws.StatusGoingAway, "Downstream service error"); err != nil {
```
jensneuse: When we call Close here, why with an error?

router/internal/wsproto/subscriptions_transport_ws.go
```diff
@@ -119,6 +121,18 @@ func (p *subscriptionsTransportWSProtocol) WriteGraphQLErrors(id string, errors
 	})
 }
 
+func (p *subscriptionsTransportWSProtocol) Close(id string) error {
```
jensneuse: It seems like we're ignoring "id"?

router/internal/wsproto/proto.go
```diff
@@ -4,8 +4,12 @@ import (
 	"encoding/json"
 	"fmt"
 	"strings"
+
+	"github.com/gobwas/ws"
 )
 
+var DownstreamErrorMessage = json.RawMessage(`[{"message":"downstream service error","extensions":{"code":"DOWNSTREAM_ERROR"}}]`)
```
jensneuse: Should we include the Subgraph name? We might not for security reasons.

router/internal/wsproto/graphql_ws.go
```diff
@@ -114,6 +116,18 @@ func (p *graphQLWSProtocol) WriteGraphQLErrors(id string, errors json.RawMessage
 	})
 }
 
+func (p *graphQLWSProtocol) Close(id string) error {
```
jensneuse: we're ignoring the id arg

router/internal/wsproto/absinthe.go
```diff
@@ -202,6 +204,18 @@ func (p *absintheWSProtocol) WriteGraphQLErrors(id string, errors json.RawMessag
 	})
 }
 
+func (p *absintheWSProtocol) Close(id string) error {
+	if err := p.conn.WriteCloser(ws.StatusGoingAway, "Downstream service error"); err != nil {
```
jensneuse: ignoring "id" arg

router-tests/websocket_test.go
```diff
@@ -2411,19 +2524,34 @@ func handleCountEmpSubscription(t *testing.T, wsWriteCh chan<- wsMessage, id str
...
-			Type: "complete",
-			ID:   id,
-		},
-		done: done,
-	}
-	if err := <-done; err != nil {
-		t.Logf("Failed to send complete message: %v", err)
+	if complete {
+		// Send complete message
+		t.Log("Sending complete message for subscription")
```
jensneuse, line 2529: should we remove log statements to keep test logs clean?

router-tests/websocket_test.go
```diff
@@ -2411,19 +2524,34 @@ func handleCountEmpSubscription(t *testing.T, wsWriteCh chan<- wsMessage, id str
...
+		done := make(chan error, 1)
+		wsWriteCh <- wsJSONMessage{
+			data: GraphQLWSDataResponse{
+				Type: "complete",
+				ID:   id,
+			},
+			done: done,
+		}
+		if err := <-done; err != nil {
+			t.Logf("Failed to send complete message: %v", err)
```
jensneuse, line 2539: remove log?

jensneuse (changes requested):

router/internal/wsproto/proto.go
endigma: Bad comment, it's not an error message

router/internal/wsproto/proto.go
endigma: done

router/internal/wsproto/proto.go
endigma: fixed

router-tests/websocket_test.go
endigma: logs only show on fail, no? or am i just ignoring this because i use a wrapper to run them

router/internal/wsproto/subscriptions_transport_ws.go
endigma: removed

router/internal/wsproto/graphql_ws.go
endigma: removed, not needed

router/internal/wsproto/absinthe.go
endigma: removed, not needed

router/internal/wsproto/proto.go
endigma: I don't think this is worth adding to scope without a customer need, also it would need options for omission or obfuscation as you mention

router/internal/wsproto/subscriptions_transport_ws.go
endigma: Not sure I follow? It returns an error?

router-tests/websocket_test.go
endigma: > For tests, the text will be printed only if the test fails or the -test.v flag is set. For benchmarks, the text is always printed to avoid having performance depend on the value of the -test.v flag.

(from: https://pkg.go.dev/testing#T.Log)

Do you still think they should be removed?

router/internal/wsproto/subscriptions_transport_ws.go
jensneuse: I think I understand now. This Close is only called when the downstream service closed, right? Maybe make this clear in a comment. I was confused as to why we send an error message when we close. Close doesn't indicate error state.

router-tests/websocket_test.go
jensneuse: if it logs only on fail, it's fine

router/internal/wsproto/subscriptions_transport_ws.go
endigma: It's a bit tricky semantically, I can add a comment to clarify

There are potentially other cases where we might want to call close with different codes eventually, but the only scenario this is used in right now is when the downstream subgraph closes without a complete message, which we _always_ transmute into a `1001: Downstream service error`, regardless of what the subgraph actually did, even if it just closes the connection with no close frame (kill -9, etc)

Whether `1001` is semantically the correct code for this I'm not 100% sure, it's what Apollo Server returns in most cases when the subgraph goes down so the thinking is to mimic that for compatibility.

There are not many codes and they are not very descriptive: https://www.rfc-editor.org/rfc/rfc6455#section-7.4.1

endigma: https://github.com/wundergraph/cosmo/pull/1888/commits/91fc81e63b9c3de9229e06313882faff3e27d1a8

StarpTech (approved):
LGTM

PR #1893, opened 2025-05-23
test: refactor router down check, move settle wait

endigma:
## Motivation and Context

Sometimes this test failed in CI because it ended too quickly for the debug watcher goroutines to get cancelled by context. This should help.

## Checklist

### DIFF
--- a/router-tests/lifecycle/supervisor_test.go
+++ b/router-tests/lifecycle/supervisor_test.go
@@ -37,10 +37,15 @@ func TestRouterSupervisor(t *testing.T) {
 	xEnv.RouterSupervisor.Stop()
 	xEnv.Shutdown()
 
-	// Let everything settle
-	time.Sleep(1 * time.Second)
+	for {
+		_, err := xEnv.MakeRequest("GET", xEnv.RouterURL+"/health/ready", nil, nil)
+		if err != nil {
+			break
+		}
 
-	// Should fail, since everything should be off now
-	err = xEnv.WaitForServer(context.Background(), xEnv.RouterURL+"/health/ready", 250, 1)
-	require.Error(t, err)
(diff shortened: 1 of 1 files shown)


router-tests/lifecycle/supervisor_test.go
```diff
@@ -37,10 +37,15 @@ func TestRouterSupervisor(t *testing.T) {
 	xEnv.RouterSupervisor.Stop()
 	xEnv.Shutdown()
 
-	// Let everything settle
-	time.Sleep(1 * time.Second)
+	for {
```
StarpTech, line 40: My question here would be why is the server not properly shutdown after `xEnv.Shutdown()` Can we change the code to wait for graceful shutdown? e..g by introducing a close method that waits until the goroutine has been shutdown.

StarpTech (changes requested):

router-tests/lifecycle/supervisor_test.go
endigma: It does normally, it's that we aren't actually blocking on the router being shut down here because of the supervisor being in the unchecked goroutine above. If this weren't a test and we could block on `RouterSupervisor.Start()` it would be fine.

`RouterSupervisor.Stop()` is non-blocking, it's just a wrapper for sending a message to the stop channel that will cause the other goroutine to shut down.

Could potentially have blocking stop? It would only be used in this test

StarpTech: Why can't we implement the test in a way that it respect a graceful shutdown of the router? Can't you wait until Start unblocked?

endigma: I can, done in latest commit

PR #1905, opened 2025-05-27
feat: gRPC plugin enhancements

Noroth:
## Motivation and Context

This pr provides integration tests and enhancements to the current plugin implementation.
Plugins will automatically restart when the forked process dies.

Using a plugin when it was disabled in the config will now return a proper error message.

## Correlated engine release

https://github.com/wundergraph/graphql-go-tools/releases/tag/v2.0.0-rc.182

## Checklist


router-tests/router_plugin_test.go
```diff
@@ -0,0 +1,116 @@
...
+			RouterConfigJSONTemplate: testenv.ConfigWithPluginsJSONTemplate,
+			Plugins: testenv.PluginConfig{
+				Enabled: true,
+				Path:    "../router/plugins",
+			},
+		},
+			func(t *testing.T, xEnv *testenv.Environment) {
+				xEnv.MakeGraphQLRequestOK(testenv.GraphQLRequest{
+					Query: `query { killService }`, // this will kill the plugin
+				})
```
endigma, line 108: Possible to show that it's dead inside the test?

router/__schemas/compose.sh
```diff
@@ -1,15 +1,9 @@
...
+# export COSMO_API_KEY=cosmo_669b576aaadc10ee1ae81d9193425705
+# export COSMO_API_URL=http://localhost:3001
 
-pnpx tsx --env-file ../../cli/.env ../../cli/src/index.ts router plugin build ../plugins/hello-world --debug
 
+#pnpm --reporter=silent wgc router compose -i ../router/_schemas/issue-167/graph.yaml > ../router/_schemas/issue-167/config.json
+#pnpm wgc router compose -i ../router/__schemas/graph.yaml -o ../router/__schemas/config.json
 pnpx tsx --env-file ../../cli/.env ../../cli/src/index.ts router compose -i graph.yaml -o config.json
-
-pnpx tsx --env-file ../../cli/.env ../../cli/src/index.ts router plugin test ../plugins/hello-world
```
endigma, line 15: Should this be as it is? With the commented out commands and etc?

router/pkg/routerplugin/grpc_plugin.go
```diff
@@ -120,43 +112,92 @@ func (s *GRPCPlugin) Start(ctx context.Context, logger *zap.Logger) error {
...
+// Name implements Plugin.
+func (p *GRPCPlugin) Name() string {
+	return p.pluginName
+}
+
+// Start implements Plugin.
+func (p *GRPCPlugin) Start(ctx context.Context, logger *zap.Logger) error {
+	if logger == nil {
+		logger = zap.NewNop()
+	}
```
endigma: When should this argument be allowed to be nil and not be programmer error? Shouldn't they pass zap.NewNop at the call site if this is the desired behaviour?

router/pkg/routerplugin/grpc_plugin.go
```diff
@@ -120,43 +112,92 @@ func (s *GRPCPlugin) Start(ctx context.Context, logger *zap.Logger) error {
...
 	}
 
+	go func() {
+		for {
+			select {
+			case <-p.done:
+				return
+			default:
+				p.handlePluginExit()
+				time.Sleep(time.Second * 2)
```
endigma: Comment explaining the sleep?

router/pkg/routerplugin/grpc_plugin_client.go
```diff
@@ -16,27 +18,114 @@ type GRPCPluginClient struct {
...
+					return false, err
+				}
+
+				if err := clientProtocol.Ping(); err != nil {
+					time.Sleep(g.config.PingInterval)
+					return true, nil
+				}
+
+				return false, nil
+			}()
```
endigma: Can we factor this out into a function like, say "CheckPluginHealth" or "PingPlugin" ?

router/pkg/routerplugin/grpc_plugin_client.go
```diff
@@ -16,27 +18,114 @@ type GRPCPluginClient struct {
...
+		}
+	}
+
+	if g.isClosed.Load() {
 		return status.Error(codes.Unavailable, "plugin is not active")
 	}
 
+	g.mu.RLock()
+	defer g.mu.RUnlock()
 	return g.cc.Invoke(ctx, method, args, reply, opts...)
```
endigma, line 130: I would suggest not returning the call result directly and perhaps wrapping the error, this is confusing on first read where defer is involved

router/pkg/routerplugin/grpc_plugin_client.go
```diff
@@ -45,6 +134,14 @@ func (g *GRPCPluginClient) NewStream(ctx context.Context, desc *grpc.StreamDesc,
...
 }
 
+func (g *GRPCPluginClient) IsPluginProcessExited() bool {
+	if g.pc == nil {
+		return true
+	}
+
+	return g.pc.Exited()
+}
+
```
endigma: `g.pc == nil || g.pc.Exited()`?

endigma (changes requested):

router/__schemas/compose.sh
Noroth: It's basically a rollback to the original script. I added an extra script file to handle the plugin itself if needed but we don't necessarily need a plugin for the demo.

router/pkg/routerplugin/grpc_plugin_client.go
Noroth: No in this case we shouldn't do anything with the error as the whole handling for this is done in the engine.

endigma: Can we do this instead then? 
```go
g.mu.RLock()
defer g.mu.RUnlock()

if err := g.cc.Invoke(ctx, method, args, reply, opts...); err != nil {
    return err
}

return nil
```

Something spooky about using the mutex and defer in such a terse way I think, doesn't read clearly

router/pkg/routerplugin/grpc_plugin_client.go
```diff
@@ -70,39 +70,40 @@ func (g *GRPCPluginClient) waitForPluginToBeActive() error {
...
-			if !shouldContinue {
+			if isActive {
 				return nil
 			}
 		}
 	}
 }
 
+// isPluginActive checks if the plugin is active by pinging it.
+// Returns true
```
endigma: Incomplete comment?

router/pkg/routerplugin/grpc_plugin_client.go
Noroth: I prefer to leave it as it is.

router-tests/router_plugin_test.go
Noroth: Added an assertion but I need to check if I should do that because it is a timing thing and could potentially make the test flaky

endigma (approved):

demo/Makefile
```diff
@@ -2,4 +2,29 @@ generate:
...
 
 bump-deps:
-	./bump-deps.sh
\ No newline at end of file
+	./bump-deps.sh
+
+PHONY: plugin-build plugin-generate plugin-build-binary plugin-build-integration
+
+plugin-build:
+	pnpx tsx --env-file ../cli/.env.example ../cli/src/index.ts router plugin build ./pkg/subgraphs/projects --debug --go-module-path github.com/wundergraph/cosmo/demo/pkg/subgraphs/projects
```
StarpTech: We shouldn't point to the example env. This is just a template.

router-tests/router_plugin_test.go
```diff
@@ -0,0 +1,125 @@
...
+	t.Run("Should fail on startup when no plugins found at a path", func(t *testing.T) {
+		t.Parallel()
+		testenv.FailsOnStartup(t, &testenv.Config{
+			RouterConfigJSONTemplate: testenv.ConfigWithPluginsJSONTemplate,
+			Plugins: testenv.PluginConfig{
+				Enabled: true,
+				Path:    "./non-existing-path",
+			},
+		}, func(t *testing.T, err error) {
+			require.Error(t, err)
```
StarpTech: This might produce false / positives in the future. I'd at least check the error message or type.

router/pkg/routerplugin/grpc_plugin.go
```diff
@@ -120,43 +117,89 @@ func (s *GRPCPlugin) Start(ctx context.Context, logger *zap.Logger) error {
...
 	}
 
+	go func() {
+		for {
+			select {
+			case <-p.done:
+				return
+			default:
+				p.ensureRunningPluginProcess()
+				// Sleep for 2 seconds to avoid busy-waiting
```
StarpTech: I don't understand the comment. You mean to over utilize the CPU? I think this can be implemented more idiomatic with a timer.

router/pkg/routerplugin/grpc_plugin.go
```diff
@@ -120,43 +117,89 @@ func (s *GRPCPlugin) Start(ctx context.Context, logger *zap.Logger) error {
...
 
 // Stop implements Plugin.
-func (s *GRPCPlugin) Stop() error {
-	s.mu.Lock()
-	defer s.mu.Unlock()
+func (p *GRPCPlugin) Stop() error {
+	p.mu.Lock()
+	defer p.mu.Unlock()
 
-	if s.disposed.Load() {
```
StarpTech, line 136: Should be placed after Load since load can return early. I always try to lock as few as possible. But since this is not a hot path, you can also keep it as it.

router/pkg/routerplugin/grpc_plugin_client.go
```diff
@@ -16,27 +18,115 @@ type GRPCPluginClient struct {
...
+
+	return true, nil
+}
+
+func (g *GRPCPluginClient) setClients(pc *plugin.Client, cc grpc.ClientConnInterface) {
+	// We need to lock here to avoid race conditions
+	// We potentially access the plugin clients during invokes
+	g.mu.Lock()
+	defer g.mu.Unlock()
+	g.pc = pc
```
StarpTech: Can we be more descriptive with the variable names?

router/pkg/routerplugin/grpc_plugin_client.go
```diff
@@ -16,27 +18,115 @@ type GRPCPluginClient struct {
...
+		return false, nil
+	}
+
+	clientProtocol, err := g.pc.Client()
+	if err != nil {
+		return false, err
+	}
+
+	if err := clientProtocol.Ping(); err != nil {
+		time.Sleep(g.config.PingInterval)
```
StarpTech, line 100: Why is this necessary?

router/pkg/routerplugin/grpc_plugin_client.go
```diff
@@ -16,27 +18,115 @@ type GRPCPluginClient struct {
...
+	if g.pc == nil {
+		return false, nil
+	}
+
+	clientProtocol, err := g.pc.Client()
+	if err != nil {
+		return false, err
+	}
+
+	if err := clientProtocol.Ping(); err != nil {
```
StarpTech, line 99: Can this hang?

router/pkg/routerplugin/grpc_plugin_client.go
```diff
@@ -16,27 +18,115 @@ type GRPCPluginClient struct {
...
 	}, nil
 }
 
+func (g *GRPCPluginClient) waitForPluginToBeActive() error {
+	timeout := time.After(g.config.ReconnectTimeout)
+	for {
+		select {
+		case <-timeout:
+			return errors.New("plugin was not active in time")
+		default:
```
StarpTech, line 72: I think this has the potential for a busy loop, when the reconnect timeout are seconds.

router/pkg/routerplugin/grpc_plugin.go
Noroth: I already changed it to time.After

router/pkg/routerplugin/grpc_plugin.go
Noroth: Yeah good point. I move the lock

router/pkg/routerplugin/grpc_plugin_client.go
Noroth: I'm not sure honestly. Technically yes but so far I couldn't provoke it. I would immediately get an error if the client can't connect to the endpoint.

The underlying implementation provides context.Background to the health endpoint in the `go-plugin` library.

router/pkg/routerplugin/grpc_plugin_client.go
Noroth: To make sure we don't have a busy loop

demo/Makefile
Noroth: Good point it was a leftover from my CI tests because when using wgc router compose it wouldn't do anything without being logged in to cosmo.

router/pkg/routerplugin/grpc_plugin_client.go
Noroth: I changed to outer loop to have an interval

router/pkg/routerplugin/grpc_plugin_client.go
Noroth: I moved the interval from the inner function to the loop

StarpTech (approved):
LGTM

faculbsz-ah:
https://github.com/wundergraph/cosmo/actions/runs/15487422191/job/43605152475#step:8:360

Some router binaries cant be built and released with this pr

@Noroth @StarpTech @endigma

endigma:
Hiya @faculbsz-ah, good catch! Fix is coming in https://github.com/wundergraph/cosmo/pull/1942.

Issue #1915, opened 2025-05-29
Support setting topologySpreadConstraints via Helm chart


lachlan-smith:
### Component(s)

helm

### Is your feature request related to a problem? Please describe.

Can't set topology spread constraints on the router when installed via the Helm chart.

### Describe the solution you'd like

Allow setting topologySpreadConstraints via Helm

lachlan-smith:
Fix proposed here: https://github.com/wundergraph/cosmo/pull/1916

jakeschurch:
@Noroth my org is on paid tier and is looking to utilize this feature. 

I would be more than happy to build off of [prior art](https://github.com/wundergraph/cosmo/pull/1916) and put in a PR to close out issue.

PR #1916, opened 2025-05-29
fix(helm): add ability to set topologySpreadConstraints on router Helm chart

lachlan-smith:
## Motivation and Context

The Helm chart currently doesn't support setting topologySpreadConstraints in the router chart. This PR adds the ability to set this.

## Checklist

- [ x ] I have discussed my proposed changes in an issue and have received approval to proceed.
- [ x ] I have followed the coding standards of the project.
- [ - ] Tests or benchmarks have been added or updated.
- [ x ] Documentation has been updated on [https://github.com/wundergraph/cosmo-docs](https://github.com/wundergraph/cosmo-docs).
- [ x ] I have read the [Contributors Guide](https://github.com/wundergraph/cosmo/blob/main/CONTRIBUTING.md).

### DIFF
--- a/helm/cosmo/charts/router/templates/deployment.yaml
+++ b/helm/cosmo/charts/router/templates/deployment.yaml
@@ -237,6 +237,10 @@ spec:
       tolerations:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      {{- if .Values.topologySpreadConstraints }}
+      topologySpreadConstraints:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
       {{- if .Values.priorityClassName }}
       priorityClassName: {{ .Values.priorityClassName }}
       {{- end }}
--- a/helm/cosmo/charts/router/values.yaml
+++ b/helm/cosmo/charts/router/values.yaml
@@ -142,6 +142,8 @@ nodeSelector: {}
 
 tolerations: []
 
+topologySpreadConstraints: []
+
 affinity: {}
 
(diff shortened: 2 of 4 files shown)


lachlan-smith:
Addresses https://github.com/wundergraph/cosmo/issues/1915

helm/cosmo/charts/router/templates/deployment.yaml
```diff
@@ -237,6 +237,10 @@ spec:
       tolerations:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      {{- if .Values.topologySpreadConstraints }}
+      topologySpreadConstraints:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
```
Noroth, line 243: This is not working the way you would expect it.

You are checking for `if`, which means the root of your scope is the root of your template. (`Capabilities, Chart, HelmVersion, etc`)
if you do `toYaml .` then you will have very random yaml values in there.

You can test this by adding a topology constraint and run `helm template foo .`

If you want to do it like this you need `{{ with }}` because that will change the root context of values to the one provided in `with`

helm/cosmo/charts/router/values.yaml
```diff
@@ -142,6 +142,8 @@ nodeSelector: {}
 
 tolerations: []
 
+topologySpreadConstraints: []
```
Noroth, line 145: Not a must, but I think including a link to the docs could be nice here.
We have done it in other places
https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/

PR #1920, opened 2025-05-30
feat: link organization to the root keycloak group

wilsonrivera:
## Motivation and Context

The motivation for this feature is guarantee that we are always referring to the correct Keycloak group when performing operations against the organization root group.

## Checklist


controlplane/src/core/bufservices/organization/leaveOrganization.ts
```diff
@@ -49,7 +49,7 @@ export function leaveOrganization(
     }
 
     // the creator of the personal org cannot leave the organization.
-    if (org.creatorUserId === (authContext.userId || req.userID)) {
+    if (org.creatorUserId === (req.userID || authContext.userId)) {
```
wilsonrivera, line 52: I have inverted this as the `authContext.userId` is always present so, if the `userId` is provided on the request, it was never being looked at

controlplane/src/core/bufservices/organization/leaveOrganization.ts
```diff
@@ -36,7 +36,7 @@ export function leaveOrganization(
 
     const orgMember = await orgRepo.getOrganizationMember({
       organizationID: authContext.organizationId,
-      userID: authContext.userId || req.userID,
+      userID: req.userID || authContext.userId,
```
wilsonrivera, line 39: I have inverted this as the `authContext.userId` is always present so, if the `userId` is provided on the request, it was never being looked at

controlplane/src/core/bufservices/organization/leaveOrganization.ts
```diff
@@ -72,27 +72,20 @@ export function leaveOrganization(
...
       throw new Error(`Organization group '${org.slug}' not found`);
     }
 
     // removing the group from the keycloak user
-    await opts.keycloakClient.client.users.delFromGroup({
-      id: orgMember.userID,
-      groupId: organizationGroup[0].id!,
+    await opts.keycloakClient.removeUserFromOrganization({
+      userID: orgMember.userID,
+      groupId: org.kcGroupId!,
```
JivusAyrus: do we need the '!' as we check if it exists above

controlplane/src/core/bufservices/organization/updateOrganizationDetails.ts
```diff
@@ -90,20 +90,13 @@ export function updateOrganizationDetails(
...
-
-      if (organizationGroup.length === 0) {
+      if (!org.kcGroupId) {
         throw new Error(`Organization group '${org.slug}' not found`);
       }
 
       await opts.keycloakClient.client.groups.update(
         {
-          id: organizationGroup[0].id!,
+          id: org.kcGroupId!,
```
JivusAyrus: same here

controlplane/src/core/bufservices/user/removeOrganizationMember.ts
```diff
@@ -91,12 +91,13 @@ export function removeOrganizationMember(
     }
 
     await opts.keycloakClient.authenticateClient();
-
-    await opts.keycloakClient.removeUserFromOrganization({
-      realm: opts.keycloakRealm,
-      userID: user.id,
-      groupName: org.slug,
-    });
+    if (org.kcGroupId) {
```
JivusAyrus: we should return an error if the kcGroupId doesnt exist right? Because, if it doesnt exist it will tell the admin that the user is removed but he/she is not removed from keycloak

controlplane/src/core/controllers/scim.ts
```diff
@@ -291,21 +310,6 @@ const plugin: FastifyPluginCallback<ScimControllerOptions> = function Scim(fasti
       }
 
       // fetching the org from keycloak
-      const organizationGroups = await opts.keycloakClient.client.groups.find({
```
JivusAyrus, line 294: should we check if a group with kcGroupId exists in keycloak?

controlplane/src/core/services/Keycloak.ts
```diff
@@ -269,27 +269,10 @@ export default class Keycloak {
...
-    const orgGroup = orgGroups.find((group) => group.name === organizationSlug);
-
-    if (!orgGroup) {
-      return;
-    }
-
-    await this.client.groups.del({
-      id: orgGroup.id!,
+  public deleteGroupById({ realm, groupId }: { realm?: string; groupId: string }) {
+    return this.client.groups.del({
```
JivusAyrus: will this throw if the group doesnt exist?

controlplane/src/core/services/Keycloak.ts
```diff
@@ -430,28 +392,18 @@ export default class Keycloak {
...
-    if (organizationGroup.length === 0) {
-      throw new Error(`Organization group '${groupName}' not found`);
-    }
-
-    const subGroups = await this.fetchAllSubGroups({
-      realm: realm || this.realm,
-      kcGroupId: organizationGroup[0].id!,
-    });
+    // Delete from the root organization group
+    await this.client.users.delFromGroup({ id: userID, groupId, realm: realm || this.realm });
```
JivusAyrus, line 419: same here

controlplane/src/core/services/Keycloak.ts
wilsonrivera: Yes it does throws

controlplane/src/core/services/Keycloak.ts
wilsonrivera: Yes, this one also throws

controlplane/src/core/bufservices/user/removeOrganizationMember.ts
wilsonrivera: If either the group or user doesn't exist in Keycloak, this would throw an exception so the removal would fail.

During the group migration, all organizations were linked to the corresponding Keycloak group and the organization creation was updated to link as soon as created, checking whether the `kcGroupId` has a value here is just a safety check.

JivusAyrus: Yep, what I meant was to return if it doesnt

controlplane/src/core/services/Keycloak.ts
JivusAyrus: Should we have a try catch and return a meaningful error?

JivusAyrus (changes requested):

controlplane/src/core/services/Keycloak.ts
wilsonrivera: I have added some try...catch where it doesn't really matter if it fails, however there are a handful of places where this could be called by the user, but I don't think adding a custom message would improve anything.

JivusAyrus: We can log here, so we find out that it errored out

JivusAyrus (approved):

PR #1924, opened 2025-06-02
feat(core): add GraphQLServerStart/Stop hooks and parameters

kaialang:
## Motivation and Context

The PR adds:
1.  A pair of router lifecycle hooks:
**ServerStartHook**: fires once when GraphQL server starts, just before it begins listening for GraphQL requests. It allows:
- Adding customer reactions to directives
- Preload query plan/normalize caches
- Pre-compile frequently used POs

**ServerStopHook**: fires once when GraphQL server stops, after it has stopped accepting new operations. It allows:
- Tear down GraphQL-specific resources, close caches
- Finish/cancel live subscriptions before process exit
- Log metrics/finish spans

2. A new **GraphQLServerParams** exposed at those hooks, it meant to be minimal, we can more fields as more use cases come in.

3. Wires the hooks:
- GraphQLServerStartHook in graphql_server.go where a new graph server is initiated right before it starts.
- GraphQLServerStopHook in graphql_server.go where the graph server shutdowns, and if/when newGraphServer returns error.

## Checklist


router/core/graph_server.go
```diff
@@ -221,6 +226,22 @@ func newGraphServer(ctx context.Context, r *Router, routerConfig *nodev1.RouterC
 
 	httpRouter := chi.NewRouter()
 
+	// open core module system: run the graphql server start hooks
+	// after the server is initialized, it allows modules to pre-warm caches,
+	// register additional directive handlers, adjust planner options, etc.
+	r.logger.Debug("Firing GraphQLServerStart hooks")
+	params := &GraphQLServerParams{
```
StarpTech, line 233: I think it would be better if we worked with interfaces here and started exposing only things based on real-world use cases, to avoid exposing internals that could introduce breaking changes along the way.

router/core/graph_server.go
```diff
@@ -221,6 +226,22 @@ func newGraphServer(ctx context.Context, r *Router, routerConfig *nodev1.RouterC
 
 	httpRouter := chi.NewRouter()
 
+	// open core module system: run the graphql server start hooks
+	// after the server is initialized, it allows modules to pre-warm caches,
+	// register additional directive handlers, adjust planner options, etc.
+	r.logger.Debug("Firing GraphQLServerStart hooks")
```
StarpTech, line 232: Should we call this per module including the module name?

router/core/graph_server.go
```diff
@@ -1495,6 +1516,16 @@ func (s *graphServer) wait(ctx context.Context) error {
 // After all requests are done, it will shut down the metric store and runtime metrics.
 // Shutdown does cancel the context after all non-hijacked requests such as WebSockets has been handled.
 func (s *graphServer) Shutdown(ctx context.Context) error {
+	// open core module system: run the graphql server stop hooks
+	// after the server stops accepting new requests, before begin waiting for in-flight requests,
+	// or tearing down metrics and other resources.
+	s.logger.Debug("Firing GraphQLServerStop hooks")
+	for _, h := range s.hookRegistry.graphQLServerStopHooks.Values() {
+		if err := h.OnGraphQLServerStop(ctx, s.graphQLServerParams, nil); err != nil {
```
StarpTech, line 1524: I think we should create distinguished structs / interfaces to avoid later the scenario where have fields in both but not all of them are set everywhere.

router/core/entities.go
```diff
@@ -14,6 +17,17 @@ type ApplicationParams struct {
 	Logger *zap.Logger
 }
 
+// GraphQLServerParams is passed to GraphQLServerStartHook/GraphQLServerStopHook
+type GraphQLServerParams struct {
```
StarpTech, line 21: Since this is really the first feedback, I'm going to start a few discussion about the general approach. This feedback can also be applied to the other PRs.

router/core/graph_server.go
```diff
@@ -1495,6 +1516,16 @@ func (s *graphServer) wait(ctx context.Context) error {
 // After all requests are done, it will shut down the metric store and runtime metrics.
 // Shutdown does cancel the context after all non-hijacked requests such as WebSockets has been handled.
 func (s *graphServer) Shutdown(ctx context.Context) error {
+	// open core module system: run the graphql server stop hooks
+	// after the server stops accepting new requests, before begin waiting for in-flight requests,
+	// or tearing down metrics and other resources.
+	s.logger.Debug("Firing GraphQLServerStop hooks")
+	for _, h := range s.hookRegistry.graphQLServerStopHooks.Values() {
```
StarpTech, line 1523: We should document which hooks fail fast and why, and which might be better to always call, e.g., for shutting down resources.

router/core/graph_server.go
```diff
@@ -221,6 +226,22 @@ func newGraphServer(ctx context.Context, r *Router, routerConfig *nodev1.RouterC
...
 	httpRouter := chi.NewRouter()
 
+	// open core module system: run the graphql server start hooks
+	// after the server is initialized, it allows modules to pre-warm caches,
+	// register additional directive handlers, adjust planner options, etc.
+	r.logger.Debug("Firing GraphQLServerStart hooks")
+	params := &GraphQLServerParams{
+		Handler:                    httpRouter,
+		Config:                     routerConfig,
+		Logger:                     r.logger,
```
StarpTech, line 236: The logger should be named. This improves the context of logs in general.

PR #1926, opened 2025-06-03
feat(core): add ApplicationStart/Stop hooks and parameters

kaialang:
…fecycle

## Motivation and Context

The PR implements https://linear.app/wundergraph/issue/MOD-6, it:
- Defines the object to be exposed at Application lifecycle hooks: ApplicationParams and ExitError 
- Integrates hoos into the RouterSupervisor.Start() (stop) flow. ensuring that:
  * OnApplicationStart hooks fire exactly once, immediately after modules are provisioned (via createRouter()), before the first router server starts.
  * OnApplicationStop hooks fire exactly once, on process exit (after the final stopRouter()), even if shutdown was driven by an error.
  * An ExitError is passed into OnApplicationStop, so modules can inspect or emit the error.

## Changes:

Supervisor

- Convert Start() to use a named err error return and defer a cleanup block that always fires OnApplicationStop(ctx, params, &exitErr).
- Populate ExitError.Code/.Err for each early-failure path (loadConfig(), createRouter(), startRouter(), stopRouter()).
- Fire OnApplicationStart(ctx, params) once after createRouter().

Hooks API

- Update ApplicationStopHook signature to accept exitErr *ExitError (containing both code & error) and ApplicationStartHook to accept params *ApplicationParams.

## Checklist

### DIFF
--- a/router/core/entities.go
+++ b/router/core/entities.go
@@ -1,0 +1,27 @@
+package core
+
+import (
+	"github.com/wundergraph/cosmo/router/pkg/config"
+	"go.uber.org/zap"
+)
+
+/* hook specific entities that are implemented by the open core module system */
+
+// ApplicationParams is passed to ApplicationStartHook/ApplicationStopHook
+type ApplicationParams struct {
+	// the global configuration
+	Config *config.Config
+	Logger *zap.Logger
+}
+
+
+/* common entities for the open core module system */
+
+// ExitError is a struct for holding the exit code and error of the router
+type ExitError struct {
+	Code int
+	Err  error
+}
+
+func (e *ExitError) Error() string { return e.Err.Error() }
+  
--- a/router/core/hooks.go
+++ b/router/core/hooks.go
@@ -1,0 +1,245 @@
+package core
+
+import (
+	"context"
+
+	"github.com/wundergraph/cosmo/router/internal/utils"
+)
+
+// Application Lifecycle Hooks
+type ApplicationLifecycleHook interface {
+	ApplicationStartHook
+	ApplicationStopHook
+}
+
+type ApplicationStartHook interface {
+	OnApplicationStart(ctx context.Context)
+}
+
+type ApplicationStopHook interface {
+	OnApplicationStop(ctx context.Context)
+}
+
+// GraphQL Server Lifecycle Hooks
+type GraphQLServerLifecycleHook interface {
+	GraphQLServerStartHook
+	GraphQLServerStopHook
+}
+
+type GraphQLServerStartHook interface {
+	OnGraphQLServerStart(ctx context.Context)
+}
+
+type GraphQLServerStopHook interface {
+	OnGraphQLServerStop(ctx context.Context)
+}
+
+// Router Lifecycle Hooks
+type RouterRequestHook interface {
+	OnRouterRequest(ctx context.Context)
+}
+
+type RouterResponseHook interface {
+	OnRouterResponse(ctx context.Context)
+}
+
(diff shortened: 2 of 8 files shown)


router/core/entities.go
```diff
@@ -0,0 +1,27 @@
...
+	// the global configuration
+	Config *config.Config
+	Logger *zap.Logger
+}
+
+
+/* common entities for the open core module system */
+
+// ExitError is a struct for holding the exit code and error of the router
+type ExitError struct {
```
janeli1, line 21: should we use the existing errorType in core package? https://github.com/wundergraph/cosmo/blob/0236bf238e0861601973a9403425139ed923b0f6/router/core/errors.go

kaialang: We can add errorType as part of ExitError struct. The errorType is internal to graphql_handler, when WriteError() writes to the http response, it maps to standard http error. The router response hook doesn't automatically have the access to the errorType. Will refactor and add it in a followup issue MOD-16.

router/core/hooks.go
```diff
@@ -13,11 +13,11 @@ type ApplicationLifecycleHook interface {
 }
 
 type ApplicationStartHook interface {
-	OnApplicationStart(ctx context.Context)
+	OnApplicationStart(ctx context.Context, params *ApplicationParams) error
 }
 
 type ApplicationStopHook interface {
-	OnApplicationStop(ctx context.Context)
+	OnApplicationStop(ctx context.Context, params *ApplicationParams, exitError *ExitError) error
```
StarpTech, line 20: For future extensibility and as a general design principle, I'd suggest to pass errors from type `error` and then unwrap them if needed.

router/core/supervisor.go
```diff
@@ -89,6 +93,12 @@ func (rs *RouterSupervisor) createRouter() error {
 	rs.routerCtx = routerCtx
 	rs.routerCancel = routerCancel
 
+	rs.hookRegistry = router.coreModuleHooks.hookRegistry
```
StarpTech, line 96: Any reason why we don't run the start and stop hooks inside the router? (As part of router `Start()` and `Shutdown()`)

router/core/supervisor.go
```diff
@@ -133,7 +143,21 @@ var (
...
-func (rs *RouterSupervisor) Start() error {
+func (rs *RouterSupervisor) Start() (err error) {
+	appCtx := context.Background()
+	started := false
+
+	defer func() {
+		// open core module system: run application stop hooks once shutdown is received
+		rs.logger.Debug("Firing ApplicationStop hooks")
+		for _, h := range rs.hookRegistry.applicationStopHooks.Values() {
+			hookErr := h.OnApplicationStop(appCtx, rs.applicationParams, &ExitError{Err: err})
```
StarpTech, line 154: This would give the hook the `appCtx` without any deadline. As above mentioned, I recommend to run them in `Start()` and `Stop()` so they are coupled to the router code and use the right context (Shutdown respects `ShutdownDelay`)

router/core/router.go
```diff
@@ -524,9 +525,30 @@ func NewRouter(opts ...Option) (*Router, error) {
 		}
 	}
 
+	// Initialize the core module system
+	// It needs to be initialized before bootstrap() is called, because we want to make sure that
+	// it is available at the application start hooks.
+	r.logger.Debug("Initializing core module system")
+	if err := r.initCoreModuleSystem(context.Background()); err != nil {
```
StarpTech, line 532: We should call it at `Start` to operate in the start context (Then people/process can cancel the provisioning of modules cleanly).

PR #1929, opened 2025-06-03
feat: circuit breaker implementation

SkArchon:
## Motivation and Context
This will add circuit breakers to the router. Circuit breakers are off by default, including metrics that are off by default.
We group subgraphs not by the name/id, but instead the routing url, and we create a circuit breaker per every single unique routing url. We use the full url, this means that 

```
subgraph1: host.com/url1
subgraph2: host.com/url1
subgraph3: host.com/url2
```
would result in two circuit breakers, since we consider the entire url.

User's can however force subgraph1 and subgraph2 to use two separate circuit breakers by specifying circuit breaker configurations at the per subgraph level.

## Example configuration
**Global config only**
```
traffic_shaping: 
  all: 
    circuit_breaker:
      enabled: true
      num_buckets: 100
      ...
```

**Subgraph disable for subgraph 5**
```
traffic_shaping:
  all:
    circuit_breaker:
      enabled: true
      ...
  subgraphs:
    subgraph5:
      circuit_breaker:
        enabled: false
```

**Subgraph overrides**
Even when the routing url is the same, this will result in separate circuit breakers being created
```
traffic_shaping:
  all:
    circuit_breaker:
      enabled: true
  subgraphs:
    subgraph1:
      circuit_breaker:
        enabled: true
    subgraph2:
      circuit_breaker:
        enabled: true
```

## Options Included
This PR includes a configurable circuit breaker with the following configurations
* **error_threshold_percentage**: The error threshold that needs to be reached to trip / open the circuit breaker.
* **request_threshold**: How many requests are needed in minimum TO START deciding if the circuit breaker should be opened or not.
* **sleep_window**: The window after the circuit breaker tripped that it definitely does not accept any requests
* **half_open_attempts**: How many failed attempts are allowed after the sleep window before preventing any requests
* **required_successful**: How many successful attempts are required to change the circuit breaker from half open to required successful
* **rolling_duration**: The duration of the sliding window
* **num_buckets**: How many buckets are allocated for the sliding window duration, requests are put into each bucket, which will be evicted eventually as the window rolls forward
* **execution_timeout**: Execution timeout, how long the circuit breaker should wait before timing out, these are counted as errors
* **max_concurrent_requests**: The maximum number of concurrent requests that are allowed on the circuit breaker, -1 results in disabling this

In addition to this the PR introduces the following metrics

* **router.circuit_breaker.state**: This shows the state of the circuit breaker, sliced by feature-flag and subgraph.
* **router.circuit_breaker.short_circuits**: This indicates the number of requests that were not even executed because of the open circuit breaker, sliced by feature-flag and subgraph

Docs PR: https://github.com/wundergraph/cosmo-docs/pull/104/files

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Introduced configurable circuit breaker support for subgraph requests, allowing granular control over error thresholds, request limits, and recovery behavior.
  * Added metrics and observability for circuit breaker state and short-circuit events, with support for OpenTelemetry and Prometheus.
  * Enabled per-subgraph circuit breaker configuration via new configuration options and schema.
  * Integrated circuit breaker logic into the request pipeline and HTTP transport for improved resilience.
  * Added circuit breaker management supporting grouping by routing URLs and feature flag scoping.

* **Tests**
  * Added comprehensive integration tests to validate circuit breaker behavior, state transitions, feature flag handling, and metrics emission.
  * Added configuration schema validation tests for circuit breaker settings.
  * Added unit tests for circuit breaker manager and routing URL grouping logic.

* **Documentation**
  * Updated configuration schema and example files to illustrate new circuit breaker settings.


router/internal/traceclient/traceclient.go
```diff
@@ -30,6 +30,7 @@ type ClientTrace struct {
 
 type ClientTraceContextKey struct{}
 type CurrentSubgraphContextKey struct{}
+type CurrentFeatureFlagContextKey struct{}
```
SkArchon: Need to look at moving this and the above SubgraphContextKey to a more appropriate package, ideally context.go in core, however this causes a cyclic dependency tree.

router/internal/circuit/breaker.go
```diff
@@ -0,0 +1,65 @@
...
+
+	// If there is no circuit defined for this subgraph
+	circuit := rt.circuitBreaker.GetCircuitBreaker(cbKey)
+	if circuit == nil {
+		return rt.roundTripper.RoundTrip(req)
+	}
+
+	preRunStatus := circuit.IsOpen()
+
+	err = circuit.Run(context.Background(), func(_ context.Context) error {
```
SkArchon: To check if we should not use `context.Background()` here

router/pkg/metric/circuitbreaker.go
```diff
@@ -0,0 +1,59 @@
...
+	w.metrics.SetCircuitBreakerState(ctx, true, nil, otelmetric.WithAttributes(w.attributes...))
+}
+
+func (w *CircuitBreakerMetricsConfig) ErrShortCircuit(ctx context.Context, _ time.Time) {
+	w.metrics.MeasureCircuitBreakerShortCircuit(ctx, nil, otelmetric.WithAttributes(w.attributes...))
+}
+
+// No-op functions required to satisfy the interface
+// We can add them if these make sense later
+func (w *CircuitBreakerMetricsConfig) Success(_ context.Context, _ time.Time, _ time.Duration) {
```
SkArchon, line 54: The interface is defined by ceb21/circuit

SkArchon:
Todo: check tests, as they seem flaky

router-tests/circuit_breaker_test.go
```diff
@@ -0,0 +1,1024 @@
...
+									if isSuccessRequest.Load() {
+										_, err := w.Write([]byte(successSubgraphJSON))
+										require.NoError(t, err)
+									} else {
+										time.Sleep(5 * time.Second)
+									}
+								})
+							},
+						},
+					},
```
endigma: Is all this breaker parameters not configurable via the normal router config? why do we need to import env, config, etc here just to provided it in routeroptions? this must have some way to come directly from config via a core.X call right?

router-tests/circuit_breaker_test.go
```diff
@@ -0,0 +1,1025 @@
...
+
+			message := xEnv.Observer().FilterMessage("Circuit breaker status changed")
+			require.Equal(t, 2, message.Len())
+
+			contextMap := message.All()[1].ContextMap()
+			require.False(t, contextMap["isOpen"].(bool))
+		})
+	})
+
+	t.Run("verify circuit breaker becoming half open after sleep window", func(t *testing.T) {
```
StarpTech: The test title is duplicated.

router-tests/circuit_breaker_test.go
```diff
@@ -0,0 +1,1025 @@
...
+
+				sendRequest(opts, "", SendFailedRequest)
+
+				// 0/5 successful requests, which means err percentage is over 90% (100%) and 5 requests
+				require.Equal(t, 1, xEnv.Observer().FilterMessage("Circuit breaker status changed").Len())
+			})
+		})
+	})
+
+	t.Run("verify circuit breaker trips separately for feature flag", func(t *testing.T) {
```
StarpTech: Why should feature flag make a difference? Ultimately, we wan't to protect the downstream service regardless how the router interact with it it.

router/internal/circuit/breaker.go
StarpTech: Why should the cb not run within the request context? If you pass req ctx, you also might to check for `ctx.Err()` inside the callback right?

router/internal/circuit/breaker.go
```diff
@@ -0,0 +1,65 @@
...
+		resp, err = rt.roundTripper.RoundTrip(req)
+		return err
+	})
+
+	postRunStatus := circuit.IsOpen()
+
+	logger := rt.loggerFunc(req)
+	if preRunStatus != postRunStatus {
+		logger.Debug("Circuit breaker status changed", zap.String("subgraph", subgraph), zap.Bool("isOpen", postRunStatus))
```
StarpTech: Do we call it `subgraph` ?

router/internal/circuit/manager.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+func (c *Manager) GetCircuitBreaker(name string) *circuit.Circuit {
+	if c == nil {
+		return nil
+	}
+
+	// Since we should not ever write into a map
+	// after the first setup we don't use any read locks
+	if circuitBreaker, ok := c.circuits[name]; ok {
```
StarpTech, line 65: A map is not safe for concurrent access even just for reads. Use `sync.Map` https://victoriametrics.com/blog/go-sync-map/

router/internal/circuit/manager.go
```diff
@@ -0,0 +1,108 @@
...
+		configuration := createConfiguration(opts.BaseConfig)
+		circuitManager.DefaultCircuitProperties = []circuit.CommandPropertiesConstructor{
+			configuration.Configure,
+		}
+	}
+
+	circuits := make(map[string]*circuit.Circuit, len(opts.Subgraphs))
+	for _, sg := range opts.Subgraphs {
+		// Base graph will start with "::"
+		sgCbName := fmt.Sprintf("%s::%s", opts.FeatureFlagName, sg.Name)
```
StarpTech: Please comment from the tests regarding ff.

router/internal/circuit/manager.go
```diff
@@ -0,0 +1,108 @@
+package circuit
+
+import (
+	"fmt"
+	"github.com/cep21/circuit/v4"
+	"github.com/cep21/circuit/v4/closers/hystrix"
+	nodev1 "github.com/wundergraph/cosmo/router/gen/proto/wg/cosmo/node/v1"
+	"github.com/wundergraph/cosmo/router/pkg/config"
```
StarpTech: This pkg should not rely on router config structs. Keep it decoupled.

router/pkg/config/config.schema.json
```diff
@@ -2911,6 +2911,47 @@
...
+        "circuit_breaker": {
+          "type": "object",
+          "description": "The retry configuration. The retry configuration is used to configure the retry behavior for the subgraphs requests. See https://cosmo-docs.wundergraph.com/router/traffic-shaping#automatic-retry for more information.",
+          "additionalProperties": false,
+          "properties": {
+            "enabled": {
+              "type": "boolean",
+              "description": "Enable the circuit breaker"
+            },
+            "request_threshold": {
```
StarpTech, line 3007: Should we enforce safe min/max?

router/internal/circuit/manager.go
```diff
@@ -0,0 +1,108 @@
...
+				circuits[sgCbName] = circuitManager.MustCreateCircuit(sgCbName, configs...)
+			}
+			continue
+		}
+
+		// This will cover the case of if a subgraph is explicitly disabled
+		if sgOptions.Enabled {
+			newConfig := createConfiguration(sgOptions)
+			configs = append(configs, newConfig.Configure(sgCbName))
+			circuits[sgCbName] = circuitManager.MustCreateCircuit(sgCbName, configs...)
```
StarpTech: How likely is it that this can fail based on user input? Should we rather return an error?

router/internal/circuit/manager.go
```diff
@@ -0,0 +1,108 @@
...
+func (c *Manager) IsEnabled() bool {
+	return c != nil && len(c.circuits) > 0
+}
+
+type ManagerOpts struct {
+	BaseConfig              *config.CircuitBreaker
+	SubgraphCircuitBreakers map[string]*config.CircuitBreaker
+	Subgraphs               []*nodev1.Subgraph
+	FeatureFlagName         string
+	MetricStore             metric.Store
```
StarpTech: Should we create a separate interface that only contains the necessary methods?

router/internal/circuit/breaker.go
```diff
@@ -0,0 +1,65 @@
...
+	// If there is no circuit defined for this subgraph
+	circuit := rt.circuitBreaker.GetCircuitBreaker(cbKey)
+	if circuit == nil {
+		return rt.roundTripper.RoundTrip(req)
+	}
+
+	preRunStatus := circuit.IsOpen()
+
+	err = circuit.Run(context.Background(), func(_ context.Context) error {
+		resp, err = rt.roundTripper.RoundTrip(req)
```
StarpTech, line 45: Does use case exists where we don't count an error as a fault? e.g. request context cancellation by the client? [Not counting user error as a fault](https://godoc.org/github.com/cep21/circuit#example-BadRequest)

router/internal/circuit/manager.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/cep21/circuit/v4/closers/hystrix"
+	nodev1 "github.com/wundergraph/cosmo/router/gen/proto/wg/cosmo/node/v1"
+	"github.com/wundergraph/cosmo/router/pkg/config"
+	"github.com/wundergraph/cosmo/router/pkg/metric"
+	"go.opentelemetry.io/otel/attribute"
+)
+
+type Manager struct {
+	// We maintain separate circuit breakers for each subgraph
+	circuits map[string]*circuit.Circuit
```
StarpTech: Why don't you use the factory? [Configuration factories](https://godoc.org/github.com/cep21/circuit#example-CommandPropertiesConstructor)

router/internal/circuit/manager.go
```diff
@@ -0,0 +1,108 @@
...
+			newConfig := createConfiguration(sgOptions)
+			configs = append(configs, newConfig.Configure(sgCbName))
+			circuits[sgCbName] = circuitManager.MustCreateCircuit(sgCbName, configs...)
+		}
+	}
+
+	return &Manager{
+		circuits: circuits,
+	}
+}
```
StarpTech, line 165: Is there nothing to cleanup after we swap the graph server on schema change?

router/internal/circuit/manager.go
endigma: If it's read heavy `sync.Map` is best, however if its small goroutine pool or read/write balanced a normal map with an RWMutex can be faster

router-tests/circuit_breaker_test.go
SkArchon: Feature flags can have separate routing urls for the same subgraph, which means that they can point to a different host which might not be reachable, etc.

router/internal/circuit/manager.go
SkArchon: Feature flags can have separate routing urls, and can fail independently (also tested this).

router/internal/circuit/manager.go
SkArchon: No, they don't really have anything that comes up when searching for "shut down", I verified that the circuits are garbage collected post restart.

router/internal/circuit/breaker.go
SkArchon: Could you elaborate on the question?

router/internal/circuit/manager.go
SkArchon: I initially had a RW, but from what I read its only not safe if there is even one write after you start to read it
> Map access is unsafe only when updates are occurring. As long as all goroutines are only reading—looking up elements in the map, including iterating through it using a for range loop—and not changing the map by assigning to elements or doing deletions, it is safe for them to access the map concurrently without synchronization.
(source: https://go.dev/doc/faq#atomic_maps)

However I think it makes sense to add add RWMutex, mainly because nothing breaks unexpectedly in case something changes down the line. Lmk wdyt.

router/internal/circuit/manager.go
SkArchon: Link seems to be broken, I am assuming you are talking about https://github.com/cep21/circuit?tab=readme-ov-file#configuration-factories?

It's just a thin wrapper that takes in an argument of the circuitName, we also need to pass the feature flag, the code would be the same just inside the factory. Didn't seem to give a big benefit imo.

Also on a separate note the library internally keeps a map, but they recommended to use it in live code, thus I maintain our own separate map.

router/internal/circuit/manager.go
SkArchon: Will return an error now

router/internal/circuit/breaker.go
SkArchon: We don't exclude any errors, all errors will trigger the circuit breaker. We can however when we want to exclude errors, by having the custom error types implement the `cep21/BadRequest` interface and return true.

router/internal/circuit/manager.go
```diff
@@ -0,0 +1,144 @@
...
+type CircuitBreakerConfig struct {
+	Enabled                    bool
+	ErrorThresholdPercentage   int64
+	RequestThreshold           int64
+	SleepWindow                time.Duration
+	HalfOpenAttempts           int64
+	RequiredSuccessfulAttempts int64
+	RollingDuration            time.Duration
+	NumBuckets                 int
+}
```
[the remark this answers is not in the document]
SkArchon: We validate this in the config json.

router-tests/circuit_breaker_test.go
SkArchon: I extract all the subgraphs (feature subgraphs and normal subgraphs) and process them once, each single subgraph gets its own circuit breaker now.

router/internal/circuit/manager.go
SkArchon: Every unique subgraph (that exists in feature flags and not) are given its own unique circuit breaker now. No more per `ff-subgraph`.

router/pkg/metric/noop_metrics.go
```diff
@@ -55,6 +55,11 @@ func (n NoopMetrics) Shutdown(ctx context.Context) error {
 	return nil
 }
 
+func (n NoopMetrics) MeasureCircuitBreakerShortCircuit(ctx context.Context, sliceAttr []attribute.KeyValue, opt otelmetric.AddOption) {
+}
+func (n NoopMetrics) SetCircuitBreakerState(ctx context.Context, state bool, sliceAttr []attribute.KeyValue, opt otelmetric.RecordOption) {
+}
```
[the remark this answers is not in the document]
SkArchon: This is the pattern used already for other similar functions, look at Provider interface and Store interface in metric_store.go

router/core/context.go
```diff
@@ -664,6 +657,7 @@ func NewSubgraphResolver(subgraphs []Subgraph) *SubgraphResolver {
 			Url:       subgraphs[i].Url,
 			UrlString: subgraphs[i].UrlString,
 		}
+		// TODO: What happens here if the URL is duplicated? CLI allows the same URL to be used
```
SkArchon: probably worth investigating separately, created a ticket

router/core/graph_server.go
```diff
@@ -340,7 +370,45 @@ func newGraphServer(ctx context.Context, r *Router, routerConfig *nodev1.RouterC
 	return s, nil
 }
 
-func (s *graphServer) buildMultiGraphHandler(ctx context.Context, baseMux *chi.Mux, featureFlagConfigs map[string]*nodev1.FeatureFlagRouterExecutionConfig) (http.HandlerFunc, error) {
+func (s *graphServer) getRoutingUrlGroupingForCircuitBreakers(routerConfig *nodev1.RouterConfig) (map[string]map[string]bool, error) {
+	routingUrlGroupings := make(map[string]map[string]bool)
+
+	overwrites, err := configureSubgraphOverwrites(routerConfig.GetEngineConfig(), routerConfig.GetSubgraphs(), s.overrideRoutingURLConfiguration, s.overrides, true)
```
SkArchon: Allow routing urls to be overwritten before creating a Cb.

router/pkg/config/config.schema.json
SkArchon: This is done.

router/internal/circuit/manager.go
SkArchon: This is done.

router/internal/circuit/manager.go
SkArchon: This is done.

router/internal/circuit/manager.go
SkArchon: This is done.

router-tests/circuit_breaker_test.go
SkArchon: this has been already done

router/internal/circuit/breaker.go
SkArchon: changed to `subgraph_name`, which aligned with a few other places.

SkArchon:
@coderabbitai pause

router/internal/circuit/breaker.go
SkArchon: Also verified if a request's context is cancelled (e.g. :- deadline exceeded, etc), it will not be treated as an error for the circuit breaker.

router/core/context.go
StarpTech: Can you elaborate?

router/core/graph_server.go
```diff
@@ -340,7 +370,45 @@ func newGraphServer(ctx context.Context, r *Router, routerConfig *nodev1.RouterC
 	return s, nil
 }
 
-func (s *graphServer) buildMultiGraphHandler(ctx context.Context, baseMux *chi.Mux, featureFlagConfigs map[string]*nodev1.FeatureFlagRouterExecutionConfig) (http.HandlerFunc, error) {
+func (s *graphServer) getRoutingUrlGroupingForCircuitBreakers(routerConfig *nodev1.RouterConfig) (map[string]map[string]bool, error) {
```
StarpTech: This requires good testing. Ideally, we can cover it by unit tests or we have to write very comprehensive integration tests.

router/internal/circuit/breaker.go
StarpTech: great, thanks for verifying.

router/internal/circuit/manager.go
```diff
@@ -0,0 +1,177 @@
...
+
+	"github.com/cep21/circuit/v4"
+	"github.com/cep21/circuit/v4/closers/hystrix"
+	"github.com/wundergraph/cosmo/router/pkg/metric"
+	"go.opentelemetry.io/otel/attribute"
+)
+
+// CircuitBreakerConfig defines the configuration for circuit breaker
+// This decouples the circuit package from the config package
+type CircuitBreakerConfig struct {
```
StarpTech, line 16: Worth to add unit tests?

router/pkg/metric/prom_metric_store.go
```diff
@@ -79,6 +79,23 @@ func (h *PromMetricStore) MeasureRequestCount(ctx context.Context, opts ...otelm
 	}
 }
 
+func (h *PromMetricStore) MeasureCircuitBreakerShortCircuit(ctx context.Context, opts ...otelmetric.AddOption) {
```
StarpTech, line 93: Collecting these needs to be configurable and disabled by default for cardinality reasons.

StarpTech (changes requested):

router/core/graph_server.go
SkArchon: Will add.

router/internal/circuit/manager.go
SkArchon: I will take a look, not for the config of course but for the manager.

router/pkg/metric/prom_metric_store.go
SkArchon: Will add.

router/core/context.go
SkArchon: Updated the comment, essentially if the same URL exists for multiple subgraphs, the previous entry in `subgraphsByURL` would be overridden, unsure if it causes any issue, but worth double checking.

router/pkg/metric/metric_store.go
```diff
@@ -163,7 +167,7 @@ type (
 
 // NewStore creates a new metrics store instance.
 // The store abstract OTEL and Prometheus metrics with a single interface.
-func NewStore(opts ...Option) (Store, error) {
+func NewStore(otlpOpts MetricOpts, promOpts MetricOpts, opts ...Option) (Store, error) {
```
SkArchon, line 167: This is the first time we actually enable/disable metrics within the same store, previously it was at the store level.

SkArchon:
Moved this from the PR description, since it was already verbose with the rabbitai description

This PR adds circuit breaker functionality to the router. After evaluating the choices were between
* https://github.com/cep21/circuit
* https://github.com/sony/gobreaker

When looking into and testing sony/gobreaker we found that it uses a fixed window, which means that it does not have a concept of "buckets" and clears the entire counts on every interval.

Whereas `cep21/circuit` is based off of hystrix and uses a rolling window approach. It also provides much more configuration options, whereas sony/gobreaker's requires us to configure the "readyToTripLogic", it's default is
```
  return counts.ConsecutiveFailures > 5
```
(See Linear ticket for more info)

StarpTech (approved):
LGTM

SkArchon:
@coderabbitai resume

SkArchon:
@coderabbitai review

SkArchon:
@coderabbitai pause

SkArchon:
@coderabbitai resolve

PR #1937, opened 2025-06-05
✨feat(auth): allow to insert additional labels in deployment

samifruit514:
## Motivation and Context

We are using some k8s operator that do some special operations based on a particular label in .metadata.labels. When adding the label in commonLabels value, im getting an `MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable` on the spec.selector attibute. It is because the template is rendering the commonLabels in .spec.selector AND .metadata.labels at the same time.

We would like to have a way to just set the .metadata.labels

### DIFF
--- a/helm/cosmo/charts/router/templates/_helpers.tpl
+++ b/helm/cosmo/charts/router/templates/_helpers.tpl
@@ -44,6 +44,15 @@ Create chart name and version as used by the chart label.
 {{- end }}
 
 {{/*
+Additional Labels that are just rendered in metadata.labels
+*/}}
+{{- define "router.additionalLabels" -}}
+{{- range $key, $value := .Values.additionalLabels -}}
+{{ $key }}: {{ quote $value }}
+{{- end }}
+{{- end }}
+
+{{/*
 Common labels
 */}}
 {{- define "router.labels" -}}
--- a/helm/cosmo/charts/router/templates/deployment.yaml
+++ b/helm/cosmo/charts/router/templates/deployment.yaml
@@ -7,6 +7,7 @@ metadata:
     kapp.k14s.io/change-group: "cosmo.apps.router.wundergraph.com/deployment"
   labels:
     {{- include "router.labels" . | nindent 4 }}
+    {{- include "router.additionalLabels" . | nindent 4 }}
 spec:
   {{- if not .Values.autoscaling.enabled }}
   replicas: {{ .Values.replicaCount }}
--- a/helm/cosmo/charts/router/values.yaml
+++ b/helm/cosmo/charts/router/values.yaml
@@ -14,6 +14,9 @@ image:
 # -- Add labels to all deployed resources
 commonLabels: {}
 
+# -- Add labels to deployment metadata.labels
+additionalLabels: {}
+
 deploymentStrategy: {}
 
 imagePullSecrets: []


Noroth:
Hi @samifruit514,
thanks for your contribution. 

Looks good so far but the doc updates are missing and the chart version needs to be bumped.
You can navigate to the helm directory and run `make docs`

samifruit514:
> Hi @samifruit514, thanks for your contribution.
> 
> Looks good so far but the doc updates are missing and the chart version needs to be bumped. You can navigate to the helm directory and run `make docs`

thanks for your quick reply. Just did the changes

helm/cosmo/charts/router/templates/_helpers.tpl
```diff
@@ -43,6 +43,15 @@ Create chart name and version as used by the chart label.
 {{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" }}
 {{- end }}
 
+{{/*
+Additional Labels that are just rendered in metadata.labels
+*/}}
+{{- define "router.additionalLabels" -}}
```
Noroth, line 49: I tested it with multiple labels and I got an error. Seems like for some reason the whitespace trimmer `-` caused it.

Can you check again.

```yaml
additionalLabels: 
  fantasy.unicorn.io/name: rainbow-router
  fantasy.unicorn.io/instance: sparkle-instance
  fantasy.unicorn.io/version: 0.197.1
  fantasy.unicorn.io/managed-by: MagicWand
  fantasy.unicorn.io/component: fairy-dust-router
```

```suggestion
{{- define "router.additionalLabels" }}
```

helm/cosmo/charts/router/templates/_helpers.tpl
```diff
@@ -43,6 +43,15 @@ Create chart name and version as used by the chart label.
 {{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" }}
 {{- end }}
 
+{{/*
+Additional Labels that are just rendered in metadata.labels
+*/}}
+{{- define "router.additionalLabels" }}
+{{- range $key, $value := .Values.additionalLabels -}}
```
Noroth: I'm sorry I made a typo here. I meant to remove the `-` after `.Values.additionalLabels`

also can you make sure locally to run `helm template .` and see if you get any more issues when you have labels in your `values.yaml`

```suggestion
{{- define "router.additionalLabels" -}}
{{- range $key, $value := .Values.additionalLabels }}
```

helm/cosmo/charts/router/templates/_helpers.tpl
samifruit514: Finally, it was not rendering the list properly because I have put a trim at the wrong place so I did the change similar to the `{{- define "router.selectorLabels" -}}` block below

helm/cosmo/charts/router/templates/_helpers.tpl
samifruit514: yeah thanks, thats how I tested it , with `helm template .` but didnt think about testing more than one label 😆 🤦

Already made the changes you are suggesting, not sure why its not there...

PR #1939, opened 2025-06-05
feat(core): add RouterRequest/Response hooks and parameters

kaialang:
## Motivation and Context

The PR introduces two new router‐level hooks that “bracket” the raw HTTP request/response cycle for /graphql endpoints. 
These hooks allow modules to add custom logic at the HTTP layer (e.g. authentication, logging, response sanitization), without stepping into the GraphQL execution pipeline itself. (A follow-up PR will add deeper GraphQL-operation-level hooks.)

The RouterResponseParameter exposes a Controller interface with methods to inspect or alter the httpResponse before it is returned to client. The altered response status code, header and body will be the final response to the client. 
This hides some of the complexity of response handling from the end user. It does come with a trade-off. Because we buffer the handler’s entire response into memory when the RouterReponseHooks are implemented , there is a small overhead—especially for very large responses. 
Please note, for batched operation, one can only alter the envelope-level response. To alter individual operation response, user should use OperationHooks instead.

```
// RouterResponseParams is passed to RouterResponseHook
type RouterResponseParams struct {
    RouterRequestParams
    // the original HTTP response
    HttpResponse *http.Response
    // Use the controller to inspect or mutate the response
    // the mutated response will be the actual response sent to the client
    Controller RouterResponseController
}

// RouterResponseController defines the things the hook can do
type RouterResponseController interface {
    // Get the status code of the response
    GetStatusCode() int
    // Set the status code of the response
    SetStatusCode(newStatusCode int) error
    // Get the body of the response
    GetBody() []byte
    // Set the body of the response
    SetBody(body []byte) error
    // Get the header map of the response
    GetHeaderMap() http.Header
    // Add a key-value pair to the response header map
    AddHeader(key, value string) error
    // Set the value of a key in the response header map
    SetHeader(key, value string) error
}
```

Added integration test for the modules and hooks.

## Checklist

### DIFF
--- a/router-tests/modules_v1/my_custom_module/authorization_module.go
+++ b/router-tests/modules_v1/my_custom_module/authorization_module.go
@@ -1,0 +1,37 @@
+package my_custom_module
+
+import (
+	"errors"
+	"context"
+
+	"github.com/wundergraph/cosmo/router/core"
+)
+
+const authorizationModuleID = "authorizationModule"
+
+type AuthorizationModule struct {
+}
+
+func (m *AuthorizationModule) MyModule() core.MyModuleInfo {
+	return core.MyModuleInfo{
+		ID: authorizationModuleID,
+		New: func() core.MyModule {
+			return &AuthorizationModule{}
+		},
+	}
+}
+
+func (m *AuthorizationModule) Provision(ctx context.Context) error {
+	return nil
+}
+
+func (m *AuthorizationModule) Cleanup() error {
+	return nil
+}
+
+func (m *AuthorizationModule) OnRouterRequest(reqContext core.RequestContext, params *core.RouterRequestParams) error {
+	return errors.New("{\"error\":\"unauthorized\"}")
+}
+
+// interface guard
+var _ core.RouterRequestHook = (*AuthorizationModule)(nil)
--- a/router-tests/modules_v1/my_custom_module/overwrite_response_module.go
+++ b/router-tests/modules_v1/my_custom_module/overwrite_response_module.go
@@ -1,0 +1,49 @@
+package my_custom_module
+
+import (
+    "context"
+
+    "github.com/wundergraph/cosmo/router/core"
+    "go.uber.org/zap"
+)
+
+const myOverwriteResponseModuleID = "myOverwriteResponseModule"
+
+type MyOverwriteResponseModule struct {
+    
+}
+
+func (m *MyOverwriteResponseModule) MyModule() core.MyModuleInfo {
+	return core.MyModuleInfo{
+		ID: myOverwriteResponseModuleID,
+		New: func() core.MyModule {
+			return &MyOverwriteResponseModule{}
+		},
+	}
+}
+
+func (m *MyOverwriteResponseModule) Provision(ctx context.Context) error {    
+	return nil
+}
+func (m *MyOverwriteResponseModule) Cleanup() error {
+	return nil
+}
+
+func (m *MyOverwriteResponseModule) OnRouterRequest(reqContext core.RequestContext, params *core.RouterRequestParams) error {
+    params.Logger.Info("Incoming /graphql request", zap.String("method", params.HttpRequest.Method))
+
+    return nil
+}
+
+func (m *MyOverwriteResponseModule) OnRouterResponse(reqContext core.RequestContext, params *core.RouterResponseParams, exitErr *core.ExitError) error {
+    status := params.Controller.GetStatusCode()
+    params.Logger.Info("Outgoing /graphql response", zap.Int("status", status))
+ 
+    params.Controller.SetStatusCode(202)
+    params.Controller.SetBody([]byte(`{"error":"graphQL partial failure"}`))
+    
+    return nil
+}
+
+// Interface guard
+var _ core.RouterLifecycleHook = (*MyOverwriteResponseModule)(nil)
--- a/router-tests/modules_v1/router_lifecycle_test.go
+++ b/router-tests/modules_v1/router_lifecycle_test.go
@@ -1,0 +1,124 @@
+package my_custom_module
+
+import (
+	"encoding/json"
+	"net/http"
+	"testing"
+
+	"go.uber.org/zap/zapcore"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/wundergraph/cosmo/router-tests/modules_v1/my_custom_module"
+	"github.com/wundergraph/cosmo/router-tests/testenv"
+	"github.com/wundergraph/cosmo/router/core"
+)
+
+func TestRouterLifecycleHooks(t *testing.T) {
+	t.Parallel()
+
+	t.Run("router_lifecycle_hooks_not_called_for_non_graphql_endpoint", func(t *testing.T) {
+		t.Parallel()
+
+		myLoggingModule := &my_custom_module.MyOverwriteResponseModule{}
+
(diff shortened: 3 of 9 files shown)


router-tests/modules_v1/my_custom_module/overwrite_response_module.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+func (m *MyOverwriteResponseModule) OnRouterRequest(reqContext core.RequestContext, params *core.RouterRequestParams) error {
+    params.Logger.Info("Incoming /graphql request", zap.String("method", params.HttpRequest.Method))
+
+    return nil
+}
+
+func (m *MyOverwriteResponseModule) OnRouterResponse(reqContext core.RequestContext, params *core.RouterResponseParams, exitErr *core.ExitError) error {
+    status := params.Controller.GetStatusCode()
```
StarpTech, line 39: Nit picking for naming. I'd like to avoid `Get*` convention. `StatusCode` is more idiomatic.

router-tests/modules_v1/my_custom_module/overwrite_response_module.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+func (m *MyOverwriteResponseModule) Provision(ctx context.Context) error {    
+	return nil
+}
+func (m *MyOverwriteResponseModule) Cleanup() error {
+	return nil
+}
+
+func (m *MyOverwriteResponseModule) OnRouterRequest(reqContext core.RequestContext, params *core.RouterRequestParams) error {
```
StarpTech, line 32: I think we should combine both parameters.

```go
OnRouterRequest(ctx context.Context, core.RouterRequestParams)
```

router-tests/modules_v1/my_custom_module/overwrite_response_module.go
```diff
@@ -0,0 +1,49 @@
...
+    params.Logger.Info("Incoming /graphql request", zap.String("method", params.HttpRequest.Method))
+
+    return nil
```
StarpTech, line 35: What would happen in case of an error?

router/core/entities.go
```diff
@@ -28,6 +28,83 @@ type GraphQLServerParams struct {
 	Logger *zap.Logger
 }
 
+// RouterRequestParams is passed to RouterRequestHook
+type RouterRequestParams struct {
+	// the raw incoming HTTP request
+	HttpRequest *http.Request
+
```
StarpTech, line 35: It should be possible to respond in any request / response hook.

router-tests/modules_v1/my_custom_module/overwrite_response_module.go
```diff
@@ -0,0 +1,49 @@
...
+func (m *MyOverwriteResponseModule) Provision(ctx context.Context) error {    
+	return nil
+}
+func (m *MyOverwriteResponseModule) Cleanup() error {
+	return nil
+}
+
+func (m *MyOverwriteResponseModule) OnRouterRequest(reqContext core.RequestContext, params *core.RouterRequestParams) error {
+    params.Logger.Info("Incoming /graphql request", zap.String("method", params.HttpRequest.Method))
+
```
StarpTech, line 34: How would we communicate to the router to abort after this middleware? Before the user has to call `next()`

Do we maintain `Aborted` flag that can be set?

```go
ctx.ResponseWriter.Header().Set("Content-Type", "application/json")
ctx.ResponseWriter.WriteHeader(status)
json.NewEncoder(ctx.ResponseWriter).Encode(body)
ctx.Aborted = true
```

router/core/router_config.go
```diff
@@ -101,6 +101,7 @@ type Config struct {
 	registrationInfo             *nodev1.RegistrationInfo
 	securityConfiguration        config.SecurityConfiguration
 	customModules                []Module
+	myCustomModules              []MyModule
```
StarpTech, line 104: Why do we need to separate them?

router-tests/modules_v1/my_custom_module/overwrite_response_module.go
```diff
@@ -0,0 +1,49 @@
...
+    return nil
+}
+
+func (m *MyOverwriteResponseModule) OnRouterResponse(reqContext core.RequestContext, params *core.RouterResponseParams, exitErr *core.ExitError) error {
+    status := params.Controller.GetStatusCode()
+    params.Logger.Info("Outgoing /graphql response", zap.Int("status", status))
+ 
+    params.Controller.SetStatusCode(202)
+    params.Controller.SetBody([]byte(`{"error":"graphQL partial failure"}`))
```
StarpTech, line 43: How would we alter an existing response? (e.g. an error from the graphql handler) My expectation is that this hook is called whenever the router responds and the user can intercept it.

A possible solution could be provide our own responseWriter and make the data available through the controller interface `Body()`

router-tests/modules_v1/router_lifecycle_test.go
```diff
@@ -0,0 +1,124 @@
...
+	"github.com/stretchr/testify/require"
+	"github.com/wundergraph/cosmo/router-tests/modules_v1/my_custom_module"
+	"github.com/wundergraph/cosmo/router-tests/testenv"
+	"github.com/wundergraph/cosmo/router/core"
+)
+
+func TestRouterLifecycleHooks(t *testing.T) {
+	t.Parallel()
+
+	t.Run("router_lifecycle_hooks_not_called_for_non_graphql_endpoint", func(t *testing.T) {
```
StarpTech, line 20: We prefer to write test titles in simple english. Go will snake_case it anyway.

router-tests/modules_v1/my_custom_module/overwrite_response_module.go
```diff
@@ -0,0 +1,49 @@
...
+    
+    return nil
```
StarpTech, line 45: How would streaming work? I feel like we should built the API around the writer interface. As close as to the responseWriter.

router-tests/modules_v1/router_lifecycle_test.go
```diff
@@ -0,0 +1,124 @@
...
+				LogLevel: zapcore.DebugLevel,
+			},
+		}, func(t *testing.T, xEnv *testenv.Environment) {
+			res, err := xEnv.MakeGraphQLRequest(testenv.GraphQLRequest{
+				Query:         `query MyQuery { employees { id } }`,
+				OperationName: json.RawMessage(`"MyQuery"`),
+			})
+			require.NoError(t, err)
+
+			requestLog := xEnv.Observer().FilterMessage("Firing RouterRequest hooks")
```
StarpTech, line 91: We should create tests that tests the implementation. Ideally, based on real-world use cases.

StarpTech (changes requested):

router-tests/modules_v1/my_custom_module/authorization_module.go
```diff
@@ -0,0 +1,37 @@
...
+
+func (m *AuthorizationModule) Provision(ctx context.Context) error {
+	return nil
+}
+
+func (m *AuthorizationModule) Cleanup() error {
+	return nil
+}
+
+func (m *AuthorizationModule) OnRouterRequest(reqContext core.RequestContext, params *core.RouterRequestParams) error {
```
StarpTech, line 32: This brings me to another topic: *race conditions.* If we encourage pre/post hooks everywhere and global struct, how can users safely share data? Do we want to move forward with the previous approach and provide safe get/set methods, or should we go with a complete different approach?

PR #1950, opened 2025-06-09
feat: allow for loading of multiple configurations

SkArchon:
## Motivation and Context

This PR Introduces the capability for users to load multiple configuration files. It uses the same field to specify the config path, however if a comma is detected, we would consider it as a list of file names.

The value from the later configuration is used to override the values from the previous configs, however note that slices are not overriden, thus complex slices, especially those that contain unique keys will be overriden.

Potential Breaking Change: If you have an existing config file name which is comma separated, ensure to rename the file to not use commas.

## Checklist


router/pkg/config/config.go
```diff
@@ -327,7 +330,7 @@ type EngineDebugConfiguration struct {
 	PrintOperationEnableASTRefs                  bool `envDefault:"false" env:"ENGINE_DEBUG_PRINT_OPERATION_ENABLE_AST_REFS" yaml:"print_operation_enable_ast_refs"`
 	PrintPlanningPaths                           bool `envDefault:"false" env:"ENGINE_DEBUG_PRINT_PLANNING_PATHS" yaml:"print_planning_paths"`
 	PrintQueryPlans                              bool `envDefault:"false" env:"ENGINE_DEBUG_PRINT_QUERY_PLANS" yaml:"print_query_plans"`
-	PrintIntermediateQueryPlans                  bool `envDefault:"false" env:"ENGINE_DEBUG_PRINT_INTERMEDIATE_QUERY_PLANS" yaml:"print_intermediate_query_plans"`
+	PrintIntermediateQueryPlans                  bool `envDefault:"false" env:"ENGINE_DEBUG_PRINT_INTERMEDIATE_QUERY_PLANS" yaml:"-"`
```
SkArchon: There is no json schema mapping for this

StarpTech: Create one

router/internal/yamlmerge/merge.go
```diff
@@ -0,0 +1,143 @@
...
+//
+//	{"one": 1, "two": 2} + {"one": 42, "three": 3}
+//	== {"one": 42, "two": 2, "three": 3}
+//
+// Sequences are replaced. For example,
+//
+//	{"foo": [1, 2, 3]} + {"foo": [4, 5, 6]}
+//	== {"foo": [4, 5, 6]}
+//
+// In non-strict mode, duplicate map keys are allowed within a single source,
```
StarpTech: This statement needs to corrected. I think we are always strict. We use yamlv3.

router/internal/yamlmerge/merge.go
```diff
@@ -0,0 +1,143 @@
+package yamlmerge
```
StarpTech, line 22: Please add a link to the previous author. This is necessary for MIT license. Please research.

router/pkg/config/config.go
```diff
@@ -1014,57 +1017,111 @@ type LoadResult struct {
 	DefaultLoaded bool
 }
 
-func LoadConfig(configFilePath string) (*LoadResult, error) {
+func LoadConfig(configFilePathString string) (*LoadResult, error) {
```
StarpTech: Please document the logic here and in the CLI flag. Does the last config file overrides the previous values?

router/pkg/config/config.go
```diff
@@ -1014,57 +1017,111 @@ type LoadResult struct {
...
+		}
+	}
+
+	if errs != nil {
+		return nil, fmt.Errorf("errors while loading config files: %w", errs)
+	}
 
-		if err := yaml.Unmarshal(configFileBytes, &cfg.Config); err != nil {
-			return nil, err
+	// If we are using the default config it means we didn't load any config files
```
StarpTech: I don't understand this comment. You can load the default config and have other files specified.

router/pkg/config/config_test.go
```diff
@@ -752,3 +754,294 @@ headers:
...
+
+listen_addr: "localhost:3007"
+`)
+
+		combinedPaths := strings.Join([]string{base, override1, override2}, ",")
+		_, err := LoadConfig(combinedPaths)
+		require.Error(t, err)
+
+		// We check given parts of the error separately since the file path is not predictable
+		require.ErrorContains(t, err, "router config validation error for")
```
StarpTech, line 970: How should a user identify which config file the error came from?

StarpTech (changes requested):

router/internal/yamlmerge/merge.go
SkArchon: We have `gopkg.in/yaml.v3` as an indirect dependency, but we use `goccy/go-yaml` for loading configs. But you are right will update.

router/pkg/config/config.go
SkArchon: If the user does not have a `config.yaml` (default config name) in the file system and also does not specify a path, while an error will be returned when attempting to read the config, it will be ignored and `cfg.DefaultLoaded` is set to true. (This is existing logic). Thus if this is the case we don't need to process any yaml's and we skip.

Updated comment to reflect a bit better.

router/pkg/config/config_test.go
SkArchon: The file name will be attached to the error message, updated the test with another assertion since `config_test` will be part of the file name.

SkArchon:
TODO: Verify / fix to make it work with hot config reloading

router/cmd/main.go
```diff
@@ -25,7 +25,7 @@ import (
 
 var (
 	overrideEnvFlag = flag.String("override-env", os.Getenv("OVERRIDE_ENV"), "Path to .env file to override environment variables")
-	configPathFlag  = flag.String("config", os.Getenv("CONFIG_PATH"), "Path to the router config file e.g. config.yaml")
+	configPathFlag  = flag.String("config", os.Getenv("CONFIG_PATH"), "Path to the router config file e.g. config.yaml, in case the path is a comma separated file list e.g. \"config.yaml,override.yaml\", the configs will be merged")
```
StarpTech: We should leverage `flag.Var` to support passing `-config` multiple times + commas separated list. This can serve as a primitive for future extensions.

```go
package main

import (
	"flag"
	"fmt"
	"strings"
)

// Simple string slice that supports both comma-separated and repeated flags
type configList []string

func (c *configList) String() string {
	return strings.Join(*c, ",")
}

func (c *configList) Set(value string) error {
	// Split by comma and add each part
	for _, path := range strings.Split(value, ",") {
		path = strings.TrimSpace(path)
		if path != "" {
			*c = append(*c, path)
		}
	}
	return nil
}

func main() {
	var configs configList
	
	flag.Var(&configs, "config", "Config file path(s) - supports comma-separated or multiple flags")
	flag.Parse()
	
	fmt.Printf("Config files: %v\n", []string(configs))
	
	// Use the configs
	for i, config := range configs {
		fmt.Printf("Config %d: %s\n", i+1, config)
	}
}
```

router/pkg/config/config_test.go
```diff
@@ -752,3 +754,307 @@ headers:
 	})
 
 }
+
+func TestConfigMerging(t *testing.T) {
```
StarpTech, line 757: Please improve the `router/docs/Configuration.md` that we should prefer `map` usage where possible to make it easier for users to merge configuration. You can highlight on the merge behaviour. Is not desired to use a list when a map is sufficient.

router-tests/integration_test.go
```diff
@@ -52,10 +52,11 @@ func TestSimpleQuery(t *testing.T) {
 }
 
 func TestConfigReload(t *testing.T) {
-	t.Parallel()
 
 	t.Run("Successfully reloads to a valid new configuration file with SIGHUP", func(t *testing.T) {
 		// Can be very slow, compiles the router binary if needed
+
+		t.Setenv("CONFIG_PATH", "demo2.config.yaml")
```
StarpTech: Please create a separate test and communicate that these tests can't be run in parallel.

StarpTech (changes requested):

PR #1952, opened 2025-06-09
fix: babel upgrade for vulnerability

SkArchon:
## Motivation and Context

Solves https://github.com/wundergraph/cosmo/security/dependabot/174

Babel was a transitive dependency to a lot of dependencies and we had multiple versions. I took the approach of removing the dependency and adding it back again, verifying if the pnpmlockfile version matches the actual version on the package.json.

## Checklist


playground/package.json
```diff
@@ -45,7 +45,7 @@
     "@radix-ui/react-scroll-area": "^1.0.5",
     "@radix-ui/react-select": "^1.2.2",
     "@radix-ui/react-separator": "^1.0.3",
-    "@radix-ui/react-slot": "^1.0.2",
+    "@radix-ui/react-slot": "1.0.2",
```
SkArchon: Needed to remove the "^" as running pnpm install otherwise results in the lockfile being updated with a different react-slot version from 1.0.2 to 1.2.0. I removed it to keep the version the same.
Screenshot shows `pnpm install` result with "^".  I believe this is ok, as if someone wants to upgrade explicitly they can change this, if not we need to upgrade the version.

Also applies to the studio dependencies

image

SkArchon: cc: @JivusAyrus @thisisnithin

pnpm-lock.yaml
```diff
@@ -693,7 +693,7 @@ importers:
         specifier: ^1.0.3
         version: 1.0.3(@types/react-dom@18.3.0)(@types/react@18.3.3)(react-dom@18.3.1(react@18.3.1))(react@18.3.1)
       '@radix-ui/react-slot':
-        specifier: ^1.0.2
+        specifier: 1.0.2
```
SkArchon, line 696: This is the lockfile change

playground/package.json
SkArchon: Discussed with @JivusAyrus, we added back the specifier

JivusAyrus (approved):

PR #1953, opened 2025-06-11
feat: add support for remote grpc services

Noroth:
## Motivation and Context

This PR adds support for calling standalone grpc subgraphs.

## Checklist


demo/graph-with-standalone.yaml
```diff
@@ -0,0 +1,41 @@
...
+  - name: mood
+    routing_url: http://localhost:4008/graphql
+    schema:
+      file: ./pkg/subgraphs/mood/subgraph/schema.graphqls
+  - name: countries
+    routing_url: http://localhost:4009/graphql
+    schema:
+      file: ./pkg/subgraphs/countries/subgraph/schema.graphqls
+  - name: projects
+    grpc:
```
StarpTech, line 37: As discussed

```yaml
  - name: remote-subgraph
    routing_url: http://localhost:50051
    grpc:
      schema_file: ./pkg/subgraphs/projects/src/schema.graphql
      proto_file: ./pkg/subgraphs/projects/generated/service.proto
      mapping_file: ./pkg/subgraphs/projects/generated/mapping.json
```

cli/src/commands/grpc/commands/generate.ts
```diff
@@ -0,0 +1,110 @@
...
+  command.description('generate a protobuf schema for a standalone grpc subgraph');
+  command.argument('[name]', 'The name of the proto service');
+
+  command.requiredOption('-i, --input <path-to-input>', 'The GraphQL schema file to generate a protobuf schema from.');
+  command.option(
+    '-o, --output <path-to-output>',
+    'The output directory for the protobuf schema. If not provided, the output directory will be the same as the input file.',
+    '',
+  );
+  command.option('--no-lock', 'Skip generating and using the lock file for deterministic field ordering');
```
StarpTech: If someone wants to start from scratch he should delete the lock file.

cli/src/commands/grpc/commands/generate.ts
```diff
@@ -0,0 +1,110 @@
...
+
+async function generateCommandAction(name: string, options: any) {
+  if (!name) {
+    program.error('A name is required for the proto service');
+  }
+
+  const spinner = Spinner();
+  spinner.start('Generating protobuf schema...');
+
+  try {
```
StarpTech: Could you use the same output format as with plugins?

StarpTech (changes requested):

cli/test/grpc.test.ts
```diff
@@ -0,0 +1,145 @@
...
+  test('should generate proto and mapping files', async () => {
+    const client: Client = {
+      platform: createPromiseClient(PlatformService, mockPlatformTransport()),
+    };
+
+    const program = new Command();
+    program.addCommand(GenerateCommand({ client }));
+
+    const tmpDir = join(tmpdir(), `grpc-test-${Date.now()}`);
+    mkdirSync(tmpDir, { recursive: true });
```
StarpTech: There are also cleanup functions that always run after a test regardless of its outcome.

cli/test/grpc.test.ts
```diff
@@ -0,0 +1,145 @@
...
+
+    const outputFile = join(tmpDir, 'output.txt');
+    writeFileSync(outputFile, 'test');
+
+      await expect(
```
StarpTech: Please test the error otherwise it can fail for anything.

StarpTech (changes requested):

cli/src/commands/grpc/commands/generate.ts
```diff
@@ -0,0 +1,130 @@
...
+import { renderResultTree } from '../../router/plugin/helper.js';
+
+export default (opts: BaseCommandOptions) => {
+  const command = new Command('generate');
+  command.description('generate a protobuf schema for a standalone grpc subgraph');
+  command.argument('[name]', 'The name of the proto service');
+  command.requiredOption('-i, --input <path-to-input>', 'The GraphQL schema file to generate a protobuf schema from.');
+  command.option(
+    '-o, --output <path-to-output>',
+    'The output directory for the protobuf schema. If not provided, the output directory will be the same as the input file.',
```
alepane21: If think you can simplify this. It is outputed with the default, so now the cli prints something like this:

```
-o, --output <path-to-output>  The output directory for the protobuf schema. If not provided, the output directory will be the same as the input file. (default: "")
```
I think you can remove the phrase about the default, so that the output will be something like this:

```
  -o, --output <path-to-output>  The output directory for the protobuf schema. (default: "")
```

cli/src/commands/grpc/commands/generate.ts
```diff
@@ -0,0 +1,130 @@
...
+    '',
+  );
+  command.option(
+    '-p, --package-name <name>',
+    'The name of the proto package. If not provided, the package name will default to "service".',
```
alepane21: same as for the comment about the output option.

cli/src/commands/grpc/commands/generate.ts
```diff
@@ -0,0 +1,130 @@
...
+    'service',
+  );
+  command.option(
+    '-g, --go-package <name>',
+    'The name of the go package. If not provided, the go package name will default to "github.com/wundergraph/cosmo/demo/test".',
```
alepane21: same as for the comment about the output option.

router/pkg/grpcconnector/grpc_remote.go
```diff
@@ -0,0 +1,108 @@
...
+// Name returns the name of the provider.
+func (g *RemoteGRPCProvider) Name() string {
+	return g.name
+}
+
+// Start initializes the gRPC client connection if it hasn't been created yet.
+// It parses the endpoint URL and creates a new insecure gRPC connection.
+func (g *RemoteGRPCProvider) Start(ctx context.Context) error {
+	if g.cc == nil {
+		ep, err := url.Parse(g.endpoint)
```
alepane21: I think that parsing the URL here, and than using the Host below, could be a mistake.
The go gRPC client supports some different behaviors based on protocol. Also we are going to lose the path, if specified.
I think we can just pass down the endpoint.

router/pkg/grpcconnector/grpc_remote.go
```diff
@@ -0,0 +1,108 @@
...
+
+		if err != nil {
+			return fmt.Errorf("failed to parse endpoint: %w", err)
+		}
+
+		clientConn, err := grpc.NewClient(ep.Host, grpc.WithTransportCredentials(insecure.NewCredentials()))
```
alepane21: I see that we haven't specified any options, and I want to be sure that the following points have been dealt, in some ways:
- timeouts
- retry

## Retry
By default the client is going to retry in some cases (https://grpc.io/docs/guides/retry/#retry-configuration) so I think that we should allow to configure this.
Also there is something important to note: it is almost always OK to retry an idempotent operation (eg: READ).
Maybe we are lucky, because we know which one are mutations and which aren't, and we can just ask the user for the configurations of queries and mutations.
Maybe, in the future, we could add a way to specify the retry for each operations.
There should be a way for a user to configure this.

## Timeouts
The timeouts could be set by the service config, that is fetched by the service, or using the config. I'm not sure what happens when no timeout is set. We should not hang on indefinitely waiting for an answer from a service.
There should be a way for a user to configure this.
##

router/core/factoryresolver.go
```diff
@@ -686,8 +686,14 @@ func toGRPCConfiguration(config *nodev1.GRPCConfiguration, pluginsEnabled bool)
 		}
 	}
 
+	// If not plugin config exists we enabled the datasource by default.
+	// If the config is a plugin:
+	//  * enable datasource when plugins are enabled.
+	//  * disable datasource when plugins are not enabled.
```
alepane21: I would change the comment:

> If no plugin config exists, we enable the datasource by default.
> If the config has a plugin:
>   * Enable the datasource when plugins are enabled.
>   * Disable the datasource when plugins are not enabled.

router-tests/router_plugin_test.go
```diff
@@ -13,6 +16,28 @@ import (
 func TestRouterPlugin(t *testing.T) {
 	t.Parallel()
 
+	t.Run("Should successfully start the router when plugins are enabled but not plugins are in the execution config", func(t *testing.T) {
```
alepane21: typo: "but no plugins"

router-tests/router_plugin_test.go
```diff
@@ -13,6 +16,28 @@ import (
...
+		testenv.Run(t, &testenv.Config{
+			Plugins: testenv.PluginConfig{
+				Enabled: true,
+			},
+		}, func(t *testing.T, xEnv *testenv.Environment) {
+			header := http.Header{
+				"Content-Type": []string{"application/json"},
+				"Accept":       []string{"application/json"},
+			}
+			body := []byte(`{"operationName": null,"query":"query Find($criteria: SearchInput!) {findEmployees(criteria: $criteria){id details {forename surname}}}","variables":{"criteria":{"nationality":"GERMAN"}}}`)
```
alepane21: Should we do an operation also? Can't we just verify with testenv.RunWithError return no error and verify that the callback is being called?

cli/src/commands/router/commands/compose.ts
```diff
@@ -87,6 +107,57 @@ type Config = {
...
+      mapping: s.mapping,
+      protoSchema: s.protoSchema,
+      version: s.version,
+      schema,
+      configurationDataByTypeName,
+    };
+    return composedSubgraphPlugin;
+  }
+
+  const composedSubgraphGRPC: ComposedSubgraphGRPC = {
```
alepane21, line 147: I would still check the kind and at the end raise an exception or a null data if the kind is wrong.

router/pkg/grpcconnector/grpc_remote.go
Noroth: Good point. I will add a ticket for that

router/core/factoryresolver.go
Noroth: I think I was tired when I was writing this comment oof

cli/src/commands/router/commands/compose.ts
Noroth: The compiler will do proper validations so we don't need that actually

router-tests/router_plugin_test.go
Noroth: updated

StarpTech (approved):
LGTM

PR #1955, opened 2025-06-11
fix: let netpoll handle closing websockets

endigma:
## Motivation and Context

We had a subtle timing dependent bug due to closing connections from within ws proto handlers, these are closed by netpoll under better conditions when left alone. 

This fixes a bug where it could almost seem as though we had partial reconnect support because the subscription was not entirely removed before clients reconnected, and also fixes an issue where the engine stats were leaking connections.

After this change, when a subgraph goes down triggering a close event, clients are fully disconnected and state is cleared.

Also some unrelated "modernize" linter fixes, just leaving the place in better condition than I found it.

## Checklist

### DIFF
--- a/router/core/websocket.go
+++ b/router/core/websocket.go
@@ -409,7 +409,7 @@ func (h *WebsocketHandler) handleUpgradeRequest(w http.ResponseWriter, r *http.R
 
 		// Export the token from the initial payload to the request header
 		if fromInitialPayloadConfig.ExportToken.Enabled {
-			var initialPayloadMap map[string]interface{}
+			var initialPayloadMap map[string]any
 			err := json.Unmarshal(handler.initialPayload, &initialPayloadMap)
 			if err != nil {
 				requestLogger.Error("Error parsing initial payload: %v", zap.Error(err))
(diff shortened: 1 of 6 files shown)


router/core/websocket.go
```diff
@@ -409,7 +409,7 @@ func (h *WebsocketHandler) handleUpgradeRequest(w http.ResponseWriter, r *http.R
 
 		// Export the token from the initial payload to the request header
 		if fromInitialPayloadConfig.ExportToken.Enabled {
-			var initialPayloadMap map[string]interface{}
```
StarpTech, line 412: Can we add a test that would catched this issue?

StarpTech (changes requested):
Can we add a test that would have catched this issue?

endigma:
> Can we add a test that would have catched this issue?

Yep, I think we can get at it from the Engine Stats side, the actual bug is _very_ hard to characterize or text externally though since it was timing dependent and mostly correct from outside the black box.

StarpTech:
Alright, give it a shot. Otherwise we can also move forward and merge it.

PR #1957, opened 2025-06-11
fix(router): provide request context of original WebSocket Upgrade request to subgraph requests

DerZade:
## Motivation and Context

Up until now requests to subgraphs that result out of WebSocket subscription request cannot use the expression context of the corresponding `Upgrade` request. This means even if authentication succeeds on the `Upgrade` request, the expression `request.auth.isAuthenticated` would evaluate to `false` for the request to the subgraph. In a similar fashion passing claims to the subgraph via headers is not possible as well.

This PR aims to fix this by providing the expression context of the original `Upgrade` request to all resulting requests to subgraphs :) 

image

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Improved support for using authentication context in header expressions for subgraph requests during WebSocket GraphQL subscriptions.

* **Tests**
  * Added a new test to verify that authentication context can be referenced in header expressions and that headers are correctly set in subgraph requests during WebSocket subscriptions.


router/core/websocket.go
```diff
@@ -995,6 +995,7 @@ func (h *WebSocketConnectionHandler) executeSubscription(registration *Subscript
 		w:                   nil,
 		r:                   registration.clientRequest,
 	})
+	reqContext.expressionContext = getRequestContext(h.request.Context()).expressionContext
```
DerZade: My experience with cosmo is limited, so I'm not sure if copying the whole expression context is a good idea here, maybe `Request` or even just `Request -> Auth` is enough.

DerZade:
@SkArchon anything I can do to move this forward? :)

SkArchon:
Hi @DerZade! I will take a look at this PR and get back to you this week

DerZade:
I rebased and added a nil check as CodeRabbit suggested.

I'm just still not sure if copying the auth part of the expression context is enough :thinking:  

I feel like also being able to use other headers (apart from auth) is a benefit. So I think copying the whole context is probably not a bad idea :)

router/core/websocket.go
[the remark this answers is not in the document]
DerZade: Implemented the `nil` check as suggested, but I'm still copying the whole expression context (second suggestion)

router/core/websocket.go
```diff
@@ -985,6 +985,10 @@ func (h *WebSocketConnectionHandler) executeSubscription(registration *Subscript
 		w:                   nil,
 		r:                   registration.clientRequest,
 	})
+
+	if origCtx := getRequestContext(h.request.Context()); origCtx != nil {
+		reqContext.expressionContext = origCtx.expressionContext
```
SkArchon: Could you use the `origCtx.expressionContext.Clone()` method

DerZade: Done in 399a0da6af6c40d39985148124f5ecf7dbdfb9f6 :)

router-tests/websocket_test.go
```diff
@@ -82,6 +83,18 @@ type CountEmpResponse struct {
...
 
+type TestAuthenticator struct{}
+
+func (a TestAuthenticator) Name() string {
+	return "Test"
+}
+
+func (a TestAuthenticator) Authenticate(ctx context.Context, p authentication.Provider) (authentication.Claims, error) {
+	return authentication.Claims{
+		"favorite_animal": "bear",
```
jensneuse: I would prefer if you check other authentication tests and put the claims into the Upgrade Request JWT or as an initial payload, just to make sure that it works end 2 end.

router/core/websocket.go
```diff
@@ -985,6 +985,10 @@ func (h *WebSocketConnectionHandler) executeSubscription(registration *Subscript
 		w:                   nil,
 		r:                   registration.clientRequest,
 	})
+
+	if origCtx := getRequestContext(h.request.Context()); origCtx != nil {
+		reqContext.expressionContext = *origCtx.expressionContext.Clone()
```
jensneuse, line 990: Not sure if we need to check if expressionContext is always set. Can it be nil?

jensneuse (changes requested):
Overall looks like a good PR and useful improvement. Just 2 minor comments.

SkArchon:
I will open up a separate [PR](https://github.com/wundergraph/cosmo/pull/2043) which will run the CI checks, as there are some limitations for CI on forks atm. Also two quick comments,
1. I noticed there was a compilation failure when trying out your change today which you have fixed a bit later, do ensure that the code compiles before requesting a review 🙏 
2. Might be a good idea if you used a separate branch other than main on your fork, especially if you want to contribute again later

router/core/websocket.go
DerZade: At least according to the code: 
https://github.com/wundergraph/cosmo/blob/29d2b7b942c5b1dc250876ba6acb798dc105c52e/router/core/context.go#L288-L305

Don't know if that is actually a case that could happen, but a sanity check won't hurt, right?

router-tests/websocket_test.go
DerZade: Fixed in  898c5bd10d

My initial thought process was that the Authenticator logic was tested thoroughly in other tests, so I wanted to keep it as simple as possible :)

DerZade:
Addressed all comments :) 

Please squash merge, since this could easily fit in one commit and there are some commits without nice messages :see_no_evil:

jensneuse (approved):

SkArchon:
Hi @DerZade we squash merge into main, since we want a clean commit history. In PR commit comments would potentially help reviewing though.

Also there is an issue in CI which my colleagues are looking into, so will need to wait until its fixed to merge this in.

PR #1959, opened 2025-06-12
feat: Support hot reloading of MCP operation files

### LINKED ISSUE #1886: Support hot reloading of MCP operation files
### Component(s)

router

hello, everyone.
currently , hot reloading for MCP operation files is not supported . as a result, i have to restart the Cosmo Router service every time the MCP operation files are updated to ensure the changes take effect .  

when MCP operation files are updated or added , they can be automatically loaded , and there is no need to restart the Cosmo Router. 

_No response_

### Additional context

_No response_

melsonic:
https://github.com/user-attachments/assets/567f3265-9fe3-456e-9877-e8afcafa22c5

## Checklist

/fixes #1886 

## Summary by CodeRabbit

* **New Features**
  * Added hot reload capability for MCP operations, enabling automatic detection and application of changes to operation files without restarting the router.
  * Introduced configuration options to enable or disable hot reload and set the reload interval via YAML, environment variables, and JSON schema.
  * Enhanced file watching to support monitoring entire directories for operation changes.

* **Bug Fixes**
  * Prevented duplicate or unintended overwriting of storage provider configurations when MCP is enabled.

* **Tests**
  * Added integration tests to verify hot reload behavior, operation updates, and to ensure no resource leaks occur during hot reload cycles.

* **Documentation**
  * Updated configuration examples and schema to reflect new hot reload options.


endigma:
Hi @melsonic, make sure to mark this ready for review when you think it's done.

For it to be merged, we will need:

1. A good implementation, of high code quality with good documentation
2. New or updated tests in router or router-tests proving the new functionality works and does not disturb other functions, unfortunately, a screen capture does not work for this, as with every new feature we need a way to prevent regressions or breakage in the future
3. A corresponding PR to https://github.com/wundergraph/cosmo-docs for any new configuration options

For this specifically, we'd need it to be off by default, and use the polling based file watcher only, not inotify. We are aware of inotify and that it is _theoretically_ a better solution for cases like this, but in practice it falls short in many environments so the router instead uses the simpler polling approach.

Thanks!

melsonic:
> Hi @melsonic, make sure to mark this ready for review when you think it's done.
> 
> For it to be merged, we will need:
> 
> 1. A good implementation, of high code quality with good documentation
> 2. New or updated tests in router or router-tests proving the new functionality works and does not disturb other functions, unfortunately, a screen capture does not work for this, as with every new feature we need a way to prevent regressions or breakage in the future
> 3. A corresponding PR to https://github.com/wundergraph/cosmo-docs for any new configuration options
> 
> For this specifically, we'd need it to be off by default, and use the polling based file watcher only, not inotify. We are aware of inotify and that it is _theoretically_ a better solution for cases like this, but in practice it falls short in many environments so the router instead uses the simpler polling approach.
> 
> Thanks!

Hi @endigma, As of now, I’ve implemented this to be applied by default. I’ll add a new config flag to allow changing that behavior.
Regarding the polling watcher behavior: currently, we can only apply a watcher to an existing file. For newly added or deleted files (if not using inotify), I’ll need to implement a new watcher functionality. Here’s the approach I’m considering:

- Create a goroutine that periodically polls the operations directory (e.g., at interval T).
- Trigger a reload if (lastModified - time.Now()) < T.

Let me know your thoughts!

endigma:
I think it should be possible to perhaps extend the polling based watcher to walk globs or directories, and use a map to keep track of seen files

New unseen files or updates to seen files could trigger a reload

melsonic:
Hi @endigma, I've refactored the code to use a polling-based watcher for detecting operation file addition/deletion changes. It's nearly ready. I have one question though.

Should we add a separate HotReloadPollInterval config option for MCP operations (like we have for the main config), or reuse the existing PollInterval?

endigma:
Enabling the watcher, setting its polling interval should be specific, yes

Let me know when you're fully done with it, as per the general merge requirements I posted above as well as you being happy with it, then I'll see about getting it into review internally!

melsonic:
Hi @endigma, this PR is ready for review. 
This PR includes changes to support Hot reloading of MCP Operations and tests to verify the functionality.
I have also created another PR to [cosmo-docs](https://github.com/wundergraph/cosmo-docs/pull/107).
However I would like to bring up a point here related to redis-cluster. It somehow gets into an unhealthy state(verified by a wundergraph member on discord) and hence couldn't verify the functionality of tests related to redis. Apart from that I ran all the tests manually and those were passing .

endigma:
Hi @melsonic, thanks for letting me know! I'll see about getting this into the review queue for next week.

melsonic:
> Hi @melsonic, thanks for letting me know! I'll see about getting this into the review queue for next week.

Sure, Thanks @endigma!

router-tests/mcp_hot_reload_test.go
```diff
@@ -0,0 +1,191 @@
...
+			MCP: config.MCPConfiguration{
+				Enabled: true,
+				HotReloadConfig: config.MCPOperationsHotReloadConfig{
+					Enabled:  true,
+					Interval: 5 * time.Second,
+				},
+			},
+		}, func(t *testing.T, xEnv *testenv.Environment) {
+
+			os.Remove(filePath)
```
endigma: You should use a `t.TempDir` to modify files during tests, not the worktree, they get automatically cleaned up after the test finishes.

router-tests/mcp_hot_reload_test.go
```diff
@@ -0,0 +1,191 @@
...
+			// initial tools count
+			initialToolsCount := len(resp.Tools)
+
+			// create new mcp operation file
+			file, err := os.Create(filePath)
+			assert.NoError(t, err)
+			defer func() {
+				file.Close()
+				os.Remove(filePath)
+			}()
```
endigma: Either use `t.Cleanup` or close it precisely when you're done with it

router/core/graph_server.go
```diff
@@ -1082,6 +1082,16 @@ func (s *graphServer) buildGraphMux(ctx context.Context,
...
+		go func() {
+			for {
+				if reloadOperations := <-s.mcpServer.ReloadOperationsChannel(); reloadOperations {
+					s.logger.Log(zap.InfoLevel, "Reloading mcp server!")
+					if mErr := s.mcpServer.Reload(executor.ClientSchema); mErr != nil {
+						return
+					}
+				}
+			}
+		}()
```
endigma: We need to make sure this goroutine is cleaned up during router shutdown, you should fix it to return when context is done and add a test similar to the one in `shutdown_test`

router/core/router.go
```diff
@@ -848,6 +848,8 @@ func (r *Router) bootstrap(ctx context.Context) error {
 			mcpserver.WithExcludeMutations(r.mcp.ExcludeMutations),
 			mcpserver.WithEnableArbitraryOperations(r.mcp.EnableArbitraryOperations),
 			mcpserver.WithExposeSchema(r.mcp.ExposeSchema),
+			mcpserver.WithHotReload(r.mcp.HotReloadConfig.Enabled),
+			mcpserver.WithHotReloadInterval(r.mcp.HotReloadConfig.Interval),
```
endigma: This should be one option with (bool, time.Duration)

router/pkg/schemaloader/loader.go
```diff
@@ -71,6 +79,35 @@ func (l *OperationLoader) LoadOperationsFromDirectory(dirPath string) ([]Operati
...
+
+			go func() {
+				if err := watchFunc(pathCtx); err != nil {
+					if !errors.Is(err, context.Canceled) {
+						l.Logger.Error("Error watching operations path", zap.Error(err))
+					} else {
+						l.Logger.Debug("Watcher context cancelled, shutting down")
+					}
+				}
+			}()
```
endigma: Leaked goroutine

endigma (changes requested):
It'll need a few more passes as well, but I have a few initial comments

router-tests/mcp_hot_reload_test.go
melsonic: Initially I tried to implement that one as it was more convenient, but had some issues, I'll again look into it & incorporate the same in my next commit.

router/core/graph_server.go
melsonic: Sure, thanks for your input.

router/pkg/schemaloader/loader.go
melsonic: I am passing the same context to all the watchFunctions under the operations directory, so in case any updation, It'll call the contextCancel function and eventually all goroutines will exit upon receiving the signal.

melsonic: There were some misunderstanding from my side, this is also fixed now.

melsonic:
Hi @endigma, please review when you have time.

router/pkg/schemaloader/loader.go
```diff
@@ -71,6 +79,35 @@ func (l *OperationLoader) LoadOperationsFromDirectory(dirPath string) ([]Operati
...
+				if err := watchFunc(ctx); err != nil {
+					if !errors.Is(err, context.Canceled) {
+						l.Logger.Error("Error watching operations path", zap.Error(err))
+					} else {
+						l.Logger.Debug("Watcher context cancelled, shutting down")
+					}
+				}
+			}()
+		}
+
```
endigma: I don't like starting a new goroutine for each new file so deep into the innards of the router. Can we perhaps modify the watcher API to have CRUD for which files it is keeping track of so the same watcher can be used persistently?

It could be very odd having many different watchers on slightly different timers watching each file in the directory.

I believe the watcher has recently been updated to support watching multiple files, another option might be upgrading this to support watching a glob of files using https://pkg.go.dev/path/filepath#Glob ?

It could re-scan the glob each tick and compare with it's store of seen files, instead of having to manually track file creation/deletion.

melsonic: Fair point @endigma :+1: 

Out of the available options, `Glob` options looks solid to me as it covers all the CRUD operations under a single umbrella. And It will also be much easier for us to look for Create/Delete operations using the glob pattern than with the currently available `paths` option. I will update you once I am ready with my changes.

melsonic: Hi @endigma,

It looks like `filepath.Glob(pattern)` doesn't support recursive globbing. 

According to the Cosmo MCP docs:
`All .graphql files in this directory (and subdirectories) will be loaded`
Hence, It should include .graphql files from subdirectories as well.

Would it be okay if I use `filepath.WalkDir` to list the files instead of `filepath.Glob`?

We could also use a shell command like `find ./operations -type f -name "*.graphql"`, but the trade-off is that we'd need to write different scripts for different platforms.

endigma: WalkDir sounds fine, don't shell out just for the sake of globbing

melsonic:
Hi @endigma, I have updated the cosmo watcher to watch CRUD operations in a directory with a file filter. In this scenario, the filefilter would help us to tell the watcher to only watch graphql files.
Please review my approach & let me know if you have any suggestions or updates.
I will also have to add the tests for this added functionality, I'll add them once you are okay with my changes.

Thank you!

router/core/graph_server.go
```diff
@@ -1077,9 +1077,22 @@ func (s *graphServer) buildGraphMux(ctx context.Context,
...
 			return nil, fmt.Errorf("failed to reload MCP server: %w", mErr)
 		}
+		go func() {
+			for {
+				select {
+				case <-s.mcpServer.ReloadOperationsChannel():
+					s.logger.Log(zap.InfoLevel, "Reloading mcp server!")
+					if mErr := s.mcpServer.Reload(ctx, executor.ClientSchema); mErr != nil {
+						return
+					}
```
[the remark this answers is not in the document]
melsonic: Hi @endigma, Please let me know if it is the expected behaviour to further retry mcpServer reload operations even after reload failure.

melsonic:
Hi @endigma, please review this MR once you have time.
Thank you!

endigma:
Hi @melsonic , you can just re-request when its ready, can you resolve the conflict? I'll put this back in the queue for review.

router-tests/mcp_hot_reload_test.go
```diff
@@ -0,0 +1,282 @@
...
+	"go.uber.org/goleak"
+)
+
+func TestMCPOperationHotReload(t *testing.T) {
+	t.Parallel()
+
+	t.Run("List Updated User Operations On Addition and Removal", func(t *testing.T) {
+
+		operationsDir := t.TempDir()
+		storageProviderId := "mcp_hot_reload_test_id"
```
endigma: `storageProviderId` should be `ID` and it seems the same across all subtests so can we move it out to test scope?

router-tests/mcp_hot_reload_test.go
```diff
@@ -0,0 +1,282 @@
...
+			},
+		}, func(t *testing.T, xEnv *testenv.Environment) {
+
+			toolsRequest := mcp.ListToolsRequest{}
+			resp, err := xEnv.MCPClient.ListTools(xEnv.Context, toolsRequest)
+			require.NoError(t, err)
+
+			initialToolsCount := len(resp.Tools)
+
+			filePath := operationsDir + "/main.graphql"
```
endigma: Using `filepath.Join` is best for this sort of thing, and can it be given a more descriptive name? This applies to all subtests.

router-tests/mcp_hot_reload_test.go
```diff
@@ -0,0 +1,282 @@
...
+					RawInputSchema: json.RawMessage(nil),
+					Annotations: mcp.ToolAnnotation{
+						Title:          "Execute operation getEmployeeNotes",
+						ReadOnlyHint:   mcp.ToBoolPtr(true),
+						IdempotentHint: mcp.ToBoolPtr(true),
+						OpenWorldHint:  mcp.ToBoolPtr(true),
+					},
+				})
+
+			}, 15*time.Second, 250*time.Millisecond)
```
endigma: Instead of using `assert.Eventually...` functions, we recently introduced an option to the watcher to directly provide a non-timing based tick source. You can check `watcher_test.go` for how to use it.

router/pkg/watcher/watcher.go
```diff
@@ -36,6 +45,31 @@ func New(options Options) (func(ctx context.Context) error, error) {
...
+				}
+				files = append(files, path)
+				return nil
+			})
+			if err != nil {
+				return []string{}, fmt.Errorf("error walking directory %s: %w", options.Directory.DirPath, err)
+			}
+		}
+		return files, nil
+	}
```
endigma: Can this be extracted and tested separately? Also, "filter" semantics usually return `true` when something should _remain_ in the list, not be removed from it.

router/pkg/watcher/watcher.go
```diff
@@ -26,8 +35,8 @@ func New(options Options) (func(ctx context.Context) error, error) {
 		return nil, errors.New("logger must be provided")
 	}
 
-	if len(options.Paths) == 0 {
-		return nil, errors.New("path must be provided")
+	if len(options.Paths) == 0 && options.Directory.DirPath == "" {
+		return nil, errors.New("paths or directory must be provided")
```
endigma: I think we should error if both `Paths` and `Directory` are provided, and I don't believe filter is necessary at the moment. This will allow you to deduplicate all the loops below and integrate Directories as just a source for a list of paths instead of separate handling.

router/pkg/watcher/watcher.go
```diff
@@ -45,6 +79,23 @@ func New(options Options) (func(ctx context.Context) error, error) {
 		}
 
 		prevModTimes := make(map[string]time.Time)
+		seenDirFilePaths := make(map[string]struct{})
```
endigma: What purpose does this serve? You maintain it as an index of the currently known file paths but don't use it for anything that prevModTimes can't be used for from what I can tell. We can already detect deleted files by looping over the prevModTimes map and comparing it with os.Stat result of the expected file. We already handle this by resetting the time so should it be created again it will be detected as an update.

endigma (changes requested):
Hi, some high level review:

- You should add unit tests for the new watcher behaviour in `watcher_test.go`
- We don't need to support `Directory` and `Paths` operating modes at the time same, we can use `Directory` as a way to build a list of paths or use a provided one, and we can rebuild the list of tracked files on every loop for the directory mode.

router-tests/mcp_hot_reload_test.go
melsonic: Yes, we can.

router-tests/mcp_hot_reload_test.go
melsonic: Sure, I'll check and update the test.

router/pkg/watcher/watcher.go
melsonic: Okay, I'll create a util function outside `New` with unit test.

router/pkg/watcher/watcher.go
melsonic: Thanks for clarifying the required design. It would definitely prevent duplicate runs from out function.

router/pkg/watcher/watcher.go
melsonic: I'll look into it.

melsonic:
Hi @endigma, I have updated the PR incorporating your comments. However, I do have these following questions.
1. Should we throw an error in case both `options.Paths` & `options.Directory` is empty?
2. Since we are only allowing either `Paths` or `Directory`, in case the newly defined function `ListDirFilePaths` returns 0 file paths, should we also throw an error from our watcher function?

router/pkg/watcher/watcher.go
```diff
@@ -67,21 +109,36 @@ func New(options Options) (func(ctx context.Context) error, error) {
...
+					_, seen := prevModTimes[path]
+
+					// Detects new files & existing file updates
+					if !seen || stat.ModTime().After(prevModTimes[path]) {
 						prevModTimes[path] = stat.ModTime()
 						changesDetected = true
 					}
 				}
 
+				if options.Directory.DirPath != "" {
```
melsonic: I am reloading the updated directory files after going through the previously listed files. The will allow us to remove deleted files from `prevModTimes` in one loop. However there is a catch, in case of a new file addition, instead of 2 ticks, it will consume 3 ticks to run the provided Callback func.

On the other hand, in case we list the updated directory files before going through `options.Paths`, we'll have to run another loop on prevModTimes to remove keys for files, that are deleted from the directory.

Currently I have implemented the first approach, let me know if the team prefers the 2nd option.

router-tests/mcp_hot_reload_test.go
melsonic: In case of mcp hot reload, we don't have direct access to the watcher config options, the watcher is defined in the `loader.go` file. So, in this particular scenario it is not possible to use non-timing based tick source.

router/pkg/watcher/watcher.go
endigma: I would personally choose option 2, looping over a map and updating it is not terribly slow and hot reloading is nowhere near a hot path. More intuitive or consistent reload behavior is more important than raw speed of a single tick handler

endigma:
Hi @melsonic, I think having 0 paths to check is fine, provided we defensively prevent anything bad from happening in this case. This would allow gracefully handling a multi-tick wide total directory swap or something like a human speed drag-out -> drag-in operation.

Generally with watcher code we prefer resilience over all, e.g not failing out when we hit an invalid state and instead just falling back to a previous working state until some iteration is usable. In this case I believe having mcp enabled but with 0 operations should not be a failure case, but I'm not sure at the moment

endigma:
And yes, watcher constructor should fail if the arguments themselves are both empty

melsonic:
Hi @endigma, I am ready with my changes, please review it once you have time.

melsonic:
Hi @endigma, Please review.

P.S: If you are busy with some other high priority task, please take your time. Commenting just in case you missed my previous review notification.

endigma:
Hi @melsonic, this is low priority internally at the moment, but rest assured it's in the queue

melsonic:
No problem @endigma, just wanted to know if it's in the queue or not.

endigma:
Unstale

melsonic:
unstale

endigma:
Hi @melsonic, we've decided internally to adopt your work as the basis for our own fix for the original issue, as such, we'll be closing this PR for now. Thanks for your contribution!

PR #1970, opened 2025-06-16
test: improved kafka integration tests

alepane21:
## Motivation and Context

Writing the redis tests without any require inside go routines and using channels instead of timeouts has reduced the flakiness of those tests.
So in this PR I tried to do the same for kafka.
Working on this task I saw, inspecting the kafka client debug log, that when the test was failing, most of the time was spent on the fetch operation, so I also reduced the FetchMaxWait with really good results (locally).

## Checklist

### DIFF
--- a/router-tests/events/kafka_events_test.go
+++ b/router-tests/events/kafka_events_test.go
@@ -109,13 +109,13 @@ func TestKafkaEvents(t *testing.T) {
 
 			produceKafkaMessage(t, xEnv, topics[0], `{"__typename":"Employee","id": 1,"update":{"name":"foo"}}`)
 
-			testenv.AwaitChannelWithT(t, KafkaWaitTimeout, subscriptionArgsCh, func(t testing.TB, args kafkaSubscriptionArgs) {
+			testenv.AwaitChannelWithT(t, KafkaWaitTimeout, subscriptionArgsCh, func(t *testing.T, args kafkaSubscriptionArgs) {
 				require.NoError(t, args.errValue)
 				require.JSONEq(t, `{"employeeUpdatedMyKafka":{"id":1,"details":{"forename":"Jens","surname":"Neuse"}}}`, string(args.dataValue))
-			}, "unable to receive message before timeout")
+			})
 
 			require.NoError(t, client.Close())
-			testenv.AwaitChannelWithT(t, KafkaWaitTimeout, clientRunCh, func(t testing.TB, err error) {
+			testenv.AwaitChannelWithT(t, KafkaWaitTimeout, clientRunCh, func(t *testing.T, err error) {
 				require.NoError(t, err)
 			}, "unable to close client before timeout")
 		})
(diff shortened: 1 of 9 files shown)


router-tests/events/kafka_events_test.go
```diff
@@ -144,65 +150,73 @@ func TestKafkaEvents(t *testing.T) {
...
-			require.Eventually(t, func() bool {
-				return counter.Load() == 2
-			}, KafkaWaitTimeout, time.Millisecond*100)
+			select {
+			case args := <-subscriptionArgsCh:
+				require.NoError(t, args.errValue)
+				require.JSONEq(t, `{"employeeUpdatedMyKafka":{"id":1,"details":{"forename":"Jens","surname":"Neuse"}}}`, string(args.dataValue))
+			case <-time.After(KafkaWaitTimeout):
+				require.Fail(t, "subscriptionArgsCh should not block")
+			}
```
endigma: Perhaps a helper that does this block like maybe `AwaitSubscriptionUpdateWithT(func (t *testing.T) {})` or something?

endigma (changes requested):

router-tests/events/kafka_events_test.go
alepane21: Good idea, but what do you thing of AwaitChannelWithT? I already refactored the tests with it!

router-tests/testenv/utils.go
```diff
@@ -7,13 +7,13 @@ import (
...
 
 	select {
 	case args := <-ch:
-		f(args)
+		f(t, args)
 	case <-time.After(timeout):
-		require.Fail(t, "clientRunCh should not block")
+		require.Fail(t, "timeout waiting for channel value", msgAndArgs...)
 	}
 }
```
endigma: Why `testing.TB` here? We are providing it always with `*testing.T`, it is the type of `t`.

router-tests/events/kafka_events_test.go
```diff
@@ -1083,9 +1083,9 @@ func produceKafkaMessage(t *testing.T, xEnv *testenv.Environment, topicName stri
 		pErrCh <- err
 	})
 
-	testenv.AwaitChannelWithT(t, KafkaWaitTimeout, pErrCh, func(pErr error) {
+	testenv.AwaitChannelWithT(t, KafkaWaitTimeout, pErrCh, func(t testing.TB, pErr error) {
 		require.NoError(t, pErr)
-	})
+	}, "unable to produce message before timeout")
```
endigma: I don't think this meaningfully improves on just `timeout waiting for channel value`, and "produce" ? this is receive timeout.

router-tests/events/kafka_events_test.go
```diff
@@ -619,7 +619,7 @@ func TestKafkaEvents(t *testing.T) {
 				line, _, gErr := reader.ReadLine()
 				require.NoError(t, gErr)
 				require.Equal(t, "", string(line))
-			})
+			}, "unable to receive message before timeout")
```
endigma: same with a lot of these, not really helpful and just makes call site ugly, just leave the default

router-tests/testenv/utils.go
```diff
@@ -7,13 +7,13 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func AwaitChannelWithT[A any, C chan A](t *testing.T, timeout time.Duration, ch C, f func(A)) {
+func AwaitChannelWithT[A any, C chan A](t *testing.T, timeout time.Duration, ch C, f func(testing.TB, A), msgAndArgs ...interface{}) {
```
endigma: can this be further restricted to a [one way channel](https://gobyexample.com/channel-directions)?

router-tests/events/kafka_events_test.go
alepane21: the message was sent from the callback of the produce action of kafka ... probably a confusing message in what was already a callback mess :D

router/pkg/config/config.schema.json
```diff
@@ -2033,6 +2033,10 @@
                         }
                       }
                     ]
+                  },
+                  "fetch_max_wait": {
+                    "type": "string",
+                    "description": "The maximum wait time for fetching messages from the Kafka broker. The period is specified as a string with a number and a unit, e.g. 10ms, 1s, 1m, 1h. The supported units are 'ms', 's', 'm', 'h'. Setting this to a higher value can help reduce the number of requests to the Kafka broker."
```
endigma: Should have `"format": "go-duration"` I think, and a default or be required?

PR #1975, opened 2025-06-18
feat: add SASL-SCRAM support and add logging of kafka client errors

alepane21:
## Motivation and Context

Some customers asked for support of authentication with SASL-SCRAM.
While developing this, I also noticed that kafka clients errors were dropped, so I also fixed that.

## Checklist

- [x] Documentation has been updated on https://github.com/wundergraph/cosmo-docs/pull/103.

### DIFF
--- a/router/pkg/pubsub/kafka/provider_builder.go
+++ b/router/pkg/pubsub/kafka/provider_builder.go
@@ -126,6 +126,10 @@ func buildKafkaOptions(eventSource config.KafkaEventSource, logger *zap.Logger) 
 		kgo.WithLogger(&kgoErrorLogger{logger: logger}),
 	}
 
+	if eventSource.FetchMaxWait > 0 {
+		opts = append(opts, kgo.FetchMaxWait(eventSource.FetchMaxWait))
+	}
+
 	if eventSource.TLS != nil && eventSource.TLS.Enabled {
 		opts = append(opts,
 			// Configure TLS. Uses SystemCertPool for RootCAs by default.
--- a/router/pkg/pubsub/kafka/provider_builder.go
+++ b/router/pkg/pubsub/kafka/provider_builder.go
@@ -99,7 +99,11 @@ func buildKafkaOptions(eventSource config.KafkaEventSource, logger *zap.Logger) 
 		)
 	}
 
+	var authSpecified bool
+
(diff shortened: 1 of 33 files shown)


router/pkg/pubsub/kafka/provider_builder.go
```diff
@@ -90,11 +106,28 @@ func buildKafkaOptions(eventSource config.KafkaEventSource) ([]kgo.Opt, error) {
 		}.AsMechanism()))
 	}
 
+	if eventSource.Authentication != nil && eventSource.Authentication.SASLSCRAM.Username != nil && eventSource.Authentication.SASLSCRAM.Password != nil && eventSource.Authentication.SASLSCRAM.Mechanism != nil {
```
Noroth: With this logic you can configure both auth methods at the same time. Is that even possible?

router/pkg/pubsub/kafka/provider_builder_test.go
```diff
@@ -50,7 +50,27 @@ func TestBuildKafkaOptions(t *testing.T) {
...
 
-		opts, err := buildKafkaOptions(cfg)
+		opts, err := buildKafkaOptions(cfg, zaptest.NewLogger(t))
+		require.NoError(t, err)
+		require.NotEmpty(t, opts)
+		// Can't directly check for SASL options, but we can verify more options are present
+		require.Greater(t, len(opts), 1)
+	})
+
+	t.Run("with sasl scram auth", func(t *testing.T) {
```
Noroth, line 60: What happens when you configure both methods?

Noroth (changes requested):

router/pkg/pubsub/kafka/provider_builder_test.go
alepane21: it shouldn't happen, the schema does not allow it (https://github.com/wundergraph/cosmo/blob/e6d149bdb8ce4e4e023bc4f95e77797bc4bde998/router/pkg/config/config.schema.json#L2013)

alepane21: I added the schema and found that the oneOf wasn't working properly :(
Now it works!

router/pkg/pubsub/kafka/provider_builder.go
alepane21: shouldn't because of schema, but still I added the check here, to be sure.

Noroth (approved):

PR #1976, opened 2025-06-18
feat: add debounce to watcher

SkArchon:
## Motivation and Context

In case a user is modifying multiple paths, we might want to wait until a subsequent tick without any changes occurs to call the watcher callback.

In cases where the user needs to update three config files for a particular task
```
Write file 1
Write file 2
<- tick 1 run
Write file 3
<- tick 2 run
```

This would cause the callback to execute on tick 1, and if validations pass, the router will be available in a valid BUT unexpected state for 1 tick. We reduce the possibility of this by debouncing, so if there are pending changes in one tick, we will wait for the next tick to run the callback, this way the user can save the third file before the second tick.

## Checklist


router/pkg/watcher/watcher.go
```diff
@@ -51,36 +51,40 @@ func New(options Options) (func(ctx context.Context) error, error) {
...
-				shouldRunCallback := false
+				changesDetected := false
+
 				for _, path := range options.Paths {
 					stat, err := os.Stat(path)
 					if err != nil {
 						ll.Debug("Target file cannot be statted", zap.String("path", path), zap.Error(err))
-
-						// Reset the mod time so we catch any new file at the target path
 						prevModTimes[path] = time.Time{}
```
endigma: Removed comment?

router/pkg/watcher/watcher_test.go
```diff
@@ -622,6 +622,130 @@ func TestWatch(t *testing.T) {
...
+		require.NoError(t, os.WriteFile(tempFile1, []byte("b2"), 0o600))
+		time.Sleep(testWaitInterval)
+
+		t.Log("Modifying file at tick 4")
+		require.NoError(t, os.WriteFile(tempFile1, []byte("b1"), 0o600))
+		time.Sleep(testWaitInterval)
+
+		t.Log("Modifying file at tick 5")
+		require.NoError(t, os.WriteFile(tempFile1, []byte("b1"), 0o600))
+		time.Sleep(testWaitInterval)
```
endigma: I really wish we could synchronize this to the actual watch interval, this seems like it will cause a problem at some point

router/pkg/watcher/watcher_test.go
```diff
@@ -622,6 +622,130 @@ func TestWatch(t *testing.T) {
 			spy.AssertCalled(t, 1)
 		}, assertTimeout, assertPollInterval)
 	})
+
+	t.Run("modify existing single file in multiple subsequent ticks", func(t *testing.T) {
+		t.Parallel()
+
+		customWatchInterval := 600 * time.Millisecond
+		intervalBuffer := 50 * time.Millisecond
+		testWaitInterval := customWatchInterval + intervalBuffer
```
endigma: can these be test level consts? `customWatchInterval` is 600ms here and 500ms in the next test, and `intervalBuffer` changes too with no comments explaining why

router/pkg/watcher/watcher_test.go
```diff
@@ -622,6 +622,130 @@ func TestWatch(t *testing.T) {
...
+
+		t.Log("Modifying file 2 at tick 2")
+		require.NoError(t, os.WriteFile(tempFile3, []byte("b2"), 0o600))
+
+		// No Modifications should have happened because we still haven't
+		// gotten a subsequent tick where no modifications happened
+		// after the first two modifications
+		require.EventuallyWithT(t, func(t *assert.CollectT) {
+			spy.AssertCalled(t, 0)
+		}, assertTimeout, assertPollInterval)
```
endigma: This should be 0 immediately, Eventually will short circuit on the first loop, no need to wait for it

router/pkg/watcher/watcher_test.go
```diff
@@ -622,6 +622,130 @@ func TestWatch(t *testing.T) {
...
+
+		t.Log("Run callback at tick 3")
+		time.Sleep(testWaitInterval)
+
+		require.EventuallyWithT(t, func(t *assert.CollectT) {
+			spy.AssertCalled(t, 1)
+		}, assertTimeout, assertPollInterval)
```
endigma: We should also assert that it _remains_ 1, not just happens to be exactly 1 at some point, this is what made the test flaky before

router/pkg/watcher/watcher_test.go
```diff
@@ -622,6 +622,130 @@ func TestWatch(t *testing.T) {
...
+			Logger:   zap.NewNop(),
+			Paths:    []string{tempFile1, tempFile2, tempFile3},
+			Callback: spy.Call,
+		})
+		require.NoError(t, err)
+
+		eg, ctx := errgroup.WithContext(ctx)
+		eg.Go(func() error {
+			return watchFunc(ctx)
+		})
```
endigma: never awaited?

router/pkg/watcher/watcher_test.go
```diff
@@ -622,6 +622,130 @@ func TestWatch(t *testing.T) {
...
+			Logger:   zap.NewNop(),
+			Paths:    []string{tempFile1},
+			Callback: spy.Call,
+		})
+		require.NoError(t, err)
+
+		eg, ctx := errgroup.WithContext(ctx)
+		eg.Go(func() error {
+			return watchFunc(ctx)
+		})
```
endigma: same as the other comment

router/pkg/watcher/watcher_test.go
endigma: Perhaps we could provide a tick source as an option to watcher, and use time.Ticker in prod and a manually controlled ticker in tests?

router/pkg/watcher/watcher_test.go
endigma: see other comment, i think we should have a different tick source for tests than in real watcher

router/pkg/watcher/watcher.go
```diff
@@ -36,8 +37,12 @@ func New(options Options) (func(ctx context.Context) error, error) {
 	ll := options.Logger.With(zap.String("component", "file_watcher"), zap.Strings("path", options.Paths))
 
 	return func(ctx context.Context) error {
-		ticker := time.NewTicker(options.Interval)
-		defer ticker.Stop()
+		ticker := time.Tick(options.Interval)
```
SkArchon: We now use time.Tick instead, since ticker.Stop is no longer needed as of go 1.23 since go gc can cleanup unreferenced ticks

router/pkg/watcher/watcher.go
```diff
@@ -36,8 +37,12 @@ func New(options Options) (func(ctx context.Context) error, error) {
...
+		// If a ticker override is provided, use that instead of the default ticker
+		// This is used only for tests
+		if options.TickerOverride != nil {
+			ticker = options.TickerOverride
+		}
```
endigma: Can we invert this so we set the ticker to `time.Tick` if not provided?

router/pkg/watcher/watcher_test.go
```diff
@@ -652,3 +818,10 @@ func TestCancel(t *testing.T) {
 
 	require.ErrorIs(t, eg.Wait(), context.Canceled)
 }
+
+// sendTick helper function which also adds jitter
+// so users don't need to manually add jitter to every test
```
endigma: Not really "jitter", it's just a 10ms sleep

router/pkg/watcher/watcher.go
```diff
@@ -10,10 +10,11 @@ import (
...
 type Options struct {
-	Interval time.Duration
-	Logger   *zap.Logger
-	Paths    []string
-	Callback func()
+	Interval       time.Duration
+	Logger         *zap.Logger
+	Paths          []string
+	Callback       func()
+	TickerOverride <-chan time.Time
```
endigma: `TickSource` ? and use time.Tick as default?

PR #1979, opened 2025-06-18
feat: expose query plan information to custom modules + bump engine

endigma:
## Motivation and Context

It may be desirable to be able to access information about the query plan in modules, e.g. for rate limiting with bucket score based on subgraphs contacted. This PR makes the following stats about the query plan available to custom modules via a new method on `OperationContext`. 

- Total subgraphs contacted
- A map of subgraph name to amount of times contacted

Incorporates graphql-go-tools [2.0.0-rc.192](https://github.com/wundergraph/graphql-go-tools/compare/v2.0.0-rc.191...v2.0.0-rc.192)

Docs https://github.com/wundergraph/cosmo-docs/pull/105

## Checklist


router/core/context.go
```diff
@@ -573,6 +578,47 @@ func (o *operationContext) ClientInfo() ClientInfo {
...
+}
+
+func (o *operationContext) QueryPlanStats() QueryPlanStats {
+	qps := QueryPlanStats{
+		TotalSubgraphFetches: 0,
+		SubgraphFetches:      make(map[string]int),
+	}
+
+	if p, ok := o.preparedPlan.preparedPlan.(*plan.SynchronousResponsePlan); ok {
+		qps.analyzePlanNode(p.Response.Fetches.QueryPlan())
```
StarpTech, line 633: Can QueryPlan ever return nil?

StarpTech (changes requested):

router/core/context.go
endigma: Only if `p.Response.Fetches` is nil, which should not be nil unless some other catastrophic failure has occurred

StarpTech: I'd suggest to do a nil check to be on the safe side. People can call this currently from several places.

endigma: If they do call it from a place where it completely fails, is it better to fail silently with an empty QPS? Maybe I should make it return (QPS, error)?

endigma: It does error handling now, instead of theoretical panic

StarpTech: Exactly, cool thx

StarpTech (approved):
LGTM

PR #1985, opened 2025-06-23
test: reduce NATS tests flakiness

alepane21:
## Motivation and Context

NATS tests were kind of flaky, this PR should improve and make them more stable.
Also I moved the edfs tests back together.

## Checklist


router-tests/events/nats_events_test.go
```diff
@@ -1616,33 +1318,66 @@ func TestNatsEvents(t *testing.T) {
...
+					eventNext, _, gErr = reader.ReadLine()
+					require.NoError(t, gErr)
+					require.Equal(t, "event: next", string(eventNext))
+					data, _, gErr = reader.ReadLine()
+					require.NoError(t, gErr)
+					require.Equal(t, fmt.Sprintf("data: {\"data\":{\"filteredEmployeeUpdated\":{\"id\":%d,\"details\":{\"forename\":\"%s\",\"surname\":\"%s\"}}}}", i, testData[i].forename, testData[i].surname), string(data))
+					line, _, gErr = reader.ReadLine()
+					require.NoError(t, gErr)
+					require.Equal(t, "", string(line))
+				}
```
endigma: Can we put this back as a `switch`? And what's the fall through? Do we just not care about if `i` is not one of these numbers? Should it fail?

If we just don't care, why do `i := 1; i < 13; i++` just to get rid of half the events? Can't we just have a const array of these 7 numbers?

router-tests/events/nats_events_test.go
```diff
@@ -1711,47 +1424,45 @@ func TestNatsEvents(t *testing.T) {
 			require.NoError(t, err)
 			err = xEnv.NatsConnectionDefault.Flush()
 			require.NoError(t, err)
-			xEnv.WaitForMessagesSent(1, NatsWaitTimeout)
-			produced.Add(1)
+			testenv.AwaitChannelWithT(t, NatsWaitTimeout, subscriptionArgsCh, func(t *testing.T, subscriptionArgs natsSubscriptionArgs) {
+				var gqlErr graphql.Errors
+				require.ErrorAs(t, subscriptionArgs.errValue, &gqlErr)
+				assert.Equal(t, "Invalid message received", gqlErr[0].Message)
+			})
```
endigma: I don't like mixing `require` and `assert` like this, you can achieve the same with something like:
```
if assert.ErrorAs(...) {
    assert.Equal(...)
}
```
or just use `require` in both cases.

endigma (changes requested):

router-tests/events/nats_events_test.go
alepane21: converted to switch, added a comment to explain better the test

router-tests/events/nats_events_test.go
alepane21: yeah, you are right, changed as you proposed

router-tests/events/nats_events_test.go
```diff
@@ -163,37 +165,23 @@ func TestNatsEvents(t *testing.T) {
 
 			surl := xEnv.GraphQLWebSocketSubscriptionURL()
 			client := graphql.NewSubscriptionClient(surl)
-			t.Cleanup(func() {
-				_ = client.Close()
-			})
-
-			var counter atomic.Uint32
+			subscriptionArgsCh := make(chan natsSubscriptionArgs, 10)
```
endigma: Remove buffer

router-tests/events/nats_events_test.go
```diff
@@ -202,37 +190,45 @@ func TestNatsEvents(t *testing.T) {
...
 			require.NoError(t, err)
-
-			require.Eventually(t, func() bool {
-				return counter.Load() == 3
-			}, NatsWaitTimeout, time.Millisecond*100)
+			testenv.AwaitChannelWithT(t, NatsWaitTimeout, subscriptionArgsCh, func(t *testing.T, args natsSubscriptionArgs) {
+				var gqlErr graphql.Errors
+				require.ErrorAs(t, args.errValue, &gqlErr)
+				require.Equal(t, "Cannot return null for non-nullable field 'Subscription.employeeUpdated.id'.", gqlErr[0].Message)
+			})
```
endigma: require.ErrorContains?

endigma (approved):

PR #2004, opened 2025-07-02
feat allow to log response payload and fix feature flag expression bug

SkArchon:
@coderabbitai summary

## Checklist

### DIFF
--- a/router-tests/structured_logging_test.go
+++ b/router-tests/structured_logging_test.go
@@ -1943,6 +1943,59 @@ func TestFlakyAccessLogs(t *testing.T) {
 			)
 		})
 
+		t.Run("validate expression evaluation for feature flag", func(t *testing.T) {
+			t.Parallel()
+
+			testenv.Run(t,
+				&testenv.Config{
+					AccessLogFields: []config.CustomAttribute{
+						{
+							Key: "url_method_expression",
+							ValueFrom: &config.CustomDynamicAttribute{
+								Expression: "request.url.method",
+							},
+						},
+					},
+					LogObservation: testenv.LogObservationConfig{
+						Enabled:  true,
+						LogLevel: zapcore.InfoLevel,
+					},
+				},
+				func(t *testing.T, xEnv *testenv.Environment) {
+					res := xEnv.MakeGraphQLRequestOK(testenv.GraphQLRequest{
+						Query: `query employees { employees { id } }`,
+						Header: map[string][]string{
+							"X-Feature-Flag": {"myff"},
+						},
+					})
+					require.JSONEq(t, employeesIDData, res.Body)
+					requestLog := xEnv.Observer().FilterMessage("/graphql")
+					requestLogAll := requestLog.All()
+					requestContext := requestLogAll[0].ContextMap()
+
+					expectedValues := map[string]interface{}{
+						"log_type":              "request",
+						"status":                int64(200),
+						"method":                "POST",
+						"path":                  "/graphql",
+						"query":                 "",
+						"ip":                    "[REDACTED]",
+						"feature_flag":          "myff",
+						"url_method_expression": "POST", // From expression
+					}
+					additionalExpectedKeys := []string{
+						"user_agent",
+						"latency",
+						"config_version",
+						"request_id",
+						"pid",
+						"hostname",
+					}
+					checkValues(t, requestContext, expectedValues, additionalExpectedKeys)
+				},
+			)
+		})
+
 		t.Run("validate expression evaluation for default value", func(t *testing.T) {
 			t.Parallel()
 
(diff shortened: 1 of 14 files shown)


router-tests/structured_logging_test.go
```diff
@@ -1943,6 +1943,59 @@ func TestFlakyAccessLogs(t *testing.T) {
...
+						"latency",
+						"config_version",
+						"request_id",
+						"pid",
+						"hostname",
+					}
+					checkValues(t, requestContext, expectedValues, additionalExpectedKeys)
+				},
+			)
+		})
```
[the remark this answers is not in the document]
SkArchon: This is added separately

router/core/graphql_handler.go
```diff
@@ -78,6 +79,7 @@ type HandlerOptions struct {
 	SubgraphErrorPropagation                    config.SubgraphErrorPropagationConfiguration
 	EngineLoaderHooks                           resolve.LoaderHooks
 	ApolloSubscriptionMultipartPrintBoundary    bool
+	ExprVisitorManager                          *expr.VisitorGroup
```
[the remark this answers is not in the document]
SkArchon: Added a nil check to the methods

router-tests/structured_logging_test.go
```diff
@@ -1943,6 +1943,59 @@ func TestFlakyAccessLogs(t *testing.T) {
 			)
 		})
 
+		t.Run("validate expression evaluation for feature flag", func(t *testing.T) {
```
StarpTech: What does it test?

router/core/engine_loader_hooks.go
```diff
@@ -148,6 +152,10 @@ func (f *engineLoaderHooks) OnFinished(ctx context.Context, ds resolve.DataSourc
 	exprCtx.Subgraph.Name = ds.Name
 	exprCtx.Subgraph.Request.Error = WrapExprError(responseInfo.Err)
 
+	if f.exprVisitorManager.IsSubgraphResponseBodyUsedInExpressions() {
```
StarpTech: I think we can move it outside of the hot path.

router/core/graphql_handler.go
```diff
@@ -195,6 +199,13 @@ func (h *GraphQLHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 			w.Header().Set("Cache-Control", "no-store, no-cache, must-revalidate")
 		}
 
+		if h.exprVisitorManager.IsResponseBodyUsedInExpressions() {
+			// String returns the unread portion of the buffer
+			// However at this point the offset is 0, which means
+			// we will get the entire string
+			reqCtx.expressionContext.Response.Body.Raw = respBuf.String()
```
StarpTech: Are you sure that this is the right place and method? We want to log the final response. This would not catch errors in the pre_handler or when something manipulates the response body via custom module

StarpTech (changes requested):

router-tests/structured_logging_test.go
SkArchon: Previously expressions did not get evaluated (worked at all) for feature flag graph muxes, it ensures that it works for feature flag graph muxes.

StarpTech: Could you describe it like this? `Should be able to use an expression...`

router/core/graphql_handler.go
SkArchon: Good point, will check

router/core/engine_loader_hooks.go
SkArchon: Could you elaborate a bit, because we have to set it before we call the subgraph access logger, which happens a bit below in the same file, https://github.com/wundergraph/cosmo/pull/2004/files#diff-82612426f570e70d58fb5e2814d47ef0d8f6eda1b5c4928c85424f432aa1dd4bR191

router/core/graphql_handler.go
SkArchon: New proposal
```
ww := middleware.NewWrapResponseWriter(w, r.ProtoMajor) // existing code

// NEW PROXIED READER (not actually a tee reader)
if responseBodyUsedInExpression {
  var respBuf bytes.Buffer
  ww.Tee(&respBuf)
}

h.handler.ServeHTTP(ww, r) // existing code

// set to response body
if responseBodyUsedInExpression {
  expr.Response.Body = respBuf.String()
}
```

Since we write to bytes.Buffer should be fast enough, but will benchmark to be sure, let me know what you think

router/core/engine_loader_hooks.go
SkArchon: Fixed

router/core/graphql_handler.go
SkArchon: It internally calls tee.Write https://github.com/go-chi/chi/blob/master/middleware/wrap_writer.go#L96 discard is never true, as we won't call Discard() on the wrap writer.

router-tests/structured_logging_test.go
SkArchon: Done

StarpTech (approved):
LGTM

PR #2011, opened 2025-07-03
fix(router): require presence of prefix value for auth header

ysmolski:
Be more strict when Authorization token is provided and configuration has only non-empty value prefixes specified. Produce an error when none of value prefixes match the value provided in the header.

@coderabbitai summary

## Checklist


router/pkg/authentication/http_header_authenticator.go
```diff
@@ -28,33 +28,50 @@ func (a *httpHeaderAuthenticator) Authenticate(ctx context.Context, p Provider)
...
+			// If claims are nil, we should return an empty Claims map to signal that the
 			// authentication was successful, but no claims were found.
 			if claims == nil {
 				claims = make(Claims)
 			}
 			return claims, nil
 		}
+		if !prefixMatchFound {
+			errs = errors.Join(errs, fmt.Errorf(
+				"header %q value %q does not start with any of the expected prefixes %v",
```
StarpTech: We should not make the secret part of the error message.

ysmolski: My bad. I have just fixed it assuming that `header` and `prefixes` variables do not contain any secrets.

StarpTech (approved):
LGTM

PR #2035, opened 2025-07-10
feat: cosmo streams v1

alepane21:
## Summary by CodeRabbit

* **New Features**
  * Introduced new hooks for customizable authorization, initial message emission, data transformation, and advanced filtering in Cosmo Streams.
  * Enabled provider-agnostic event processing supporting multiple event sources and richer event metadata.
  * Enhanced developer experience with examples and AsyncAPI-based event schema documentation for type-safe integrations.

## Checklist


alepane21:
@coderabbitai summary

rfc/cosmo-streams-v1.md
```diff
@@ -0,0 +1,584 @@
...
+    ProviderId() string
+    // the subscription configuration is specific for each provider
+    SubscriptionConfiguration() []byte
+}
+
+type RequestContext interface {
+    Authentication() *core.Authentication
+}
+
+type SubscriptionContext interface {
```
StarpTech: According to the new module design we have established the following pattern

- Interface - `SubscriptionOnStartHookContext`

rfc/cosmo-streams-v1.md
```diff
@@ -0,0 +1,584 @@
...
+
+### Example: Check if the client is allowed to subscribe to the stream
+
+```go
+// the structs are reported only with the fields that are used in the example
+type StreamContext interface {
+    ProviderType() string
+    ProviderId() string
+    // the subscription configuration is specific for each provider
+    SubscriptionConfiguration() []byte
```
StarpTech: Can we make it typed? Via generics or type assertion.

rfc/cosmo-streams-v1.md
```diff
@@ -0,0 +1,584 @@
...
+            ProviderID: receivedEventConfiguration.ProviderID,
+            Subject: receivedEventConfiguration.Subject,
+            Data: dataForStream,
+            Metadata: receivedEventConfiguration.Metadata,
+        }
+        data, err := json.Marshal(publishAndRequestEventConfiguration)
+        if err != nil {
+            return fmt.Errorf("error marshalling data: %w", err)
+        }
+        event.SetData(data)
```
StarpTech: Instead of setting data on each event, it would be more convenient if we would introduce some sort of data pipeline pattern.

```go
// StreamBatchEventHook processes a batch of stream events (inbound or outbound).
//
// Return:
//   - empty slice: drop all events.
//   - non-empty slice: emit those events (can grow, shrink, or reorder the batch).
// err != nil: abort the subscription with an error.
type StreamBatchEventHook interface {
    OnStreamEvents(ctx StreamContext, events []core.StreamEvent) ([]core.StreamEvent, error)
}

func (m *MyModule) OnStreamEvents(
    ctx core.StreamContext,
    events []core.StreamEvent,
) ([]core.StreamEvent, error) {
    // 1) DENY: check auth and return nil to drop all if unauthorized
    //    if !ctx.RequestContext().Authentication().HasScope("stream:read") {
    //        return nil, nil
    //    }

    // 2) MAP: transform each event’s payload to your internal shape
    //    for _, evt := range events { ... }

    // 3) FILTER: drop unwanted events based on metadata or content
    //    if shouldDrop(evt) { continue }

    // 4) BATCHING: split or combine events, e.g., grouping related updates
    //    if needsSplitting(evt) { ... }

    // Finally, return the new batch of events to deliver
    return events, nil
}

func (m *MyModule) Module() core.ModuleInfo { /*…*/ }

```

In this proposal, the user does not need to work with `setData` and multiple events at the same can be handled. Kafka for instance is batching protocol.

rfc/cosmo-streams-v1.md
```diff
@@ -0,0 +1,584 @@
...
+type PublishAndRequestEventConfiguration struct {
+    ProviderID string          `json:"providerId"`
+    Subject    string          `json:"subject"`
+    Data       json.RawMessage `json:"data"`
+    Metadata   map[string]string `json:"metadata"`
+}
+
+type MyModule struct {}
+
+func (m *MyModule) SubscriptionOnStart(ctx SubscriptionContext) error {
```
StarpTech: This is much better. This can also be used for subscriptions in general.

rfc/cosmo-streams-v1.md
```diff
@@ -0,0 +1,584 @@
...
+type PublishAndRequestEventConfiguration struct {
+    ProviderID string          `json:"providerId"`
+    Subject    string          `json:"subject"`
+    Data       json.RawMessage `json:"data"`
+    Metadata   map[string]string `json:"metadata"`
+}
+
+type MyModule struct {}
+
+func (m *MyModule) StreamOnEventToSend(ctx StreamContext, event core.StreamEvent) error {
```
StarpTech: Very similar to the proposal above. I'd recommend a more direct API. This does not take into account queue specific semantics like headers etc.. The challenge here is to provide a simple API but with the ability to leverage provider specific capabilities later.

```go
// StreamPublishEventHook is invoked just before any events are sent
// to the provider’s message queue. You can transform, drop, split,
// or reorder events in one place.
//
// Return:
//   - empty slice: drop all outbound events.
//   - non-empty slice: emit those events (can grow, shrink, or reorder).
// err != nil: abort the publish with an error.
type StreamPublishEventHook interface {
    OnPublishEvents(ctx StreamContext, events []core.StreamEvent) ([]core.StreamEvent, error)
}

func (m *MyModule) OnPublishEvents(
    ctx core.StreamContext,
    events []core.StreamEvent,
) ([]core.StreamEvent, error) {
    // 1) MAP: modify payloads for external systems
    //    for i, evt := range events { ... events[i] = mappedEvt }

    // 2) FILTER: drop events that shouldn’t be published
    //    if shouldDrop(evt) { continue }

    // 3) BATCH/SPLIT: group or split messages as needed
    //    if needsSplitting(evt) { ... }

    // Finally, return the modified batch to be sent to the queue
    return events, nil
}
```

rfc/cosmo-streams-v1.md
```diff
@@ -0,0 +1,584 @@
...
+            EmployeeId string `json:"EmployeeId"`
+            OtherField string `json:"OtherField"`
+        }
+        dataToSend.EmployeeId = data.Id
+        dataToSend.OtherField = "Custom value"
+        dataToSendMarshalled, err := json.Marshal(dataToSend)
+        if err != nil {
+            return fmt.Errorf("error marshalling data: %w", err)
+        }
+        publishAndRequestEventConfiguration := nats.PublishAndRequestEventConfiguration{
```
StarpTech: This requires finding a common abstraction between all providers. Is this feasible and DX friendly or should we work with generics? I'd experiment how the API can look like. Maybe we can find a good approach and stay with the API.

Alternatively, I could also think of a complete different approach e.g. registering a provider specific adapter in `provision` where the mapping, filtering etc can happen. This could allow for better usage of generics and allow a more direct api with the specific provider api.

```go
// Wrap provider messages without type assertions:
type StreamEvent[T any] struct {
    Subject string  // NATS subject or Kafka topic
    Record  T       // *nats.Msg or *kafka.Message
}

// Adapter: OnStart emits initial events; FromProvider decodes inbound → 0…N T’s;
// ToProvider encodes 0…N T’s → core.StreamEvent
type StreamAdapter[T any] struct {
    OnStart      func(ctx StreamContext) ([]T, error)
    FromProvider func(ctx StreamContext, evt StreamEvent[T]) ([]T, error)
    ToProvider   func(ctx StreamContext, items []T) []StreamEvent[T]
}

func (m *MyModule) Provision(ctx ModuleContext) error {
    ctx.RegisterStreamAdapter[*nats.Msg](StreamAdapter[*nats.Msg]{
        // 1) authenticate & emit a “ready” event
        OnStart: func(ctx StreamContext) ([]*nats.Msg, error) {
            if !ctx.RequestContext().Authentication().HasScope("stream:start") {
                return nil, fmt.Errorf("unauthorized")
            }
            init := &nats.Msg{
                Subject: ctx.StreamContext().ProviderId(),
                Data:    []byte(`{"type":"ready"}`),
            }
            return []*nats.Msg{init}, nil
        },
        // 2) inbound: deny/filter/map/batch
        FromProvider: func(ctx StreamContext, evt StreamEvent[*nats.Msg]) ([]*nats.Msg, error) {
            msg := evt.Record
            // → DENY if missing auth
            if msg.Header.Get("X-Auth") == "" {
                return nil, nil
            }
            // → BATCH: emit original + trace
            trace := *msg
            trace.Data = []byte(fmt.Sprintf(`{"traceOf":"%s"}`, msg.Subject))
            return []*nats.Msg{msg, &trace}, nil
        },
        // 3) outbound: map & wrap back into StreamEvent
        ToProvider: func(ctx StreamContext, items []*nats.Msg) []StreamEvent[*nats.Msg] {
            out := make([]StreamEvent[*nats.Msg], 0, len(items))
            for _, m := range items {
                // → MAP outbound: tag with environment
                m.Header.Set("X-Env", "prod")
                out = append(out, StreamEvent[*nats.Msg]{Subject: m.Subject, Record: m})
            }
            return out
        },
    })

    return nil
}
```

rfc/cosmo-streams-v1.md
```diff
@@ -0,0 +1,584 @@
...
+    │
+    └─▶ core.StreamOnEventReceivedHandler (Data mapping)
+    │
+    └─▶ core.StreamOnEventFilterHandler (Filtering)
+    │
+    └─▶ "Deliver event to client"
+
+A mutation is sent from the client
+    │
+    └─▶ core.StreamOnEventToSendHandler (Data mapping)
```
StarpTech: This was not clear to me. The mutation use case was not explained in the beginning. The hook also does not indicate it. I would not build a hook around and it keep it generic. A user can act on data in/out. If a user decides to emit data from the router, we should expose this API in the regular module hooks. For instance, in the `Middleware` the user has access to the full parsed GraphQL operation. The user can decide on his own in what condition he wants to emit it.

rfc/cosmo-streams-v1.md
alepane21: @StarpTech We can, but that means that at each new provider we will have to work also on the hooks system. I'm not against it, but do you think is better to have a typed configuration and a bit more work on each new provider?

rfc/cosmo-streams-v1.md
alepane21: @StarpTech I really like this idea, thanks!

rfc/cosmo-streams-v1.md
alepane21: Fixed, thanks!

rfc/cosmo-streams-v1.md
alepane21: I really like this. I've added to the stream context the "Direction" so that the custom module can operate only on one of the two ways.

rfc/cosmo-streams-v1.md
alepane21: Using the same hook defined in the previous comment, we also fixed this issue. And also filtering!

rfc/cosmo-streams-v1.md
StarpTech: I'd try to provide the best developer experience. If that means to put more effort into manage them, yes.

rfc/cosmo-streams-v1.md
alepane21: The only common field between the Subscriptions is the ProviderID and the only common field between events is the data.
So I put the common field in the interface, and every custom module, if it needs some not common fields, can cast to the specific event type.
I think this is easier to write and read than the stream adapter pattern, and will allow more flexibility.

rfc/cosmo-streams-v1.md
alepane21: Well, I think we solved this with the new common hook than!
The issue of having this code in the middleware was that we don't have access to the pubsub configuration of that field, and adding it would require some work.

rfc/cosmo-streams-v1.md
alepane21: Well I think that with the last changes we are using the types in a better way!

rfc/cosmo-streams-v1.md
StarpTech: Would love to see a concrete example in context of multiple providers.

rfc/cosmo-streams-v1.md
StarpTech: I don't think so. We have provided hook to filter in/out events but we didn't show a way to emit events on the message bus.

rfc/cosmo-streams-v1.md
```diff
@@ -0,0 +1,583 @@
...
+type StreamBatchEventHook interface {
+    OnStreamEvents(ctx StreamBatchEventHookContext, events []StreamEvent) ([]StreamEvent, error)
+}
+
+func (m *MyModule) OnStreamEvents(
+    ctx StreamBatchEventHookContext,
+    events []StreamEvent,
+) ([]StreamEvent, error) {
+    // we only rewrite the data for outbound events
+    if ctx.Direction() == StreamBatchDirectionInbound {
```
StarpTech: I'm not a fan of handling this via an attribute. It requires the developer to understand the API. If not understood, the developer might apply the logic for both in and out. For simplicity and better separation of concerns, I would provide two hooks, as shown in my comments.

https://github.com/wundergraph/cosmo/pull/2035#discussion_r2202477725

rfc/cosmo-streams-v1.md
```diff
@@ -0,0 +1,583 @@
...
+    └─▶ core.StreamBatchEventHook (Data mapping, Filtering)
+    │
+    └─▶ "Send event to provider"
+```
+
+### Data Flow
+
+We will need to change the format of the event data sent within the router. Today we use the data that will be sent to the provider directly, but we will need to add a structure where we can include additional fields (metadata, etc.) in the event.
+
+## Implementation Details
```
StarpTech, line 565: It would also be great to explain how the development workflow could look like.

1. Defining an edfs directive with an ID, topics...
2. Implement a hook and filter based on ID to be provider specific
3. How could we ensure message and go types are consistent e.g. https://www.asyncapi.com ?

StarpTech (changes requested):

rfc/cosmo-streams-v1.md
alepane21: Fixed!

rfc/cosmo-streams-v1.md
alepane21: I've added two complete examples (with the graph), but I'm not sure about the third point.

rfc/cosmo-streams-v1.md
```diff
@@ -0,0 +1,882 @@
...
+// This is the new hook that will be called once at subscription start
+func (m *MyModule) SubscriptionOnStart(ctx SubscriptionOnStartHookContext) error {
+    // get the operation name and variables that we need
+    opName := ctx.RequestContext().Operation().Name()
+    opVarId := ctx.RequestContext().Operation().Variables().GetInt("id")
+    
+    // check if the operation name is the one expected by the module
+    if opName == "employeeSub" {
+        // create the event to emit using the operation variables
+        evt := &NatsEvent{
```
StarpTech, line 159: How does the user know at this point that the provider is NATS? There's no check for it. Additionally, the user intends to write events to an existing stream. The subject and providerID are already defined and are implementation details.

rfc/cosmo-streams-v1.md
```diff
@@ -0,0 +1,882 @@
...
+
+type MyModule struct {}
+
+// This is the new hook that will be called each time a batch of events is going to be sent to the provider
+func (m *MyModule) OnPublishEvents(
+    ctx StreamPublishEventHookContext,
+    events []StreamEvent,
+) ([]StreamEvent, error) {
+    // create a new slice of events that we will return with the events with the new format
+    newEvents := make([]StreamEvent, 0, len(events))
```
StarpTech, line 359: I’d keep the RFC free of implementation logic and instead explain what’s possible. Including those details makes it harder to read.

rfc/cosmo-streams-v1.md
```diff
@@ -0,0 +1,882 @@
...
+                }
+                dataToSend.EmployeeId = dataReceived.Id
+                dataToSend.OtherField = "Custom value"
+                dataToSendMarshalled, err := json.Marshal(dataToSend)
+                if err != nil {
+                    return events, fmt.Errorf("error marshalling data: %w", err)
+                }
+
+                // create the new event
+                newEvent := &NatsEvent{
```
StarpTech: Same here. You assume the user only has NATS streams. How do you distinguish between multiple providers? Those details are more important than a full implementation.

rfc/cosmo-streams-v1.md
```diff
@@ -0,0 +1,882 @@
...
+                    ProviderID: natsEvent.ProviderID,
```
StarpTech: The stream should operate within a provider and subject context. It’s not possible to support dynamic subjects across all providers. For now, I’d assume in the RFC that the provider and subjects are predefined.

rfc/cosmo-streams-v1.md
alepane21: Well in this case there is the conversion of the event to NatsEvent that is only going to be possibile if the event is generated by the NATS provider. And there is also the check on the .Subject.

rfc/cosmo-streams-v1.md
alepane21: This is what is going to happen, but still you have to check that the current "batch" that you are receiving is of the right provider type and id.

alepane21: I'll try to clean it up a bit.

rfc/cosmo-streams-v1.md
StarpTech: But this is not very explicit. It would be cleaner, if you could show how to execute code only for a particular subscription by provider. In your example, you would iterate over all events from all subscriptions although you're only interested in NATS. 

This brings me to a different case: how can a user selectively filter and map events for a particular edfs subscription? How would they identify it in code? Should we add an ID field to the directive?

rfc/cosmo-streams-v1.md
StarpTech: It would help, if you could explain how a User identify the subscription, provider etc to make the right decision on the correct stream.

rfc/cosmo-streams-v1.md
alepane21: added the check, thanks

rfc/cosmo-streams-v1.md
alepane21: Ok I cleaned the data structure and now you can access the provider type and id from the configuration in the SubscriptionEventConfiguration() (for the subscriptions) and PublishEventConfiguration().

rfc/cosmo-streams-v1.md
alepane21: Mmm it is a good point. You have access to the OperationContext(), but there is only the general name. Maybe I can expose also the subscription name as defined in the graph? They are already unique in the graph.

rfc/cosmo-streams-v1.md
alepane21: So I think we can close this conversation, since this feature will be covered with another RFC

rfc/cosmo-streams-v1.md
alepane21: logic removed!

rfc/cosmo-streams-v1.md
alepane21: Ok, I've added an appendix about how we could integrate AsyncAPI. Is nothing too fancy, but it would add some structure to the workflow and "fix" the data structure to what is explicited in the async spec.

rfc/cosmo-streams-v1.md
alepane21: I've added te FieldName() to the SubscriptionEventConfiguration() and the PublishEventConfiguration(), to allow the developers to check if is the same subscription they wanted. I think it could be a good solution.
Also added in the example with the init message.

rfc/cosmo-streams-v1.md
```diff
@@ -0,0 +1,1143 @@
...
+    //check if the provider type is the one expected by the module
+    if ctx.SubscriptionEventConfiguration().ProviderType() != "nats" {
+        return nil
+    }
+    
+    // check if the operation name is the one expected by the module
+    if opName == "employeeSub" {
+        // create the event to emit using the operation variables
+        evt := &NatsEvent{
+            Data: []byte(fmt.Sprintf("{\"id\": \"%d\", \"__typename\": \"Employee\"}", opVarId)),
```
StarpTech, line 160: Can we abstract this into a helper function?

rfc/cosmo-streams-v1.md
```diff
@@ -0,0 +1,1143 @@
...
+}
+```
+
+### Proposal
+
+Using the new `SubscriptionOnStart` hook that we introduced for the previous requirement, we can emit the initial message on subscription start. We will also need access to operation variables, which are currently not available in the request context.
+
+To emit the message, I propose adding a new method to the stream context, `WriteEvent`, which will emit the event to the stream at the lowest level. The message will pass through all hooks, making it behave like any other event received from the provider.
+
+The `StreamEvent` contains the data as used by the provider. This allows the hooks system to be provider-agnostic, so adding a new provider will not require changes to the hooks system. To use specific fields, the hook can cast the event to the specific type for the provider. If the custom modules only need to read the data, they can use the `Data()`/`SetData()` methods without casting the event.
```
StarpTech: Old reference to `setData`

rfc/cosmo-streams-v1.md
```diff
@@ -0,0 +1,1143 @@
...
+    ctx StreamPublishEventHookContext,
+    events []StreamEvent,
+) ([]StreamEvent, error) {
+    // check if the provider ID is the one expected by the module
+    if ctx.PublishEventConfiguration().ProviderID() != "my-provider-id" {
+        return events, nil
+    }
+
+    // check if the provider type is the one expected by the module
+    if ctx.PublishEventConfiguration().ProviderType() != "nats" {
```
StarpTech: Using "enums" ?

rfc/cosmo-streams-v1.md
```diff
@@ -0,0 +1,1143 @@
...
+### Stream Lifecycle
+
+```
+One or more batched events are received from the provider
+    │
+    └─▶ core.StreamBatchEventHook (Data mapping, Filtering)
+    │
+    └─▶ "Deliver events to client"
+
+One or more batched events are sent to the provider
```
StarpTech: Use `publish` instead of `sent` might make it clearer across the RFC.

StarpTech (changes requested):

rfc/cosmo-streams-v1.md
alepane21: fixed!

rfc/cosmo-streams-v1.md
alepane21: I kept the SetData/Data() because it is common field between all the adapters. Do you think this is ugly?

StarpTech: Not ugly but based on the proposed hooks, events are added / sliced according to your needs. What would be the concrete use case, where someone wants to read or set data without operating in a concrete provider, subscription?

rfc/cosmo-streams-v1.md
alepane21: Sure, I could introduce the enums in the pubsub package. It would be for sure better!

rfc/cosmo-streams-v1.md
alepane21: Something like this?

```go
evt := nats.NewNatsEvent("{\"id\": \"%d\", \"__typename\": \"Employee\"}", opVarId))
evt.SetMetadata(map[string]string{
    "entity-id": fmt.Sprintf("%d", opVarId),
})
```

rfc/cosmo-streams-v1.md
alepane21: exactly!

rfc/cosmo-streams-v1.md
StarpTech: I don't think this should be handled via metadata. Let's reiterate on this when we have a working POC.

rfc/cosmo-streams-v1.md
alepane21: Removed. You are right, there is no usage for these methods right now.

rfc/cosmo-streams-v1.md
```diff
@@ -0,0 +1,1004 @@
+# RFC Cosmo Streams V1
+
+Based on customer feedback, we've identified the need for more customizable stream behavior. The key areas for customization include:
+- **Authorization**: Implementing authorization checks at the start of subscriptions
+- **Initial message**: Sending an initial message to clients upon subscription start
```
jensneuse, line 5: I would rephrase this to something like:

Our goal is that the Router can autonomously emit events, e.g. on Subscription start.

This gives us more flexibility in terms of the solution. It's not just about the start. It's about 2 things:

1. being able to hook into the full lifecycle of a Subscription & Stream
2. allowing the user to emit events at any lifecycle event

rfc/cosmo-streams-v1.md
```diff
@@ -0,0 +1,1004 @@
...
+- **Data mapping**: Transforming data to align with internal specifications
```
jensneuse: What does this mean?

I guess what you're trying to say is:

Our goal is to map Stream events to Federation compatible Router events.

rfc/cosmo-streams-v1.md
```diff
@@ -0,0 +1,1004 @@
...
+- **Event filtering**: Filtering events using custom logic
+
+Let's explore how we can address each of these requirements.
+
+## Authorization
+
+To support authorization, we need a hook that enables two key decisions:
```
jensneuse: I think it's about 3 decisions.

1. can a client/user subscribe at all?
2. what topics should they be able to subscribe to?
3. are they allowed to consume an event from the stream?

rfc/cosmo-streams-v1.md
```diff
@@ -0,0 +1,1004 @@
...
+    Subjects            []string             `json:"subjects"`
+    StreamConfiguration *StreamConfiguration `json:"streamConfiguration,omitempty"`
+}
+
+type MyModule struct {}
+
+// This is a custom function that will be used to check if the client is allowed to subscribe to the stream
+func customCheckIfClientIsAllowedToSubscribe(ctx SubscriptionOnStartHookContext) bool {
+    // check if the field name is the one expected by the module
+    if ctx.SubscriptionEventConfiguration().FieldName() != "employeeUpdates" {
```
jensneuse: Is this FieldName or RootFieldName?

It might be slightly confusing to just say FieldName because a subscription can have multiple sub fields.

rfc/cosmo-streams-v1.md
```diff
@@ -0,0 +1,1004 @@
...
+    }
+
+    // get the specific configuration for the provider to make more advanced checks
+    cfg, ok := ctx.SubscriptionEventConfiguration().(*NatsSubscriptionEventConfiguration)
+    if !ok {
+        return true
+    }
+
+    providerId := cfg.ProviderID
+    clientScopes := ctx.RequestContext().Authentication().Scopes()
```
jensneuse: Will this panic if Authentication returns nil?

rfc/cosmo-streams-v1.md
```diff
@@ -0,0 +1,1004 @@
...
+    
+    // add checks here on client scopes, provider ID, etc.
+
+    return false
+}
+
+// This is the new hook that will be called once at subscription start
+func (m *MyModule) SubscriptionOnStart(ctx SubscriptionOnStartHookContext) error {
+    // check if the client is allowed to subscribe to the stream
+    if !customCheckIfClientIsAllowedToSubscribe(ctx) {
```
jensneuse, line 73: Overall I very much like this approach. It's very flexible.

rfc/cosmo-streams-v1.md
```diff
@@ -0,0 +1,1004 @@
...
+### Proposal
+
+Add a new hook to the subscription lifecycle, `SubscriptionOnStartHandler`, that will be called once at subscription start.
+
+The hook arguments are:
+* `ctx SubscriptionContext`: The subscription context, which contains the request context and, optionally, the stream context
+
+`RequestContext` already exists and requires no changes, but `SubscriptionContext` is new.
+
+The hook should return an error if the client is not allowed to subscribe to the stream, preventing the subscription from starting.
```
jensneuse, line 106: Are we able to give the user options here?

- return a GraphQL error but keep the WebSocket connection alive
- customize the GraphQL error, e.g. with a code field and custom message
- kill the WS connection maybe?

rfc/cosmo-streams-v1.md
```diff
@@ -0,0 +1,1004 @@
...
+
+The `StreamContext.SubscriptionEventConfiguration()` contains the subscription configuration as used by the provider. This allows the hooks system to be provider-agnostic, so adding a new provider will not require changes to the hooks system. To use specific fields, the hook can cast the configuration to the specific type for the provider.
+
+## Initial Message
+
+When starting a subscription, the client sends a query to the server containing the operation name and variables. The client must then wait for the broker to send the initial message. This waiting period can lead to a poor user experience, as the client cannot display anything until the initial message is received. To address this, we can emit an initial message on subscription start.
+
+To emit an initial message on subscription start, we need access to the stream context (to get the provider type and ID) and the query that the client sent. The variables are particularly important, as they allow the module to use them in the initial message. For example, if someone starts a subscription with employee ID 100 as a variable, the custom module can include that ID in the initial message.
```
jensneuse, line 119: One thing we want to avoid is emitting an event to all subscribers of a stream. Keep in mind that there can be multiple subscribers on the same stream. One joins first, emits an initial message through middleware, then waits. Second subscriber joins, emits an initial message through middleware. In this case, we have 2 subscribers but just one trigger. In this case we don't want to emit the event to all subscribers when a new subscriber joins an existing trigger. It would create a lot of noise and unnecessary events.

rfc/cosmo-streams-v1.md
```diff
@@ -0,0 +1,1004 @@
...
+    if opName == "employeeSub" {
+        // create the event to emit using the operation variables
+        evt := &NatsEvent{
+            Data: []byte(fmt.Sprintf("{\"id\": \"%d\", \"__typename\": \"Employee\"}", opVarId)),
+            Metadata: map[string]string{
+                "entity-id": fmt.Sprintf("%d", opVarId),
+            },
+        }
+        // emit the event to the stream, that will be received by the client
+        ctx.StreamContext().WriteEvent(evt)
```
jensneuse, line 166: As pointed out above, we need to make sure that this event gets only emitted to this particular subscriber, not all subscribers on the same trigger.

If we want to emit an event globally, it wouldn't make sense to do so on a single Router in any way. That's because if we have multiple Routers, we would only reach client connected to our particular Router.

rfc/cosmo-streams-v1.md
```diff
@@ -0,0 +1,1004 @@
...
+
+The `StreamEvent` contains the data as used by the provider. This allows the hooks system to be provider-agnostic, so adding a new provider will not require changes to the hooks system. To use events, the hook has to cast the event to the specific type for the provider.
+
+This change will require adding a new type in each provider package to represent the event with additional fields (metadata, etc.). This is a significant change, but it is necessary to support additional data in events, anyway, even if we don't expose them to the custom modules.
+
+Emitting the initial message with this hook ensures that the client will receive the message before the first event from the provider is received.
+
+## Data Mapping
+
+The current approach for emitting and reading data from the stream is not flexible enough. We need to be able to map data from an external format to the internal format, and vice versa.
```
jensneuse, line 196: Be more explicit about this. E.g. if we stick to Kafka, what exactly do we want to map?

rfc/cosmo-streams-v1.md
```diff
@@ -0,0 +1,1004 @@
...
+    natsConfig := ctx.SubscriptionEventConfiguration().(*nats.SubscriptionEventConfiguration)
+    if natsConfig.Subjects[0] != "topic-with-internal-data-format" {
+        return events, nil
+    }
+
+    // create a new slice of events that we will return with the events with the new format
+    newEvents := make([]StreamEvent, 0, len(events))
+    for _, evt := range events {
+        // check if the event is the one expected by the module
+        if natsEvent, ok := evt.(*NatsEvent); ok {
```
jensneuse, line 260: Can you show all the options we'd have available for Kafka?

rfc/cosmo-streams-v1.md
```diff
@@ -0,0 +1,1004 @@
...
+            // here you can umarshal the old data and map it to the new format
+            // for example:
+            // var dataReceived struct {
+            //     EmployeeId string `json:"EmployeeId"`
+            // }
+            // err := json.Unmarshal(natsEvent.Data, &dataReceived)
+
+            // create the new event
+            newEvent := &NatsEvent{
```
jensneuse, line 371: I'd like to see how this will be used to map from e.g. a Kafka Header to a __typename and \@key field

rfc/cosmo-streams-v1.md
```diff
@@ -0,0 +1,1004 @@
...
+The hook arguments of `StreamPublishEventHook` are:
+* `ctx StreamPublishEventHookContext`: The stream context, which contains the provider ID and the publish configuration
+* `events []StreamEvent`: The events that are going to be sent to the provider
+
+The hook will return a new slice of events that will be used to emit the events to the provider.
+The hook will also return an error if one of the events cannot be processed, preventing the events from being processed.
+
+#### Do we need two new hooks?
+
+Another possible solution for mapping outward data would be to use the existing middleware hooks `RouterOnRequestHandler` or `RouterMiddlewareHandler` to intercept the mutation, access the stream context, and emit the event to the stream. However, this would require exposing a stream context in the request lifecycle, which is difficult. It would also require coordination to ensure that an event emitted on the stream is sent only after the subscription starts.
```
jensneuse, line 421: I think it would be beneficial if we could expose a hook on any context that allows the a Router middleware to publish events to a stream.

E.g. after successfully executing a mutation, such a hook could be used to emit an event to all clients across all Routers. To make this easy, we could create a method on the Request Context with all available stream clients that allows us to publish events on the streams.

rfc/cosmo-streams-v1.md
```diff
@@ -0,0 +1,1004 @@
...
+    natsConfig := ctx.SubscriptionEventConfiguration().(*nats.SubscriptionEventConfiguration)
+    if natsConfig.Subjects[0] != "topic-with-internal-data-format" {
+        return events, nil
+    }
+
+    // create a new slice of events that we will return with the events that are allowed to be received by the client
+    newEvents := make([]StreamEvent, 0, len(events))
+
+    // get the client's allowed entities IDs
+    clientAllowedEntitiesIds, found := ctx.RequestContext().Authentication().Claims()["allowedEntitiesIds"]
```
jensneuse, line 486: The name "OnStreamEvents" indicates that this func is called for a batch of events from the stream. However, this line seems to contradict  as it shows RequestContext.

To be clear, will this func run for every subscription? Or for the trigger?

Maybe we should separate the two. One hook on the stream (trigger), one hook for all events on the Subscription?

rfc/cosmo-streams-v1.md
```diff
@@ -0,0 +1,1004 @@
...
+    ProviderTypeKafka ProviderType = "kafka"
+    ProviderTypeRedis ProviderType = "redis"
+}
+
+```
+
+## Appendix 2, Using AsyncAPI for Event Data Structure
+
+As a side note, it is important to find ways to document the data that is arriving and going out of the cosmo streams engine. This could allow some automatic code generation starting from the schema and the events data.
+As an example, we are going to explore how AsyncAPI could be used to generate the data structures for the custom modules and assure the messages format.
```
jensneuse, line 820: It's a very good idea which you might even take one step further.

1. Allow the user to define their streams using AsyncAPI
2. generate fully typesafe hooks with all events, etc...

rfc/cosmo-streams-v1.md
alepane21: Changed to RootFieldName, thanks!

rfc/cosmo-streams-v1.md
alepane21: Right now this RFC is only covering the "initial message" need. That means that the message will be delivered only to the specific subscription that started the subscription.
Do you think that we will also have to cover the emit events at any lifecycle event with the cosmo streams v1 release, or it could be delayed?

rfc/cosmo-streams-v1.md
alepane21: I changed it to "Transforming events data from the format that could be used by the external system to/from Federation compatible Router events"

rfc/cosmo-streams-v1.md
alepane21: That was my idea, but I missed to wrote about this behaviour. I've added this part to the spec: "The message will be received only by the client that subscribed to the stream, and not by the other clients that subscribed to the same stream."

rfc/cosmo-streams-v1.md
alepane21: I explicited the behaviour in the comment. That was how I intended the feature to work.

rfc/cosmo-streams-v1.md
alepane21: This function is going to be called for each client that subscribed, before delivering the events to the client.
Initially I had like two different hooks:
- OnStreamEvents, called one time for each chunk of events and without access to the request context (so, with two subscriptions active, it would still be called only one time)
- OnStreamEventsDelivery, called one time for each chunk of events being delivered to the clients (so, with two subscriptions active, it would be called two times).

But adding two different hooks looked a bit useless. Maybe could be a good thing on the performance side, but on the developer side it will be two very similar things to choose between...

rfc/cosmo-streams-v1.md
alepane21: This would for sure make the devs life easier and hard to use the wrong event!
I've added an Outlook section to the appendix, thanks!

rfc/cosmo-streams-v1.md
alepane21: whops, yes it could, I've added a check on the examples

rfc/cosmo-streams-v1.md
alepane21: I think we can discuss this here https://github.com/wundergraph/cosmo/pull/2035#discussion_r2204360073
I kept this feature out of the RFC because I thought it was different from the initial message feature, and it could be introduced with a next RFC after the v1 release. But I'm ready to include the feature in this RFC if you think it is important 💪 .

rfc/cosmo-streams-v1.md
alepane21: I've added the third point and referenced to the Filtering proposal, that is going to address it.

rfc/cosmo-streams-v1.md
alepane21: Good point. I've added a StreamHookError that, if returned, could allow the dev to customize the behaviour.

rfc/cosmo-streams-v1.md
alepane21: Added some details

rfc/cosmo-streams-v1.md
alepane21: added a comment with a kafka example

rfc/cosmo-streams-v1.md
alepane21: I've added the code in the example 1 for nats, for Kafka it would be the same thing!

rfc/cosmo-streams-v1.md
alepane21: We will use only the OnStreamEvents, if with the POC we see some performance issue, we will bring this back.

rfc/cosmo-streams-v1.md
alepane21: We will implement this later.

rfc/cosmo-streams-v1.md
alepane21: We will implement this later.

PR #2059, opened 2025-07-17
feat: add SubscriptionOnStartHandler

alepane21:
## Summary by CodeRabbit

* **New Features**
  * Introduced a flexible hooks system for stream subscriptions, enabling custom authorization, initial event emission, event filtering, and data mapping across NATS, Kafka, and Redis providers.
  * Added support for executing custom logic at subscription start, including emitting initial events and controlling subscription lifecycle.
  * Unified event and configuration abstractions for consistent handling across different pubsub providers.
  * Added a new module to facilitate implementing subscription start hooks with logging and callbacks.

* **Bug Fixes**
  * Improved consistency and structure in event data handling for NATS, Kafka, and Redis integrations.

* **Tests**
  * Added extensive tests covering subscription start hooks, event emission, authorization checks, and subscription lifecycle management.

* **Chores**
  * Updated dependencies, removed unused indirect packages, and enhanced mock generation for subscription event updaters across pubsub providers.

Depends on engine PR: https://github.com/wundergraph/graphql-go-tools/pull/1243

## How to use the new hook?

The hook implemented in this PR is the `SubscriptionOnStartHandler` as defined here: https://github.com/wundergraph/cosmo/blob/ceac25583049c027f289f4508d6477355298fc65/router/core/subscriptions_modules.go#L107

#### Query

On the demo graph, we will create a module that add some logic to the following query

```graphql
employeeUpdatedMyKafka(employeeID: $employeeID)
```

#### Module code that implement the new hook

As an example, this module will emit a message if the subscription is started with employeeID == 1.

```go
const myModuleID = "startSubscriptionModule"

type StartSubscriptionModule struct {}

func (m *StartSubscriptionModule) SubscriptionOnStart(ctx core.SubscriptionOnStartHookContext) (bool, error) {
    if ctx.SubscriptionEventConfiguration().RootFieldName() != "employeeUpdatedMyKafka" {
      return false, nil
    }
    employeeId := ctx.RequestContext().Operation().Variables().GetInt64("employeeID")
    if employeeId != 1 {
        return false, nil
    }
    ctx.WriteEvent(&kafka.Event{
        Data: []byte(`{"id": 1, "__typename": "Employee"}`),
    })
    return false, nil
}

func (m *StartSubscriptionModule) Module() core.ModuleInfo {
	return core.ModuleInfo{
		// This is the ID of your module, it must be unique
		ID: myModuleID,
		// The priority of your module, lower numbers are executed first
		Priority: 1,
		New: func() core.Module {
			return &StartSubscriptionModule{}
		},
	}
}
```

## Checklist

### DIFF
--- a/router/go.sum
+++ b/router/go.sum
@@ -321,8 +321,6 @@ github.com/wk8/go-ordered-map/v2 v2.1.8 h1:5h/BUHu93oj4gIdvHHHGsScSTMijfx5PeYkE/
 github.com/wk8/go-ordered-map/v2 v2.1.8/go.mod h1:5nJHM5DyteebpVlHnWMV0rPz6Zp7+xBAnxjb1X5vnTw=
 github.com/wundergraph/astjson v0.0.0-20250106123708-be463c97e083 h1:8/D7f8gKxTBjW+SZK4mhxTTBVpxcqeBgWF1Rfmltbfk=
 github.com/wundergraph/astjson v0.0.0-20250106123708-be463c97e083/go.mod h1:eOTL6acwctsN4F3b7YE+eE2t8zcJ/doLm9sZzsxxxrE=
-github.com/wundergraph/graphql-go-tools/v2 v2.0.0-rc.229 h1:VCfCX/xmpBGQLhTHJMHLugzJrXJk/smjLRAEruCI0HY=
-github.com/wundergraph/graphql-go-tools/v2 v2.0.0-rc.229/go.mod h1:g1IFIylu5Fd9pKjzq0mDvpaKhEB/vkwLAIbGdX2djXU=
 github.com/wundergraph/graphql-go-tools/v2 v2.0.0-rc.229.0.20250930144208-ddc652f78bbb h1:stBTAle5FyytsTNxYeCwNzYlyhKzlS4he6f7/y6O3qE=
 github.com/wundergraph/graphql-go-tools/v2 v2.0.0-rc.229.0.20250930144208-ddc652f78bbb/go.mod h1:g1IFIylu5Fd9pKjzq0mDvpaKhEB/vkwLAIbGdX2djXU=
 github.com/yosida95/uritemplate/v3 v3.0.2 h1:Ed3Oyj9yrmi9087+NczuL5BwkIc4wvTb5zIM+UJPGz4=


alepane21:
@coderabbitai review

alepane21:
@coderabbitai full review

alepane21:
@coderabbitai full review

alepane21:
@coderabbitai full review

alepane21:
@coderabbitai full review

alepane21:
@coderabbitai full review

alepane21:
@coderabbitai full review

router/pkg/pubsub/datasource/hookeddatasource.go
```diff
@@ -0,0 +1,30 @@
...
+	return
+}
+
+func (h *HookedSubscriptionDataSource) Start(ctx *resolve.Context, input []byte, updater resolve.SubscriptionUpdater) error {
+	return h.SubscriptionDataSource.Start(ctx, input, NewSubscriptionEventUpdater(updater))
+}
+
+func (h *HookedSubscriptionDataSource) UniqueRequestID(ctx *resolve.Context, input []byte, xxh *xxhash.Digest) (err error) {
+	return h.SubscriptionDataSource.UniqueRequestID(ctx, input, xxh)
+}
```
Noroth: Can we get some comments about this type and functions and also does it have to be exported?

router/pkg/pubsub/kafka/engine_datasource.go
```diff
@@ -9,31 +9,94 @@ import (
...
 type SubscriptionDataSource struct {
 	pubSub Adapter
 }
 
+func (s *SubscriptionDataSource) SubscriptionEventConfiguration(input []byte) datasource.SubscriptionEventConfiguration {
+	var subscriptionConfiguration SubscriptionEventConfiguration
+	err := json.Unmarshal(input, &subscriptionConfiguration)
+	if err != nil {
+		return nil
+	}
```
Noroth:
```suggestion
	if err := json.Unmarshal(input, &subscriptionConfiguration); err != nil {
		return nil
	}
```

router/pkg/pubsub/kafka/engine_datasource.go
```diff
@@ -9,31 +9,94 @@ import (
...
+
+	return fmt.Sprintf(`{"topic":"%s", "event": {"data": %s, "key": "%s", "headers": %s}, "providerId":"%s"}`, s.Topic, s.Event.Data, s.Event.Key, headersBytes, s.ProviderID()), nil
 }
 
 type SubscriptionDataSource struct {
 	pubSub Adapter
 }
 
+func (s *SubscriptionDataSource) SubscriptionEventConfiguration(input []byte) datasource.SubscriptionEventConfiguration {
```
Noroth: If we for some reason encounter an error here we will never know that it happened?

router/pkg/pubsub/kafka/engine_datasource.go
```diff
@@ -54,14 +117,18 @@ func (s *SubscriptionDataSource) UniqueRequestID(ctx *resolve.Context, input []b
...
 
-func (s *SubscriptionDataSource) Start(ctx *resolve.Context, input []byte, updater resolve.SubscriptionUpdater) error {
-	var subscriptionConfiguration SubscriptionEventConfiguration
-	err := json.Unmarshal(input, &subscriptionConfiguration)
-	if err != nil {
-		return err
+func (s *SubscriptionDataSource) Start(ctx *resolve.Context, input []byte, updater datasource.SubscriptionEventUpdater) error {
+	subConf := s.SubscriptionEventConfiguration(input)
+	if subConf == nil {
+		return fmt.Errorf("no subscription configuration found")
```
Noroth: No formatting here and in line 128. Use `errors.New()`

router/pkg/pubsub/nats/engine_datasource.go
```diff
@@ -9,38 +9,90 @@ import (
...
 type SubscriptionSource struct {
 	pubSub Adapter
 }
 
+func (s *SubscriptionSource) SubscriptionEventConfiguration(input []byte) datasource.SubscriptionEventConfiguration {
+	var subscriptionConfiguration SubscriptionEventConfiguration
+	err := json.Unmarshal(input, &subscriptionConfiguration)
+	if err != nil {
+		return nil
+	}
```
Noroth:
```suggestion
	if err := json.Unmarshal(input, &subscriptionConfiguration); err != nil {
		return nil
	}
```

router/pkg/pubsub/nats/engine_datasource.go
```diff
@@ -9,38 +9,90 @@ import (
...
 	// e.g. {"id":$$0$$,"update":$$1$$}
-	return fmt.Sprintf(`{"subject":"%s", "data": %s, "providerId":"%s"}`, s.Subject, s.Data, s.ProviderID)
+	return fmt.Sprintf(`{"subject":"%s", "event": {"data": %s}, "providerId":"%s"}`, s.Subject, s.Event.Data, s.ProviderID()), nil
 }
 
 type SubscriptionSource struct {
 	pubSub Adapter
 }
 
+func (s *SubscriptionSource) SubscriptionEventConfiguration(input []byte) datasource.SubscriptionEventConfiguration {
```
Noroth: Same here, silent error.

router/pkg/pubsub/pubsub.go
```diff
@@ -49,6 +49,10 @@ func (e *ProviderNotDefinedError) Error() string {
 	return fmt.Sprintf("%s provider with ID %s is not defined", e.ProviderTypeID, e.ProviderID)
 }
 
+type Hooks struct {
+	SubscriptionOnStart []pubsub_datasource.SubscriptionOnStartFn
+}
```
Noroth, line 57: Add some comments please. Maybe also explain why we have a struct wrapper for `[]pubsub_datasource.SubscriptionOnStartFn`

router/pkg/pubsub/datasource/hookeddatasource.go
alepane21: Thanks to your comment, I went on a thought train and realized that this structure revealed an opportunity to remove boilerplate code across all implementations!
I've now introduced a PubSubSubscriptionDataSource using generics, which also handles the hooks behaviour. It can be shared across all implementations that previously had to re-implement an almost identical SubscriptionDataSource.

router/pkg/pubsub/pubsub.go
alepane21: It will be helpful because with the next PR I'll add two more hooks, without this structure we will have to add a param to pubsub methods for each hook.

router/pkg/pubsub/nats/engine_datasource.go
alepane21: fixed!

router/pkg/pubsub/nats/engine_datasource.go
alepane21: fixed

router/pkg/pubsub/kafka/engine_datasource.go
alepane21: fixed

router/pkg/pubsub/kafka/engine_datasource.go
alepane21: fixed

router/pkg/pubsub/kafka/engine_datasource.go
alepane21: fixed

router/pkg/pubsub/datasource/subscription_datasource.go
```diff
@@ -0,0 +1,75 @@
...
+	subscriptionOnStartFns []SubscriptionOnStartFn
+}
+
+func (s *PubSubSubscriptionDataSource[C]) SubscriptionEventConfiguration(input []byte) (SubscriptionEventConfiguration, error) {
+	var subscriptionConfiguration C
+	err := json.Unmarshal(input, &subscriptionConfiguration)
+	if err != nil {
+		return nil, err
+	}
+	return subscriptionConfiguration, nil
```
Noroth: nit: I think you could either remove the if or inline the error

```suggestion
	var subscriptionConfiguration C
	err := json.Unmarshal(input, &subscriptionConfiguration)
	return subscriptionConfiguration, err
```

Noroth (approved):

adr/cosmo-streams-v1.md
```diff
@@ -0,0 +1,398 @@
...
+
+- **Authorization**: Implementing authorization checks at the start of subscriptions
+- **Initial message**: Sending an initial message to clients upon subscription start
+- **Data mapping**: Transforming events data from the format that could be used by the external system to/from Federation compatible Router events
+- **Event filtering**: Filtering events using custom logic
+
+## Backwards Compatibility
+
+The new hooks can be integrated in the router in a fully backwards compatible way.
+When the new module system will be released, some changes will be needed.
```
StarpTech: Could you explain what changes?

adr/cosmo-streams-v1.md
```diff
@@ -0,0 +1,398 @@
...
+    // the subscription event configuration
+    SubscriptionEventConfiguration() SubscriptionEventConfiguration
+    // write an event to the stream of the current subscription
+    WriteEvent(event core.StreamEvent)
+}
+
+type SubscriptionOnStartHandler interface {
+    // OnSubscriptionOnStart is called once at subscription start
+    // Returning an error will result in a GraphQL error being returned to the client, could be customized returning a StreamHookError.
+    SubscriptionOnStart(ctx SubscriptionOnStartHookContext) error
```
StarpTech, line 89: The implementation appears different. We return a `(bool, error)` tuple, which caused some confusion. I initially thought that returning `true` with no error would indicate a successful subscription but it's not.
I recommend implementing it as discussed here: using the presence of an error to indicate whether the user wants to abort the operation. This approach clearly communicates the user's intent.

StarpTech, line 89: I experienced during testing that the websocket preflight request was made although I returned `(true, nil)` in `SubscriptionOnStart`. As a User, I'd expect that nothing is happening before I decided to continue with the subscription in the hook.

This result in a nil pointer exception because `resp` of the `OnOriginResponse` was nil.

StarpTech (changes requested):

adr/cosmo-streams-v1.md
alepane21: Done!

adr/cosmo-streams-v1.md
alepane21: Changed back to how it is in the ADR.

alepane21: ## Preflight behaviour

The hook is attached to the subscription, so it is only triggered after the websocket has been established and the subscription payload is sent to the websocket. It is only after that that I have the data useful for the SubscriptionOnStart hook (operation name, content and variables, field name, etc). Without the query I don't know a lot.
I think that if a customer want to put some check before the subscription is even started, should use the RouterOnRequestHandler.
I'm obviously open to change if you think that how it is working now is not intuitive.

## Panics

I've created a test that replicate the issue and I'm working on it

alepane21: I fixed the issue about the origin getting called. It was an oversight on my side in the engine :(.
Now I'm going to wait to start the datasource subscription until the initial hooks is done.
Before I could cancel the connection to the datasource in a random moment after the initial hooks were processed.
The fix is in the related engine repo https://github.com/wundergraph/graphql-go-tools/pull/1243.
I also added a test in the router-tests for cover for this issue.

This was a really great finding, thanks!

router/core/subscriptions_modules.go
```diff
@@ -0,0 +1,165 @@
...
+		message:         message,
+		statusCode:      statusCode,
+		code:            code,
+		closeSubscription: closeSubscription,
+	}
+}
+
+type SubscriptionOnStartHookContext interface {
+	// the request context
+	RequestContext() RequestContext
```
StarpTech: Whatever we expose here needs to be tested. E.g., the ResponseWriter wouldn't work in subscriptions. Ideally, in our first iteration, we are purely use case driven and expose only functionality and data that is necessary to enable these.

router/core/subscriptions_modules.go
```diff
@@ -0,0 +1,165 @@
...
+			subscriptionEventConfiguration: subConf,
+			writeEventHook:                 resolveCtx.TryEmitSubscriptionUpdate,
+		}
+
+		err := fn(hookCtx)
+
+		// Check if the error is a StreamHookError and should close the connection
+		var streamHookErr *StreamHookError
+		close := false
+		if errors.As(err, &streamHookErr) {
```
StarpTech: I think any error should close the subscription. Why do we want to do it selectively?

router/pkg/pubsub/kafka/adapter.go
```diff
@@ -88,31 +87,44 @@ func (p *ProviderAdapter) topicPoller(ctx context.Context, client *kgo.Client, u
 				r := iter.Next()
 
 				p.logger.Debug("subscription update", zap.String("topic", r.Topic), zap.ByteString("data", r.Value))
-				updater.Update(r.Value)
+				headers := make(map[string][]byte)
```
StarpTech, line 103: Do we already have the requirement for this? If not, I'd reduce the implementation to the absolute minimum while allowing for further extensions in the future.

router/pkg/pubsub/redis/engine_datasource_factory.go
```diff
@@ -59,26 +61,46 @@ func (c *EngineDataSourceFactory) ResolveDataSourceInput(eventData []byte) (stri
...
 
 // ResolveDataSourceSubscription returns the subscription data source
-func (c *EngineDataSourceFactory) ResolveDataSourceSubscription() (resolve.SubscriptionDataSource, error) {
-	return &SubscriptionDataSource{
-		pubSub: c.RedisAdapter,
-	}, nil
+func (c *EngineDataSourceFactory) ResolveDataSourceSubscription() (datasource.SubscriptionDataSource, error) {
+	return datasource.NewPubSubSubscriptionDataSource[*SubscriptionEventConfiguration](
+		c.RedisAdapter,
+		func(ctx *resolve.Context, input []byte, xxh *xxhash.Digest) error {
```
StarpTech, line 77: Why did we move this outside of the provider specific implementation?

alepane21: Because the hook is managed inside the PubSubSubscriptionDataSource. Having it in the datasource allows the implemented providers (and the future ones!) to just return the PubSubSubscriptionDataSource to implements the hooks.
An alternative solution that I thought was to embed this struct inside another one that was implementing the hooks, but looked like an overly complicated solution.

router/pkg/pubsub/kafka/adapter.go
alepane21: Yes, one of the requirements is to read/write kafka headers.

router/core/subscriptions_modules.go
alepane21: Mmm I see, so you would prefer an ad-hoc version (I'm thinking of a `SubscriptionRequestContext`) with only things that make sense?

router/core/subscriptions_modules.go
alepane21: The idea was that this would allow the customer to customize the client's behaviour: if an error with closeConnection set to false was returned, the client could have just write something to the UI (like "no initial data found, waiting for the first real data") and then, when a message with the data arrived, the client could just read it.

In the real world, the client will simply close the subscription on the first error it receives, so this behaviour makes no sense :(

I'll remove it to avoid complexity and confusion.

Issue #2056, opened 2025-07-17
Ability to configure the min size for response compression


mrpahdan:
### Component(s)

router

### Is your feature request related to a problem? Please describe.

When making a request to the router, if the response is under 4KB, the response returned is not compressed. If the router receives a lot of small requests, there's a lot of unnecessary network traffic. This is an issue for router hosted in various cloud providers where data transfer costs through gateways and across networks incurs a cost.

### Describe the solution you'd like

I would like to introduce a new configuration item called: `response_compression_min_size`.

It would sit under the yaml: traffic_shaping.router as such:
```
traffic_shaping:
  router:
    response_compression_min_size: 1KB
```

This configuration item would be consumed [here](https://github.com/wundergraph/cosmo/blob/3f4dea6196f5efcd175eab8885230b83b445d24a/router/core/graph_server.go#L286C10-L286C17)

The default value should be set to 4KB so the same setting is applied.

see proposed solution: https://github.com/wundergraph/cosmo/commit/dd443fb310db10c93650d5f118aacfae43a0c737

SkArchon:
Looks good to me, you can proceed ahead.

PR #2060, opened 2025-07-17
feat: ability to configure the min size for response compression

mrpahdan:
Add ability to configure the min size for response compression.

## Checklist

Resolve issue: https://github.com/wundergraph/cosmo/issues/2056

Documentation update: https://github.com/wundergraph/cosmo-docs/pull/124

## Summary by CodeRabbit

* **New Features**
  * Added a configurable option to set the minimum response size required to enable HTTP response compression.
  * Updated configuration files and schema to support the new compression size setting, with a default value of 4KB.  
* **Tests**
  * Introduced tests to verify response compression behavior with custom minimum size settings.

### DIFF
--- a/router/internal/requestlogger/requestlogger.go
+++ b/router/internal/requestlogger/requestlogger.go
@@ -2,15 +2,16 @@ package requestlogger
 
 import (
 	"crypto/sha256"
-	"fmt"
+	"encoding/hex"
+	"net"
+	"net/http"
+	"time"
+
 	"github.com/wundergraph/cosmo/router/internal/errors"
 	"github.com/wundergraph/cosmo/router/internal/expr"
 	"github.com/wundergraph/cosmo/router/pkg/config"
 	"github.com/wundergraph/cosmo/router/pkg/logging"
 	rtrace "github.com/wundergraph/cosmo/router/pkg/trace"
-	"net"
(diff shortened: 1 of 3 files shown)


router/pkg/config/config.schema.json
```diff
@@ -1468,6 +1468,15 @@
               "type": "boolean",
               "description": "When enabled, the router will check incoming requests for a 'Content-Encoding' header and decompress the body accordingly. Currently only gzip is supported",
               "default": true
+            },
+            "response_compression_min_size": {
+              "type": "string",
+              "format": "bytes-string",
+              "description": "The minimum size of the response body in bytes to enable response compression. The size is specified as a string with a number and a unit, e.g. 10KB, 1MB, 1GB. The supported units are 'B', 'KB', 'MB', 'GB'.",
+              "default": "4KB",
```
SkArchon: Let's set the value to 4KiB, as otherwise it's actually a breaking change since the new value is 4000, but the old default is 4096. You will need to update the config test snapshots as well.

SkArchon: I don't think we need to specify the supported units in the description, but the examples could incorporate "5 Kib", we could also mention the difference between KB and KiB in the docs.

PR #2067, opened 2025-07-21
feat: authorization using secrets

SkArchon:
@coderabbitai summary

## Checklist


router/pkg/authentication/jwks_token_decoder.go
```diff
@@ -42,37 +42,72 @@ type JWKSConfig struct {
...
+			jwk, err := jwkset.NewJWKFromKey([]byte(c.Secret), jwkOptions)
+			if err != nil {
+				return nil, err
+			}
+
+			err = given.KeyWrite(ctx, jwk)
+			if err != nil {
+				return nil, err
+			}
 		}
```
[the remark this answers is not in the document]
SkArchon: Will fix the error wrapping, the empty validations are handled via the json schema.

router/pkg/config/config.schema.json
```diff
@@ -1634,11 +1634,28 @@
...
                     "format": "http-url"
                   },
+                  "secret": {
+                    "type": "string",
+                    "description": "The secret of the JWKs"
+                  },
+                  "algorithm": {
+                    "type": "string",
+                    "description": "The algorithm used",
+                    "enum": [
```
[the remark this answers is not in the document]
SkArchon: This needs to be potentially backwards compatible, thus I added a warning

StarpTech:
@SkArchon please update the title. This would not describe the feature accurately.

router-tests/authentication_test.go
```diff
@@ -753,6 +753,175 @@ func TestAuthorization(t *testing.T) {
 			require.Equal(t, employeesExpectedData, string(data))
 		})
 	})
+
+	t.Run("authenticate when multiple jwks are present", func(t *testing.T) {
```
StarpTech: What is the actual expectation of this test? Could you make it clearer?

SkArchon: Renamed, essentially, this test ensures that even if authentication with some configurations fails, as long as one configuration succeeds, authentication still succeeds.

router-tests/authentication_test.go
```diff
@@ -753,6 +753,175 @@ func TestAuthorization(t *testing.T) {
...
+			data, err := io.ReadAll(res.Body)
+			require.NoError(t, err)
+			require.Equal(t, employeesExpectedData, string(data))
+		})
+	})
+
+}
+
+func TestNonHttpAuthorization(t *testing.T) {
+	t.Run("valid HS256 token", func(t *testing.T) {
```
Noroth: Can we get more descriptive texts for tests? something like `should succeed with valid token` etc to make sure you can read the test name and directly understand what your expectation is instead of reading the test code?

router/pkg/config/config.schema.json
```diff
@@ -1634,11 +1634,28 @@
...
+                  "key_id": {
+                    "type": "string",
+                    "description": "The secret of the JWKs"
+                  },
                   "algorithms": {
                     "type": "array",
                     "description": "The allowed algorithms for the keys that are retrieved from the JWKs. An empty list means that all algorithms are allowed.",
                     "items": {
-                      "type": "string",
+                      "type": "string"  ,
```
Noroth: formatting issue?

router/pkg/config/config.schema.json
```diff
@@ -1634,11 +1634,28 @@
...
                       "enum": [
                         "HS256",
                         "HS384",
```
Noroth: Should we exclude symmetric algorithms from the jwks server?

SkArchon: I removed it from here. Though this will only break for users who have explicitly specify HS* algorithms.

I did not remove it from the list on the validation_store as I believe this would be a runtime breaking change (and users might not be aware of it), as we only get to know the algorithm when the http endpoint is queried. I think instead we should make specifying algorithms explicit. I will however open a ticket for this so we can align on this separately.

router-tests/authentication_test.go
SkArchon: Done

router/pkg/config/config.schema.json
SkArchon: Done

PR #2074, opened 2025-07-23
feat: add audience validation

SkArchon:
## Summary by CodeRabbit

* **New Features**
  * Added support for specifying and validating JWT audiences in JWKS authentication configuration.
  * Audience validation now enforced for JWT tokens, improving security by ensuring tokens are intended for the configured audiences.

* **Bug Fixes**
  * Improved error handling for duplicate JWKS URLs and key IDs in configuration.

* **Tests**
  * Expanded test coverage for audience validation and duplicate JWKS/key ID detection in authentication setup.

* **Documentation**
  * Updated configuration schema and example files to include the new "audiences" field for JWKS.

> [!WARNING]  
> Potential Breaking Change: If you have the same KID manually specified (in the config) or are using the same URL, the router won't startup.

## Checklist

### DIFF
--- a/router-tests/authentication_test.go
+++ b/router-tests/authentication_test.go
@@ -2465,6 +2465,8 @@ func TestAudienceValidation(t *testing.T) {
 	})
 
 	t.Run("authentication validation is ignored when expected aud is not provided", func(t *testing.T) {
+		t.Parallel()
+		
 		tokenAudiences := []bool{true, true}
 
 		authServer, err := jwks.NewServer(t)
--- a/router-tests/authentication_test.go
+++ b/router-tests/authentication_test.go
@@ -683,6 +683,27 @@ func TestAuthenticationWithCustomHeaders(t *testing.T) {
 func TestHttpJwksAuthorization(t *testing.T) {
 	t.Parallel()
 
+	t.Run("startup should fail when duplicate URLs are specified", func(t *testing.T) {
+		t.Parallel()
+
+		authServer, err := jwks.NewServer(t)
+		require.NoError(t, err)
+		t.Cleanup(authServer.Close)
+
+		_, err = authentication.NewJwksTokenDecoder(NewContextWithCancel(t), zap.NewNop(), []authentication.JWKSConfig{
+			{
+				URL:             authServer.JWKSURL(),
+				RefreshInterval: 2 * time.Second,
+			},
+			{
+				URL:             authServer.JWKSURL(),
+				RefreshInterval: 2 * time.Second,
+			},
+		})
+
+		require.ErrorContains(t, err, "duplicate JWK URL found")
+	})
+
 	t.Run("authentication should fail with no token", func(t *testing.T) {
 		t.Parallel()
 
@@ -765,7 +786,10 @@ func TestHttpJwksAuthorization(t *testing.T) {
 		t.Cleanup(authServer2.Close)
 		require.NoError(t, err)
 
-		token, err := authServer2.Token(nil)
+		// aud claim
+		token, err := authServer2.Token(map[string]any{
+			"aud": "https://example.com",
+		})
 		require.NoError(t, err)
 
 		authenticators := ConfigureAuthWithJwksConfig(t, []authentication.JWKSConfig{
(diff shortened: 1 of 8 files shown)


router/pkg/authentication/jwks_token_decoder.go
```diff
@@ -132,8 +201,19 @@ func NewJwksTokenDecoder(ctx context.Context, logger *zap.Logger, configs []JWKS
...
+	for _, item := range tokenAudiences {
+		set[item] = struct{}{}
+	}
+	for _, item := range expectedAudiences {
+		if _, found := set[item]; found {
+			return true
+		}
+	}
+	return false
 }
```
jensneuse:
```suggestion
func hasAudience(tokenAudiences, expectedAudiences []string) bool {
	for _, expected := range expectedAudiences {
		for _, token := range tokenAudiences {
			if token == expected {
				return true
			}
		}
	}
	return false
}
```

jensneuse (commented):
One minor feedback. Please also check with @StarpTech

router-tests/authentication_test.go
```diff
@@ -2028,6 +2074,436 @@ func TestAuthenticationOverWebsocket(t *testing.T) {
...
+			require.Equal(t, http.StatusUnauthorized, res.StatusCode)
+			require.Equal(t, "", res.Header.Get(xAuthenticatedByHeader))
+			data, err := io.ReadAll(res.Body)
+			require.NoError(t, err)
+			require.JSONEq(t, unauthorizedExpectedData, string(data))
+		})
+
+	})
+
+	t.Run("authentication validation is ignored when expected aud is not provided", func(t *testing.T) {
```
StarpTech: Typo? You mean "Audience is not validated when an empty set is provided,"

SkArchon: Resolved

router/pkg/authentication/jwks_token_decoder.go
SkArchon: I would prefer to keep is it as it is since the proposed solution is O(N^2)

router/pkg/authentication/jwks_token_decoder.go
```diff
@@ -46,16 +47,26 @@ type JWKSConfig struct {
...
 	Algorithm string
 	KeyId     string
-}
 
-func NewJwksTokenDecoder(ctx context.Context, logger *zap.Logger, configs []JWKSConfig) (TokenDecoder, error) {
+	Audiences []string
+}
 
-	remoteJWKSets := make(map[string]jwkset.Storage)
+type AudKey struct {
```
Noroth: I think this type doesn't have to be exported as it isn't used anywhere else

router/pkg/authentication/jwks_token_decoder.go
```diff
@@ -46,16 +47,26 @@ type JWKSConfig struct {
...
+	kid string
+	url string
+}
 
-	given := jwkset.NewMemoryStorage()
+func NewJwksTokenDecoder(ctx context.Context, logger *zap.Logger, configs []JWKSConfig) (TokenDecoder, error) {
+	audiencesMap := make(map[AudKey][]string, len(configs))
```
Noroth: We could optimize this a bit as we are iterating on every request.
Imagine something like (high level)

```go
type set[T comparable] map[T]struct{}{}
type audienceStore map[AudKey]set[string]
```

Then you can use the `set` in the `hasAudience` function directly

router/pkg/authentication/jwks_token_decoder.go
SkArchon: Done

router/pkg/authentication/jwks_token_decoder.go
SkArchon: Done, didn't make it a generic type though and made it a more specific type of `audienceSet`

router/pkg/authentication/jwks_token_decoder.go
SkArchon: Refactored this further, as per @Noroth's suggestion.

Noroth (approved):
LGTM

PR #2076, opened 2025-07-23
fix: prevent use of hs3256 hs symmetric keys as urls and disallow http urls by default

SkArchon:
@coderabbitai summary

## Checklist


SkArchon:
@coderabbitai review

router/pkg/authentication/jwks_token_decoder.go
```diff
@@ -56,18 +57,33 @@ type AudKey struct {
 	url string
 }
 
-func NewJwksTokenDecoder(ctx context.Context, logger *zap.Logger, configs []JWKSConfig) (TokenDecoder, error) {
+func NewJwksTokenDecoder(ctx context.Context, logger *zap.Logger, configs []JWKSConfig, allowInsecureJwksUrls bool) (TokenDecoder, error) {
```
Noroth: I have the feeling that instead of adding a bool flag to the parameters, we could have a more descriptive config type
```go
type JwksTokenDecoderConfig struct {
  Logger *zap.Logger
  Configs []JWKSConfig
  AllowInsecureJwksUrls
}

func NewJwksTokenDecoder(ctx context.Context, config JwksTokenDecoderConfig) (TokenDecoder, error) {
```

Noroth: I just noticed that the naming of `JWKS` is different than in other places. I know this has been like this before but maybe we can refactor it to `JwksConfig`

SkArchon: Done

SkArchon: Actually upon reading https://go.dev/wiki/CodeReviewComments#initialisms, it should be the opposite, except for when we want it to be unexported, I renamed a few places from "Jwks" to "JWKS".

SkArchon:
Discussed with @StarpTech, we will not do this change.

PR #2077, opened 2025-07-23
fix: improve gzip request decompression middleware

endigma:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of compressed request bodies to ensure proper resource cleanup and prevent potential decompression attacks.
  * Adjusted middleware order to enforce request size limits after decompression, enhancing security for compressed requests.

* **Tests**
  * Added tests to verify that oversized plain and compressed request bodies are correctly rejected.
  * Refactored existing compression middleware tests for improved clarity and reliability.

## Checklist

### DIFF
--- a/router/core/graph_server.go
+++ b/router/core/graph_server.go
@@ -237,11 +237,12 @@ func newGraphServer(ctx context.Context, r *Router, routerConfig *nodev1.RouterC
 		)
 	})))
 
-	// Request traffic shaping related middlewares
-	httpRouter.Use(rmiddleware.RequestSize(int64(s.routerTrafficConfig.MaxRequestBodyBytes)))
 	if s.routerTrafficConfig.DecompressionEnabled {
 		httpRouter.Use(rmiddleware.HandleCompression(s.logger))
 	}
+
+	// Request traffic shaping related middlewares, happens after decompression to prevent unbounded decompression attacks
+	httpRouter.Use(rmiddleware.RequestSize(int64(s.routerTrafficConfig.MaxRequestBodyBytes)))
 
 	httpRouter.Use(middleware.RequestID)
 	httpRouter.Use(middleware.RealIP)
--- a/router/internal/middleware/compression.go
+++ b/router/internal/middleware/compression.go
@@ -2,6 +2,7 @@ package middleware
 
 import (
 	"compress/gzip"
+
 	"net/http"
 	"strings"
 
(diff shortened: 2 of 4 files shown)


router/internal/middleware/compression.go
```diff
@@ -32,14 +33,19 @@ func HandleCompression(logger *zap.Logger) func(http.Handler) http.Handler {
 					return
 				}
 
+				originalBody := r.Body
```
StarpTech, line 36: Don't we also need to reset the contentLength values after body compression?

```go
// Content-Length is no longer valid after decompression
r.Header.Del("Content-Length")
r.ContentLength = -1
```

endigma: Yes, I fixed this actually but it got lost in git somehow

endigma: Fixed in latest

endigma:
@coderabbitai review

StarpTech:
> I think this solution will not prevent scenarios with excessively large requests that could potentially cause the router to be oom killed in the worst case.

We have this already in place. @endigma is the order right?

endigma:
> > I think this solution will not prevent scenarios with excessively large requests that could potentially cause the router to be oom killed in the worst case.
> 
> We have this already in place. @endigma is the order right?

I fixed the order in the PR, it now goes `Limit(Decompress(r.Body))`, so the limited reader wraps the gzip reader instead of vice versa

StarpTech (approved):
LGTM

router/core/graph_server.go
```diff
@@ -237,12 +237,13 @@ func newGraphServer(ctx context.Context, r *Router, routerConfig *nodev1.RouterC
...
 	})))
 
-	// Request traffic shaping related middlewares
-	httpRouter.Use(rmiddleware.RequestSize(int64(s.routerTrafficConfig.MaxRequestBodyBytes)))
 	if s.routerTrafficConfig.DecompressionEnabled {
 		httpRouter.Use(rmiddleware.HandleCompression(s.logger))
 	}
 
+	// Request traffic shaping related middlewares, happens after decompression to prevent unbounded decompression attacks
+	httpRouter.Use(rmiddleware.RequestSize(int64(s.routerTrafficConfig.MaxRequestBodyBytes)))
```
ysmolski, line 245: I have to agree with @Noroth. This  middleware won't be able to check because previous middleware `HandleCompression` would blow because of gzip bomb.

And to test this fix, you might want to recreate a gzip bomb in the testing code. Your current test is not aggressive enough.

ysmolski: Jesse has explained to me how little I know about our middleware. It was good trap.

PR #2078, opened 2025-07-23
fix(router): check persistent query hashes to be well-formed

ysmolski:
The OperationKit.unmarshalOperation method checks if non-empty sha256 hash in GraphQLRequestExtensions.PersistedQuery.Sha256Hash is valid. It rejects hex strings that could not be sha256 hash and raises en error that it cannot parse the body.

Tests now deal only with valid hash values.


ysmolski:
@coderabbitai review

ysmolski:
It is ready for review. Please take a look.

router-tests/persisted_operations_test.go
```diff
@@ -16,18 +16,57 @@ import (
...
 
+func TestPersistedOperationInvalidHash(t *testing.T) {
+	t.Parallel()
+
+	malformed := []string{
+		"malformed",
+		"some-words-free-style-not-allowed",
+		"../../../path/not/ok",
+		// too short
+		"000000000000000000000000000000000000000000000000000000000000000",
```
StarpTech: Should we simplify this? For better readability.

```go
strings.Repeat("0", 42)
```

router/core/operation_processor.go
```diff
@@ -192,6 +192,21 @@ type GraphQLRequestExtensionsPersistedQuery struct {
 	Sha256Hash string `json:"sha256Hash"`
 }
 
+// isWellFormed verifies if Sha256Hash string is valid and well-formed.
+func (pq *GraphQLRequestExtensionsPersistedQuery) isWellFormed() bool {
```
StarpTech: I think we can simplify it. DecodeString validates it internally

```go
func (pq *GraphQLRequestExtensionsPersistedQuery) isWellFormed() bool {
	if len(pq.Sha256Hash) != 64 {
		return false
	}
	_, err := hex.DecodeString(pq.Sha256Hash)
	return err == nil
}
```

router-tests/persisted_operations_test.go
ysmolski: Thanks! Done.

router/core/operation_processor.go
ysmolski: `DecodeString` does an allocation and much more than we need. Are we okay to sacrifice performance here?

router/core/operation_processor.go
```diff
@@ -308,6 +327,13 @@ func (o *OperationKit) unmarshalOperation() error {
 			}
 		}
 		if o.parsedOperation.GraphQLRequestExtensions.PersistedQuery != nil {
+			if !o.parsedOperation.GraphQLRequestExtensions.PersistedQuery.isValidHash() {
+				return &httpGraphqlError{
+					message:    "persistedQuery.sha256Hash must be a valid sha256 hash",
```
jensneuse: I would expect this error message to surface somewhere in a test. Currently it looks like it's swallowed or not yet tested.

jensneuse (changes requested):

router/core/operation_processor.go
StarpTech: I don't think it is impactful. My thoughts here are using std library. Internally, Decode looks very optimized with proper error handling. It's just a suggestion. I'mm ok with both approaches.

router/core/operation_processor.go
StarpTech: `persistedQuery.` is internal. Not sure, how meaningful this error message is to the user. Maybe a more fluent error description would be better here.

ysmolski: Messages from all the errors returned from the `unmarchalOperation` are rewritten by this calling block:

```
	} else if req.Method == http.MethodPost {
		if err := operationKit.UnmarshalOperationFromBody(httpOperation.body); err != nil {
			return &httpGraphqlError{
				message:    "error parsing request body",
				statusCode: http.StatusBadRequest,
			}
		}
```
`UnmarshalOperationFromBody` returns errors from `unmarchalOperation`
It means that those internal errors act like a kind of documentation right now.

ysmolski: That is the reason why we see only `"error parsing request body"` errors in the tests.

ysmolski: Please see https://github.com/wundergraph/cosmo/pull/2078#discussion_r2235644495 . All errors in the `unmarshalOperation` are rewritten by the callee. I am not brave enough to expose them all and break so many tests. :-)

ysmolski: Anyway, I have changed the error message to be more human friendly: "persisted query does not have a valid sha256 hash"

ysmolski: I had a chat with Jens. I am going to look into the wider problem of errors being rewritten in mentioned methods.

router/core/operation_processor.go
```diff
@@ -192,6 +192,25 @@ type GraphQLRequestExtensionsPersistedQuery struct {
 	Sha256Hash string `json:"sha256Hash"`
 }
 
+// isValidHash verifies if the Sha256Hash string is valid and well-formed.
+func (pq *GraphQLRequestExtensionsPersistedQuery) isValidHash() bool {
+	if len(pq.Sha256Hash) != 64 {
+		return false
+	}
+	// Iterating over bytes of the string
+	for i := 0; i < len(pq.Sha256Hash); i++ {
```
devsergiy: maybe we could use `!slices.Contains`?

router/core/operation_processor.go
ysmolski: @jensneuse 

I have changed how we handle errors returned by `operationKit.UnmarshalOperation*` methods. 
Now the client will see the error returned by the "UnmarshalOperationFromBody". 
I added couple of tested for "UnmarshalOperationFromURL".

Please review.

ysmolski: @StarpTech  I have fixed the errors handling, please check it out.

router/core/operation_processor.go
ysmolski: It is not in the hot-path, so I have swapped it with DecodeString.

StarpTech (approved):
LGTM

router-tests/batch_test.go
```diff
@@ -78,7 +78,7 @@ func TestBatch(t *testing.T) {
 				}, nil)
 				require.NoError(t, err)
 				require.Equal(t, http.StatusBadRequest, res.Response.StatusCode)
-				require.JSONEq(t, `{"errors":[{"message":"error parsing request body"}]}`, res.Body)
+				require.JSONEq(t, `{"errors":[{"message":"invalid request body: json: cannot unmarshal array into Go value of type core.GraphQLRequest"}]}`, res.Body)
```
jensneuse, line 81: can we exclude details like "Go value of type core.GraphQLRequest" and return something more generic? In this case, it seems that we expected an object but got an array. But we could also be more generic and just return that the JSON provided is not a well formed GraphQL Request.

router-tests/graphql_over_get_test.go
```diff
@@ -46,6 +46,35 @@ func TestOperationsOverGET(t *testing.T) {
...
+		testenv.Run(t, &testenv.Config{}, func(t *testing.T, xEnv *testenv.Environment) {
+			res, err := xEnv.MakeGraphQLRequestOverGET(testenv.GraphQLRequest{
+				OperationName: []byte(`Find`),
+				Query:         `query Find($criteria: SearchInput!) {findEmployees(criteria: $criteria){id details {forename surname}}}`,
+				Variables:     []byte(`{"criteria":{   "nationality":"GERMAN"}  }   `),
+				Extensions:    []byte(`true`),
+			})
+			require.NoError(t, err)
+			require.Equal(t, http.StatusBadRequest, res.Response.StatusCode)
+			require.Equal(t, `{"errors":[{"message":"invalid GET request: error parsing extensions: json: cannot unmarshal bool into Go value of type core.GraphQLRequestExtensions"}]}`, res.Body)
```
jensneuse: alternative error message:

```
invalid GET request: error parsing extensions: json: cannot unmarshal bool into extensions field of type Object.
```

I think that telling the user about "core.GraphQLRequestExtensions" will not be helpful.

router-tests/graphql_over_http_test.go
```diff
@@ -74,7 +74,7 @@ func TestGraphQLOverHTTPCompatibility(t *testing.T) {
 			require.Equal(t, res.Header.Get("Content-Type"), "application/json; charset=utf-8")
 			data, err := io.ReadAll(res.Body)
 			require.NoError(t, err)
-			require.Equal(t, `{"errors":[{"message":"error parsing request body"}]}`, string(data))
+			require.Equal(t, `{"errors":[{"message":"invalid request body: error parsing extensions: json: cannot unmarshal bool into Go value of type core.GraphQLRequestExtensions"}]}`, string(data))
```
jensneuse: same

jensneuse (changes requested):

router-tests/batch_test.go
ysmolski: I did some trimming on error messages exluding parts that are not relevant to a user.

router-tests/graphql_over_get_test.go
ysmolski: I settled on the middle ground: `invalid GET request: error parsing extensions: json: cannot unmarshal bool`

router-tests/batch_test.go
ysmolski: Also I had to remove the "sanity" parsing check for Extensions, because it is not needed as for variables. See comments in the code.

router-tests/graphql_over_http_test.go
ysmolski: Done

PR #2087, opened 2025-07-28
feat: batch and stream hooks

alepane21:
## Summary by CodeRabbit

* **New Features**
  * Added support for custom event hooks during stream publishing and batch processing, enabling advanced filtering and modification of events in real time.
  * Introduced new modules and utilities for enhanced Kafka and GraphQL subscription testing, including configurable batch and publish hooks.

* **Refactor**
  * Unified and extended hook management for pub/sub providers, allowing more flexible and centralized event handling.
  * Improved configuration and provider management for Kafka, NATS, and Redis integrations, simplifying adapter usage and setup.
  * Changed internal provider collections from slices to maps for efficient access and management.

* **Bug Fixes**
  * Enhanced error handling and type safety for event publishing and subscription flows across all supported providers.
  * Added error propagation and subscription closure handling in event update flows.
  * Improved logging and error messages for pub/sub operations.

* **Tests**
  * Expanded and updated test coverage for batch and publish hooks, and improved test utilities for Kafka, NATS, and Redis.
  * Replaced local Kafka and Redis test helpers with centralized utility functions for consistency.
  * Updated mocks to align with new interfaces and method signatures, removing deprecated mocks.

* **Chores**
  * Removed unused mock implementations and streamlined mock usage across the codebase.
  * Centralized Kafka and Redis event helper functions into a shared package for reuse.

## How to use the new hook?

The hooks implemented in this PR are StreamBatchEventHook and StreamPublishEventHook. 
StreamBatchEventHook is defined here:
https://github.com/wundergraph/cosmo/blob/3d829f871e7c0c0830bd46478b2b38f69407e32e/router/core/subscriptions_modules.go#L188

StreamPublishEventHook is defined here:
https://github.com/wundergraph/cosmo/blob/3d829f871e7c0c0830bd46478b2b38f69407e32e/router/core/subscriptions_modules.go#L201

#### Subscription
On the demo graph, we will create a module that filters out messages based on the kafka header
```graphql
employeeUpdatedMyKafka(employeeID: $employeeID)
```

#### Module code that implement the new hook

```go
const myModuleID = "streamBatchModule"

type StreamBatchModule struct {}

func (m *StreamBatchModule) OnStreamEvents(ctx core.StreamBatchEventHookContext, events []datasource.StreamEvent) ([]datasource.StreamEvent, error) {
        if ctx.SubscriptionEventConfiguration().RootFieldName() != "employeeUpdatedMyKafka" {
	        return events, nil
        }
        newEvents := []datasource.StreamEvent{}
        for _, event := range events {
	        evt, ok := event.(*kafka.Event)
	        if !ok {
		        continue
	        }
	        if string(evt.Headers["destination"]) != "external-clients" {
		        continue
	        }
	        newEvents = append(newEvents, event)
        }
        
        return newEvents, nil
}

func (m *StreamBatchModule) Module() core.ModuleInfo {
	return core.ModuleInfo{
		// This is the ID of your module, it must be unique
		ID: myModuleID,
		// The priority of your module, lower numbers are executed first
		Priority: 1,
		New: func() core.Module {
			return &StreamBatchModule{}
		},
	}
}
```

#### Mutation
On the demo graph, we will create a module that will add the $employeeId value in a kafka header.
```graphql
updateEmployeeMyKafka(employeeID: $employeeId, update: $update)
```

#### Module code that implement the new hook

```go

const myModuleID = "publishModule"

type PublishModule struct {}

func (m *PublishModule) OnPublishEvents(ctx core.StreamPublishEventHookContext, events []datasource.StreamEvent) ([]datasource.StreamEvent, error) {
        if ctx.PublishEventConfiguration().RootFieldName() != "updateEmployeeMyKafka" {
	        return events, nil
        }
        
        employeeID := ctx.RequestContext().Operation().Variables().GetInt("employeeID")
        
        newEvents := []datasource.StreamEvent{}
        for _, event := range events {
	        evt, ok := event.(*kafka.Event)
	        if !ok {
		        continue
	        }
	        if evt.Headers == nil {
		        evt.Headers = map[string][]byte{}
	        }
	        evt.Headers["x-employee-id"] = []byte(strconv.Itoa(employeeID))
	        newEvents = append(newEvents, event)
        }
        return newEvents, nil
}

func (m *PublishModule) Module() core.ModuleInfo {
	return core.ModuleInfo{
		// This is the ID of your module, it must be unique
		ID: myModuleID,
		// The priority of your module, lower numbers are executed first
		Priority: 1,
		New: func() core.Module {
			return &PublishModule{}
		},
	}
}
```

## Checklist


alepane21:
@coderabbitai full review

alepane21:
@coderabbitai full review

router/core/subscriptions_modules.go
```diff
@@ -106,11 +135,16 @@ func (e *EngineEvent) GetData() []byte {
 	return e.Data
 }
 
+func (e *EngineEvent) Clone() datasource.StreamEvent {
+	e2 := *e
```
dkorittki: I suppose you want to do a full memory copy of e right? This would only do a shallow copy, so only the struct and the structs Data slice metadata of e.Data would be copied but the actual byte locations remains the same.

https://go.dev/play/p/Ikwk_549ZkZ

Could use `bytes.Clone()` instead.

Found other places having a `Clone` method in the same style as well but did not make a seperate issue of it.

router/core/subscriptions_modules.go
```diff
@@ -176,13 +210,121 @@ func NewEngineSubscriptionOnStartHook(fn func(ctx SubscriptionOnStartHookContext
...
 			authentication: requestContext.Authentication(),
 			writeEventHook: resolveCtx.Updater,
 		}
 
 		return fn(hookCtx)
 	}
 }
+
+type StreamReceiveEventHookContext interface {
+	// Request is the original request received by the router.
```
dkorittki: Maybe add some more doc to tell that this request is the subscription start request from users?

router/pkg/pubsub/datasource/datasource.go
```diff
@@ -9,7 +9,7 @@ type SubscriptionDataSource interface {
 	SubscriptionEventConfiguration(input []byte) (SubscriptionEventConfiguration, error)
 	Start(ctx *resolve.Context, input []byte, updater resolve.SubscriptionUpdater) error
 	UniqueRequestID(ctx *resolve.Context, input []byte, xxh *xxhash.Digest) (err error)
-	SetSubscriptionOnStartFns(fns ...SubscriptionOnStartFn)
+	SetHooks(hoks Hooks)
```
dkorittki: Typo on `hoks`

router/pkg/pubsub/datasource/factory.go
```diff
@@ -9,16 +9,18 @@ import (
...
-	Event                  E
-	SubscriptionOnStartFns []SubscriptionOnStartFn
+	Providers       map[string]Provider
+	ProviderBuilder PB
+	Event           E
+	Hooks           Hooks
 }
 
-func NewPlannerConfig[PB ProviderBuilder[P, E], P any, E any](providerBuilder PB, event E, subscriptionOnStartFns []SubscriptionOnStartFn) *PlannerConfig[PB, P, E] {
+func NewPlannerConfig[PB ProviderBuilder[P, E], P any, E any](providerBuilder PB, event E, providers map[string]Provider, hooks Hooks) *PlannerConfig[PB, P, E] {
```
dkorittki, line 18: More a question than anything else: Why have there been no `providers` before this change? Shouldn't this be something already present in EDFS?

router/pkg/pubsub/datasource/subscription_event_updater.go
```diff
@@ -1,34 +1,115 @@
...
+		processedEvents, err := applyStreamEventHooks(
+			ctx,
+			s.subscriptionEventConfiguration,
+			events,
+			s.hooks.OnReceiveEvents,
+		)
+		// updates the events even if the hooks fail
+		// if a hook doesn't want to send the events, it should return no events!
+		for _, event := range processedEvents {
+			s.eventUpdater.UpdateSubscription(subId, event.GetData())
```
dkorittki, line 45: Are you sure this works? I followed the code back to see what happens inside the engine. Whhen this method here is called, in the engine an `subscriptionEventKindUpdateSubscription` event is thrown but there it seems that still every subscription of the trigger is updated, instead of a single one.

router/pkg/pubsub/kafka/adapter.go
```diff
@@ -155,52 +160,71 @@ func (p *ProviderAdapter) Subscribe(ctx context.Context, conf datasource.Subscri
...
+// Publish publishes the given events to the Kafka topic in a non-blocking way.
 // Publish errors are logged and returned as a pubsub error.
-// The event is written with a dedicated write client.
-func (p *ProviderAdapter) Publish(ctx context.Context, event PublishEventConfiguration) error {
+// The events are written with a dedicated write client.
+func (p *ProviderAdapter) Publish(ctx context.Context, conf datasource.PublishEventConfiguration, events []datasource.StreamEvent) error {
+	pubConf, ok := conf.(*PublishEventConfiguration)
+	if !ok {
+		return datasource.NewError("invalid event type for Kafka adapter", nil)
+	}
```
dkorittki, line 190: I wonder if we can remove the interface `datasource.PublishEventConfiguration` and use the type `PublishEventConfiguration` directly for all providers? Basically one type for all providers, instead of an interface, whichh needs to be type asserted in the providers.

router/pkg/pubsub/kafka/engine_datasource.go
```diff
@@ -43,13 +53,47 @@ func (s *SubscriptionEventConfiguration) RootFieldName() string {
...
+	if headers == nil {
+		headers = make(map[string][]byte)
+	}
+
+	headersBytes, err := json.Marshal(headers)
+	if err != nil {
+		return "", err
+	}
+
+	return fmt.Sprintf(`{"topic":"%s", "event": {"data": %s, "key": "%s", "headers": %s}, "providerId":"%s", "rootFieldName":"%s"}`, p.Topic, p.Event.Data, p.Event.Key, headersBytes, p.Provider, p.FieldName), nil
```
dkorittki, line 92: Can you go into detail why we cannot use a json marshaller for this? This looks to me like injections could be possible

router/pkg/pubsub/nats/adapter.go
```diff
@@ -206,45 +203,83 @@ func (p *ProviderAdapter) Subscribe(ctx context.Context, conf datasource.Subscri
 	return nil
 }
 
-func (p *ProviderAdapter) Publish(_ context.Context, event PublishAndRequestEventConfiguration) error {
+func (p *ProviderAdapter) Publish(_ context.Context, conf datasource.PublishEventConfiguration, events []datasource.StreamEvent) error {
```
dkorittki: I have a hard time understanding what exactly `datasource.PublishEventConfiguration` represents. Is it a temporary config for a group of events? Before this change the topic name was part of the event type, which feels more natural to me, as its a part of the event. What was the reason to invent a publish event configuration?

router/pkg/pubsub/nats/adapter.go
```diff
@@ -206,45 +203,83 @@ func (p *ProviderAdapter) Subscribe(ctx context.Context, conf datasource.Subscri
...
 	)
 
 	if p.client == nil {
 		return datasource.NewError("nats client not initialized", nil)
 	}
 
-	log.Debug("publish", zap.ByteString("data", event.Event.Data))
+	if len(events) == 0 {
+		return nil
+	}
```
dkorittki: I don't think an early return is needed here. Besides the log message below nothing is done in the function. The for loop down below is simply skipped and nil is returned. Should not matter performance wise.

router/pkg/pubsub/nats/adapter.go
```diff
@@ -206,45 +203,83 @@ func (p *ProviderAdapter) Subscribe(ctx context.Context, conf datasource.Subscri
...
+		if err != nil {
+			log.Error(
+				"publish error",
+				zap.Error(err),
+				zap.String("provider_id", pubConf.ProviderID()),
+				zap.String("provider_type", string(pubConf.ProviderType())),
+				zap.String("field_name", pubConf.RootFieldName()),
+			)
+			return datasource.NewError(fmt.Sprintf("error publishing to NATS subject %s", pubConf.Subject), err)
+		}
```
dkorittki, line 269: This current approach allows for partial failures. If somewhere inbetween a message can't be send, then some messages made it to nats and some don't. Not sure if it was like this in EDFS as well but as it looks right now NATS does not support batches anyway. Though there is something coming: https://github.com/nats-io/nats-architecture-and-design/pull/346 Might make sense to have a discussion about wether we want to support this when its ready?

router/pkg/pubsub/redis/adapter.go
```diff
@@ -13,15 +13,18 @@ import (
...
-	// Publish publishes the given event to the specified channel
-	Publish(ctx context.Context, event PublishEventConfiguration) error
+	Subscribe(ctx context.Context, conf datasource.SubscriptionEventConfiguration, updater datasource.SubscriptionEventUpdater) error
+	// Publish publishes the given events to the specified channel
+	Publish(ctx context.Context, conf datasource.PublishEventConfiguration, events []datasource.StreamEvent) error
 	// Startup initializes the adapter
 	Startup(ctx context.Context) error
 	// Shutdown gracefully shuts down the adapter
 	Shutdown(ctx context.Context) error
 }
```
dkorittki: If I compare this to other pubsub adapters this interface is missing the type embedding of `datasource.Adapter`

dkorittki (commented):
I have reviewed all non-test code first because of the size of this PR. I think it makes sense to discuss this first, then tests might have to be adjusted second.

I have a general question regarding the current method of pushing a message batch to brokers. In this approach we abort on first problems, with parts of messages being sent to the broker while others not. Did this change in the way EDFS did it before? Generally I would think batch events should be sent transactionally but I am lacking the historical context here. Just wanted to bring this topic up for discussion.

router/pkg/pubsub/datasource/subscription_event_updater.go
dkorittki: My bad, cant read

router/pkg/pubsub/nats/adapter.go
alepane21: The adapter interface now supports batching.
So I think that having a single PublishEventConfiguration with inside more than one events would have been weird, so I decided to separate the publish configurations and the events data.
Let me know if you need a better explanation for this, thanks.

router/pkg/pubsub/kafka/engine_datasource.go
alepane21: As we discussed, and as is reported at the start of the method, this is needed because the JSON could be invalid at this point when using templating to calculate the topic name. Injection should be avoided by the code valuing this struct.

router/pkg/pubsub/kafka/adapter.go
alepane21: PublishEventConfiguration has different specific fields for each adapter, so we can't generalize it. We could search for some common names between the actual adapters, but future adapters could have different fields needed.

dkorittki:
I believe that in the ADR the StreamHookError type is not whats in the code:

In ADR
```
// StreamHookError is used to customize the error messages and the behavior
type StreamHookError struct {
    HttpError core.HttpError
    CloseSubscription bool
}
```

In Code
```
type StreamHookError struct {
	err        error
	message    string
	statusCode int
	code       string
}
```
router/core/subscriptions_modules.go

dkorittki:
CI tests are failing because Bitnami moved their docker images. It's fixed upstream. Can you merge the topic branch into this?

```
Error response from daemon: manifest for bitnami/kafka:3.7.0 not found: manifest unknown: manifest unknown
  Warning: Docker pull failed with exit code 1, back off 4.098 seconds before retry.
  /usr/bin/docker pull bitnami/kafka:3.7.0
  Error response from daemon: manifest for bitnami/kafka:3.7.0 not found: manifest unknown: manifest unknown
  Warning: Docker pull failed with exit code 1, back off 4.072 seconds before retry.
  /usr/bin/docker pull bitnami/kafka:3.7.0
  Error response from daemon: manifest for bitnami/kafka:3.7.0 not found: manifest unknown: manifest unknown
```

router-tests/events/utils.go
```diff
@@ -2,55 +2,72 @@ package events
...
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
 	defer cancel()
 
 	pErrCh := make(chan error)
 
 	xEnv.KafkaClient.Produce(ctx, &kgo.Record{
 		Topic: xEnv.GetPubSubName(topicName),
 		Value: []byte(message),
-	}, func(_ *kgo.Record, err error) {
+	}, func(record *kgo.Record, err error) {
```
dkorittki: `record` is not used here, can stay as `_`

router-tests/modules/stream_publish_test.go
```diff
@@ -0,0 +1,314 @@
...
+
+		cfg := config.Config{
+			Graph: config.Graph{},
+			Modules: map[string]interface{}{
+				"publishModule": stream_publish.PublishModule{
+					Callback: func(ctx core.StreamPublishEventHandlerContext, events []datasource.StreamEvent) ([]datasource.StreamEvent, error) {
+						for _, event := range events {
+							evt, ok := event.(*kafka.Event)
+							if !ok {
+								continue
```
dkorittki, line 69: Maybe it makes sense to let the test fail here? We are not expecting anythinig else than kafka events here and if we receive something else it means there is a flaw in code, which we could detect here

router-tests/modules/stream_publish_test.go
```diff
@@ -0,0 +1,314 @@
...
+			LogObservation: testenv.LogObservationConfig{
+				Enabled:  true,
+				LogLevel: zapcore.InfoLevel,
+			},
+		}, func(t *testing.T, xEnv *testenv.Environment) {
+			events.KafkaEnsureTopicExists(t, xEnv, time.Second, "employeeUpdated")
+			resOne := xEnv.MakeGraphQLRequestOK(testenv.GraphQLRequest{
+				Query: `mutation { updateEmployeeMyKafka(employeeID: 3, update: {name: "name test"}) { success } }`,
+			})
+			require.JSONEq(t, `{"data": {"updateEmployeeMyKafka": {"success": false}}}`, resOne.Body)
```
dkorittki, line 140: Maybe also check the http status code returned? When lookinig at line 118 you set a 500 error. This should be the one returned to the mutation caller, right? Maybe you can set a more unusual error code like 507 or something so we can be sure that the error code returned is not a hardcoded 500 error.

router-tests/modules/stream_publish_test.go
```diff
@@ -0,0 +1,314 @@
...
+			})
+			require.JSONEq(t, `{"data": {"updateEmployeeMyRedis": {"success": false}}}`, resOne.Body)
+
+			requestLog := xEnv.Observer().FilterMessage("Publish Hook has been run")
+			assert.Len(t, requestLog.All(), 1)
+
+			requestLog2 := xEnv.Observer().FilterMessage("error applying publish event hooks")
+			assert.Len(t, requestLog2.All(), 1)
+
+			require.Len(t, records, 1)
```
dkorittki, line 249: Also check the response body here? Like you did on the nats test

router-tests/modules/stream_publish_test.go
```diff
@@ -0,0 +1,314 @@
...
+							return events, nil
+						}
+
+						employeeID := ctx.Operation().Variables().GetInt("employeeID")
+
+						newEvents := []datasource.StreamEvent{}
+						for _, event := range events {
+							evt, ok := event.(*kafka.Event)
+							if !ok {
+								continue
```
dkorittki, line 271: Maybe throw an error here since this is not expected to happen

router-tests/modules/stream_receive_test.go
```diff
@@ -0,0 +1,533 @@
...
+		errValue  error
+	}
+
+	t.Run("Test Receive hook is called", func(t *testing.T) {
+		t.Parallel()
+
+		cfg := config.Config{
+			Graph: config.Graph{},
+			Modules: map[string]interface{}{
+				"streamReceiveModule": stream_receive.StreamReceiveModule{},
```
dkorittki, line 42: Currently this hook does nothing but I think its a good idea to do "something", so we can see the hook is actually being called. Maybe log something we can fetch.

router-tests/modules/stream_publish_test.go
dkorittki, line 69: Maybe fail the test here, since it is not expected that this happens

router-tests/modules/stream_publish_test.go
dkorittki, line 271: Maybe fail the test here, since it is not expected that this happens

router-tests/modules/stream_receive_test.go
```diff
@@ -0,0 +1,533 @@
...
+
+		cfg := config.Config{
+			Graph: config.Graph{},
+			Modules: map[string]interface{}{
+				"streamReceiveModule": stream_receive.StreamReceiveModule{
+					Callback: func(ctx core.StreamReceiveEventHandlerContext, events []datasource.StreamEvent) ([]datasource.StreamEvent, error) {
+						for _, event := range events {
+							evt, ok := event.(*kafka.Event)
+							if !ok {
+								continue
```
dkorittki, line 120: Maybe fail the test here, since it is not expected that this happens

router/pkg/pubsub/datasource/pubsubprovider_test.go
```diff
@@ -7,8 +7,59 @@ import (
...
+type testEvent struct {
+	data []byte
+}
+
+func (e *testEvent) GetData() []byte {
+	return e.data
+}
+
+func (e *testEvent) Clone() StreamEvent {
+	e2 := *e
```
dkorittki: Do a deep copy instead of a shallow one

router/pkg/pubsub/datasource/subscription_event_updater_test.go
```diff
@@ -0,0 +1,669 @@
...
+	}
+
+	updater.Update(events)
+
+	// Assert that Update and UpdateSubscription were not called on the eventUpdater
+	mockUpdater.AssertNotCalled(t, "Update")
+	mockUpdater.AssertNotCalled(t, "UpdateSubscription")
+}
+
+func TestSubscriptionEventUpdater_Update_WithMultipleHooks_Success(t *testing.T) {
```
dkorittki, line 152: I would redesign this test a bit. Instead of creating two individual channels you could create only one and let both hooks insert into it. On the bottom of the test you could then verify that you received both events and also make sure the order of events arrived is correct (even from first hook is first, other one is second). Also  I think I would do more like five events instead of two. This way we ensure correct ordering in arrived events is not coincidence.

Further down I have seen some more tests using hooks having each their own channel where you could also use a single channel for all of them and test ordering as well. I will not highlight them to not clutter this review unnecessarily.

router/pkg/pubsub/kafka/engine_datasource.go
```diff
@@ -22,6 +26,18 @@ func (e *Event) GetData() []byte {
...
 }
 
+func (e *Event) Clone() datasource.StreamEvent {
+	e2 := *e
+	e2.Data = slices.Clone(e.Data)
+	e2.Headers = make(map[string][]byte, len(e.Headers))
+	for k, v := range e.Headers {
+		e2.Headers[k] = slices.Clone(v)
+	}
+	return &e2
```
dkorittki, line 36:
```suggestion
	e2 := *e
	e2.Data = bytes.Clone(e.Data)
	e2.Headers = make(map[string][]byte, len(e.Headers))
	for k, v := range e.Headers {
		e2.Headers[k] = bytes.Clone(v)
	}
	return &e2
```

Found that for the other two implementations as well. It probably works with slices.Clone() as well but its not intended for byte slices. Thats what bytes.Clone() is for.

router/pkg/pubsub/redis/engine_datasource.go
```diff
@@ -20,6 +23,11 @@ func (e *Event) GetData() []byte {
 	return e.Data
 }
 
+func (e *Event) Clone() datasource.StreamEvent {
+	e2 := *e
```
dkorittki: Found that one here, seems untouched

router-tests/modules/stream_publish_test.go
alepane21: I think that is better not to do this. Checking inside a go routine is not a good practice, so we would have to send all the events to a channel and than read from there the events and check them in the main thread ... it would be too work for what is (I think, but happy to be convinced otherwise!) just a nitpick.

router-tests/modules/stream_publish_test.go
alepane21: Same as before, failing inside a go routine is bad practice so I would be against it if is not something that we really want to test!

alepane21: Same as before, failing inside a go routine is bad practice so I would be against it if is not something that we really want to test!

router-tests/modules/stream_receive_test.go
alepane21: Mmm not sure about this. If no custom behaviour is specified in the test, we are still emitting the log "Stream Hook has been run", in the base module (and is what we are testing in this test a bit below this line). Did you wanted to test that the custom code of the custom module is running? Isn't this already being checked in the tests that we have after this? We are only checking if the custom module is running...

router-tests/modules/stream_publish_test.go
alepane21: Same as before, failing inside a go routine is bad practice so I would be against it if is not something that we really want to test!

router-tests/modules/stream_receive_test.go
alepane21: Same as before, failing inside a go routine is bad practice so I would be against it if is not something that we really want to test!

router/pkg/pubsub/kafka/engine_datasource.go
dkorittki: After discussing this together: bytes.Clone and slices.Clone do the same thing under the hood even though at some point it was different. So we opted to go for the generic solution and use the slices package.

router/pkg/pubsub/datasource/subscription_event_updater_test.go
dkorittki: After discussing this we came to the conclusion that its fine as is, no need to change anything.

router-tests/modules/stream_receive_test.go
dkorittki: Yeah I got this wrong. I thought this is the actual body of the handler but in fact its only the handler type which is being registered here.

PR #2090, opened 2025-07-29
fix: trim query operation name length [WIP]

SkArchon:
@coderabbitai summary

## Checklist


router/core/operation_planner.go
```diff
@@ -53,6 +53,20 @@ func (p *OperationPlanner) preparePlan(ctx *operationContext) (*planWithMetaData
 		return nil, &reportError{report: &report}
 	}
```
devsergiy: why do we need to do it here? could we do it during parse?

router/core/operation_processor.go
```diff
@@ -278,6 +280,14 @@ func (o *OperationKit) UnmarshalOperationFromURL(url *url.URL) error {
 	return o.unmarshalOperation()
 }
 
+func GetOperationNameTrimmed(operationName string, limit int) (bool, string) {
+	if len(operationName) <= limit {
+		return false, operationName
+	}
+	operationName = operationName[:limit]
+	return true, operationName
+}
```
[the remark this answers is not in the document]
SkArchon: Added the limit fix, we dont need to worry about byte vs string as we dont allow chars with more than 1 code point for graphql operation names.

router/pkg/config/config.schema.json
```diff
@@ -2457,6 +2457,11 @@
             }
           }
         },
+        "operation_name_trim_limit": {
+          "type": "integer",
+          "description": "The limit of the operation name length",
+          "default": "256"
+        },
```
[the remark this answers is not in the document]
SkArchon: Its not expected to be inside complexity_limits.

router/core/operation_planner.go
SkArchon: This creates another document by parsing `ctx.content` which contains the normalised document as a raw string with the original untrimmed name.

In the `planner.selectOperation` function it cross checks the parsed ctx.context doc with the trimmed operation name, which results in a mismatch.

SkArchon: Sidenote, I was thinking if it wouls make sense to pass in a function to the engine which can be called when parsing the operation name, so that it will ignore the extra bytes?

router/core/operation_processor.go
```diff
@@ -278,6 +280,14 @@ func (o *OperationKit) UnmarshalOperationFromURL(url *url.URL) error {
 	return o.unmarshalOperation()
 }
 
+func GetOperationNameTrimmed(operationName string, limit int) (bool, string) {
+	if limit <= 0 || len(operationName) <= limit {
+		return false, operationName
+	}
+	operationName = operationName[:limit]
+	return true, operationName
+}
```
endigma: Nit: Can we do something like take a hash of the full operation name and append it to the truncated one? This way we don't potentially cut off the differentiator and make two operations have the same name?

router/core/operation_planner.go
```diff
@@ -53,6 +53,20 @@ func (p *OperationPlanner) preparePlan(ctx *operationContext) (*planWithMetaData
...
+	for i := range doc.RootNodes {
+		if doc.RootNodes[i].Kind != ast.NodeKindOperationDefinition {
+			continue
+		}
+		currName := doc.OperationDefinitionNameString(doc.RootNodes[i].Ref)
+		if trimmed, _ := GetOperationNameTrimmed(currName, ctx.operationTrimLimit); trimmed {
+			name := doc.OperationDefinitions[doc.RootNodes[i].Ref].Name
+			doc.OperationDefinitions[doc.RootNodes[i].Ref].Name.End = name.Start + uint32(ctx.operationTrimLimit)
+		}
+	}
```
endigma: Does this only apply to Root nodes? Could this happen with a long alias, directive, variable maybe? Isn't anything ugc that ends up in metrics vulnerable?

router/pkg/config/config.go
```diff
@@ -408,6 +408,7 @@ type SecurityConfiguration struct {
 	ComplexityLimits            *ComplexityLimits           `yaml:"complexity_limits"`
 	DepthLimit                  *QueryDepthConfiguration    `yaml:"depth_limit"`
 	ParserLimits                ParserLimitsConfiguration   `yaml:"parser_limits"`
+	OperationNameTrimLimit      int                         `yaml:"operation_name_trim_limit" envDefault:"256"` // 0 is disabled
```
endigma: Please add to `full.yaml`

router/pkg/config/config.schema.json
```diff
@@ -2457,6 +2457,12 @@
             }
           }
         },
+        "operation_name_trim_limit": {
+          "type": "integer",
+          "description": "The limit of the operation name length",
```
endigma: Document 0==disabled in the docstring

router-tests/security_test.go
```diff
@@ -109,3 +113,125 @@ func TestParserHardLimits(t *testing.T) {
...
+		}, func(t *testing.T, xEnv *testenv.Environment) {
+			xEnv.MakeGraphQLRequestOK(testenv.GraphQLRequest{
+				Query: "query " + queryName + " { employees { id } }",
+			})
+			sn := exporter.GetSpans().Snapshots()
+
+			require.Equal(t, "Operation - Execute", sn[7].Name())
+			operationName := getOperationName(sn[7].Attributes())
+			require.NotNil(t, operationName)
+			require.Equal(t, operationName.AsString(), queryName)
```
endigma: (expected, actual) is the format for testify parameters

router-tests/security_test.go
```diff
@@ -109,3 +113,125 @@ func TestParserHardLimits(t *testing.T) {
...
```
endigma: (expected, actual)

router/core/context.go
```diff
@@ -533,6 +533,8 @@ type operationContext struct {
 	validationTime    time.Duration
 	planningTime      time.Duration
 	normalizationTime time.Duration
+
+	operationTrimLimit int
```
endigma: `operationNameTrimLimit`? it's this most everywhere else

router/core/operation_planner.go
jensneuse: If you trim the name, it would mean that the name is now no longer referencing the name used in the GraphQL document (query part of the JSON). As such, we would now have a planning error. I think the right thing to do is not to trim, but to verify the length at parsing time. If the length exceeds the configured limit, we should stop the request pipeline and return a bad request.

router/core/operation_processor.go
```diff
@@ -278,6 +280,14 @@ func (o *OperationKit) UnmarshalOperationFromURL(url *url.URL) error {
 	return o.unmarshalOperation()
 }
 
+func GetOperationNameTrimmed(operationName string, limit int) (bool, string) {
```
endigma: Nit
```suggestion
func TrimOperationName(operationName string, limit int) (bool, string) {
```

endigma (changes requested):

router/core/operation_planner.go
StarpTech: The idea of trimming came from the requirement to not introduce breaking changes. This is unlikely and if there are complication, we can also return an error. The limit needs to be very reasonable like 256. I know that some tools generate long operation names.

SkArchon: The change I did makes it so that we match the documents name with the trimmed name for planning.

However the original query is still present, thus if reparsed would need to be manually modified, based on the context I have right now I think to make sure nothing is missed in the future when something is added, its best if this change lives in the engine (cc: @devsergiy), that way anytime a document is parsed we would set the `EndBytes` to the trimmed index for the operation name, if it exists.

StarpTech: It sounds complicated. Let's choose a high enough limit (512) and return an error. Following @jensneuse approach here.

router/core/operation_processor.go
SkArchon: Changed approach to blocking, this has been removed now

router/core/operation_planner.go
SkArchon: Changed approach to blocking, this has been removed now

router/core/operation_processor.go
SkArchon: Changed approach to blocking, this has been removed now

router/core/operation_planner.go
SkArchon: Changed to block.

router/core/operation_processor.go
```diff
@@ -527,6 +536,13 @@ func (o *OperationKit) isIntrospectionQuery() (result bool, err error) {
 	return false, nil
 }
 
+func (o *OperationKit) isOperationNameLimitExceeded(operationName string) bool {
+	if o.operationProcessor.operationNameLimit == 0 {
+		return false
+	}
+	return len(operationName) > o.operationProcessor.operationNameLimit
+}
+
```
endigma: Is operation name the only ugc field that can be long enough to cause issues? field aliases, directives maybe also affected?

router/core/operation_processor.go
```diff
@@ -487,6 +489,13 @@ func (o *OperationKit) isIntrospectionQuery() (result bool, err error) {
 			ref := possibleOperationDefinitionRefs[i]
 			name := o.kit.doc.OperationDefinitionNameString(ref)
 
+			if o.isOperationNameLimitExceeded(name) {
+				return false, &httpGraphqlError{
+					message:    "operation name too large",
```
endigma: I suggest including both the maximum and the length that was too long in the error, if possible

router/pkg/config/config.go
```diff
@@ -408,6 +408,7 @@ type SecurityConfiguration struct {
 	ComplexityLimits            *ComplexityLimits           `yaml:"complexity_limits"`
 	DepthLimit                  *QueryDepthConfiguration    `yaml:"depth_limit"`
 	ParserLimits                ParserLimitsConfiguration   `yaml:"parser_limits"`
+	OperationNameLimit          int                         `yaml:"operation_name_limit" envDefault:"512"` // 0 is disabled
```
endigma: Should be `OperationNameLengthLimit`, just "Limit" is very generic and could refer to many ways of counting

router/pkg/config/config.schema.json
```diff
@@ -2457,6 +2457,12 @@
             }
           }
         },
+        "operation_name_limit": {
+          "type": "integer",
+          "description": "The max allowed limit of the operation name length, 0 would disable the limit validation.",
```
endigma:
```suggestion
          "description": "The maximum allowed length of the operation name, 0 allows any length.",
```

router-tests/security_test.go
```diff
@@ -109,3 +110,221 @@ func TestParserHardLimits(t *testing.T) {
...
+				require.NoError(t, err)
+				require.JSONEq(t, `{"errors":[{"message":"operation name too large"}]}`, resGet.Body)
+				require.Equal(t, http.StatusBadRequest, resGet.Response.StatusCode)
+			})
+		})
+
+		t.Run("with small query name and small operation name", func(t *testing.T) {
+			t.Parallel()
+
+			trimSize := 7
```
endigma: trimSize -> lengthLimit or etc?

router/core/operation_processor.go
SkArchon: Will need to check, including if this affects aqps.

router/core/operation_processor.go
SkArchon: Will change to `fmt.Sprintf("operation name of length %d exceeds max length of %d", len(name), o.operationProcessor.operationNameLimit)`

router/core/operation_processor.go
SkArchon: Apq still caches the raw query, since we can only reject it post parsing. I did check aliases from the rest, and they are not restricted either, however not a perf test. Will open a ticket to check it after otel.

router-tests/security_test.go
```diff
@@ -109,3 +111,255 @@ func TestParserHardLimits(t *testing.T) {
...
+				resGet, err := xEnv.MakeGraphQLRequestOverGET(testenv.GraphQLRequest{
+					Query: "query " + query1Name + " { employees { id } } query " + query2Name + " { employees { id } }",
+				})
+				require.NoError(t, err)
+				require.JSONEq(t, `{"data":{"employees":[{"id":1},{"id":2},{"id":3},{"id":4},{"id":5},{"id":7},{"id":8},{"id":10},{"id":11},{"id":12}]}}`, resGet.Body)
+				require.Equal(t, http.StatusOK, resGet.Response.StatusCode)
+			})
+		})
+
+		t.Run("with large queries with max length of 0 where the validation is not ebaled", func(t *testing.T) {
```
endigma: typo

endigma (approved):
looks good

PR #2097, opened 2025-08-01
fix: generate plans for subscriptions

ysmolski:
Implement plans for subscriptions. Represent them the same way as Apollo [does](https://github.com/apollographql/router/blob/e56887bc76f4842ed34185d1ea85a6b308722bc6/apollo-federation/src/query_plan/display.rs#L52):

    QueryPlan {
      Subscription {
        Primary: {
          Fetch(service: "messages") {
            {
              messageInConversation(id: $id) {
                id
                text
                sentTime
              }
            }
          }
        },
        Rest: {
          Fetch(service: "employees") {
            {
              ...
            }
          }
        },
      }
    }

"Primary" key contains the subscription that is setup to receive messages from subgraph. 
"Rest" key contains the queries that willl be executed for each received message by the router.

Printing Query Plans was broken for PubSub subscriptions. 
From now on, for this type we return empty Fetch node for "Primary" key.

## Summary by CodeRabbit

* **New Features**
  * Added support for logging query plans of subscription operations, enhancing visibility during development and debugging.
  * Introduced new GraphQL subscription queries and corresponding query plans in test data, including real-time current time updates and employee update subscriptions.

* **Bug Fixes**
  * Improved formatting of subscription query strings in test fixtures for better readability.

* **Tests**
  * Refactored test structure for plan generator to improve maintainability and scalability.
  * Expanded test coverage with new queries and plans for subscription scenarios.

* **Chores**
  * Updated a dependency to a more precise version for improved reliability.

## Checklist


StarpTech:
Please make it a draft if the intent is not to merge it.

ysmolski:
Redoing representation to follow Apollo's, mentioned in the title message.

ysmolski:
> LGTM @ysmolski two things concerns me:
> 
>     * we are going to change the results of the query plans for all the query with subscription, and this means that if a customer is going to use the query plan to check for problems, we are going to trigger it; this is a change for the better, but we should comunicate that we are changing the query plans (maybe we can see this as a breaking change)
> 
>     * we should open an issue to fix the output also for EDFS sources, or if it is easy fix it now.

I don't think it is a breaking change because Query Planning was not changed. The only thing we changed is how QueryPlans are printed. Also before subscriptions were totally empty, and it was considered a bug. Thus, this change has fixed the bug and did not break any contract.

As for EDFS sources, it is not clear if it is possible to generate query plans for the trigger, so EDFS change won't be included in this PR. I want to keep that EDFS test as a placeholder for the future work (if any).

@devsergiy if you feel up to it and have time, please have a look. I have followed Apollo's representation after all.

router/pkg/plan_generator/testdata/plans/base/4.graphql
```diff
@@ -0,0 +1,26 @@
+QueryPlan {
+  Subscription {
+    Primary: {
+      Fetch(service: "2-nats-3") {
```
jensneuse, line 4: It might be a bit tricky, but doable, to know what fields were selected from the EDFS source. We could split this into another story and add the root selections that belong to the EDFS source, just to make it easier to understand what's going on.

devsergiy (approved):

router/pkg/plan_generator/testdata/plans/base/4.graphql
ysmolski: Created a separate story for this.

alepane21 (approved):

PR #2130, opened 2025-08-11
feat: option to redact query param variables in access logs

SkArchon:
## Motivation

Right now we print the entire raw query params, however in certain cases, especially when sending queries over GET, sensitive values can be logged. As we currently log the raw `query` in the access logs by default. This adds an option to ignore certain query parameters. In the case there are skip list entries and we cannot parse the query params, we will reset the query to "" for logging.

```
access_logs:
  router:
    ignore_query_params_list:
      - "variables"
```

@coderabbitai summary

## Checklist

### DIFF
--- a/router/internal/requestlogger/requestlogger.go
+++ b/router/internal/requestlogger/requestlogger.go
@@ -5,6 +5,7 @@ import (
 	"encoding/hex"
 	"net"
 	"net/http"
+	"net/url"
 	"time"
 
 	"github.com/wundergraph/cosmo/router/internal/errors"
@@ -109,6 +110,12 @@ func WithDefaultOptions() Option {
 	}
 }
 
+func WithIgnoreQueryParamsList(ignoreList []string) Option {
+	return func(r *handler) {
+		r.accessLogger.ignoreQueryParamsList = ignoreList
+	}
+}
+
 func New(logger *zap.Logger, opts ...Option) func(h http.Handler) http.Handler {
 	return func(h http.Handler) http.Handler {
 		r := &handler{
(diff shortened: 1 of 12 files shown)


router/internal/requestlogger/requestlogger.go
```diff
@@ -206,6 +213,23 @@ func (al *accessLogger) getRequestFields(r *http.Request) []zapcore.Field {
 		}
 	}
 
+	if query != "" && len(al.ignoreQueryParamsList) > 0 {
```
StarpTech, line 216: How about subgraphs logs?

router/pkg/config/config.schema.json
```diff
@@ -694,6 +694,13 @@
           "properties": {
             "fields": {
               "$ref": "#/$defs/context_fields"
+            },
+            "ignore_query_params_list": {
```
StarpTech, line 698: We should skip logging `variables` by default to prevent leaking informations

StarpTech (changes requested):

router/internal/requestlogger/requestlogger.go
SkArchon: While the code is common for subgraph access logger, it is not enabled. This is as I only saw an option to forward query params to the subgraph in case of an ws upgrade request, however this does not trigger the subgraph access logger.

router/pkg/config/config.schema.json
SkArchon: Will add

router/internal/requestlogger/requestlogger.go
StarpTech: ok

router/pkg/config/config.schema.json
SkArchon: resolved

StarpTech (approved):
LGTM

PR #2133, opened 2025-08-12
feat: always rewrite abstract fragments for gRPC

ysmolski:
Integrates a feature of the engine, that always rewrites inline fragments for gRPC datasource.
Integration test required adding new query field: "nodesById(id: ID!) [Node!]!".

## Summary by CodeRabbit

* **New Features**
  * GraphQL schema expanded with many new list-pattern fields and a new nodesById query; Projects service now exposes an endpoint returning heterogeneous Node results.

* **Chores**
  * Bumped a GraphQL dependency to a newer prerelease for more reproducible builds.
  * Added post-generation JSON formatting and a new demo build target to integrate formatted configs.

* **Tests**
  * Added a test validating nodesById with interfaces and inline fragments.


router-tests/testenv/testdata/configWithPlugins.json
```diff
@@ -1 +1 @@
...HBigObject","xFieldOnHBigObject","yFieldOnHBigObject","zFieldOnHBigObject"]},{"typeName":"IBigObject","fieldNames":["aFieldOnIBigObject","bFieldOnIBigObject","cFieldOnIBigObject","dFieldOnIBigObject","eFieldOnIBigObject","fFieldOnIBigObject","gFieldOnIBigObject","hFieldOnIBigObject","iFieldOnIBigObject","jFieldOnIBigObject","kFieldOnIBigObject","lFieldOnIBigObject","mFieldOnIBigObject","nFieldOnIBigObject","oFieldOnIBigObject","pFieldOnIBigObject","qFieldOnIBigObject","rFieldOnIBigObject","sFieldOnIBigObject","tFieldOnIBigObject","uFieldOnIBigObject","vFieldOnIBigObject","wFieldOnIBigObject","xFieldOnIBigObject","yFieldOnIBigObject","zFieldOnIBigObject"]},{"typeName":"JBigObject","fieldNames":["aFieldOnJBigObject","bFieldOnJBigObject","cFieldOnJBigObject","dFieldOnJBigObject","eFieldOnJBigObject","fFieldOnJBigObject","gFieldOnJBigObject","hFieldOnJBigObject","iFieldOnJBigObject","jFieldOnJBigObject","kFieldOnJBigObject","lFieldOnJBigObject","mFieldOnJBigObject","nFieldOnJBigObject","oFieldOnJBigObject","pFieldOnJBigObject","qFieldOnJBigObject","rFieldOnJBigObject","sFieldOnJBigObject","tFieldOnJBigObject","uFieldOnJBigObject","vFieldOnJBigObject","wFieldOnJBigObject","xFieldOnJBigO
```
Noroth, line 1: I think it makes sense to also apply prettier formatting to this file.

demo/pkg/subgraphs/projects/src/schema.graphql
```diff
@@ -26,6 +26,8 @@ type Query {
   # query to simulate that the service goes down
   killService: Boolean!
   panic: Boolean!
+
+  interfaceNamed: [Named!]!
```
Noroth: Wouldn't `Node` suffice here? Node is being used on all types and you could get any of the types.
Something like `nodesById(id: ID!): [Node!]!

ysmolski: That may be a good idea. Let me try it.

router-tests/testenv/testdata/configWithPlugins.json
ysmolski: Done

demo/pkg/subgraphs/projects/src/schema.graphql
ysmolski: I think your suggestion has worked! Please have a look.

demo/pkg/subgraphs/projects/generated/service.proto.lock.json
```diff
@@ -92,8 +92,13 @@
...
         "killService": 14,
-        "panic": 15
-      }
+        "panic": 15,
+        "nodesById": 18
+      },
+      "reservedNumbers": [
+        16,
+        17
+      ]
```
Noroth: Something looks wrong here. Why are the numbers reserved?

might also be another issue because Query, Mutation and Subscription should probably not be tracked in the lock file as we are only generating endpoint from it.

ysmolski: I have cleaned it from the failed attempt to add interface and weird query before.

PR #2155, opened 2025-08-20
feat(router): more intuitive subgraph traffic shaping config inheritance

endigma:
@coderabbitai summary

## Checklist

### DIFF
--- a/router/core/router.go
+++ b/router/core/router.go
@@ -1816,8 +1816,12 @@ func DefaultFileUploadConfig() *config.FileUpload {
 	}
 }
 
-func NewTransportRequestOptions(cfg config.GlobalSubgraphRequestRule) *TransportRequestOptions {
-	defaults := DefaultTransportRequestOptions()
+// NewTransportRequestOptions creates a new TransportRequestOptions instance with the given configuration and defaults.
+// If defaults is nil, it uses the global default values.
+func NewTransportRequestOptions(cfg config.GlobalSubgraphRequestRule, defaults *TransportRequestOptions) *TransportRequestOptions {
+	if defaults == nil {
+		defaults = DefaultTransportRequestOptions()
+	}
 
 	return &TransportRequestOptions{
 		RequestTimeout:         or(cfg.RequestTimeout, defaults.RequestTimeout),
@@ -1850,13 +1854,15 @@ func DefaultTransportRequestOptions() *TransportRequestOptions {
 }
 
 func NewSubgraphTransportOptions(cfg config.TrafficShapingRules) *SubgraphTransportOptions {
+	allRequestOptions := NewTransportRequestOptions(cfg.All, nil)
+
 	base := &SubgraphTransportOptions{
-		TransportRequestOptions: NewTransportRequestOptions(cfg.All),
+		TransportRequestOptions: allRequestOptions,
 		SubgraphMap:             map[string]*TransportRequestOptions{},
 	}
 
 	for k, v := range cfg.Subgraphs {
-		base.SubgraphMap[k] = NewTransportRequestOptions(*v)
+		base.SubgraphMap[k] = NewTransportRequestOptions(*v, allRequestOptions)
 	}
 
 	return base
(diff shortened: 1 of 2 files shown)


router/core/router.go
```diff
@@ -1850,13 +1854,15 @@ func DefaultTransportRequestOptions() *TransportRequestOptions {
...
 	base := &SubgraphTransportOptions{
-		TransportRequestOptions: NewTransportRequestOptions(cfg.All),
+		TransportRequestOptions: allRequestOptions,
 		SubgraphMap:             map[string]*TransportRequestOptions{},
 	}
 
 	for k, v := range cfg.Subgraphs {
-		base.SubgraphMap[k] = NewTransportRequestOptions(*v)
+		base.SubgraphMap[k] = NewTransportRequestOptions(*v, allRequestOptions)
 	}
```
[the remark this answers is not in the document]
endigma: This is prohibited by JSON schema

router/core/router_test.go
```diff
@@ -259,11 +261,20 @@ func TestTrafficShapingRules(t *testing.T) {
...
-	assert.Equal(t, subgraphDialTimeout, router.subgraphTransportOptions.SubgraphMap["some-subgraph"].DialTimeout)
-	assert.Equal(t, defaults.MaxIdleConns, router.subgraphTransportOptions.SubgraphMap["some-subgraph"].MaxIdleConns)
+	subgraphRequestOptions := router.subgraphTransportOptions.SubgraphMap["some-subgraph"]
+
+	// Subgraph specific configurations
+	assert.Equal(t, subgraphRequestTimeout, subgraphRequestOptions.RequestTimeout)
+	assert.Equal(t, subgraphDialTimeout, subgraphRequestOptions.DialTimeout)
+
+	// Inherit from `all`
+	assert.Equal(t, allMaxConnsPerHost, subgraphRequestOptions.MaxConnsPerHost)
```
StarpTech: Please verify all options

router/core/router.go
```diff
@@ -1850,13 +1854,15 @@ func DefaultTransportRequestOptions() *TransportRequestOptions {
 }
 
 func NewSubgraphTransportOptions(cfg config.TrafficShapingRules) *SubgraphTransportOptions {
+	allRequestOptions := NewTransportRequestOptions(cfg.All, nil)
```
SkArchon, line 1857: From how I understand this, we previously copied over the default value when the subgraph specific value exists, however now we apply the value from "all" (with the defaults applied to all first).

PR looks ok but I believe this could break backward compatiblity, if someone expects the default value to be applied, INSTEAD of the value from all.

Should we introduce a flag to enable this?

SkArchon (changes requested):
PR seems ok to me, however comment on backwards compatibility

router/core/router.go
endigma: I think the current behaviour is unintuitive enough that if we "break" this more users are likely to get expected results with the new version than unexpected ones.

I don't think a flag would improve things:
- An enabled-by-default flag would be work equivalent to just changing the config to have the value they want
- A disabled-by-default one would leave it "broken" for the vast majority of users

If a user wants the default value instead of the one from `all`, they should manually specify it.

StarpTech (approved):
LGTM

router/core/router.go
SkArchon: Don't disagree its unintuitive, but "this more users are likely to get expected results with the new version than unexpected ones." is only an assumption right?

If we have a flag it would have to be "disabled" by default ofcourse, also if we don't have a flag it might be worth informing customers explicitly? Either way just want to be sure we have an understanding of what impact we decide to introduce.

Might be worth getting @Noroth / @StarpTech opinion also.

SkArchon (approved):

router/core/router.go
StarpTech: Fixing this is more important than ensuring default value behaviour. The inherent values should already be reasonable, if not the same

PR #2158, opened 2025-08-20
feat: add span for each grpc invocation

SkArchon:
@coderabbitai summary

## Checklist

### DIFF
--- a/router-tests/router_plugin_test.go
+++ b/router-tests/router_plugin_test.go
@@ -2,6 +2,7 @@ package integration
 
 import (
 	"fmt"
+	"github.com/wundergraph/cosmo/router/pkg/otel"
 	"testing"
 	"time"
 
@@ -95,21 +96,23 @@ func TestRouterPlugin(t *testing.T) {
 }
 
 func TestVerifyTelemetryForRouterPluginRequests(t *testing.T) {
-	exporter := tracetest.NewInMemoryExporter(t)
+	t.Parallel()
 
-	testenv.Run(t,
-		&testenv.Config{
-			TraceExporter:            exporter,
-			RouterConfigJSONTemplate: testenv.ConfigWithPluginsJSONTemplate,
-			Plugins: testenv.PluginConfig{
-				Enabled: true,
-				Path:    "../router/plugins",
+	t.Run("query projects simple", func(t *testing.T) {
+		t.Parallel()
+
+		exporter := tracetest.NewInMemoryExporter(t)
+
+		testenv.Run(t,
+			&testenv.Config{
+				TraceExporter:            exporter,
+				RouterConfigJSONTemplate: testenv.ConfigWithPluginsJSONTemplate,
+				Plugins: testenv.PluginConfig{
+					Enabled: true,
+					Path:    "../router/plugins",
+				},
 			},
-		},
-		func(t *testing.T, xEnv *testenv.Environment) {
-			t.Run("query projects simple", func(t *testing.T) {
-				t.Parallel()
-
+			func(t *testing.T, xEnv *testenv.Environment) {
 				queryName := "query sample"
 				response := xEnv.MakeGraphQLRequestOK(testenv.GraphQLRequest{
 					Query: fmt.Sprintf(`%s { projects { id name } }`, queryName),
@@ -119,7 +122,7 @@ func TestVerifyTelemetryForRouterPluginRequests(t *testing.T) {
 				require.Equal(t, expected, response.Body)
 
 				snapshots := exporter.GetSpans().Snapshots()
-				require.Len(t, snapshots, 8)
+				require.Len(t, snapshots, 9)
 
 				queryNameInstances := 0
 				for _, sn := range snapshots {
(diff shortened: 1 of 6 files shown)


router/pkg/grpcconnector/grpcplugin/grpc_plugin.go
```diff
@@ -97,7 +107,10 @@ func (p *GRPCPlugin) fork() error {
 	}
 
 	pluginCmd := exec.Command(filePath)
-	grpccommon.PrepareCommand(pluginCmd, p.startupConfig)
+	err = grpccommon.PrepareCommand(pluginCmd, p.startupConfig)
```
SkArchon, line 110: @Noroth I noticed there were a bunch of places where errors were unhandled, I only noticed it since goland highlights such cases.

router/pkg/grpcconnector/grpccommon/grpc_plugin_client.go
```diff
@@ -133,6 +146,11 @@ func (g *GRPCPluginClient) SetClients(pluginClient *plugin.Client, clientConn gr
 
 // Invoke implements grpc.ClientConnInterface.
 func (g *GRPCPluginClient) Invoke(ctx context.Context, method string, args any, reply any, opts ...grpc.CallOption) error {
+	ctx, span := g.tracer.Start(ctx,
+		"GRPC Plugin Client - Invoke",
+		trace.WithAttributes(g.getTraceAttributes(ctx)...))
+	defer span.End()
+
```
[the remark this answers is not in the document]
SkArchon: Records the error, tracer and getTraceAttributes wont be nil.

router-tests/router_plugin_test.go
```diff
@@ -131,7 +134,48 @@ func TestVerifyTelemetryForRouterPluginRequests(t *testing.T) {
...
+				require.Contains(t, span1.Attributes(), otel.WgOperationName.String("projects"))
+				require.Len(t, span1.Attributes(), 10)
+
+				span2 := snapshots[6]
+				require.Equal(t, "GRPC Plugin Client - Invoke", span2.Name())
+				require.Contains(t, span2.Attributes(), otel.WgOperationProtocol.String("grpc"))
+				require.Contains(t, span1.Attributes(), otel.WgOperationType.String("query"))
+				require.Contains(t, span1.Attributes(), otel.WgOperationName.String("projects"))
+				require.Len(t, span2.Attributes(), 10)
+			})
```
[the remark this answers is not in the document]
SkArchon: fixed

router/pkg/grpcconnector/grpccommon/grpc_plugin_client.go
```diff
@@ -133,13 +146,20 @@ func (g *GRPCPluginClient) SetClients(pluginClient *plugin.Client, clientConn gr
 
 // Invoke implements grpc.ClientConnInterface.
 func (g *GRPCPluginClient) Invoke(ctx context.Context, method string, args any, reply any, opts ...grpc.CallOption) error {
+	ctx, span := g.tracer.Start(ctx,
+		"GRPC Plugin Client - Invoke",
```
StarpTech: In HTTP we follow `query <type> <operation_name>`, we can do the same here since gRPC is an internal detail. People can identify the typ e via protocol attribute.

router/core/graph_server.go
```diff
@@ -1537,6 +1543,39 @@ func (s *graphServer) setupConnector(ctx context.Context, config *nodev1.EngineC
 
 		startupConfig := newGRPCStartupParams(s.traceConfig, s.ipAnonymization)
 
+		tracer := s.tracerProvider.Tracer("wundergraph/cosmo/router/engine/grpc", oteltrace.WithInstrumentationVersion("0.0.1"))
+
+		getTracingAttributes := func(ctx context.Context) []attribute.KeyValue {
```
StarpTech: Can we put this into a separate function?

router/pkg/grpcconnector/grpcplugin/grpc_plugin.go
StarpTech: This is a rare case, but in that case, it should return immediately.

router/pkg/grpcconnector/grpccommon/grpc_plugin_client.go
SkArchon: Will take a look.

router/core/graph_server.go
SkArchon: I think should be able to, either as its own function, or a function that returns a function. Will take a look.

router/pkg/grpcconnector/grpcplugin/grpc_plugin.go
SkArchon: The thing is I saw this in a bunch of places, I was thinking might be a good single rule to add for golanglintci (We could do it so that it only applies to the PR diff). Either way I will ticket this to take a look.

router/pkg/grpcconnector/grpccommon/grpc_plugin_client.go
SkArchon: Done, however when testing a bit more I noticed that it's different from normal HTTP subgraphs, because if we can aggregate queries to a subgraphs, it would be done and sent in one request to the subgraph. However in GRPC, we don't do this and invoke it multiple times. Thus this graph
image

router/core/graph_server.go
SkArchon: Done

StarpTech (approved):
LGTM

PR #2183, opened 2025-09-02
feat: add timings per client fetch for GraphQL http

SkArchon:
This PR Adds timings per datasource fetch for Graphql (HTTP) this would mean one fetch per subgraph call (or per subgraph log entry)

We didn't add a flag to conditionally since adding a timing adds about 2 microseconds on average. And adding a flag roughly costs about 300 nanoseconds (as per my local machine).

User's can compare how long (roughly, since we still have some go code to go through especially in the case of GraphQL/HTTP) the actual datasource call took with the total latency provided in the subgraph access log.

```
17:15:55 PM INFO requestlogger/subgraphlogger.go:68 /graphql {..., "latency": 0.006611209, ..."fetch_duration": 0.005611209}
```

Looking at `latency - fetchDuration` will give how long non fetch part of the router/engine per subgraph call takes.

Config
```
access_logs:
  subgraphs:
    enabled: true
    fields:
      - key: awesome
        value_from:
          expression: "subgraph.request.clientTrace.fetchDuration"
```

**GRPC Skipped**
This PR does not add fetch timings for GRPC datasources, as we do not process timings for logs because "Request" is nil for responseInfo for GRPC requests. The fun
```
if responseInfo.Request != nil {
  RequestFields(...)
```
This will be worked on as a separate PR, re factoring `RequestFields`.

## Summary by CodeRabbit

- **New Features**
  - Records per-data-source fetch duration and surfaces it in client traces and structured logs (including multi-subgraph and conditional cases).

- **Refactor**
  - Duration values standardized to native time units; telemetry attributes now report durations in seconds.

- **Tests**
  - Added/updated tests to validate fetch-duration and connection-acquire duration presence, formats, and telemetry across engine fetch spans and multi-subgraph requests.

## Checklist


router/internal/traceclient/traceclient.go
```diff
@@ -115,7 +116,7 @@ func (t *TraceInjectingRoundTripper) processConnectionMetrics(ctx context.Contex
 		duration := trace.ConnectionAcquired.Time.Sub(trace.ConnectionGet.Time)
 		connAcquireTime := float64(duration) / float64(time.Millisecond)
 
-		exprContext.Subgraph.Request.ClientTrace.ConnectionAcquireDuration = connAcquireTime
+		exprContext.Subgraph.Request.ClientTrace.ConnectionAcquireDuration = duration
```
SkArchon, line 117: This changes it to float seconds, right now its float miliseconds, and uses time.Duration, instead of int. This is different from "latency" which is float seconds, which I think is a bug. cc: @StarpTech

router/internal/traceclient/traceclient.go
```diff
@@ -113,9 +114,7 @@ func (t *TraceInjectingRoundTripper) processConnectionMetrics(ctx context.Contex
 
 	if trace.ConnectionGet != nil && trace.ConnectionAcquired != nil {
 		duration := trace.ConnectionAcquired.Time.Sub(trace.ConnectionGet.Time)
-		connAcquireTime := float64(duration) / float64(time.Millisecond)
-
-		exprContext.Subgraph.Request.ClientTrace.ConnectionAcquireDuration = connAcquireTime
+		exprContext.Subgraph.Request.ClientTrace.ConnectionAcquireDuration = duration
```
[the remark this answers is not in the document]
SkArchon: This is fine as it is existing behaviour

router/demo.config.yaml
```diff
@@ -19,4 +19,29 @@ events:
...
+      - endpoint: "http://localhost:4319"
+        exporter: http
+
+access_logs:
+  subgraphs:
+    enabled: true
+    fields:
+      - key: fetch_duration
+        value_from:
+          expression: "subgraph.request.clientTrace.fetchDuration"
```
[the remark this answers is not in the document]
SkArchon: This file should not be committed, cleaned up

router-tests/structured_logging_test.go
```diff
@@ -2867,6 +2868,121 @@ func TestFlakyAccessLogs(t *testing.T) {
...
+			testenv.Run(t, &testenv.Config{
+				SubgraphAccessLogsEnabled: true,
+				SubgraphAccessLogFields: []config.CustomAttribute{
+					{
+						Key: "data_source_fetch_duration",
+						ValueFrom: &config.CustomDynamicAttribute{
+							Expression: "subgraph.request.clientTrace.dataSourceFetchDuration",
+						},
+					},
+				},
```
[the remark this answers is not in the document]
SkArchon: Updated

router/internal/expr/expr.go
```diff
@@ -132,7 +133,8 @@ type SubgraphResponse struct {
 }
 
 type ClientTrace struct {
-	ConnectionAcquireDuration float64 `expr:"connAcquireDuration"`
+	DataSourceFetchDuration   time.Duration `expr:"dataSourceFetchDuration"`
```
StarpTech: I'd try to be free of engine specific. Something like `RequestDuration`

router/internal/traceclient/traceclient.go
StarpTech: What if the connection could not be obtained?

router/core/engine_loader_hooks.go
```diff
@@ -153,6 +157,12 @@ func (f *engineLoaderHooks) OnFinished(ctx context.Context, ds resolve.DataSourc
 	exprCtx.Subgraph.Name = ds.Name
 	exprCtx.Subgraph.Request.Error = WrapExprError(responseInfo.Err)
 
+	if value := ctx.Value(rcontext.FetchTimingKey); value != nil {
```
StarpTech, line 160: Why do we need to do this in a different place? Can't we unify this in a single custom transport?

StarpTech (changes requested):

router/internal/traceclient/traceclient.go
SkArchon: It will default to the zero value, which is 0, this is as GotConn hook is only called on successful acquires. People will need to ignore 0, will add it to the docs.

PR #2192, opened 2025-09-08
feat(router): add introspection authentication bypass feature

dkorittki:
@coderabbitai summary

## Checklist

Marked this PR as draft since I want to create a docs PR before this is here is ready for review. 

# Motivation and Context

Linear ticket: https://linear.app/wundergraph/issue/ENG-7980/allow-to-exclude-introspection-endpoint-from-auth

This pull requests adds the ability to configure the router to bypass authentication, if the operation is identified as an introspection query. Optionally a separate, static secret for introspection can be configured, which has to be passed via `Authorization` header instead of the usual JWT token when an introspection query is sent.

The motivation behind this feature is that it allows users to configure GraphQL client tooling without the need to acquire valid auth tokens first, which sometimes is not easy in this scenario. It's meant to be used locally. **This should not be enabled in production**.

To enable this feature I added a new section to the router configuration called `introspection`:

```yaml
introspection:
  enabled: true # deprecates the introspection_enabled config parameter
  skip_authentication: true # bypasses auth for introspection queries, if enabled (default: false)
  token: "demo" # optionally require a secret passed via Authorization header for introspection queries
```

# Changes

- Config parameter `introspection_enabled` is marked as deprecated
- New config parameter `introspection` with three child parameters, as described above, added
- `AccessController` is extended to know wether introspection auth bypass is enabled and what the secret is
- `AccessController` provides a method to check wether the incoming operation is an introspection query (using a temporary `OperationKit`) and based on it's configuration, if it should be bypassed
- `PreHandler` first checks via `AccessController` if authentication should be bypassed and skips it if true
- Modified existing auth tests in `router-tests` to disable this feature for all existing tests (to reflect the default config)
- Extended auth tests in `router-tests` to test authentication when this feature is enabled

### DIFF
--- a/router-tests/authentication_test.go
+++ b/router-tests/authentication_test.go
@@ -24,11 +24,13 @@ import (
 )
 
 const (
-	employeesQuery                = `{"query":"{ employees { id } }"}`
-	employeesQueryRequiringClaims = `{"query":"{ employees { id startDate } }"}`
-	employeesExpectedData         = `{"data":{"employees":[{"id":1},{"id":2},{"id":3},{"id":4},{"id":5},{"id":7},{"id":8},{"id":10},{"id":11},{"id":12}]}}`
-	unauthorizedExpectedData      = `{"errors":[{"message":"unauthorized"}]}`
-	xAuthenticatedByHeader        = "X-Authenticated-By"
+	employeesQuery                  = `{"query":"{ employees { id } }"}`
+	employeesQueryRequiringClaims   = `{"query":"{ employees { id startDate } }"}`
+	employeesExpectedData           = `{"data":{"employees":[{"id":1},{"id":2},{"id":3},{"id":4},{"id":5},{"id":7},{"id":8},{"id":10},{"id":11},{"id":12}]}}`
+	unauthorizedExpectedData        = `{"errors":[{"message":"unauthorized"}]}`
+	xAuthenticatedByHeader          = "X-Authenticated-By"
+	simpleIntrospectionQuery        = `{"query":"{ __type(name: \"Query\") { name } }"}`
+	simpleIntrospectionExpectedData = `{"data":{"__type":{"name":"Query"}}}`
 )
 
 func TestAuthentication(t *testing.T) {
@@ -41,7 +43,7 @@ func TestAuthentication(t *testing.T) {
 
 		testenv.Run(t, &testenv.Config{
 			RouterOptions: []core.Option{
-				core.WithAccessController(core.NewAccessController(authenticators, false)),
+				core.WithAccessController(core.NewAccessController(authenticators, false, false, "")),
 			},
 		}, func(t *testing.T, xEnv *testenv.Environment) {
 			// Operations without token should succeed
@@ -62,7 +64,7 @@ func TestAuthentication(t *testing.T) {
 		authenticators, _ := ConfigureAuth(t)
 		testenv.Run(t, &testenv.Config{
 			RouterOptions: []core.Option{
-				core.WithAccessController(core.NewAccessController(authenticators, false)),
+				core.WithAccessController(core.NewAccessController(authenticators, false, false, "")),
 			},
 		}, func(t *testing.T, xEnv *testenv.Environment) {
 			// Operations with an invalid token should fail
@@ -86,7 +88,7 @@ func TestAuthentication(t *testing.T) {
 		authenticators, authServer := ConfigureAuth(t)
 		testenv.Run(t, &testenv.Config{
 			RouterOptions: []core.Option{
-				core.WithAccessController(core.NewAccessController(authenticators, false)),
+				core.WithAccessController(core.NewAccessController(authenticators, false, false, "")),
 			},
 		}, func(t *testing.T, xEnv *testenv.Environment) {
 			// Operations with a token should succeed
@@ -112,7 +114,7 @@ func TestAuthentication(t *testing.T) {
 		authenticators, _ := ConfigureAuth(t)
 		testenv.Run(t, &testenv.Config{
 			RouterOptions: []core.Option{
-				core.WithAccessController(core.NewAccessController(authenticators, false)),
+				core.WithAccessController(core.NewAccessController(authenticators, false, false, "")),
 			},
 		}, func(t *testing.T, xEnv *testenv.Environment) {
 			res, err := xEnv.MakeRequest(http.MethodPost, "/graphql", nil, strings.NewReader(employeesQueryRequiringClaims))
@@ -130,7 +132,7 @@ func TestAuthentication(t *testing.T) {
 		authenticators, authServer := ConfigureAuth(t)
 		testenv.Run(t, &testenv.Config{
 			RouterOptions: []core.Option{
-				core.WithAccessController(core.NewAccessController(authenticators, false)),
+				core.WithAccessController(core.NewAccessController(authenticators, false, false, "")),
 			},
 		}, func(t *testing.T, xEnv *testenv.Environment) {
 			// Operations with a token should succeed
@@ -155,7 +157,7 @@ func TestAuthentication(t *testing.T) {
 		authenticators, authServer := ConfigureAuth(t)
 		testenv.Run(t, &testenv.Config{
 			RouterOptions: []core.Option{
-				core.WithAccessController(core.NewAccessController(authenticators, false)),
+				core.WithAccessController(core.NewAccessController(authenticators, false, false, "")),
 			},
 		}, func(t *testing.T, xEnv *testenv.Environment) {
 			// Operations with a token should succeed
@@ -182,7 +184,7 @@ func TestAuthentication(t *testing.T) {
 		authenticators, authServer := ConfigureAuth(t)
 		testenv.Run(t, &testenv.Config{
 			RouterOptions: []core.Option{
-				core.WithAccessController(core.NewAccessController(authenticators, false)),
+				core.WithAccessController(core.NewAccessController(authenticators, false, false, "")),
 			},
 		}, func(t *testing.T, xEnv *testenv.Environment) {
 			// Operations with a token should succeed
@@ -208,7 +210,7 @@ func TestAuthentication(t *testing.T) {
 		authenticators, authServer := ConfigureAuth(t)
 		testenv.Run(t, &testenv.Config{
 			RouterOptions: []core.Option{
-				core.WithAccessController(core.NewAccessController(authenticators, false)),
+				core.WithAccessController(core.NewAccessController(authenticators, false, false, "")),
 			},
 		}, func(t *testing.T, xEnv *testenv.Environment) {
 			// Operations with an token should succeed
@@ -235,7 +237,7 @@ func TestAuthentication(t *testing.T) {
 		authenticators, authServer := ConfigureAuth(t)
 		testenv.Run(t, &testenv.Config{
 			RouterOptions: []core.Option{
-				core.WithAccessController(core.NewAccessController(authenticators, false)),
+				core.WithAccessController(core.NewAccessController(authenticators, false, false, "")),
 				core.WithAuthorizationConfig(&config.AuthorizationConfiguration{
 					RejectOperationIfUnauthorized: true,
 				}),
@@ -266,7 +268,7 @@ func TestAuthentication(t *testing.T) {
 		authenticators, authServer := ConfigureAuth(t)
 		testenv.Run(t, &testenv.Config{
 			RouterOptions: []core.Option{
-				core.WithAccessController(core.NewAccessController(authenticators, false)),
+				core.WithAccessController(core.NewAccessController(authenticators, false, false, "")),
 				core.WithAuthorizationConfig(&config.AuthorizationConfiguration{
 					RejectOperationIfUnauthorized: true,
 				}),
@@ -295,7 +297,7 @@ func TestAuthentication(t *testing.T) {
 		authenticators, _ := ConfigureAuth(t)
 		testenv.Run(t, &testenv.Config{
 			RouterOptions: []core.Option{
-				core.WithAccessController(core.NewAccessController(authenticators, false)),
+				core.WithAccessController(core.NewAccessController(authenticators, false, false, "")),
 				core.WithAuthorizationConfig(&config.AuthorizationConfiguration{
 					RejectOperationIfUnauthorized: true,
 				}),
@@ -320,7 +322,7 @@ func TestAuthentication(t *testing.T) {
 		authenticators, _ := ConfigureAuth(t)
 		testenv.Run(t, &testenv.Config{
 			RouterOptions: []core.Option{
-				core.WithAccessController(core.NewAccessController(authenticators, false)),
+				core.WithAccessController(core.NewAccessController(authenticators, false, false, "")),
 				core.WithAuthorizationConfig(&config.AuthorizationConfiguration{
 					RejectOperationIfUnauthorized: true,
 				}),
@@ -343,7 +345,7 @@ func TestAuthentication(t *testing.T) {
 		authenticators, authServer := ConfigureAuth(t)
 		testenv.Run(t, &testenv.Config{
 			RouterOptions: []core.Option{
-				core.WithAccessController(core.NewAccessController(authenticators, false)),
+				core.WithAccessController(core.NewAccessController(authenticators, false, false, "")),
 			},
 		}, func(t *testing.T, xEnv *testenv.Environment) {
 			// Operations with a token should succeed
@@ -370,7 +372,7 @@ func TestAuthentication(t *testing.T) {
 		authenticators, authServer := ConfigureAuth(t)
 		testenv.Run(t, &testenv.Config{
 			RouterOptions: []core.Option{
-				core.WithAccessController(core.NewAccessController(authenticators, false)),
+				core.WithAccessController(core.NewAccessController(authenticators, false, false, "")),
 			},
 		}, func(t *testing.T, xEnv *testenv.Environment) {
 			// Operations with a token should succeed
@@ -397,7 +399,7 @@ func TestAuthentication(t *testing.T) {
 		authenticators, authServer := ConfigureAuth(t)
 		testenv.Run(t, &testenv.Config{
 			RouterOptions: []core.Option{
-				core.WithAccessController(core.NewAccessController(authenticators, false)),
+				core.WithAccessController(core.NewAccessController(authenticators, false, false, "")),
 			},
 		}, func(t *testing.T, xEnv *testenv.Environment) {
 			// Operations with a token should succeed
@@ -426,7 +428,7 @@ func TestAuthentication(t *testing.T) {
 		authenticators, _ := ConfigureAuth(t)
 		testenv.Run(t, &testenv.Config{
 			RouterOptions: []core.Option{
-				core.WithAccessController(core.NewAccessController(authenticators, false)),
+				core.WithAccessController(core.NewAccessController(authenticators, false, false, "")),
 			},
 		}, func(t *testing.T, xEnv *testenv.Environment) {
 			res, err := xEnv.MakeRequest(http.MethodPost, "/graphql", nil, strings.NewReader(`
@@ -446,7 +448,7 @@ func TestAuthentication(t *testing.T) {
 		authenticators, _ := ConfigureAuth(t)
 		testenv.Run(t, &testenv.Config{
 			RouterOptions: []core.Option{
-				core.WithAccessController(core.NewAccessController(authenticators, false)),
+				core.WithAccessController(core.NewAccessController(authenticators, false, false, "")),
 			},
 		}, func(t *testing.T, xEnv *testenv.Environment) {
 			res, err := xEnv.MakeRequest(http.MethodPost, "/graphql", nil, strings.NewReader(`
@@ -466,7 +468,7 @@ func TestAuthentication(t *testing.T) {
 		authenticators, _ := ConfigureAuth(t)
 		testenv.Run(t, &testenv.Config{
 			RouterOptions: []core.Option{
-				core.WithAccessController(core.NewAccessController(authenticators, false)),
+				core.WithAccessController(core.NewAccessController(authenticators, false, false, "")),
 			},
 		}, func(t *testing.T, xEnv *testenv.Environment) {
 			res, err := xEnv.MakeRequest(http.MethodPost, "/graphql", nil, strings.NewReader(`
@@ -486,7 +488,7 @@ func TestAuthentication(t *testing.T) {
 		authenticators, authServer := ConfigureAuth(t)
 		testenv.Run(t, &testenv.Config{
 			RouterOptions: []core.Option{
-				core.WithAccessController(core.NewAccessController(authenticators, false)),
+				core.WithAccessController(core.NewAccessController(authenticators, false, false, "")),
 			},
 		}, func(t *testing.T, xEnv *testenv.Environment) {
 			token, err := authServer.Token(nil)
@@ -511,7 +513,7 @@ func TestAuthentication(t *testing.T) {
 		authenticators, authServer := ConfigureAuth(t)
 		testenv.Run(t, &testenv.Config{
 			RouterOptions: []core.Option{
-				core.WithAccessController(core.NewAccessController(authenticators, false)),
+				core.WithAccessController(core.NewAccessController(authenticators, false, false, "")),
 			},
 		}, func(t *testing.T, xEnv *testenv.Environment) {
 			token, err := authServer.Token(map[string]any{
@@ -538,7 +540,7 @@ func TestAuthentication(t *testing.T) {
 		authenticators, authServer := ConfigureAuth(t)
 		testenv.Run(t, &testenv.Config{
 			RouterOptions: []core.Option{
-				core.WithAccessController(core.NewAccessController(authenticators, false)),
+				core.WithAccessController(core.NewAccessController(authenticators, false, false, "")),
 			},
 		}, func(t *testing.T, xEnv *testenv.Environment) {
 			token, err := authServer.Token(map[string]any{
@@ -565,7 +567,7 @@ func TestAuthentication(t *testing.T) {
 		authenticators, authServer := ConfigureAuth(t)
 		testenv.Run(t, &testenv.Config{
 			RouterOptions: []core.Option{
-				core.WithAccessController(core.NewAccessController(authenticators, false)),
+				core.WithAccessController(core.NewAccessController(authenticators, false, false, "")),
 			},
 		}, func(t *testing.T, xEnv *testenv.Environment) {
 			token, err := authServer.Token(map[string]any{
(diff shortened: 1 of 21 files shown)


dkorittki:
fyi image scan fails because our distroless go image is vulnerable right now

https://github.com/GoogleContainerTools/distroless/issues/1863

Will wait for upstream fix and then rerun CI step, should fetch the updated image automatically and then it should work. 

EDIT: vulnerability fixed upstream.

router/core/access_controller.go
```diff
@@ -44,3 +49,64 @@ func (a *AccessController) Access(w http.ResponseWriter, r *http.Request) (*http
...
+		return false
+	}
+
+	if isIntrospection {
+		if a.isValidIntrospectionToken(r) {
+			return true
+		}
+	}
+
+	return false
```
Noroth: return isIntrospection && a.isValidIntrospectionToken ?

router/core/graphql_prehandler.go
```diff
@@ -360,29 +360,33 @@ func (h *PreHandler) Handler(next http.Handler) http.Handler {
...
 
-			r = validatedReq
+				authenticateSpan.End()
 
-			requestContext.expressionContext.Request.Auth = expr.LoadAuth(r.Context())
+				r = validatedReq
+
+				requestContext.expressionContext.Request.Auth = expr.LoadAuth(r.Context())
+			}
 		}
```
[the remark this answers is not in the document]
dkorittki: This is a valid finding. However, I am not sure if it's the right approach to close the span like suggested by coderabbit. How about we simply not start it if we don't authenticate the request?

router/core/access_controller.go
dkorittki: fixed

router/pkg/config/testdata/config_full.json
```diff
@@ -459,6 +459,11 @@
   },
   "PlaygroundEnabled": true,
   "IntrospectionEnabled": true,
+  "IntrospectionConfig": {
+    "Enabled": true,
+    "SkipAuthentication": false,
+    "Token": "aTLQw7rckY3N19Fo7ynYgbwaJFKMrXxI"
+  },
```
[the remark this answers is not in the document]
dkorittki: fixed

router/pkg/config/fixtures/full.yaml
```diff
@@ -20,6 +20,10 @@ playground:
   path: '/my-playground'
   concurrency_limit: 1500
 introspection_enabled: true
+introspection:
+  enabled: true
+  skip_authentication: false
+  token: 'aTLQw7rckY3N19Fo7ynYgbwaJFKMrXxI'
```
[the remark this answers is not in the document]
dkorittki: fixed

router/pkg/config/config.go
```diff
@@ -1009,7 +1015,8 @@ type Config struct {
 	ControlplaneURL               string                      `yaml:"controlplane_url" envDefault:"https://cosmo-cp.wundergraph.com" env:"CONTROLPLANE_URL"`
 	PlaygroundConfig              PlaygroundConfig            `yaml:"playground,omitempty"`
 	PlaygroundEnabled             bool                        `yaml:"playground_enabled" envDefault:"true" env:"PLAYGROUND_ENABLED"`
-	IntrospectionEnabled          bool                        `yaml:"introspection_enabled" envDefault:"true" env:"INTROSPECTION_ENABLED"`
+	IntrospectionEnabled          bool                        `yaml:"introspection_enabled" envDefault:"true" env:"INTROSPECTION_ENABLED"` // deprecated, use IntrospectionConfiguration instead
+	IntrospectionConfig           IntrospectionConfiguration  `yaml:"introspection,omitempty"`
```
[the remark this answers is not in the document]
dkorittki: fixed

router/core/access_controller.go
```diff
@@ -44,3 +49,58 @@ func (a *AccessController) Access(w http.ResponseWriter, r *http.Request) (*http
...
+		return false
+	}
+
+	operationKit, err := operationProcessor.NewKit()
+	if err != nil {
+		return false
+	}
+	defer operationKit.Free()
+
+	err = operationKit.UnmarshalOperationFromBody(body)
```
SkArchon: I believe introspection queries sent from GET requests are not handled?

Take a look here: https://github.com/wundergraph/cosmo/blob/a52c2153304610e116274752a8269e41be887359/router/core/graphql_prehandler.go#L523

SkArchon: I am also not 100% sure of the performance implication of creating a new kit and parsing the body twice for skipping auth.
We parse it a bit below again in `h.handleOperation`. See: https://github.com/wundergraph/cosmo/blob/a52c2153304610e116274752a8269e41be887359/router/core/graphql_prehandler.go#L504

I am thinking if we are duplicating the logic in `h.handleOperation` it might be worth considering if we can duplicate the auth logic instead (or change the ordering carefully) to skip parsing twice (and also duplicate the GET / POST parsing conditions), this would mean

Right now 
 - in case of skipAuth = true: parse twice, always parse before auth
 - in case of skipAuth = false: parse once
  
 After
 - in case of skipAuth = true: parse once, always parse before auth
 - in case of skipAuth = false: parse once

cc: @devsergiy

router/core/supervisor_instance.go
```diff
@@ -156,6 +161,7 @@ func optionsFromResources(logger *zap.Logger, config *config.Config) []Option {
 		WithOverrides(config.Overrides),
 		WithLogger(logger),
 		WithIntrospection(config.IntrospectionEnabled),
+		WithIntrospectionConfig(config.IntrospectionConfig),
```
SkArchon: I notice we only use this to access the `config.Enabled` flag here and nowhere else
https://github.com/wundergraph/cosmo/pull/2192/files#diff-aa58824b56f499a6a7d39dd7d905130a1aed0683be3aa6826a63dfd5d5047edcR227-R232

Maybe we could simplify it and pass in two parameters to `WithIntrospection`? Only if that would simplify things.

dkorittki: Yeah currently it's only meant to override the original config parameter (introspection_enabled) on the router until we eventually remove the old parameter. Will implement your suggestion, thanks!

router/core/access_controller.go
dkorittki: Yeah, tbh I am not really happy with parsing it twice either. I decided to go with it because it thought it's better than the alternative (parsing before authentication), because doing so also means we have to fetch persisted operations from the CDN before parsing, right?

https://github.com/wundergraph/cosmo/blob/a52c2153304610e116274752a8269e41be887359/router/core/graphql_prehandler.go#L594

To me it does not sound like a great thing to do before we authenticated the request. And since parsing (and invoking the operation kit) only happens twice when this not-meant-for-production feature is enabled, I did go for it. That's basically why I did it this way but maybe I am totally wrong here. Please let me know!

router/pkg/config/config.go
```diff
@@ -983,6 +983,12 @@ type PluginRegistryConfiguration struct {
 	URL string `yaml:"url" env:"URL" envDefault:"cosmo-registry.wundergraph.com"`
 }
 
+type IntrospectionConfiguration struct {
+	Enabled            bool   `yaml:"enabled" envDefault:"true" env:"INTROSPECTION_ENABLED"`
+	SkipAuthentication bool   `yaml:"skip_authentication" envDefault:"false" env:"INTROSPECTION_SKIP_AUTHENTICATION"`
+	Token              string `yaml:"token" env:"INTROSPECTION_TOKEN"`
+}
```
endigma, line 1004: Could this be like the below example? Is the non-token authentication even required by the ticket? We could potentially alleviate some of the concerns with duplicating the operationkit and PO and etc if all requests are authenticated _somehow_
```yml
introspection:
  enabled: true
  authentication_mode: full | token | skip
  token: "wasd" (required in JSON schema if authentication mode == token)
```

router/pkg/config/config.schema.json
```diff
@@ -1362,7 +1362,31 @@
...
+        "skip_authentication": {
+          "type": "boolean",
+          "description": "If set to true introspection requests will skip the auth.",
+          "default": false
+        },
+        "token": {
+          "type": "string",
+          "description": "Optionally, you can set a token to authenticate the introspection requests.",
+          "minLength": 32
+        }
```
endigma: We should document/verify how these two options interact, or make it more rigid as I suggested elsewhere

router/core/router.go
```diff
@@ -222,6 +222,15 @@ func NewRouter(opts ...Option) (*Router, error) {
...
 	}
 
+	// handle introspection config deprecation
+	// if either the old deprecated or the new config is set to false, introspection is disabled
+	if !r.introspection {
+		r.introspectionConfig.Enabled = r.introspection
+		r.logger.Warn("The introspection_enabled option is deprecated. Use the introspection.enabled option in the config instead.")
+	} else if !r.introspectionConfig.Enabled {
+		r.introspection = r.introspectionConfig.Enabled
+	}
```
endigma, line 232: Nit: I think elsewhere we use the JSON schema path format like `/introspection/enabled`? If so we should be consistent.

endigma (changes requested):
I have some thoughts about duplicating/moving the operationkit, but here's some other things that should come first.

router/core/supervisor_instance.go
dkorittki: fixed

router/core/graphql_prehandler.go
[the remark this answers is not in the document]
dkorittki: > How about we simply not start it if we don't authenticate the request?

Implemented it like this

router/core/access_controller.go
```diff
@@ -1,12 +1,28 @@
 package core
 
 import (
+	"crypto/subtle"
 	"errors"
+	"fmt"
 	"net/http"
 
 	"github.com/wundergraph/cosmo/router/pkg/authentication"
 )
```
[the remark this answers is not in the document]
dkorittki: Does not make sense. We expect the user to not put a Bearer prefix into the Authorization header for this but simply the static secret as is. If he does put a Bearer prefix in the header I would consider this an invalid secret and hence let this method return `false`. The same goes for whitespaces at the front/end.

router-tests/authentication_test.go
```diff
@@ -2504,6 +2698,246 @@ func TestAudienceValidation(t *testing.T) {
...
+	})
+
+	t.Run("introspection query with auth skip and introspection token", func(t *testing.T) {
+		t.Parallel()
+
+		// since we are skipping auth for introspection and have not configured
+		// a dedicated token for introspection queries, this query must succeed
+
+		authenticators, _ := ConfigureAuth(t)
+		token := "0aG3TVhkZ2lPy8ULUPQLXZ4JFPfpxk"
```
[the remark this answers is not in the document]
dkorittki: fixed

router/core/access_controller.go
```diff
@@ -44,3 +74,68 @@
...
+// isValidIntrospectionToken safely validates the configured introspection token
+// against the Authorization header of the request.
+func (a *AccessController) isValidIntrospectionToken(r *http.Request) bool {
+	authHeader := r.Header.Get("Authorization")
+	if authHeader == "" {
+		return false
+	}
+
+	return subtle.ConstantTimeCompare([]byte(authHeader), []byte(a.introspectionAuthSkipToken)) == 1
+}
```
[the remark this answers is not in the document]
dkorittki: Quote from further above:

> Does not make sense. We expect the user to not put a Bearer prefix into the Authorization header for this but simply the static secret as is. If he does put a Bearer prefix in the header I would consider this an invalid secret and hence let this method return false. The same goes for whitespaces at the front/end.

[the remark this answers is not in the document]
dkorittki: added the guard. It should not be possible to configure an empty token due to config schema but still can't hurt to check for it here as well.

router/pkg/config/config.go
dkorittki: Great idea! Implemented that

router/pkg/config/config.schema.json
dkorittki: I added it to the description and will update my pr on the Cosmo docs as well

router/core/access_controller.go
```diff
@@ -44,3 +75,23 @@ func (a *AccessController) Access(w http.ResponseWriter, r *http.Request) (*http
...
+		}
+
+		authHeader = strings.TrimSpace(authHeader)
+
+		// guard to prevent bypass when token is unset
+		if a.introspectionAuthSkipToken == "" {
+			return false
+		}
+
+		return subtle.ConstantTimeCompare([]byte(authHeader), []byte(a.introspectionAuthSkipToken)) == 1
```
SkArchon: Could you explain what we are trying to achieve here, what I understood is that there is a static token we can define that gets compared? Why do we need this

router/core/router.go
```diff
@@ -211,17 +211,26 @@ func NewRouter(opts ...Option) (*Router, error) {
...
 	}
 
+	// handle introspection config deprecation
+	// if either the old deprecated or the new config is set to false, introspection is disabled
+	if !r.introspection {
+		r.introspectionConfig.Enabled = r.introspection
+		r.logger.Warn("The introspection_enabled option is deprecated. Use the /introspection/enabled option in the config instead.")
+	} else if !r.introspectionConfig.Enabled {
+		r.introspection = r.introspectionConfig.Enabled
+	}
```
SkArchon, line 232: I was thinking that maybe you could move this logic into `WithIntrospection`, but it's not a big deal.

router/core/access_controller.go
SkArchon: nvm answered it, saw it in the ticket

router/core/access_controller.go
```diff
@@ -16,15 +33,29 @@ var (
...
+		return nil, fmt.Errorf("invalid introspection auth mode: %s", introspectionAuthMode)
 	}
+
+	return &AccessController{
+		authenticationRequired:     authenticationRequired,
+		authenticators:             authenticators,
+		introspectionAuthMode:      introspectionAuthMode,
+		introspectionAuthSkipToken: introspectionAuthSkipToken,
+	}, nil
 }
```
endigma: I think the parameters for this function are now enough that there should likely be an options type, 4 unnamed params of mixed types is a lot.

router/core/access_controller.go
```diff
@@ -1,12 +1,29 @@
...
+const (
+	// IntrospectionAuthModeFull requires normal authentication for introspection queries.
+	IntrospectionAuthModeFull IntrospectionAuthMode = "full"
+
+	// IntrospectionAuthModeToken requires a specific introspection token for introspection queries.
+	IntrospectionAuthModeToken IntrospectionAuthMode = "token"
+
+	// IntrospectionAuthModeSkip bypasses authentication for introspection queries.
+	IntrospectionAuthModeSkip IntrospectionAuthMode = "skip"
+)
```
endigma: It's good practice with the Go string enums like this to have a named default like

```go
const IntrospectionAuthModeInvalid IntrospectionAuthMode = ""
```

router/core/access_controller.go
```diff
@@ -16,15 +33,29 @@ var (
...
+// NewAccessController creates a new AccessController.
+// It returns an error if the introspection auth mode is invalid.
+func NewAccessController(
+	authenticators []authentication.Authenticator,
+	authenticationRequired bool,
+	introspectionAuthMode IntrospectionAuthMode,
+	introspectionAuthSkipToken string) (*AccessController, error) {
+	if introspectionAuthMode != IntrospectionAuthModeFull && introspectionAuthMode != IntrospectionAuthModeToken && introspectionAuthMode != IntrospectionAuthModeSkip {
+		return nil, fmt.Errorf("invalid introspection auth mode: %s", introspectionAuthMode)
 	}
```
endigma: Total nitpick, but you could attach a validator to the type and invert this, it would be faster. As-is you always check 4 inequalities instead of =<4 equalities

router/core/graphql_prehandler.go
```diff
@@ -655,6 +675,33 @@ func (h *PreHandler) handleOperation(req *http.Request, variablesParser *astjson
 		engineParseSpan.End()
 	}
 
+	if h.accessController != nil {
+		// Based on the authentication result, the introspection config,
+		// and wether this is an introspection query,
+		// we decide here if we need to abort the request or not.
+		isIntrospection, err := operationKit.isIntrospectionQuery()
+		if err != nil {
+			requestContext.logger.Debug("failed to check if operation is introspection, treat it like non-introspection operation", zap.Error(err))
```
endigma: I think this is worth `logger.Error`

router/core/graphql_prehandler.go
```diff
@@ -655,6 +675,33 @@ func (h *PreHandler) handleOperation(req *http.Request, variablesParser *astjson
...
+					}
+				}
+			} else {
+				return &httpGraphqlError{
+					message:    "unauthorized",
+					statusCode: http.StatusUnauthorized,
+				}
+			}
+		}
+	}
```
endigma: Can you swap and invert the outer if with the inner one and return early instead of nesting?

endigma (changes requested):
Can you add a nice loud warning log in startup if it's set to skip? "Only for development do not enable in production" type thing

router/core/access_controller.go
dkorittki: fixed

router/core/access_controller.go
dkorittki: fixed

router/core/access_controller.go
dkorittki: fixed

router/core/graphql_prehandler.go
dkorittki: fixed

router/core/graphql_prehandler.go
dkorittki: fixed

router/pkg/config/config.schema.json
```diff
@@ -1362,7 +1362,42 @@
...
+      "type": "object",
+      "description": "",
+      "additionalProperties": false,
+      "properties": {
+        "enabled": {
+          "type": "boolean",
+          "description": "",
+          "default": true
+        },
+        "authentication_mode": {
```
StarpTech: I find these variations confusing. We either check for auth or not and allow optionally a token to validate. We should always avoid creating configurations that depend on each other on the same level. Have you thought about simplifying it?

```yaml
introspection:
  enabled: true
  skip_auth: false
  token: ".."
```

router/pkg/config/config.schema.json
```diff
@@ -1362,7 +1362,42 @@
     "introspection_enabled": {
       "type": "boolean",
       "description": "Enable the GraphQL introspection. The GraphQL introspection allows you to query the schema of the GraphQL API. The default value is true. If the value is false, the GraphQL introspection is disabled. In production, it is recommended to disable the introspection.",
-      "default": true
+      "default": true,
+      "deprecated": true,
+      "deprecationMessage": "introspection_enabled is deprecated. Please use the introspection.enabled configuration instead."
```
StarpTech: Always uppercase in the beginning

StarpTech (changes requested):

router/core/access_controller.go
```diff
@@ -44,3 +87,23 @@ func (a *AccessController) Access(w http.ResponseWriter, r *http.Request) (*http
...
+		if authHeader == "" {
+			return false
+		}
+
+		authHeader = strings.TrimSpace(authHeader)
+
+		// guard to prevent bypass when token is unset
+		if a.introspectionAuthSkipToken == "" {
+			return false
+		}
```
endigma: Can we fail this at setup instead (or as well?)

router/pkg/config/config.schema.json
endigma: Funnily enough, I suggested the opposite of this to make it the way it is now because the interactions between `token` and `skip_auth` are not clear in this way.

Perhaps:


```yaml
introspection:
  enabled: true
  skip_auth: true <- use no auth
  token_auth: "..." <- use token auth
```

Where `skip_auth` and `token_auth` are mutually exclusive enforced by schema?

endigma (changes requested):

router/pkg/config/config.schema.json
StarpTech: Got it. How about?

```yaml
introspection:
  enabled: true
  secret: "..." # empty by default
  
authentication:
  ignore_introspection: true # false by default
```

dkorittki: Yeah its hard to let the user know intuitively what the config parameters are meaning and how they interact without pointing to the docs. What about this approach:

```yaml
introspection:
  enabled: true
  skip_auth: false
  skip_auth_token: ".."
```

Thats @StarpTech ´s approach with a renamed `token` parameter. The downside is requiring a token to skip auth is still some sort of auth. So it reads a bit confusing ("skip auth by using auth"). But I think to the user it brings across the meaning. At least as long as they read it as `-->introspection<--.skip_auth_token`. 

Me personally I am not too invested and I think both approaches will work, just have to settle for one.

router/core/access_controller.go
dkorittki: There´s a schema check which ensures no empty tokens are provided when auth mode == token

router/pkg/config/config.schema.json
StarpTech: I agree with your `"skip auth by using auth"`. I'd use the `authentication ` section to let the user skip auth for introspection. That's why we have the section for and practically that's what we will skip. Optionally, a User can then define a secret / token to protect introspection calls. IMO this is the least confusing approach.

dkorittki: Yeah I agree, I will go with your suggestion from above
https://github.com/wundergraph/cosmo/pull/2192#discussion_r2362360334

router/core/supervisor_instance.go
```diff
@@ -57,7 +57,26 @@ func newRouter(ctx context.Context, params RouterResources, additionalOptions ..
...
+			AuthenticationRequired:   cfg.Authorization.RequireAuthentication,
+			SkipIntrospectionQueries: cfg.Authentication.IgnoreIntrospection,
+			IntrospectionSkipSecret:  cfg.IntrospectionConfig.Secret,
+		})
+		if err != nil {
+			return nil, fmt.Errorf("could not create access controller: %w", err)
+		}
+
+		options = append(options, WithAccessController(accessController))
+	}
```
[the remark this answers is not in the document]
dkorittki: We do not need an access controller, if authentication is disabled via config (i.e. `len(authenticators) == 0`). More important, there must not be an access controller then because other code parts (the prehandler) use `accessController == nil` to determine if auth should or should not be performed. By always creating an access controller we would perform authentication even if we should not.

router/core/supervisor_instance.go
```diff
@@ -57,7 +57,26 @@ func newRouter(ctx context.Context, params RouterResources, additionalOptions ..
...
+
+	if cfg.Authentication.IgnoreIntrospection && cfg.IntrospectionConfig.Secret == "" {
+		params.Logger.Warn("introspection operations are not authenticated. " +
+			"Consider setting /introspection/secret or set /authentication/ignore_introspection to false. Do not use this in production.")
+	}
+
+	if !cfg.Authentication.IgnoreIntrospection && cfg.IntrospectionConfig.Secret != "" {
+		params.Logger.Warn("/introspection/secret is ignored because /authentication/ignore_introspection is false.")
 	}
```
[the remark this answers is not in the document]
dkorittki: Obsolete since because of my comment [here](https://github.com/wundergraph/cosmo/pull/2192#discussion_r2372870649) we will not have `introspectionSkip`

router/pkg/config/config.schema.json
```diff
@@ -1357,12 +1357,31 @@
...
+          "description": "",
+          "default": true
+        },
+        "secret": {
+          "type": "string",
+          "description": "A dedicated secret to authenticate introspection requests independently from the regular authentication. Need to passed by introspection queries via the 'Authorization' header. Works only when authentication.ignore_introspection is true.",
+          "minLength": 32
+        }
+      }
     },
```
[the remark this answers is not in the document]
dkorittki: fixed

router/pkg/config/config.go
```diff
@@ -1019,7 +1025,8 @@ type Config struct {
 	ControlplaneURL               string                      `yaml:"controlplane_url" envDefault:"https://cosmo-cp.wundergraph.com" env:"CONTROLPLANE_URL"`
 	PlaygroundConfig              PlaygroundConfig            `yaml:"playground,omitempty"`
 	PlaygroundEnabled             bool                        `yaml:"playground_enabled" envDefault:"true" env:"PLAYGROUND_ENABLED"`
-	IntrospectionEnabled          bool                        `yaml:"introspection_enabled" envDefault:"true" env:"INTROSPECTION_ENABLED"`
+	IntrospectionEnabled          bool                        `yaml:"introspection_enabled" envDefault:"true"` // deprecated, use IntrospectionConfiguration instead
+	IntrospectionConfig           IntrospectionConfiguration  `yaml:"introspection,omitempty"`
```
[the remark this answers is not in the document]
dkorittki: The deprecation strategy is to migrate these calls to the new config parameter once we get rid of the old one.

router/core/graphql_prehandler.go
```diff
@@ -362,27 +372,38 @@ func (h *PreHandler) Handler(next http.Handler) http.Handler {
...
 			if err != nil {
-				requestContext.SetError(err)
-				requestLogger.Debug("Failed to authenticate request", zap.Error(err))
-
-				// Mark the root span of the router as failed, so we can easily identify failed requests
-				rtrace.AttachErrToSpan(routerSpan, err)
-				rtrace.AttachErrToSpan(authenticateSpan, err)
-
-				authenticateSpan.End()
+				// Auth failed but but introspection queries might be allowed to skip auth.
```
StarpTech: Typo

router/core/supervisor_instance.go
```diff
@@ -57,7 +57,26 @@ func newRouter(ctx context.Context, params RouterResources, additionalOptions ..
...
 
 	if len(authenticators) > 0 {
-		options = append(options, WithAccessController(NewAccessController(authenticators, cfg.Authorization.RequireAuthentication)))
+		accessController, err := NewAccessController(AccessControllerOptions{
+			Authenticators:           authenticators,
+			AuthenticationRequired:   cfg.Authorization.RequireAuthentication,
+			SkipIntrospectionQueries: cfg.Authentication.IgnoreIntrospection,
+			IntrospectionSkipSecret:  cfg.IntrospectionConfig.Secret,
+		})
```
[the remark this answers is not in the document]
dkorittki: `cfg.IntrospectionConfig.SkipAuthentication` does not exist anymore since at least the latest commit.

router/core/supervisor_instance.go
```diff
@@ -57,7 +57,26 @@ func newRouter(ctx context.Context, params RouterResources, additionalOptions ..
...
+		if err != nil {
+			return nil, fmt.Errorf("could not create access controller: %w", err)
+		}
+
+		options = append(options, WithAccessController(accessController))
+	}
+
+	if cfg.Authentication.IgnoreIntrospection && cfg.IntrospectionConfig.Secret == "" {
+		params.Logger.Warn("introspection operations are not authenticated. " +
+			"Consider setting /introspection/secret configuration parameter or set /authentication/ignore_introspection to false. Do not use this in production.")
```
StarpTech: Please use dot syntax to refer to configuration `foo.bar`

router/core/graphql_prehandler.go
dkorittki: Fixed

router/core/supervisor_instance.go
dkorittki: Fixed

endigma (approved):
Lgtm

PR #2210, opened 2025-09-12
feat(router): add option to block on missing client headers

dkorittki:
@coderabbitai summary

## Checklist

# Motivation and Context

Cosmo Router supports two special http request header configuration parameters for identifying clients by their requests:
https://cosmo-docs.wundergraph.com/studio/analytics/client-identification#client-identification

Sometimes users want to reject requests, which do not add these headers. This is currently not supported and that is, what this PR will change. Right now this PR only contains the configuration changes as a basis for discussion. After discussion I will start implementing it.

The proposal is to let the user decide wether to reject requests, which do not contain the `Name`, the `Version` or both headers. The user can also decide what http response code should be used, the default being `400`.

### DIFF
--- a/router/pkg/config/fixtures/full.yaml
+++ b/router/pkg/config/fixtures/full.yaml
@@ -473,6 +473,10 @@ router_config_path: 'latest.json'
 client_header:
   name: 'Client-Name'
   version: 'Client_Version'
+  enforce:
+    on_name: false
+    on_version: false
+    response_code: 400
 
 apollo_compatibility_flags:
   enable_all: false
--- a/router/pkg/config/config.go
+++ b/router/pkg/config/config.go
@@ -927,8 +927,15 @@ type ApolloCompatibilityFlag struct {
 }
 
 type ClientHeader struct {
-	Name    string `yaml:"name,omitempty"`
-	Version string `yaml:"version,omitempty"`
+	Name    string                           `yaml:"name,omitempty"`
+	Version string                           `yaml:"version,omitempty"`
+	Enforce ClientHeaderEnforceConfiguration `yaml:"enforce,omitempty"`
+}
+
+type ClientHeaderEnforceConfiguration struct {
+	OnName       bool `yaml:"on_name"`
+	OnVersion    bool `yaml:"on_version"`
+	ResponseCode int  `yaml:"response_code" envDefault:"400"`
 }
 
 type CacheWarmupSource struct {
--- a/router/pkg/config/config.schema.json
+++ b/router/pkg/config/config.schema.json
@@ -3051,6 +3051,28 @@
         "version": {
           "type": "string",
           "description": "The custom client version header."
+        },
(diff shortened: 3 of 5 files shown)


router/pkg/config/fixtures/full.yaml
```diff
@@ -473,6 +473,10 @@ router_config_path: 'latest.json'
 client_header:
   name: 'Client-Name'
   version: 'Client_Version'
+  enforce:
+    on_name: false
+    on_version: false
+    response_code: 400
```
endigma, line 479:
```suggestion
  require:
    name: false
    version: false
    response_code: 400
```

endigma: I think this makes more sense? Do we need to let the response code be customized? End users can always do this or any other behaviour they like with a custom module.

jensneuse: I like your suggestion on name and version. That said, I'd recommend to set a default response status code and allow the user to define a custom_error_status_code (or similar) and a custom_error_message_expression (or similar) to customize the error message and code. Then we're kind of safe. If you want to keep it simpler, just do name and version, but in this case we cannot customize the status and message in a post middleware because we don't know what error case happened, right?

dkorittki: Thanks for your feedback @endigma @jensneuse

Really prefer `require` over `enforce` 👍 

> That said, I'd recommend to set a default response status code and allow the user to define a custom_error_status_code (or similar)

Thats what `enforce.response_code` is supposed to be. If the user doesn't define it via config, it defaults to `400` but maybe the parameter name is confusing and doesn't imply this? I will use `custom_error_status_code` to make this more clear.

> End users can always do this or any other behaviour they like with a custom module.

Yeah the reason to provide this via config is so users don't have to write middleware for it and we have to find a balance on how sophisticated our config-based solution should be before it becomes to complex, wich is when users should opt for a custom module middleware instead. Imo I think allowing custom response codes is okay for this config-based approach.

> [...] and a custom_error_message_expression (or similar) to customize the error message [...]

Hmm, can you explain the idea behind a custom error message expression a bit? Should it look something like this?

```yaml
enforce:
  custom_error_message_expression: 'Missing required header {{.MissingHeaderName}}'
```

Albeit I think this might get a bit too sophisticated for a config-based approach and it might be a better idea to resort to a `core.RouterOnRequestHandler` or similar custom module for something so complex. What do you think @jensneuse ?

jensneuse: Naming is fine. You can leave the custom error message expression out if you think it's out of scope.
We are using this approach in many cases: https://cosmo-docs.wundergraph.com/router/configuration/template-expressions

We've built a standard around creating the expr lang context. Using this approach everywhere makes it very powerful and flexible. But like I said, I don't see this as a strict requirement.

PR #2216, opened 2025-09-15
fix: jwt validation blocks on multiple requests

SkArchon:
We right now have the following line in our JWT configuration
```
RefreshUnknownKID: rate.NewLimiter(rate.Every(5*time.Minute), 1),
```
This means whenever a **valid formed** token (does not mean it is valid for auth) is sent to the router, it will try to find the correct JWKs entry based on the key id, but if it cannot find the key id it will on demand re-query the JWKs endpoint OUTSIDE of the refresh cycle within that request, hoping that the key id has been added to the JWKs. However to prevent misuse of this feature, there is a rate limit, so that if multiple requests are sent with an unknown keyid, we don't call the JWKs endpoint asap. However this blocks each request for 5 minutes cumulatively from my testing (because the burst is 1).

Cumulatively meaning,
if we send 6 requests at once
```
* 1st response is returned immediately
* 2nd response is returned in 5 * 1 minutes
* 3rd response is returned in 5 * 2 minutes
* 4th response is returned in 5 * 3 minutes
* 5th response is returned in 5 * 4 minutes
* 6th response is returned in 5 * 5 minutes
```

What happens now.
* We refresh the jwks from the jwks endpoint per defined refresh cycle (1m, etc)
* If the keyid is not present in the jwks, after the first request, subsequent requests will be blocked for 5 mins minimum
* If by chance the keyid has been added to the jwks outside of the refresh cycle, setting RefreshUnknownKID would help

What happens with this PR
* We still refresh the jwks in the refresh cycle
* User can configure refreshing unknown kids, the following is an example with the defaults

```
authentication:
  jwt:
    jwks:
      - refresh_interval: 10m
        url: "https://www.googleapis.com/oauth2/v3/certs"
        refresh_unknown_kid:
          enabled: false # These are defaults
          max_wait: 2m 
          interval: 30s
          burst: 2
```

Also note on how max wait  behaves with 2m and 30 second interval (0.5m = 30s) is the following

```
* 1st response is returned immediately
* 2nd response is returned in 0.5 * 1 minutes
* 3rd response is returned in 0.5 * 2 minutes
* 4th response is returned in 0.5 * 3 minutes
* 5th response is returned in 0.5 * 4 minutes
* 6th response is returned in 0.5 * 5 minutes, exceeds max_wait of 2 minutes, returns immediately
* 7th response is returned in 0.5 * 7 minutes, exceeds max_wait of 2 minutes, returns immediately
```

## Summary by CodeRabbit

* **New Features**
  * Configurable rate limiting for JWKS refreshes on unknown JWT KIDs (enable, interval, burst, max-wait).

* **Configuration**
  * Added refresh_unknown_kid block to JWT JWKS settings in schema and examples.

* **Public API**
  * JWKS configuration now exposes allowed signing algorithms and unknown-KID refresh settings.
  * Test signing helper signature updated to accept an additional flag.

* **Tests**
  * Expanded tests for unknown-KID behavior, timing, bursts, max-wait, and concurrent scenarios.
  * Added test helper to run asynchronous assertions with timeouts.

## Checklist

### DIFF
--- a/studio/CHANGELOG.md
+++ b/studio/CHANGELOG.md
@@ -4,6 +4,12 @@ Binaries are attached to the github release otherwise all images can be found [h
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+# [0.133.0](https://github.com/wundergraph/cosmo/compare/studio@0.132.1...studio@0.133.0) (2025-09-22)
+
+### Features
+
+* fix navigation sometimes not working when switching graph ([#2231](https://github.com/wundergraph/cosmo/issues/2231)) ([77396ff](https://github.com/wundergraph/cosmo/commit/77396ffa63e2b17acca78ef57b71a51dcde78887)) (@wilsonrivera)
+
 ## [0.132.1](https://github.com/wundergraph/cosmo/compare/studio@0.132.0...studio@0.132.1) (2025-09-22)
 
 **Note:** Version bump only for package studio
--- a/studio/package.json
+++ b/studio/package.json
@@ -1,6 +1,6 @@
 {
   "name": "studio",
-  "version": "0.132.1",
+  "version": "0.133.0",
   "private": true,
   "license": "Apache-2.0",
   "description": "WunderGraph Cosmo Studio",
--- a/studio/src/components/dashboard/graph-command-group.tsx
+++ b/studio/src/components/dashboard/graph-command-group.tsx
@@ -67,6 +67,15 @@ interface GraphLinkProps {
   setNamespace(namespace: string): void;
 }
 
+const defaultGraphTemplate = `/[organizationSlug]/[namespace]/graph/[slug]`;
+const graphAreasWithParameters: readonly string[] = [
+  'change-log',
+  'checks',
+  'compositions',
+  'feature-flags',
+  'proposals'
+];
+
 function GraphCommandItem({
   name,
   namespace,
@@ -82,16 +91,22 @@ function GraphCommandItem({
 
   const pathname = useMemo(
     () => {
-      const segment = router.pathname.split('/')[3]?.toLowerCase();
+      const segmentSplit = router.pathname.split('/');
+      const segment = segmentSplit[3]?.toLowerCase();
       if (isSubgraph) {
         return segment === 'subgraph'
           ? router.pathname
           : `/[organizationSlug]/[namespace]/subgraph/[subgraphSlug]`;
       }
 
-      return segment === 'graph'
-        ? router.pathname
-        : `/[organizationSlug]/[namespace]/graph/[slug]`;
+      if (segment !== 'graph') {
+        return defaultGraphTemplate;
+      }
+
+      const areaSegment = segmentSplit[5]?.toLowerCase();
+      return areaSegment && graphAreasWithParameters.includes(areaSegment) && segmentSplit.length > 5
+        ? `${defaultGraphTemplate}/${areaSegment}`
+        : router.pathname;
     },
     [router.pathname, isSubgraph],
   );
--- a/studio/src/components/dashboard/namespace-selector.tsx
+++ b/studio/src/components/dashboard/namespace-selector.tsx
@@ -6,8 +6,7 @@ import { useMemo, useState } from "react";
 import Link from "next/link";
 import { cn } from "@/lib/utils";
 import * as React from "react";
-import { CheckIcon } from "@radix-ui/react-icons";
-import { CaretSortIcon } from "@radix-ui/react-icons";
+import { CheckIcon, CaretSortIcon } from "@radix-ui/react-icons";
 import { docsBaseURL } from "@/lib/constants";
 import { WorkspaceCommandWrapper } from "./workspace-command-wrapper"
 import { useCurrentOrganization } from "@/hooks/use-current-organization";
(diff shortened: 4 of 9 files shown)


router-tests/jwks/jwks.go
```diff
@@ -50,11 +50,19 @@ func (s *Server) Token(claims map[string]any) (string, error) {
...
-	if !ok {
+	if !useInvalidKid && !ok {
 		return "", jwt.ErrInvalidKey
+	} else if useInvalidKid {
+		// If we don't care about the kid we don't care about the provider
+		// we just get the first random provider provided
+		for _, pr := range s.providers {
+			provider = pr
+			break
+		}
```
endigma: This should fail more loudly if len(s.providers) > 1, I don't think randomly selecting a provider from a list is _ever_ desirable and will just make bug reports for us.

router-tests/jwks/jwks.go
```diff
@@ -50,11 +50,19 @@ func (s *Server) Token(claims map[string]any) (string, error) {
 	return "", jwt.ErrInvalidKey
 }
 
-func (s *Server) TokenForKID(kid string, claims map[string]any) (string, error) {
+func (s *Server) TokenForKID(kid string, claims map[string]any, useInvalidKid bool) (string, error) {
```
endigma: `useInvalidKid` -> `allowInvalidKID` ? naming and capitalization for acronyms

router-tests/testenv/testenv.go
```diff
@@ -2532,6 +2532,27 @@ func (e *Environment) WaitForConnectionCount(desiredCount uint64, timeout time.D
...
+	}()
+
+	select {
+	case <-done:
+		return
+	case <-ctx.Done():
+		e.t.Fatalf("test timed out, want %d", timeout)
+		return
+	}
+}
```
endigma: We have like `AwaitChannelWithT` somewhere, maybe we can use that semantic for this instead of being an environment method?

router-tests/jwks/jwks.go
endigma: Also, when is this value set to true outside tests?

SkArchon: What do you mean outside of tests? This is to simulate a test of what happens when there is a valid formed token with an invalid KID. This value is false. when you want to set an invalid KID to the header.

endigma: Ok, can we maybe just have two functions? This is confusing as it's not in a test package but seems to only be used in tests.

router-tests/authentication_test.go
```diff
@@ -828,6 +828,44 @@ func TestHttpJwksAuthorization(t *testing.T) {
...
+			},
+		}, func(t *testing.T, xEnv *testenv.Environment) {
+			// This token has a static keyid
+			token, err := authServer.TokenForKID("wg_static_kid", nil, true)
+			require.NoError(t, err)
+
+			maxDuration := 5 * time.Second
+
+			testenv.AwaitFunc(t, maxDuration, func() {
+				for range 5 {
```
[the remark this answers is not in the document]
SkArchon: in go 1.25 you can

router-tests/jwks/jwks.go
```diff
@@ -50,11 +50,19 @@ func (s *Server) Token(claims map[string]any) (string, error) {
 	return "", jwt.ErrInvalidKey
 }
 
-func (s *Server) TokenForKID(kid string, claims map[string]any) (string, error) {
+func (s *Server) TokenForKID(kid string, claims map[string]any, useInvalidKID bool) (string, error) {
 	provider, ok := s.providers[kid]
-	if !ok {
+	if !useInvalidKID && !ok {
```
devsergiy: will it be more readable like this
```
if useInvalidKey {
} else if !ok {
}
```
double negation hard to understnad

devsergiy: @coderabbitai could you suggest how to simplify it?

SkArchon: Resolved

StarpTech:
Great PR description! I recommend making this configurable. I agree that, by default, we should be pessimistic and abort requests where the KID can't be found. That means only the `refreshInterval` is activated. Additionally, a user can define enabling `refresh_unknown_kid`, including a rate limit (disabled by default), if their case requires them to fetch them proactively. This could make sense to avoid avalanche effects in big router deployments.

Example

```yaml
authentication:
  jwt:
    jwks:
      - refresh_interval: 10m
        url: "https://www.googleapis.com/oauth2/v3/certs"
        refresh_unknown_kid:
          enabled: true # Disabled by default
          max_wait: 10s # RateLimitWaitMax (Something reasonable like 10s)
          interval: 1m # rate.Every (Something reasonable like 1m)
          burst: 10 # burst of rate.Every (Something reasonable like 2)
```

_RateLimitWaitMax_ is an option of `jwkset.HTTPClientOptions`

router/pkg/authentication/jwks_token_decoder.go
```diff
@@ -91,12 +99,18 @@ func NewJwksTokenDecoder(ctx context.Context, logger *zap.Logger, configs []JWKS
 
 			audiencesMap[key] = getAudienceSet(c.Audiences)
 
+			// Configure the rate limiter for refreshing unknown KIDs
+			var refreshLimiter *rate.Limiter
+			if c.RefreshUnknownKID.Enabled {
+				refreshLimiter = rate.NewLimiter(rate.Every(c.RefreshUnknownKID.Interval), c.RefreshUnknownKID.Burst)
+			}
+
```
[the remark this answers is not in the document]
SkArchon: This is guarded from the config schema.

PR #2219, opened 2025-09-16
chore: replace del with rm

miklosbarabas:
- to support portability in Unix-like environments

## Summary by CodeRabbit

* **Chores**
  * Standardized pre-build cleanup across packages for more consistent cross-environment builds.
  * Expanded root cleanup to remove common artifacts (build outputs, caches, generated files) for cleaner installs.
  * Removed an unnecessary development dependency to simplify tooling.
  * No changes to application features or end-user runtime behavior.

## Checklist


endigma:
Should we use [rimraf](https://www.npmjs.com/package/rimraf) for this?

miklosbarabas:
> Should we use [rimraf](https://www.npmjs.com/package/rimraf) for this?

@endigma: I think that would invalidate the the point of removing a dependency just to delete directories

package.json
```diff
@@ -23,7 +23,7 @@
     "generate": "pnpm buf generate --path proto/wg/cosmo/platform --path proto/wg/cosmo/notifications --path proto/wg/cosmo/common --path proto/wg/cosmo/node --template buf.ts.gen.yaml && pnpm -r run --filter './connect' build",
     "husky": "husky install",
     "lint:fix": "pnpm run -r --parallel lint:fix",
-    "clean": "del-cli '**/node_modules/' '**/**/dist/' '**/**/gen/' '**/**/.next' '**/**/tsconfig.tsbuildinfo' '**/**/.eslintcache'",
+    "clean": "rm -rf node_modules **/node_modules dist **/dist gen **/gen .next **/.next tsconfig.tsbuildinfo **/tsconfig.tsbuildinfo .eslintcache **/.eslintcache",
```
endigma, line 26: Do we need both the `x and **/x` variants? I think doublestar matches on `./`

endigma: image

miklosbarabas: It's for compatibility reasons, globbing tends to work differently in different shells.
By the look of your screenshot, you're might be using zsh, but for instance in bash both the `x and **/x` variants would be needed by default if globstar is not enabled.

endigma: Should we use a consistent alternative like `find` then? We wouldn't want the subfolders that are meant to be deleted by the doublestar to be left if the shell isn't behaving

miklosbarabas: using find would require a longer argument, but would essentially achieve the same as the current command. i dont see what you mean by the shell isn't behaving.

StarpTech (approved):
LGTM

codepunkt:
`nitpick`

`del-cli` is still a dependency in `cli`, `composition`, `connect`, `controlplane` and `shared`, but seemingly unused.

StarpTech:
@miklosbarabas https://github.com/wundergraph/cosmo/pull/2219#issuecomment-3478331570

PR #2222, opened 2025-09-16
fix(router): enable safelisting on websocket subscriptions

dkorittki:
## Summary by CodeRabbit

- New Features
  - Persisted operations over WebSocket: pre-parse SHA-256 computation and validation, safelist enforcement, optional logging of unknown persisted operations, rejection of mismatched hashes, cache-poisoning prevention, and client-name propagation.

- Tests
  - Expanded WebSocket tests for safelist/unknown-hash scenarios, logging observation, mismatched-hash rejection, cache-poisoning protection, and client-name propagation.
  - Added subscription test data for end-to-end validation.

## Checklist

# Motivation and Context

Users are confused that safelisting does not work for subscriptions. After some digging I found that it does work but only on HTTP2 streams, not websocket connections. This is because for websockets, we haven't implemented safelisting, although persisted operations are in fact supported. This PR adds safelisting in a similar fashion to how its done for HTTP connections in prehandler, including hash calculation.

Now, when safelisting is enabled and you provide a query whoms hash is unknown, you get this error:

Query with payload whoms hash is unknown
```json
{
  "id": "sub-1",
  "type": "subscribe",
  "payload": {
    "query": "subscription { currentTime { unixTime timeStamp } }"
  }
}
```

Query which references the unknown hash directly
```json
{
  "id": "sub-1",
  "type": "subscribe",
  "payload": {
    "extensions": {
      "persistedQuery": {
        "version": 1,
        "sha256Hash": "9a41d21da2823195ad42c11d51e9ad3345824abdabf567b3615a235843a1fcc7"
      }
    }
  }
}
```

It would also be the same for an operation containing both the query body and hash extension.

Feedback on websocket:
```json
[{"message":"operation '9a41d21da2823195ad42c11d51e9ad3345824abdabf567b3615a235843a1fcc7' for client 'my-client' not found"}]
```

Note: The subscription is still registered, even if we faced this error. That's due to the [error handling of websocket parsing](https://github.com/wundergraph/cosmo/blob/9a55be87e97c2f130cb783873f069eb5bda9ceb1/router/core/websocket.go#L935-L941). Not sure if thats a good or bad thing tbh.

I also added support for `persisted_operations.log_unknown` in a similar fashion to the HTTP prehandler: log it bu if safelisting is disabled let it pass.

For safelisting to work I had to add hash calculation in front of it, similar to how its done in the HTTP prehandler.
Alongside that I added a verification of provided hash vs calculated hash. **Thats actually kind of a breaking change**.
No matter the safelist settings, these queries will now get blocked (look at `sha256Hash`). Before this PR it would have passed.

```
curl -X POST http://localhost:3002/graphql \
  -H "Content-Type: application/json" \
  -d '{
    "query": "subscription { currentTime { unixTime timeStamp } }",
    "extensions": {
      "persistedQuery": {
        "version": 1,
        "sha256Hash": "1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef"
      }
    }
  }'
```


router/core/websocket.go
```diff
@@ -836,11 +837,40 @@ func (h *WebSocketConnectionHandler) parseAndPlan(registration *SubscriptionRegi
...
 			return nil, nil, err
 		}
+
+		if h.operationBlocker.safelistEnabled || h.operationBlocker.logUnknownOperationsEnabled {
+			// Set the request hash to the parsed hash, to see if it matches a persisted operation
+			operationKit.parsedOperation.GraphQLRequestExtensions.PersistedQuery = &GraphQLRequestExtensionsPersistedQuery{
+				Sha256Hash: operationKit.parsedOperation.Sha256Hash,
+			}
+		}
+	}
```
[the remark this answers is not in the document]
dkorittki: This is actually true and I found that out during development. This is basically a copy of how its done in the PreHandler, so we face the same issue there. That is, if it even is an issue. Maybe it was a conscious decision. Would be great if human reviewer could provide some feedback here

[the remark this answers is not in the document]
endigma: @dkorittki you can move it to validate before generating a new one, the validation is to prevent poisoning attacks, it's irrelevant if we add our own hash.

dkorittki:
@endigma can have a look again? implemented your suggestions

router-tests/websocket_test.go
```diff
@@ -2069,6 +2070,238 @@ func TestWebSockets(t *testing.T) {
...
+				Payload: []byte(`{"query":"subscription { employeeUpdated(employeeID: 1) { id } }"}`),
+			})
+			require.NoError(t, err)
+			var res testenv.WebSocketMessage
+			err = testenv.WSReadJSON(t, conn, &res)
+			require.NoError(t, err)
+			require.Equal(t, "error", res.Type)
+			require.Equal(t, "1", res.ID)
+			require.JSONEq(t, `[{"message":"operation '9a41d21da2823195ad42c11d51e9ad3345824abdabf567b3615a235843a1fcc7' for client 'my-client' not found"}]`,
+				string(res.Payload))
```
endigma, line 2097: JSONEq works on bytes I believe

router-tests/websocket_test.go
```diff
@@ -2069,6 +2070,238 @@ func TestWebSockets(t *testing.T) {
...
+					"query": "%s",
+					"extensions": {
+						"persistedQuery": {
+							"version": 1,
+							"sha256Hash": "%s"
+						}
+					}
+				}`, query2, hashOfQuery2)),
+			})
+			require.NoError(t, err)
```
endigma, line 2287: This should send _only_ the hash, not both. The goal is to make sure none of the caches were poisoned which is subverted in some places by the query body being present.

endigma (changes requested):

router-tests/websocket_test.go
dkorittki: Just checked, its string

endigma: No I mean I think it can handle just bytes as the second parameter? Not sure though and it's a nit at best

router-tests/websocket_test.go
dkorittki: fixed

router-tests/websocket_test.go
dkorittki: Yeah I checked the function signature of JSONEq and it requires string arguments for `expected` and `actual`

```go
func JSONEq(t TestingT, expected string, actual string, msgAndArgs ...interface{})
```

router-tests/websocket_test.go
```diff
@@ -2069,6 +2070,236 @@ func TestWebSockets(t *testing.T) {
...
+			require.Equal(t, "2", res2.ID)
+
+			// we expect the response to look like what we asked for via query2
+			require.Contains(t, string(res2.Payload), "currentTime")
+			require.Contains(t, string(res2.Payload), "unixTime")
+
+			require.NoError(t, conn.Close())
+		})
+	})
+
```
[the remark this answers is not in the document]
endigma: The project uses go1.25, it cannot be built with toolchains lower than 1.23.

PR #2234, opened 2025-09-22
chore: router ci improvement

miklosbarabas:
- improved redis usage from the ci
- renamed TestNatsEvents to TestFlakyNatsEvents to be retried on failure
- commented out suspected faulty test

@coderabbitai summary

## Checklist

### DIFF
--- a/.github/workflows/router-ci.yaml
+++ b/.github/workflows/router-ci.yaml
@@ -187,11 +187,50 @@ jobs:
           --health-interval 10s
           --health-timeout 5s
           --health-retries 5
-        credentials:
-          username: ${{secrets.DOCKER_USERNAME}}
-          password: ${{secrets.DOCKER_PASSWORD}}
         ports:
           - 6379:6379
+      redis-0:
+        image: bitnamilegacy/redis-cluster:7.2
+        options: >-
+          --health-cmd "redis-cli ping"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+        env:
+          ALLOW_EMPTY_PASSWORD: "yes"
+          REDIS_NODES: "redis-0 redis-1 redis-2"
+          REDIS_CLUSTER_REPLICAS: "0"
+          REDIS_CLUSTER_CREATOR: "no"
+        ports:
+          - 7001:6379
+      redis-1:
+        image: bitnamilegacy/redis-cluster:7.2
+        options: >-
+          --health-cmd "redis-cli ping"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+        env:
+          ALLOW_EMPTY_PASSWORD: "yes"
+          REDIS_NODES: "redis-0 redis-1 redis-2"
+          REDIS_CLUSTER_REPLICAS: "0"
+          REDIS_CLUSTER_CREATOR: "no"
+        ports:
+          - 7002:6379
+      redis-2:
+        options: >-
+          --health-cmd "redis-cli ping"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+        image: bitnamilegacy/redis-cluster:7.2
+        env:
+          ALLOW_EMPTY_PASSWORD: "yes"
+          REDIS_NODES: "redis-0 redis-1 redis-2"
+          REDIS_CLUSTER_REPLICAS: "0"
+          REDIS_CLUSTER_CREATOR: "yes"
+        ports:
+          - 7003:6379
       kafka:
         image: bitnamilegacy/kafka:3.7.0
         options: >-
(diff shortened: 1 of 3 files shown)


.github/workflows/router-ci.yaml
```diff
@@ -187,11 +187,50 @@ jobs:
           --health-interval 10s
           --health-timeout 5s
           --health-retries 5
-        credentials:
-          username: ${{secrets.DOCKER_USERNAME}}
-          password: ${{secrets.DOCKER_PASSWORD}}
```
endigma, line 192: Isn't this important to avoid rate limiting from dockerhub?

.github/workflows/router-ci.yaml
```diff
@@ -187,11 +187,50 @@ jobs:
...
+          --health-timeout 5s
+          --health-retries 5
+        image: bitnamilegacy/redis-cluster:7.2
+        env:
+          ALLOW_EMPTY_PASSWORD: "yes"
+          REDIS_NODES: "redis-0 redis-1 redis-2"
+          REDIS_CLUSTER_REPLICAS: "0"
+          REDIS_CLUSTER_CREATOR: "yes"
+        ports:
+          - 7003:6379
```
endigma, line 233: `image` before `options` would make the order consistent on all 3 services.

router-tests/Makefile
```diff
@@ -11,7 +12,7 @@ test-deps:
 	$(MAKE) -C ../demo plugin-build-ci
 
 test: test-deps
-	gotestsum -f $(FORMAT) -- -ldflags=-extldflags=-Wl,-ld_classic $(test_params) -race $(test_target)
+	gotestsum --rerun-fails="$(test_retry_count)" --packages="$(test_target)" -f $(FORMAT) -- -ldflags=-extldflags=-Wl,-ld_classic $(test_params) -race $(test_target)
```
endigma, line 15: The retry count should be disabled for non-flagged-flaky tests, is that possible?

miklosbarabas: @endigma: possibly, but didnt take the effort yet to separate them out, as the main goal for flaky tests would be to resolve the root cause of the flakiness. what benefit do you see from splitting them? happy to tackle it if we think it's useful and possible

.github/workflows/router-ci.yaml
miklosbarabas: I don't see how would this be needed for the container which doesn't pull anything from docker.

.github/workflows/router-ci.yaml
miklosbarabas: fixed

router-tests/Makefile
endigma: Because if we have retry and a new flaky test or a test _becomes_ flaky it can fly under the radar

.github/workflows/router-ci.yaml
endigma: `redis:7` is from https://hub.docker.com/_/redis ?

miklosbarabas: Yes, AFAIK it would be pulled from Docker Hub by default, if not specified otherwise explicitly.

router-tests/Makefile
miklosbarabas: Fixed

router-tests/events/nats_events_test.go
```diff
@@ -62,7 +62,7 @@ type natsSubscriptionArgs struct {
 	errValue  error
 }
 
-func TestNatsEvents(t *testing.T) {
+func TestFlakyNatsEvents(t *testing.T) {
```
SkArchon: I think you might need to do the same for kafka and redis tests. Also Ale refactored these tests a while back so he might have some input @alepane21

ysmolski (approved):

router-tests/events/nats_events_test.go
ysmolski: @miklosbarabas I am not sure about this though. There are two specific tests that fail on the first run and almost never on 2nd, 3rd run when there are no other parallel tests running. This all related to performance of tests and logic there. Are we okay marking them with Flaky, so they never will be properly fixed?

miklosbarabas: @SkArchon: thanks for highlighting it 
@alepane21: what's your take, is there any other known flaky test that we should flag now?

miklosbarabas: @endigma could you re-review please?

endigma (approved):
LGTM

router-tests/batch_test.go
```diff
@@ -24,7 +24,7 @@ import (
 	"go.opentelemetry.io/otel/trace"
 )
 
-func TestBatch(t *testing.T) {
+func TestFlakyBatch(t *testing.T) {
```
alepane21: All tests here are flaky? If not, I’ll move only the flaky ones into this function and leave the others in the non-flaky function. The goal is to eliminate flaky functions altogether one day.

router-tests/events/nats_events_test.go
alepane21: I would set as flaky only `TestNatsEvents/subscribe_sse_with_filter`

alepane21 (changes requested):

router-tests/batch_test.go
miklosbarabas: ill rename it back and have couple of runs to see if it's one specific

router-tests/events/nats_events_test.go
miklosbarabas: > I would set as flaky only `TestNatsEvents/subscribe_sse_with_filter`

@alepane21 : 👍  done

router-tests/batch_test.go
miklosbarabas: @alepane21 ive extracted and flagged only the specific flaky tests now, as you asked for. Could you re-review?

miklosbarabas: done, please check again

PR #2235, opened 2025-09-24
feat: retrieve operation body from gqlmetrics instead of traces

endigma:
## Summary by CodeRabbit

- **New Features**
  - Added a tracing option to include original and normalized GraphQL operation content as span attributes (disabled by default; configurable via telemetry/tracing). New attribute keys exposed for richer trace metadata.

- **Studio**
  - Trace Details: separately fetches operation content, adds dedicated loading/error states, and shows side-by-side viewers for operation and variables with a “Play in Playground” link and validation/truncation handling.

- **Tests**
  - Added telemetry tests validating presence/absence of operation content attributes when the new tracing flag is toggled.

- **Chores**
  - Collector pipeline updated to strip operation body attributes by default; config/schema and defaults updated to surface the new flag.

## Checklist


otelcollector/otel-config.yaml
```diff
@@ -19,6 +19,12 @@ processors:
       - key: wg.federated_graph.id
         from_context: auth.federated_graph_id
         action: insert
+  attributes/delete_operation_bodies:
+    actions:
+      - key: wg.operation.normalized_content
+        action: delete
+      - key: wg.operation.original_content
+        action: delete
```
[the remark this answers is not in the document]
endigma: This is intentional, the flag is false by default, and the attributes are only for use with other OTEL endpoints.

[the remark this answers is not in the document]
endigma: This otel collector goes to the studio which does not display this attribute, if the end user reconfigures the tracing to go to Tempo, Jaeger, etc then the attribute can be used.

JivusAyrus (approved):

router/core/graphql_prehandler.go
```diff
@@ -520,14 +526,15 @@ func (h *PreHandler) handleOperation(req *http.Request, variablesParser *astjson
 	requestContext := httpOperation.requestContext
 
 	// Handle the case when operation information are provided as GET parameters
-	if req.Method == http.MethodGet {
+	switch req.Method {
+	case http.MethodGet:
```
Noroth, line 530: Are we even able to get anything else than GET or POST at this stage?

router/pkg/config/config.go
```diff
@@ -79,6 +79,8 @@ type Tracing struct {
 	ResponseTraceHeader ResponseTraceHeader `yaml:"response_trace_id"`
 	Attributes          []CustomAttribute   `yaml:"attributes"`
 
+	EnableOperationBodyAttributes bool `yaml:"enable_operation_body_attributes" envDefault:"false" env:"TRACING_ENABLE_OPERATION_BODY_ATTRIBUTES"`
```
Noroth: I personally don't like `Enable`. I would rather call it `Append`. 

To me it sounds better when you have `tacing->append attributes` instead of `enable attributes` but up to you

router/debug.config.yaml
```diff
@@ -20,6 +20,24 @@ watch_config:
...
+        endpoint: http://localhost:4318
+        headers:
+          Authorization: Bearer ${GRAPH_API_TOKEN}
+        batch_timeout: '10s'
+        export_timeout: '30s'
+      - exporter: http
+        endpoint: http://localhost:4319
+        batch_timeout: '10s'
+        export_timeout: '30s'
+
```
Noroth, line 40: Why would you specifically enable tracing but not metrics when adding defaults to the debug config?

endigma: Metrics are enabled by default and work with Grafana/Studio fine, tracing needs to be explicitly configured to go to both studio and jaeger, however this could be removed entirely

router/core/graphql_prehandler.go
endigma: This is the same behavior as before, just a staticcheck linter fix

wilsonrivera (approved):

router/pkg/config/config.go
endigma: I think enable is more suggestive of its default state and the fact that it's mostly for debugging, not something you'd want enabled all the time.

Noroth: I disagree with that. The naming confuses me when I just read the key.

Noroth (approved):

router/pkg/config/config.schema.json
```diff
@@ -857,6 +857,11 @@
               "default": false,
               "description": "Specifies that the router span should be treated as a root Span. Any existing parent span context will be ignored when defining the Span's trace identifiers. This is useful when the router must be the entry point of the trace."
             },
+            "enable_operation_body_attributes": {
```
StarpTech: Can we come up with a more concrete name to describe this behavior? We also never used `Enable` in that case. Maybe we should stick with the current convention.

studio/src/components/analytics/trace-details.tsx
```diff
@@ -98,107 +121,114 @@ export const TraceDetails = ({ ast }: { ast: GraphQLSchema | null }) => {
...
+                      <TooltipTrigger asChild>
+                        <Button variant="outline" size="icon" className="relative" asChild>
+                          <Link
+                            href={`/${organizationSlug}/${namespace}/graph/${slug}/playground?operation=${encodeURIComponent(
+                              content || '',
+                            )}&variables=${encodeURIComponent(
+                              variables || JSON.stringify(extractVariablesFromGraphQL(content, ast)),
+                            )}`}
+                          >
+                            <PlayIcon className="h-5" />
```
StarpTech, line 201: Do you have a custom formatter, or did we forget to format?

endigma: I'm just using prettier, and this isn't a reformat it's been indented by one to go inside an error boundary

router/pkg/config/config.schema.json
endigma: How about `IncludeOperationContentAttributes`?

StarpTech: I was referring more to whether "OperationBodyAttributes" is meaningful enough. Does people understand that this will attach the operation to the normalization and parsing span? I also would imply "adding" something by just setting it to true, hence `OperationContentAttributes`

endigma: I think it's niche enough an option that we could call it just OperationContentAttributes, the docs for it will have more keywords for search, it's a tough concept to explain in few enough words for a config option

StarpTech: ok

PR #2236, opened 2025-09-25
feat: view query plan as a json

SkArchon:
**What happens now**
Right now we can view query plan from the plan generator in the router using a textual pretty printed format

**With this Pr**
This PR introduces a "format" attribute, which is "text" by default, upon being set to "json", this would cause query plans to be generated as a raw json. This is NOT similar to the apollo jsons, to make it similar would require separate work.

## Summary by CodeRabbit

- New Features
  - CLI adds --print-format to choose plan output as text or json (case-insensitive). Defaults to text, validates input and errors on invalid values. Plan generation now emits either human-readable text or JSON.

- Tests
  - Expanded tests cover both JSON and text outputs, validate per-file outputs and parsing, and improve temporary directory cleanup and test isolation.

## Checklist

### DIFF
--- a/router/cmd/plan_generator.go
+++ b/router/cmd/plan_generator.go
@@ -38,8 +38,8 @@ func PlanGenerator(args []string) {
 	f.BoolVar(&cfg.OutputReport, "print-report", false, "write a report.json file, with all the query plans and errors sorted by file name")
 	f.BoolVar(&cfg.FailOnPlanError, "fail-on-error", false, "if at least one plan fails, the command exit code will be 1")
 	f.BoolVar(&cfg.FailFast, "fail-fast", false, "stop as soon as possible if a plan fails")
-	f.StringVar(&cfg.LogLevel, "log-level", "warn", "log level to use (debug, info, warn, error, panic, fatal)")
-	f.Func("print-format", "output format (text|json)", func(s string) error {
+	f.StringVar(&cfg.LogLevel, "log-level", "warn", "the log level to use (debug, info, warn, error, panic, fatal)")
+	f.Func("print-format", "output format of the query plan (text, json) (default \"text\")", func(s string) error {
 		switch value := core.PlanOutputFormat(strings.ToLower(s)); value {
 		case core.PlanOutputFormatUnset:
 			cfg.OutputFormat = core.PlanOutputFormatText


router/cmd/plan_generator.go
```diff
@@ -36,6 +36,7 @@ func PlanGenerator(args []string) {
 	f.BoolVar(&cfg.FailOnPlanError, "fail-on-error", false, "if at least one plan fails, the command exit code will be 1")
 	f.BoolVar(&cfg.FailFast, "fail-fast", false, "stop as soon as possible if a plan fails")
 	f.StringVar(&cfg.LogLevel, "log-level", "warn", "log level to use (debug, info, warn, error, panic, fatal)")
+	f.BoolVar(&cfg.Raw, "raw", false, "get the raw json output of the plan generator")
```
devsergiy: maybe name it `json`?

SkArchon:
converting to draft again for a few changes

router/pkg/plan_generator/plan_generator_test.go
```diff
@@ -331,4 +331,107 @@ func TestPlanGenerator(t *testing.T) {
 		assert.NoError(t, err)
 		assert.Equal(t, errMsg, writtenResults.Error)
 	})
+
+	t.Run("generates raw json plans when Raw is enabled", func(t *testing.T) {
+		tempDir, err := os.MkdirTemp("", "plans-")
```
devsergiy: maybe use fixtures instead of temp files?

SkArchon: was following the same pattern for existing tests, will take a look

router/cmd/plan_generator.go
endigma: I would suggest doing something like

-o --output plain (or pretty, etc) | json

So we don't end up boxing ourselves into having two output formats or complex flags

endigma (changes requested):

router/cmd/plan_generator.go
SkArchon: makes sense

router/pkg/plan_generator/plan_generator_test.go
endigma: Also, if you do use a tempdir, you can use the function on `t` to avoid having to clean it up manually

router/cmd/plan_generator.go
devsergiy: maybe -f --format? output more like location

devsergiy (approved):

router/pkg/plan_generator/plan_generator_test.go
```diff
@@ -33,7 +34,9 @@ func TestPlanGenerator(t *testing.T) {
 	t.Run("checks queries path exists", func(t *testing.T) {
 		tempDir, err := os.MkdirTemp("", "plans-")
 		require.NoError(t, err)
-		defer os.RemoveAll(tempDir)
+		defer func() {
+			_ = os.RemoveAll(tempDir)
+		}()
```
endigma: You can use `t.TempDir` to avoid this boilerplate

router/core/plan_generator.go
```diff
@@ -53,6 +53,13 @@ type Planner struct {
 	operationValidator *astvalidation.OperationValidator
 }
 
+type PlanOutputFormat string
+
+const (
+	PlanOutputFormatText PlanOutputFormat = "text"
+	PlanOutputFormatJSON PlanOutputFormat = "json"
+)
```
endigma, line 62: It's good in cases like this to make a `PlanOutputFormatUnset/Invali PlanOutputFormat = ""` to handle the zero value for the string gracefully

router/pkg/plan_generator/plan_generator_test.go
```diff
@@ -227,7 +244,9 @@ func TestPlanGenerator(t *testing.T) {
 	t.Run("will not fail on warnings and results should return the warnings and generate results file", func(t *testing.T) {
 		tempDir, err := os.MkdirTemp("", "plans-")
 		require.NoError(t, err)
-		defer os.RemoveAll(tempDir)
+		t.Cleanup(func() {
+			_ = os.RemoveAll(tempDir)
+		})
```
endigma: `t.TempDir`

router/cmd/plan_generator.go
```diff
@@ -36,6 +39,17 @@ func PlanGenerator(args []string) {
 	f.BoolVar(&cfg.FailOnPlanError, "fail-on-error", false, "if at least one plan fails, the command exit code will be 1")
 	f.BoolVar(&cfg.FailFast, "fail-fast", false, "stop as soon as possible if a plan fails")
 	f.StringVar(&cfg.LogLevel, "log-level", "warn", "log level to use (debug, info, warn, error, panic, fatal)")
+	f.Func("print-format", "output format (text|json)", func(s string) error {
+		switch value := core.PlanOutputFormat(strings.ToLower(s)); value {
+		case "":
```
endigma: Please add a named constant for `""`

endigma (changes requested):

router/cmd/plan_generator.go
```diff
@@ -36,6 +39,17 @@ func PlanGenerator(args []string) {
 	f.BoolVar(&cfg.FailOnPlanError, "fail-on-error", false, "if at least one plan fails, the command exit code will be 1")
 	f.BoolVar(&cfg.FailFast, "fail-fast", false, "stop as soon as possible if a plan fails")
 	f.StringVar(&cfg.LogLevel, "log-level", "warn", "log level to use (debug, info, warn, error, panic, fatal)")
+	f.Func("print-format", "output format (text|json)", func(s string) error {
```
endigma: Nit: inconsistent helptext

PR #2247, opened 2025-09-30
fix(router): Use proper OTEL span type for `Authenticate` middleware step

Blacksmoke16:
@coderabbitai summary

## Checklist

Follow up to https://github.com/wundergraph/cosmo/pull/456#discussion_r2389253235. It was unexpected for the `Authenticate` middleware step to have a [span type](https://opentelemetry.io/docs/concepts/signals/traces/#client) of `server` due to this step not representing an incoming remote call. In fact since authentication results in an outgoing HTTP request, it should in fact be `client`. It cannot be `internal` because it's crossing a process boundary (the HTTP request).

### DIFF
--- a/router/core/graphql_prehandler.go
+++ b/router/core/graphql_prehandler.go
@@ -356,7 +356,7 @@ func (h *PreHandler) Handler(next http.Handler) http.Handler {
 		// If we have authenticators, we try to authenticate the request
 		if h.accessController != nil {
 			_, authenticateSpan := h.tracer.Start(r.Context(), "Authenticate",
-				trace.WithSpanKind(trace.SpanKindServer),
+				trace.WithSpanKind(trace.SpanKindClient),
 				trace.WithAttributes(requestContext.telemetry.traceAttrs...),
 			)
 


Blacksmoke16:
Not stale, just pending a review :slightly_smiling_face:

Blacksmoke16:
@SkArchon bump

Aenimus:
@Blacksmoke16 Hi, sorry for the delay.

I think at minimum we will need some tests to prove that we don't make assumptions on the kind in the control plane.

Could you add some relevant tests, please?

The WunderGraph Team

CC: @SkArchon @miklosbarabas @StarpTech

Blacksmoke16:
@Aenimus I guess one follow-up question: how would you actually write a test for that? If there was a prior assumption I'd have expected some existing test(s) to fail. I can of course write a test that asserts this span's kind is what we expect and that it doesn't affect root span detection or anything tho.

The UI just exposes whatever the kind is cosmetically, and ClickHouse views are based on attributes so no change in behavior there either.

StarpTech:
Hi @Blacksmoke16, please look into https://github.com/wundergraph/cosmo/blob/d6a443e448470dac7391a4e68210d0fcdd95741d/router-tests/telemetry/telemetry_test.go#L9231. There, we test authentication and verify telemetry; however, we never checked for the Span kind. You can choose add the assertion there.

PR #2266, opened 2025-10-08
fix: improve access logs

SkArchon:
This PR:
* Introduces a `add_stacktrace` flag for `access_logs` which is `true` by default, which when enable would allow any access logs with error to print a stacktrace along with it
* Introduces a `level` flag for `access_logs` which is `info` by default, when set to `warn` for example, any access logs above `warn` wonted be printed

Note:
We only previously introduced stack traces for panics, we keep that behaviour even if stack traces are disabled.

## Summary by CodeRabbit

* **New Features**
  * Access logs: configurable log level (debug/info/warn/error/panic/fatal) and optional stack traces per config.
  * Per-request log level handler so requests and subgraph entries are logged at appropriate levels.
  * Subgraph access logger supports error-level entries.

* **Bug Fixes**
  * Errors now logged at error level instead of info.
  * Broken-pipe errors omit stack traces and include a broken_pipe indicator.

* **Tests / Docs**
  * Expanded tests and updated config/schema defaults for access_logs (level, add_stacktrace).

## Checklist

### DIFF
--- a/router/core/supervisor_instance.go
+++ b/router/core/supervisor_instance.go
@@ -75,12 +75,12 @@ func newRouter(ctx context.Context, params RouterResources, additionalOptions ..
 		}
 
 		var level zapcore.Level
-		accessLogLevel := strings.ToUpper(cfg.AccessLogs.Level)
-		if accessLogLevel == "" {
-			accessLogLevel = zapcore.InfoLevel.CapitalString()
-		}
-		if err := level.Set(accessLogLevel); err != nil {
-			return nil, fmt.Errorf("could not parse log level: %w for access logs", err)
+		if cfg.AccessLogs.Level == "" {
+			level = zapcore.InfoLevel
+		} else {
+			if err := level.Set(strings.ToUpper(cfg.AccessLogs.Level)); err != nil {
+				return nil, fmt.Errorf("could not parse log level: %w for access logs", err)
+			}
 		}
 
 		if cfg.AccessLogs.Output.File.Enabled {


router/pkg/config/config.go
```diff
@@ -869,11 +869,13 @@ type AutomaticPersistedQueriesConfig struct {
...
 
 type AccessLogsConfig struct {
-	Enabled   bool                      `yaml:"enabled" env:"ACCESS_LOGS_ENABLED" envDefault:"true"`
-	Buffer    AccessLogsBufferConfig    `yaml:"buffer,omitempty" env:"ACCESS_LOGS_BUFFER"`
-	Output    AccessLogsOutputConfig    `yaml:"output,omitempty" env:"ACCESS_LOGS_OUTPUT"`
-	Router    AccessLogsRouterConfig    `yaml:"router,omitempty" env:"ACCESS_LOGS_ROUTER"`
-	Subgraphs AccessLogsSubgraphsConfig `yaml:"subgraphs,omitempty" env:"ACCESS_LOGS_SUBGRAPH"`
+	Enabled       bool                      `yaml:"enabled" env:"ACCESS_LOGS_ENABLED" envDefault:"true"`
+	Level         string                    `yaml:"level" env:"ACCESS_LOGS_LEVEL" envDefault:"info"`
+	AddStacktrace bool                      `yaml:"add_stacktrace" env:"ACCESS_LOGS_ADD_STACKTRACE" envDefault:"true"`
```
devsergiy: should it be enabled by default? or is it only for errors?

router/pkg/config/config.schema.json
```diff
@@ -624,6 +624,17 @@
...
         },
+        "level": {
+          "type": "string",
+          "enum": ["debug", "info", "warn", "error", "panic", "fatal"],
+          "description": "The log level for access logs. The log level is used to control the verbosity of the access logs. The default value is 'info'.",
+          "default": "info"
+        },
+        "add_stacktrace": {
+          "type": "boolean",
+          "description": "Add stack traces to access logs. When enabled, stack traces will be included in access log entries. The default value is true.",
```
devsergiy: should we mention that this is only for error log level?

router/pkg/config/config.go
SkArchon: Its false now

router/pkg/config/config.schema.json
SkArchon: Updated

PR #2272, opened 2025-10-10
fix: support requireFetchReasons on interface fields

ysmolski:
When an interface field is marked,
and a user requests that field in an operation, 
then fetch reasons for this field will be extended 
with fetch reasons for all the implementing types.

In general, a marked interface field leads to all 
implementation's fields being used for propagation.

I have removed special config needed for testing fetch reasons.
Instead I added those directives to the demo package.

Bumped the shoenig/go-m1cpu package to get rid of warnings from xcode 26.

## Summary by CodeRabbit

- New Features
  - Added a repeatable GraphQL directive (openfed__requireFetchReasons) usable on fields, interfaces, and objects.
  - Router now includes fetch_reasons data in response extensions when the directive is applied (e.g., on employee and Query fields).

- Tests
  - Expanded integration tests to verify rendered subgraph queries and the presence of fetch_reasons in extensions.

- Chores
  - Upgraded dependencies, including graphql-go-tools to v2.0.0-rc.230 and go-m1cpu to v0.1.7.


router-tests/testenv/testenv.go
```diff
@@ -91,8 +91,6 @@ var (
 	ConfigWithPluginsJSONTemplate string
 	//go:embed testdata/configWithGRPC.json
 	ConfigWithGRPCJSONTemplate string
-	//go:embed testdata/configWithRequireFetchReasons.json
-	ConfigWithRequireFetchReasonsJSONTemplate string
```
Aenimus, line 95: @ysmolski you do intend to remove this?

ysmolski: Yeah, I have removed the file completely. I added directives to the original config.json instead

PR #2273, opened 2025-10-10
feat(router): implement Cosmo Streams v1

dkorittki:
## Summary by CodeRabbit

* **New Features**
  * Cosmo Streams v1: hook-driven subscription lifecycle (on-start, inbound receive, outbound publish), provider-agnostic event model, mutable/immutable event support, and multi-event publish across NATS/Kafka/Redis.
* **Configuration**
  * New handlers settings to tune receive concurrency and handler timeouts; runtime option to supply streams handler configuration.
* **Bug Fixes**
  * Improved streams error mapping and subscription error propagation to GraphQL responses.
* **Tests**
  * Extensive unit & integration coverage for hooks, cross-provider flows, concurrency, timeouts, error and panic recovery.
* **Documentation**
  * ADR/RFC and examples for Streams v1, hook usage and event transformation.

## Checklist

# Context

This pull request implements Cosmo Streams v1. It rebrands and extends EDFS by the ability to add hooks in form of new custom modules:

- `SubscriptionOnStart`: A handler running before a subscription is established
- `OnReceiveEvents`: A handler running whenever events are received from a message provider, before being passed to clients
- `OnPublishEvents`: A handler running before events are getting sent to message providers

These custom modules extend EDFS´s capabilities by allowing custom authorization logic, event filtering, etc.

The preview docs explains this in more detail. It's a good place to see what the hooks are capable of and how to use them.
https://wundergraphinc-topics-streams-v1.mintlify.app/overview

# Related pull requests

There are two other pull requests, which contribute to the initial release of Cosmo Streams.

Cosmo Docs: https://github.com/wundergraph/cosmo-docs/pull/187
Engine: https://github.com/wundergraph/graphql-go-tools/pull/1309

> [!IMPORTANT]  
> This pull request must not be merged before the engine pull request has been merged. The router´s go.mod in this pull request currently points to a version of the engine from the Cosmo Streams branch and once we have an official version of the engine with these changes included, this pull requests gets updated to use that.

### DIFF
--- a/controlplane/CHANGELOG.md
+++ b/controlplane/CHANGELOG.md
@@ -4,6 +4,12 @@ Binaries are attached to the github release otherwise all images can be found [h
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+# [0.171.0](https://github.com/wundergraph/cosmo/compare/controlplane@0.170.2...controlplane@0.171.0) (2025-11-22)
+
+### Features
+
+* improve SCIM implementation ([#2351](https://github.com/wundergraph/cosmo/issues/2351)) ([cafd1fc](https://github.com/wundergraph/cosmo/commit/cafd1fc7e1a6857a7abb4d1632b5f42e54f09cf2)) (@wilsonrivera)
+
 ## [0.170.2](https://github.com/wundergraph/cosmo/compare/controlplane@0.170.1...controlplane@0.170.2) (2025-11-21)
 
 **Note:** Version bump only for package controlplane
--- a/controlplane/migrations/0133_foamy_vargas.sql
+++ b/controlplane/migrations/0133_foamy_vargas.sql
@@ -1,0 +1,2 @@
+ALTER TABLE "organization_invitations" ADD COLUMN "last_sent_at" timestamp with time zone;--> statement-breakpoint
+ALTER TABLE "organization_members" ADD COLUMN "active" boolean DEFAULT true NOT NULL;
--- a/controlplane/migrations/meta/0133_snapshot.json
+++ b/controlplane/migrations/meta/0133_snapshot.json
@@ -1,0 +1,8651 @@
+{
+  "id": "781ded1a-9a7d-4d9e-b443-51651c9f154e",
+  "prevId": "3904e4bd-f2e7-4364-a807-c3000b7d0f56",
+  "version": "7",
+  "dialect": "postgresql",
+  "tables": {
+    "public.api_key_permissions": {
+      "name": "api_key_permissions",
+      "schema": "",
+      "columns": {
+        "id": {
+          "name": "id",
+          "type": "uuid",
+          "primaryKey": true,
+          "notNull": true,
+          "default": "gen_random_uuid()"
+        },
+        "api_key_id": {
+          "name": "api_key_id",
+          "type": "uuid",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "permission": {
+          "name": "permission",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        }
+      },
+      "indexes": {
+        "akp_api_key_id_idx": {
+          "name": "akp_api_key_id_idx",
+          "columns": [
+            {
+              "expression": "api_key_id",
+              "isExpression": false,
+              "asc": true,
+              "nulls": "last"
+            }
+          ],
+          "isUnique": false,
+          "concurrently": false,
+          "method": "btree",
+          "with": {}
+        }
+      },
+      "foreignKeys": {
+        "api_key_permissions_api_key_id_api_keys_id_fk": {
+          "name": "api_key_permissions_api_key_id_api_keys_id_fk",
+          "tableFrom": "api_key_permissions",
+          "tableTo": "api_keys",
+          "columnsFrom": [
+            "api_key_id"
+          ],
+          "columnsTo": [
+            "id"
+          ],
+          "onDelete": "cascade",
+          "onUpdate": "no action"
+        }
+      },
+      "compositePrimaryKeys": {},
+      "uniqueConstraints": {},
+      "checkConstraints": {}
+    },
+    "public.api_key_resources": {
+      "name": "api_key_resources",
+      "schema": "",
+      "columns": {
+        "id": {
+          "name": "id",
+          "type": "uuid",
+          "primaryKey": true,
+          "notNull": true,
+          "default": "gen_random_uuid()"
+        },
+        "api_key_id": {
+          "name": "api_key_id",
+          "type": "uuid",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "target_id": {
+          "name": "target_id",
+          "type": "uuid",
+          "primaryKey": false,
+          "notNull": false
+        }
+      },
+      "indexes": {
+        "akr_api_key_id_idx": {
+          "name": "akr_api_key_id_idx",
+          "columns": [
+            {
+              "expression": "api_key_id",
+              "isExpression": false,
+              "asc": true,
+              "nulls": "last"
+            }
+          ],
+          "isUnique": false,
+          "concurrently": false,
+          "method": "btree",
+          "with": {}
+        },
+        "akr_target_id_idx": {
+          "name": "akr_target_id_idx",
+          "columns": [
+            {
+              "expression": "target_id",
+              "isExpression": false,
+              "asc": true,
+              "nulls": "last"
+            }
+          ],
+          "isUnique": false,
+          "concurrently": false,
+          "method": "btree",
+          "with": {}
+        }
+      },
+      "foreignKeys": {
+        "api_key_resources_api_key_id_api_keys_id_fk": {
+          "name": "api_key_resources_api_key_id_api_keys_id_fk",
+          "tableFrom": "api_key_resources",
+          "tableTo": "api_keys",
+          "columnsFrom": [
+            "api_key_id"
+          ],
+          "columnsTo": [
+            "id"
+          ],
+          "onDelete": "cascade",
+          "onUpdate": "no action"
+        },
+        "api_key_resources_target_id_targets_id_fk": {
+          "name": "api_key_resources_target_id_targets_id_fk",
+          "tableFrom": "api_key_resources",
+          "tableTo": "targets",
+          "columnsFrom": [
+            "target_id"
+          ],
+          "columnsTo": [
+            "id"
+          ],
+          "onDelete": "set null",
+          "onUpdate": "no action"
+        }
+      },
+      "compositePrimaryKeys": {},
+      "uniqueConstraints": {},
+      "checkConstraints": {}
+    },
+    "public.api_keys": {
+      "name": "api_keys",
+      "schema": "",
+      "columns": {
+        "id": {
+          "name": "id",
+          "type": "uuid",
+          "primaryKey": true,
+          "notNull": true,
+          "default": "gen_random_uuid()"
+        },
+        "user_id": {
+          "name": "user_id",
+          "type": "uuid",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "organization_id": {
+          "name": "organization_id",
+          "type": "uuid",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "name": {
+          "name": "name",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "key": {
+          "name": "key",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "group_id": {
+          "name": "group_id",
+          "type": "uuid",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "last_used_at": {
+          "name": "last_used_at",
+          "type": "timestamp with time zone",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "created_at": {
+          "name": "created_at",
+          "type": "timestamp with time zone",
+          "primaryKey": false,
+          "notNull": true,
+          "default": "now()"
+        },
+        "expires_at": {
+          "name": "expires_at",
+          "type": "timestamp with time zone",
+          "primaryKey": false,
+          "notNull": false
+        }
+      },
+      "indexes": {
+        "apikey_name_idx": {
+          "name": "apikey_name_idx",
+          "columns": [
+            {
+              "expression": "name",
+              "isExpression": false,
+              "asc": true,
+              "nulls": "last"
+            },
+            {
+              "expression": "organization_id",
+              "isExpression": false,
+              "asc": true,
+              "nulls": "last"
+            }
+          ],
+          "isUnique": true,
+          "concurrently": false,
+          "method": "btree",
+          "with": {}
+        },
+        "ak_user_id_idx": {
+          "name": "ak_user_id_idx",
+          "columns": [
+            {
+              "expression": "user_id",
+              "isExpression": false,
+              "asc": true,
+              "nulls": "last"
+            }
+          ],
+          "isUnique": false,
+          "concurrently": false,
+          "method": "btree",
+          "with": {}
+        },
+        "ak_organization_id_idx": {
+          "name": "ak_organization_id_idx",
+          "columns": [
+            {
+              "expression": "organization_id",
+              "isExpression": false,
+              "asc": true,
+              "nulls": "last"
+            }
+          ],
+          "isUnique": false,
+          "concurrently": false,
+          "method": "btree",
+          "with": {}
+        }
+      },
+      "foreignKeys": {
+        "api_keys_user_id_users_id_fk": {
+          "name": "api_keys_user_id_users_id_fk",
+          "tableFrom": "api_keys",
+          "tableTo": "users",
+          "columnsFrom": [
+            "user_id"
+          ],
+          "columnsTo": [
+            "id"
+          ],
+          "onDelete": "cascade",
+          "onUpdate": "no action"
+        },
+        "api_keys_organization_id_organizations_id_fk": {
+          "name": "api_keys_organization_id_organizations_id_fk",
+          "tableFrom": "api_keys",
+          "tableTo": "organizations",
+          "columnsFrom": [
+            "organization_id"
+          ],
+          "columnsTo": [
+            "id"
+          ],
+          "onDelete": "cascade",
+          "onUpdate": "no action"
+        },
+        "api_keys_group_id_organization_groups_id_fk": {
+          "name": "api_keys_group_id_organization_groups_id_fk",
+          "tableFrom": "api_keys",
+          "tableTo": "organization_groups",
+          "columnsFrom": [
+            "group_id"
+          ],
+          "columnsTo": [
+            "id"
+          ],
+          "onDelete": "set null",
+          "onUpdate": "no action"
+        }
+      },
+      "compositePrimaryKeys": {},
+      "uniqueConstraints": {
+        "api_keys_key_unique": {
+          "name": "api_keys_key_unique",
+          "nullsNotDistinct": false,
+          "columns": [
+            "key"
+          ]
+        }
+      },
+      "checkConstraints": {}
+    },
+    "public.audit_logs": {
+      "name": "audit_logs",
+      "schema": "",
+      "columns": {
+        "id": {
+          "name": "id",
+          "type": "uuid",
+          "primaryKey": true,
+          "notNull": true,
+          "default": "gen_random_uuid()"
+        },
+        "organization_id": {
+          "name": "organization_id",
+          "type": "uuid",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "organization_slug": {
+          "name": "organization_slug",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "action": {
+          "name": "action",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "audit_action": {
+          "name": "audit_action",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "auditable_type": {
+          "name": "auditable_type",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "auditable_display_name": {
+          "name": "auditable_display_name",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "target_id": {
+          "name": "target_id",
+          "type": "uuid",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "target_type": {
+          "name": "target_type",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "target_display_name": {
+          "name": "target_display_name",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "target_namespace_id": {
+          "name": "target_namespace_id",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "target_namespace": {
+          "name": "target_namespace",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "actor_id": {
+          "name": "actor_id",
+          "type": "uuid",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "actor_display_name": {
+          "name": "actor_display_name",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "actor_type": {
+          "name": "actor_type",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "api_key_name": {
+          "name": "api_key_name",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "created_at": {
+          "name": "created_at",
+          "type": "timestamp with time zone",
+          "primaryKey": false,
+          "notNull": true,
+          "default": "now()"
+        }
+      },
+      "indexes": {
+        "auditlogs_organization_idx": {
+          "name": "auditlogs_organization_idx",
+          "columns": [
+            {
+              "expression": "organization_id",
+              "isExpression": false,
+              "asc": true,
+              "nulls": "last"
+            }
+          ],
+          "isUnique": false,
+          "concurrently": false,
+          "method": "btree",
+          "with": {}
+        },
+        "auditlogs_created_at_idx": {
+          "name": "auditlogs_created_at_idx",
+          "columns": [
+            {
+              "expression": "created_at",
+              "isExpression": false,
+              "asc": true,
+              "nulls": "last"
+            }
+          ],
+          "isUnique": false,
+          "concurrently": false,
+          "method": "btree",
+          "with": {}
+        }
+      },
+      "foreignKeys": {},
+      "compositePrimaryKeys": {},
+      "uniqueConstraints": {},
+      "checkConstraints": {}
+    },
+    "public.billing_plans": {
+      "name": "billing_plans",
+      "schema": "",
+      "columns": {
+        "id": {
+          "name": "id",
+          "type": "text",
+          "primaryKey": true,
+          "notNull": true
+        },
+        "active": {
+          "name": "active",
+          "type": "boolean",
+          "primaryKey": false,
+          "notNull": true,
+          "default": true
+        },
+        "name": {
+          "name": "name",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "price": {
+          "name": "price",
+          "type": "integer",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "features": {
+          "name": "features",
+          "type": "json",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "stripe_price_id": {
+          "name": "stripe_price_id",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "weight": {
+          "name": "weight",
+          "type": "integer",
+          "primaryKey": false,
+          "notNull": true,
+          "default": 0
+        }
+      },
+      "indexes": {},
+      "foreignKeys": {},
+      "compositePrimaryKeys": {},
+      "uniqueConstraints": {},
+      "checkConstraints": {}
+    },
+    "public.billing_subscriptions": {
+      "name": "billing_subscriptions",
+      "schema": "",
+      "columns": {
+        "id": {
+          "name": "id",
+          "type": "text",
+          "primaryKey": true,
+          "notNull": true
+        },
+        "organization_id": {
+          "name": "organization_id",
+          "type": "uuid",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "metadata": {
+          "name": "metadata",
+          "type": "json",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "status": {
+          "name": "status",
+          "type": "status",
+          "typeSchema": "public",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "price_id": {
+          "name": "price_id",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "quantity": {
+          "name": "quantity",
+          "type": "integer",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "cancel_at_period_end": {
+          "name": "cancel_at_period_end",
+          "type": "boolean",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "cancel_at": {
+          "name": "cancel_at",
+          "type": "timestamp with time zone",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "canceled_at": {
+          "name": "canceled_at",
+          "type": "timestamp with time zone",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "current_period_start": {
+          "name": "current_period_start",
+          "type": "timestamp with time zone",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "current_period_end": {
+          "name": "current_period_end",
+          "type": "timestamp with time zone",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "created_at": {
+          "name": "created_at",
+          "type": "timestamp with time zone",
+          "primaryKey": false,
+          "notNull": true,
+          "default": "now()"
+        },
+        "ended_at": {
+          "name": "ended_at",
+          "type": "timestamp with time zone",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "trial_start": {
+          "name": "trial_start",
+          "type": "timestamp with time zone",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "trial_end": {
+          "name": "trial_end",
+          "type": "timestamp with time zone",
+          "primaryKey": false,
+          "notNull": false
+        }
+      },
+      "indexes": {
+        "billsubs_organization_id_idx": {
+          "name": "billsubs_organization_id_idx",
+          "columns": [
+            {
+              "expression": "organization_id",
+              "isExpression": false,
+              "asc": true,
+              "nulls": "last"
+            }
+          ],
+          "isUnique": false,
+          "concurrently": false,
+          "method": "btree",
+          "with": {}
+        }
+      },
+      "foreignKeys": {
+        "billing_subscriptions_organization_id_organizations_id_fk": {
+          "name": "billing_subscriptions_organization_id_organizations_id_fk",
+          "tableFrom": "billing_subscriptions",
+          "tableTo": "organizations",
+          "columnsFrom": [
+            "organization_id"
+          ],
+          "columnsTo": [
+            "id"
+          ],
+          "onDelete": "cascade",
+          "onUpdate": "no action"
+        }
+      },
+      "compositePrimaryKeys": {},
+      "uniqueConstraints": {},
+      "checkConstraints": {}
+    },
+    "public.cache_warmer_operations": {
+      "name": "cache_warmer_operations",
+      "schema": "",
+      "columns": {
+        "id": {
+          "name": "id",
+          "type": "uuid",
+          "primaryKey": true,
+          "notNull": true,
+          "default": "gen_random_uuid()"
+        },
+        "organization_id": {
+          "name": "organization_id",
+          "type": "uuid",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "federated_graph_id": {
+          "name": "federated_graph_id",
+          "type": "uuid",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "operation_content": {
+          "name": "operation_content",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "operation_hash": {
+          "name": "operation_hash",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "operation_persisted_id": {
+          "name": "operation_persisted_id",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "operation_name": {
+          "name": "operation_name",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "client_name": {
+          "name": "client_name",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "client_version": {
+          "name": "client_version",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "created_at": {
+          "name": "created_at",
+          "type": "timestamp with time zone",
+          "primaryKey": false,
+          "notNull": true,
+          "default": "now()"
+        },
+        "planning_time": {
+          "name": "planning_time",
+          "type": "real",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "is_manually_added": {
+          "name": "is_manually_added",
+          "type": "boolean",
+          "primaryKey": false,
+          "notNull": true,
+          "default": false
+        },
+        "created_by_id": {
+          "name": "created_by_id",
+          "type": "uuid",
+          "primaryKey": false,
+          "notNull": false
+        }
+      },
+      "indexes": {
+        "cwo_organization_id_idx": {
+          "name": "cwo_organization_id_idx",
+          "columns": [
+            {
+              "expression": "organization_id",
+              "isExpression": false,
+              "asc": true,
+              "nulls": "last"
+            }
+          ],
+          "isUnique": false,
+          "concurrently": false,
+          "method": "btree",
+          "with": {}
+        },
+        "cwo_federated_graph_id_idx": {
+          "name": "cwo_federated_graph_id_idx",
+          "columns": [
+            {
+              "expression": "federated_graph_id",
+              "isExpression": false,
+              "asc": true,
+              "nulls": "last"
+            }
+          ],
+          "isUnique": false,
+          "concurrently": false,
+          "method": "btree",
+          "with": {}
+        },
+        "cwo_created_by_id_idx": {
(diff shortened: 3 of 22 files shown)


router/pkg/pubsub/kafka/engine_datasource.go
```diff
@@ -6,32 +6,174 @@
...
+	if headers == nil {
+		headers = make(map[string][]byte)
+	}
+
+	headersBytes, err := json.Marshal(headers)
+	if err != nil {
+		return "", err
+	}
+
+	return fmt.Sprintf(`{"topic":"%s", "event": {"data": %s, "key": "%s", "headers": %s}, "providerId":"%s", "rootFieldName":"%s"}`, p.Topic, p.Event.Data, p.Event.Key, headersBytes, p.Provider, p.FieldName), nil
```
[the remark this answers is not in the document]
dkorittki: Yes, this one bugs me as well but we cannot use a json marshaller to create the output because we don't return json here. It returns json-like data with custom placeholders, which later on in the code gets replaced to then be valid json. These placeholders break json-compatibility. The output looks like

```
{
  "topic":"employeeUpdated",
  "event": {
    "data": {"employeeID":$$0$$,"update":$$1$$},
    "key": "",
    "headers": {}
  },
  "providerId":"my-kafka",
  "rootFieldName":"updateEmployeeMyKafka"
}
```

This is not valid json and hence we cannot create this format with a json marshaller, because it will return the error
`error(*encoding/json.SyntaxError) *{msg: "invalid character '$' looking for beginning of value", Offset: 0}`.

The field `p.Event.Data` is whats providing the data containing placeholders `{"employeeID":$$0$$,"update":$$1$$}`.

The security thread in this function is, that we do not escape the input data the output is build from. This is
- p.Topic --> The topic named in the schema EDFS directive, not via user operation input
- p.Event.Data --> Gets properly escaped before this function gets called
- p.Event.Key --> Kafka message key, can be set by a custom module, not via user GraphQL operation input
- headersBytes --> Kafka message headers, can be set by a custom module, not via user operation input
- p.Provider --> ProviderID from router configuration, not via user operation input
- p.FieldName --> Root field name of mutation, set by schema, not via user operation input

So for an attacker there seem to not be an easy way to exploit this. However, the GraphQL schema owner could maybe set their topic names, headers, keys, providerID´s in a way that this could escape or break in a way. So it still makes sense to marshall everything we can before orchestrating the final output. I will improve this function.

router/go.mod
```diff
@@ -31,7 +31,7 @@ require (
 	github.com/tidwall/gjson v1.18.0
 	github.com/tidwall/sjson v1.2.5
 	github.com/twmb/franz-go v1.16.1
-	github.com/wundergraph/graphql-go-tools/v2 v2.0.0-rc.237
+	github.com/wundergraph/graphql-go-tools/v2 v2.0.0-rc.237.0.20251110152155-423a60c6a33e
```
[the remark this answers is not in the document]
dkorittki: Will be done once engine pull request got merged and new version of engine is available.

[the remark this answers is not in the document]
dkorittki: Keeping this issue unresolved until the engine got updated and used in the routers go.mod

router/pkg/pubsub/redis/engine_datasource.go
```diff
@@ -6,33 +6,135 @@ import (
...
+type MutableEvent struct {
+	Data json.RawMessage `json:"data"`
+}
+
+func (e *MutableEvent) GetData() []byte {
+	if e == nil {
+		return nil
+	}
+	return e.Data
+}
```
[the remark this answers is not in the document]
dkorittki: It's intentional like this and part of the hook interface design. A `MutableEvent` is designed to be mutable, while a `Event` is immutable. Since `Event` is immutable, we cannot return the original data slice as it would mean a hook developer could modify the content of `Event`.

router/pkg/pubsub/redis/engine_datasource.go
```diff
@@ -6,33 +6,135 @@ import (
...
+	return &PublishEventConfiguration{
+		Provider:  p.Provider,
+		Channel:   p.Channel,
+		FieldName: p.FieldName,
+	}
+}
+
+func (p *publishData) MarshalJSONTemplate() (string, error) {
+	return fmt.Sprintf(`{"channel":"%s", "event": {"data": %s}, "providerId":"%s", "rootFieldName":"%s"}`, p.Channel, p.Event.Data, p.Provider, p.FieldName), nil
 }
```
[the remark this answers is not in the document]
dkorittki: I'm resolving the issue because we have multiple other comments for this issue on this pull request already and I would like to keep it at a central place.

router/pkg/pubsub/nats/engine_datasource.go
```diff
@@ -6,39 +6,163 @@ import (
...
+		FieldName: p.FieldName,
+	}
+}
+
+func (p *publishData) MarshalJSONTemplate() (string, error) {
 	// The content of the data field could be not valid JSON, so we can't use json.Marshal
 	// e.g. {"id":$$0$$,"update":$$1$$}
-	return fmt.Sprintf(`{"subject":"%s", "data": %s, "providerId":"%s"}`, s.Subject, s.Data, s.ProviderID)
+	return fmt.Sprintf(`{"subject":"%s", "event": {"data": %s}, "providerId":"%s", "rootFieldName":"%s"}`, p.Subject, p.Event.Data, p.Provider, p.FieldName), nil
+}
```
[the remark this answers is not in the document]
dkorittki: I'm resolving the issue because we have multiple other comments for this issue on this pull request already and I would like to keep it at a central place.

router/pkg/pubsub/kafka/engine_datasource.go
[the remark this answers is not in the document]
dkorittki: Fixed. We now use json marshall as many input fields as we can, with the exception of `p.Event.Data`, which contains the template values with dollar signs as delimiters. But this field gets properly escaped and validated before it enters this function, so we are safe to skip it here.

I ensured that the method stays fast, because it is called with every GraphQL mutation operation on an EDFS backed mutation. It's around as fast or slightly faster than the old implementation at the expense of some memory sacrifices but I don't think this matters at all in the end.

```
OLD:

goos: darwin
goarch: arm64
pkg: github.com/wundergraph/cosmo/router/pkg/pubsub/kafka
cpu: Apple M4 Pro
Benchmark_MarshalJSONTemplate-14    	 3520440	       335.1 ns/op	     393 B/op	       8 allocs/op
PASS


NEW:

goos: darwin
goarch: arm64
pkg: github.com/wundergraph/cosmo/router/pkg/pubsub/kafka
cpu: Apple M4 Pro
Benchmark_MarshalJSONTemplate-14    	 3752730	       328.3 ns/op	     520 B/op	      10 allocs/op
PASS

```

router/pkg/pubsub/datasource/pubsubprovider.go
```diff
@@ -2,15 +2,65 @@ package datasource
...
+func (p *PubSubProvider) applyPublishEventHooks(ctx context.Context, cfg PublishEventConfiguration, events []StreamEvent) (currentEvents []StreamEvent, err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			if p.Logger != nil {
+				p.Logger.
+					WithOptions(zap.AddStacktrace(zapcore.ErrorLevel)).
+					Error("[Recovery from handler panic]",
+						zap.Any("error", r),
+					)
+			}
```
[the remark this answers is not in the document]
dkorittki: Fixed

router/pkg/pubsub/datasource/subscription_event_updater.go
```diff
@@ -0,0 +1,160 @@
...
+		// Timeout exceeded, some subscription updates may still be running.
+		// We can't stop them but we will also not wait for them, basically abandoning them.
+		// They will continue to hold their semaphore slots until they complete,
+		// which means the next Update() call will have fewer available slots.
+		// Also since we will process the next batch of events while having abandoned updaters,
+		// those updaters might eventually push their events to the subscription late,
+		// which means events might arrive out of order.
+		s.logger.Warn("Timeout exceeded during subscription updates, events may arrive out of order")
+	}
+}
```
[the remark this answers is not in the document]
dkorittki: Fixed, I opted for only reading the event slice in the `updateSubscription` and avoid any writes. I do not use the deep copy solution suggested here, since this is what we intentionally want to avoid. It would cause a lot of allocations as this would have to be done for every active subscription these events have to be written to.

router/pkg/pubsub/datasource/subscription_event_updater.go
```diff
@@ -0,0 +1,160 @@
...
+
+func (s *subscriptionEventUpdater) Complete() {
+	s.eventUpdater.Complete()
+}
+
+func (s *subscriptionEventUpdater) Close(kind resolve.SubscriptionCloseKind) {
+	s.eventUpdater.Close(kind)
+	close(s.semaphore)
+}
```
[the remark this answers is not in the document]
dkorittki: Fixed, I removed the close of the semaphore.

router/pkg/pubsub/datasource/subscription_event_updater.go
```diff
@@ -0,0 +1,158 @@
...
+	for _, event := range events {
+		if event == nil {
+			continue
+		}
+		s.eventUpdater.UpdateSubscription(subID, event.GetData())
+	}
+
+	// In case there was an error we close the affected subscription.
+	if err != nil {
+		s.eventUpdater.CloseSubscription(resolve.SubscriptionCloseKindNormal, subID)
```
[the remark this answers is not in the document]
dkorittki: Fixed, loop is stopped when an error has been returned by a hook and a test has been added.

dkorittki:
Fixed all complaints from coderrabbit and Github on actual router code. I will take care of unit test / integration test / rfc complaints tomorrow.

router-tests/modules/start_subscription_test.go
```diff
@@ -0,0 +1,662 @@
...
+
+			testenv.AwaitChannelWithT(t, time.Second*10, clientRunCh, func(t *testing.T, err error) {
+				require.NoError(t, err)
+			}, "unable to close client before timeout")
+
+			require.Empty(t, originResponseCalled)
+
+			requestLog := xEnv.Observer().FilterMessage("SubscriptionOnStart Hook has been run")
+			assert.Len(t, requestLog.All(), 1)
+		})
```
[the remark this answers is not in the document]
dkorittki: I talked to @alepane21 . The test name is old. There was the expectation at the beginning, that the `OnOriginResponse` hook is called even if `SubscriptionOnStart` returned an error. Later on it during development it became clear, that this does not make sense because the connection to the origin is only started after we ran the `SubscriptionOnStart` hook. If that hook returns an error no connection to the origin is established and thus the `OnOriginResponse` is expected to not be called. So the test itself asserts the right thing but the test name was not updated. Fixed it by updating the test name.

router-tests/modules/start-subscription/module.go
```diff
@@ -0,0 +1,61 @@
...
+func (m *StartSubscriptionModule) SubscriptionOnStart(ctx core.SubscriptionOnStartHandlerContext) error {
+
+	m.Logger.Info("SubscriptionOnStart Hook has been run")
+
+	if m.Callback != nil {
+		return m.Callback(ctx)
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
dkorittki: Fixed

router-tests/modules/stream-publish/module.go
```diff
@@ -0,0 +1,49 @@
...
+
+func (m *PublishModule) OnPublishEvents(ctx core.StreamPublishEventHandlerContext, events datasource.StreamEvents) (datasource.StreamEvents, error) {
+	m.Logger.Info("Publish Hook has been run")
+
+	if m.Callback != nil {
+		return m.Callback(ctx, events)
+	}
+
+	return events, nil
+}
```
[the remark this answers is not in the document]
dkorittki: Fixed

router/pkg/pubsub/pubsub_test.go
```diff
@@ -61,13 +62,17 @@ func TestBuild_OK(t *testing.T) {
...
+		OnPublishEvents: datasource.OnPublishEventsHooks{Handlers: []datasource.OnPublishEventsFn(nil)},
+	}).Return(nil)
 
 	mockBuilder.On("TypeID").Return("nats")
 	mockBuilder.On("BuildProvider", natsEventSources[0]).Return(mockPubSubProvider, nil)
 
 	// ctx, kafkaBuilder, config.Providers.Kafka, kafkaDsConfsWithEvents
 	// Execute the function
-	providers, dataSources, err := build(ctx, mockBuilder, natsEventSources, dsConfs, rmetric.NewNoopStreamMetricStore())
+	providers, dataSources, err := build(ctx, mockBuilder, natsEventSources, dsConfs, rmetric.NewNoopStreamMetricStore(), datasource.Hooks{})
```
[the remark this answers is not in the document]
dkorittki: Fixed. The odd thing was this test passed even though two arguments need to be passed to `BuildProvider()`. The generated mock code was broken and worked with either 1 or 2 arguments regardless. I regenerated the mock code with mockery and that fixed the broken mock code. I also updated the test function to include `mock.Anything` as the second parameter, which is consistend with other tests in that file.

adr/cosmo-streams-v1.md
```diff
@@ -0,0 +1,456 @@
...
+func (e StreamEvents) Unsafe() []StreamEvent             // returns the underlying slice
+
+func NewStreamEvents(evts []StreamEvent) StreamEvents
+
+// StreamEvent is a generic immutable event.
+// Every provider will have it's distinct implementation with additionals fields.
+// Common to all providers is that their events have a payload.
+type StreamEvent interface {
+    // GetData returns a copy of payload data of the event
+    GetData() []byte
```
jensneuse, line 58:
```suggestion
    Data() []byte
```

would be more idiomatic

router/pkg/pubsub/pubsub_test.go
```diff
@@ -61,13 +62,17 @@ func TestBuild_OK(t *testing.T) {
 	}
 
 	mockPubSubProvider.On("ID").Return("provider-1")
+	mockPubSubProvider.On("SetHooks", datasource.Hooks{
+		OnReceiveEvents: datasource.OnReceiveEventsHooks{Handlers: []datasource.OnReceiveEventsFn(nil)},
+		OnPublishEvents: datasource.OnPublishEventsHooks{Handlers: []datasource.OnPublishEventsFn(nil)},
+	}).Return(nil)
```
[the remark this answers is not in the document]
dkorittki: Fixed

router/pkg/pubsub/pubsub_test.go
```diff
@@ -236,13 +241,17 @@ func TestBuild_ShouldNotInitializeProviderIfNotUsed(t *testing.T) {
 	}
 
 	mockPubSubUsedProvider.On("ID").Return("provider-2")
+	mockPubSubUsedProvider.On("SetHooks", datasource.Hooks{
+		OnReceiveEvents: datasource.OnReceiveEventsHooks{Handlers: []datasource.OnReceiveEventsFn(nil)},
+		OnPublishEvents: datasource.OnPublishEventsHooks{Handlers: []datasource.OnPublishEventsFn(nil)},
+	}).Return(nil)
```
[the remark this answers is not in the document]
dkorittki: Fixed

router-tests/modules/stream-receive/module.go
```diff
@@ -0,0 +1,51 @@
+package batch
```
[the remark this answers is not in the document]
dkorittki: Fixed

router-tests/modules/start_subscription_test.go
```diff
@@ -0,0 +1,662 @@
...
+
+			xEnv.WaitForSubscriptionCount(1, time.Second*10)
+
+			require.NoError(t, client.Close())
+			testenv.AwaitChannelWithT(t, time.Second*10, clientRunCh, func(t *testing.T, err error) {
+				require.NoError(t, err)
+
+			}, "unable to close client before timeout")
+
+			requestLog := xEnv.Observer().FilterMessage("SubscriptionOnStart Hook has been run")
```
jensneuse: how about using a mock implementation that increases a counter and assert on the number of hook calls instead of filtering on a message?

router-tests/modules/start_subscription_test.go
```diff
@@ -0,0 +1,662 @@
...
+			Graph: config.Graph{},
+			Modules: map[string]interface{}{
+				"startSubscriptionModule": start_subscription.StartSubscriptionModule{
+					Callback: func(ctx core.SubscriptionOnStartHandlerContext) error {
+						if ctx.SubscriptionEventConfiguration().RootFieldName() != "employeeUpdatedMyKafka" {
+							return nil
+						}
+						ctx.WriteEvent((&kafka.MutableEvent{
+							Key:  []byte("1"),
+							Data: []byte(`{"id": 1, "__typename": "Employee"}`),
```
jensneuse: Can we access the id from the ctx here?

router-tests/modules/start_subscription_test.go
```diff
@@ -0,0 +1,662 @@
...
+
+		callbackCalled := make(chan bool)
+
+		cfg := config.Config{
+			Graph: config.Graph{},
+			Modules: map[string]interface{}{
+				"startSubscriptionModule": start_subscription.StartSubscriptionModule{
+					Callback: func(ctx core.SubscriptionOnStartHandlerContext) error {
+						callbackCalled <- true
+						return core.NewHttpGraphqlError("subscription closed", http.StatusText(http.StatusOK), http.StatusOK)
```
jensneuse: Is the intention here to gracefully stop a subscription or is this just returning a generic error? If yes, could we find a more idiomatic way to close a subscription?

adr/cosmo-streams-v1.md
```diff
@@ -0,0 +1,456 @@
...
+    Logger() *zap.Logger
+    // Operation is the GraphQL operation
+    Operation() OperationContext
+    // Authentication is the authentication for the request
+    Authentication() authentication.Authentication
+    // SubscriptionEventConfiguration is the subscription event configuration (will return nil for engine subscription)
+    SubscriptionEventConfiguration() datasource.SubscriptionEventConfiguration
+    // WriteEvent writes an event to the stream of the current subscription
+    // It returns true if the event was written to the stream, false if the event was dropped
+    WriteEvent(event datasource.StreamEvent) bool
```
jensneuse: I think we should have much stronger naming differentiation between WriteEvent and NewEvent.
My understanding from reading the tests is that WriteEvent goes through the event broker (e.g. Kafka) so it's fan out to all consumers vs. NewEvent is just local to the Subscription but I'm not entirely sure.
I've searched through the ADR and it's not clear what the differentiation is.
Can you clarify?

router-tests/modules/stream_publish_test.go
```diff
@@ -0,0 +1,345 @@
...
+	"github.com/wundergraph/cosmo/router/core"
+	"github.com/wundergraph/cosmo/router/pkg/config"
+	"github.com/wundergraph/cosmo/router/pkg/pubsub/datasource"
+	"github.com/wundergraph/cosmo/router/pkg/pubsub/kafka"
+)
+
+func TestPublishHook(t *testing.T) {
+	t.Parallel()
+
+	t.Run("Test Publish hook can't assert to mutable types", func(t *testing.T) {
```
jensneuse, line 28: I'm putting this comment just on this test but my general feedback on the tests in router-tests is that it could be useful to add comments and the tests to give a bit more context on how each test adds variance in the test suite. The titles are rather short and I understand why, but as someone from outside, it's a bit hard to understand why each test is relevant and what particular (edge) case it's covering.

router/core/graphql_handler.go
```diff
@@ -400,6 +400,22 @@ func (h *GraphQLHandler) WriteError(ctx *resolve.Context, err error, res *resolv
 		if isHttpResponseWriter {
 			httpWriter.WriteHeader(http.StatusInternalServerError)
 		}
+	case errorTypeHttpError:
+		var httpErr *httpGraphqlError
+		if !errors.As(err, &httpErr) {
+			response.Errors[0].Message = "Internal server error"
```
jensneuse: got test coverage for this?

router/core/graphql_handler.go
```diff
@@ -400,6 +400,22 @@ func (h *GraphQLHandler) WriteError(ctx *resolve.Context, err error, res *resolv
...
+			return
+		}
+		response.Errors[0].Message = httpErr.Message()
```
jensneuse: what if response.Errors == nil or len < 1?

router/core/router.go
```diff
@@ -253,6 +253,14 @@ func NewRouter(opts ...Option) (*Router, error) {
 		r.metricConfig = rmetric.DefaultConfig(Version)
 	}
 
+	if r.subscriptionHooks.onReceiveEvents.maxConcurrentHandlers == 0 {
+		r.subscriptionHooks.onReceiveEvents.maxConcurrentHandlers = 100
```
jensneuse, line 257: is this configurable via config.go?

router-tests/modules/start_subscription_test.go
dkorittki: Can you specify what id you mean? The one from `Data`? Unfortunately no, thats not possible at the moment

router/core/router.go
dkorittki: Yes, you can set this via

```yaml
events:
  subscription_hooks:
    on_receive_events:
      max_concurrent_handlers: 100
```

But when thinkin about this I think I will change this to

```yaml
events:
  handlers:
    on_receive_events:
      max_concurrent_handlers: 100
```

Because the `subscription_hooks` section is meant for configuring all Cosmo Streams handlers, not just the ones for subscriptions.  The `OnPublishEvent` hook is running for mutations instead of subscriptions. So the name `subscription_hooks` does not fit well.

adr/cosmo-streams-v1.md
dkorittki: Ah, I see the confusion.

`WriteEvent` creates an event inside the router. For the router it looks like an event coming straight from a broker. It resolves the fields asked for by the operation and returns the results to the client. The method expects an event object, which you can create with `NewEvent`.

I will update the godoc and ADR to make this more clear.

router-tests/modules/start_subscription_test.go
dkorittki: The intention is to give hook developers the option to abort the subscription during initialization phase. So suppose return from the hook like this

```go
return core.NewHttpGraphqlError("Forbidden", http.StatusText(http.StatusForbidden), http.StatusForbidden)
```

On the connection, this is what you get

```
 > {"type":"connection_init","payload":{}}
 < {"type":"connection_ack"}
 > {"id":"1","type":"subscribe","payload":{"query":"subscription { employeeUpdatedMyKafka(employeeID:1) { id details { forename surname } } }"}}
 < {"id":"1","type":"error","payload":[{"message":"Forbidden","extensions":{"statusCode":403,"code":"Forbidden"}}]}
Disconnected (code: 1000, reason: "Normal closure")
```

So first an error message is returned after a subscription start (not at connection_init), than the websockt connection is closed normally by the router. Do you have any specific ideas in mind on how to close the connection instead?

adr/cosmo-streams-v1.md
dkorittki: I could do it but only at some expense. We would need to change the `Data` field of all concrete structs that implement the `StreamEvent` interface. Those are

- [`kafka.MutableEvent`](https://github.com/wundergraph/cosmo/pull/2273/files#diff-10b19360ceda026f0014394dcc14ba4055a4aab1011bda0054bee83d7d3fd4abR62-R66)
- [`nats.MutableEvent`](https://github.com/wundergraph/cosmo/pull/2273/files#diff-58d5ec5831d5307b2c956ed572a9b86599db994942f64bcf86ef9c83351383b2R42-R45)
- [`redis.MutableEvent`](https://github.com/wundergraph/cosmo/pull/2273/files#diff-10b75112e883a8fd36aec1f9bc0a811aedb4bfb173c030a175e6ca8632c5735aR35-R37)

Lets take the Kafka implementation as an example:

```go
type MutableEvent struct {
	Key     []byte            `json:"key"` // only on kafka implementation
	Data    json.RawMessage   `json:"data"`
	Headers map[string][]byte `json:"headers"` // only on kafka & nats implementation
}
```

Because of the `Data` field I can't have a method `Data()` on that struct as it will name clash.

Option A would be to make the fields private

```go
type MutableEvent struct {
	key     []byte
	data    json.RawMessage
	headers map[string][]byte
}
```

and expose all fields via methods `Key()`, `SetKey()`, `Data`, `SetData()`, `Headers`, `SetHeaders()` to let hook developers write to the fields. Certainly possible but not very Go idiomatic. As a hook developer I would expect to have direct access to fields I can write on.

Another idea: Rename the `Data` field to something else like `EventData`:

```go
type MutableEvent struct {
	Key          []byte            `json:"key"` // only on kafka implementation
	EventData    json.RawMessage   `json:"data"`
	Headers      map[string][]byte `json:"headers"` // only on kafka & nats implementation
}
```

Then we can have a `Data()` method but there would be confusion for hook developers:

```go
event.EventData = []byte("new value")

// is the same as
event.SetData([]byte("new value"))
```

So essentially we have two differently named ways to access the same property of an event ("Data" vs "EventData")

What is your take on this?

router-tests/modules/start_subscription_test.go
jensneuse: could we return a special error?

```go
return core.CloseSubscription{
    Reason: "my custom message"
}
```

You can then do `errors.As/errors.Is` to react.

adr/cosmo-streams-v1.md
jensneuse: So, WriteEvent is not really writing an event to the broker, which I thought it would, but instead it's just emitting a new event, but not to other Routers through the broker, only to the current Router? Can we make it clear that this event is only for the current subscription? or is it for all Router local subscriptions? and could it also be delivered back to the broker to fan it out? Maybe we want to offer the user options. In any way, we want to be very explicit in regards to naming such that the side effects are very clear. "WriteEvent" is very generic and can mean a lot of things.

adr/cosmo-streams-v1.md
jensneuse: keep it as is

adr/cosmo-streams-v1.md
dkorittki: > So, WriteEvent is not really writing an event to the broker, which I thought it would, but instead it's just emitting a new event, but not to other Routers through the broker, only to the current Router?

Yes, it's only emitting an event inside the router. No broker will ever receive that message.

> Can we make it clear that this event is only for the current subscription? or is it for all Router local subscriptions?

The event is local to the subscription the hook is being called for. Other subscriptions will never see this event. I will address this in godoc and Cosmo docs.

> and could it also be delivered back to the broker to fan it out? Maybe we want to offer the user options.

It's a good idea but at the moment we don't support this. Right now the main goal of the `WriteEvent` method was to provide hook developers a way to push initial data to new subscription clients, so they can build up state without waiting for an actual event to arrive from the broker.

But I do see how it can be valuable to push events to brokers when a new client connects. Not sure how complicated it is to add but I believe it's not something we can easily add to this hook. Probably involves some more work. Will look into it to see if we could easily add this.

Overall I agree the method sounds too vague and doesn't really tell what it's actually doing. Gonna rename the method and clarify in godoc and Cosmo Streams what's actually happening.

router/core/graphql_handler.go
dkorittki: `response` always gets created with a len of 1 around line 300

```go
response := GraphQLErrorResponse{
	Errors: make([]graphqlError, 1),
	Data:   nil,
}
```

router-tests/modules/stream_publish_test.go
dkorittki: Added a short descriptive comment to each new router-test.
Commit: [chore: add description to router-tests + delete obsolete test](https://github.com/wundergraph/cosmo/pull/2273/commits/736ef093475cb1baf457f09ffdeff996c8d712db)

adr/cosmo-streams-v1.md
dkorittki: Renamed the method to `EmitLocalEvent` with a more detailed godoc.
Commit: [chore: rename WriteEvent to EmitLocalEvent](https://github.com/wundergraph/cosmo/pull/2273/commits/65d5d5561fc645aa07cb0aea8836c7d5d8c03c94)

router-tests/modules/start_subscription_test.go
dkorittki: After discussion with Jens, I now have added a dedicated type for this: `core.StreamHandlerError`.
Commit: [chore: reintroduce StreamHandlerError](https://github.com/wundergraph/cosmo/pull/2273/commits/f93145f34cbfa3ac59a49ab921cf26b5a3396a33)

router/core/graphql_handler.go
dkorittki: Thats now obsolete because hook developers should not return errors via `core.NewHttpGraphqlError()` anymore (it would lead into this switch case). They should use `core.StreamHandlerError`.
More Info: https://github.com/wundergraph/cosmo/pull/2273#discussion_r2541740238

And yes, we have tests covering the error handling when returning `core.StreamHandlerError`.

router-tests/modules/start_subscription_test.go
dkorittki: Done, using a counter on the custom module now, which increases whenever the hook is called.
Commit: [chore: use counter to count hook calls](https://github.com/wundergraph/cosmo/pull/2273/commits/e1bf8cd4f66cb3d831da78dfd93271844841811e)

router/core/graphql_handler.go
```diff
@@ -400,6 +414,22 @@ func (h *GraphQLHandler) WriteError(ctx *resolve.Context, err error, res *resolv
...
+		if httpErr.ExtensionCode() != "" || httpErr.StatusCode() != 0 {
+			response.Errors[0].Extensions = &Extensions{
+				Code:       httpErr.ExtensionCode(),
+				StatusCode: httpErr.StatusCode(),
+			}
+		}
+		if isHttpResponseWriter {
+			httpWriter.WriteHeader(httpErr.StatusCode())
+		}
```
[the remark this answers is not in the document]
dkorittki: That switch case is unreachable now, since we don't use errorTypeHttpError in hooks anymore (replaced by `core.StreamHandlerError`). I removed that code.

router/core/graphql_handler.go
```diff
@@ -390,6 +390,20 @@ func (h *GraphQLHandler) WriteError(ctx *resolve.Context, err error, res *resolv
...
+			// We could set response.Errors[0].Extensions, too
+			if isHttpResponseWriter {
+				httpWriter.WriteHeader(http.StatusInternalServerError)
+			}
+			return
+		}
+		response.Errors[0].Message = errStreamHandlerError.Message
+		if isHttpResponseWriter {
+			httpWriter.WriteHeader(http.StatusOK)
+		}
```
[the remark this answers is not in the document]
dkorittki: Fixed

router/core/router.go
dkorittki: Fixed

router-tests/modules/start_subscription_test.go
```diff
@@ -0,0 +1,695 @@
...
+
+			clientRunCh := make(chan error)
+			go func() {
+				clientRunCh <- client.Run()
+			}()
+
+			xEnv.WaitForSubscriptionCount(1, time.Second*10)
+			<-callbackCalled
+			xEnv.WaitForSubscriptionCount(0, time.Second*10)
+
```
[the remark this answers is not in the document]
dkorittki: Fixed

router-tests/modules/stream_receive_test.go
```diff
@@ -0,0 +1,969 @@
...
+
+			var subscriptionOne struct {
+				employeeUpdatedMyKafka struct {
+					ID      float64 `graphql:"id"`
+					Details struct {
+						Forename string `graphql:"forename"`
+						Surname  string `graphql:"surname"`
+					} `graphql:"details"`
+				} `graphql:"employeeUpdatedMyKafka(employeeID: 3)"`
+			}
```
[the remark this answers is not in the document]
dkorittki: It's okay to use the same `subscriptionOne` struct, since this only describes the query both clients use. I renamed that variable to `subscriptionQuery` to avoid confusion.

router-tests/modules/stream_publish_test.go
```diff
@@ -0,0 +1,357 @@
...
+			resOne := xEnv.MakeGraphQLRequestOK(testenv.GraphQLRequest{
+				Query: `mutation { updateEmployeeMyRedis(id: 3, update: {name: "name test"}) { success } }`,
+			})
+			require.JSONEq(t, `{"data": {"updateEmployeeMyRedis": {"success": false}}}`, resOne.Body)
+
+			assert.Equal(t, int32(1), customModule.HookCallCount.Load())
+
+			require.Len(t, records, 1)
+		})
+	})
```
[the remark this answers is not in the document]
dkorittki: Fixed

router-tests/modules/stream_receive_test.go
```diff
@@ -0,0 +1,969 @@
...
+			// Wait for server to close the subscription connection
+			xEnv.WaitForSubscriptionCount(0, Timeout)
+
+			// Verify that client.Run() completed when server closed the connection
+			testenv.AwaitChannelWithT(t, Timeout, clientRunCh, func(t *testing.T, err error) {
+				require.NoError(t, err)
+			}, "client should have completed when server closed connection")
+
+			xEnv.WaitForTriggerCount(0, Timeout)
+		})
```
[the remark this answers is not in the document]

router/core/graphql_handler.go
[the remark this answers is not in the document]
dkorittki: All `httpGraphqlError` instances you mentioned are handled by the writeOperationError function (in router/core/errors.go:317), which is called from the prehandler before the GraphQL handler's ServeHTTP method is ever invoked, which is where the `WriteError` method we just touched is called. Please analyze more carefully, if the code is actually reachable.

But you are right about code I missed to remove. The getErrorType function should not check for httpGraphqlError. In fact, we don't need the httpGraphqlError in general anymore. This type got introduced by this PR and is now completely replaced `StreamHandlerError`. I will remove the mentioned code.

router-tests/modules/start_subscription_test.go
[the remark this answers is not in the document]
dkorittki: Using an unbuffered channel is done on purpose. It ensures the hook, who writes to the channel is blocked until the test itself reads from the channel (line 260). We need this blocking so we can measure in line 259 the client connected and in line 263 that the client got disconnected due to the error being returned by the hook.

router/core/graphql_handler.go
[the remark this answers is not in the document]
dkorittki: Fixed

adr/cosmo-streams-v1.md
jensneuse: You should update the ADR, right?

dkorittki:
Waiting for Engine PR to get merged and new engine version to be released. Then I pull latest changes from main, update router go.mod, then its ready for merge.

router/pkg/pubsub/kafka/adapter.go
```diff
@@ -161,57 +179,85 @@ func (p *ProviderAdapter) Subscribe(ctx context.Context, event SubscriptionEvent
...
 
-	p.writeClient.Produce(ctx, &kgo.Record{
-		Topic: event.Topic,
-		Value: event.Data,
-	}, func(record *kgo.Record, err error) {
-		defer wg.Done()
-		if err != nil {
-			pErr = err
+	for _, streamEvent := range events {
+		kafkaEvent, ok := streamEvent.Clone().(*MutableEvent)
```
StarpTech: Does it need to be cloned?

router/pkg/pubsub/kafka/adapter.go
```diff
@@ -161,57 +179,85 @@ func (p *ProviderAdapter) Subscribe(ctx context.Context, event SubscriptionEvent
...
+		}
+
+		p.writeClient.Produce(ctx, &kgo.Record{
+			Key:     kafkaEvent.Key,
+			Topic:   pubConf.Topic,
+			Value:   kafkaEvent.Data,
+			Headers: headers,
+		}, func(record *kgo.Record, err error) {
+			defer wg.Done()
+			if err != nil {
```
StarpTech, line 238: Should we use an err group?

router/core/graphql_handler.go
```diff
@@ -390,6 +390,19 @@ func (h *GraphQLHandler) WriteError(ctx *resolve.Context, err error, res *resolv
 		if isHttpResponseWriter {
 			httpWriter.WriteHeader(http.StatusInternalServerError)
 		}
+	case errorTypeEDFSHookError:
```
StarpTech: Should we avoid EDFS terms for the future?

router/pkg/pubsub/datasource/subscription_datasource.go
```diff
@@ -0,0 +1,107 @@
...
+
+	return s.pubSub.Subscribe(ctx.Context(), conf, NewSubscriptionEventUpdater(conf, s.hooks, updater, logger, s.eventBuilder))
+}
+
+func (s *PubSubSubscriptionDataSource[C]) SubscriptionOnStart(ctx resolve.StartupHookContext, input []byte) (err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			s.logger.
+				WithOptions(zap.AddStacktrace(zapcore.ErrorLevel)).
+				Error("[Recovery from handler panic]",
```
StarpTech, line 64: Log the error with the Error api

router/pkg/pubsub/datasource/subscription_event_updater.go
```diff
@@ -0,0 +1,161 @@
...
+	subscriptions := s.eventUpdater.Subscriptions()
+	wg := sync.WaitGroup{}
+	updaterCtx, cancel := context.WithDeadline(context.Background(), time.Now().Add(s.timeout))
+	defer cancel()
+
+	done := make(chan struct{})
+
+	go func() {
+		for subCtx, subId := range subscriptions {
+			s.semaphore <- struct{}{} // Acquire slot, blocks if all slots are taken
```
StarpTech: Maybe using https://pkg.go.dev/golang.org/x/sync/semaphore?

router/pkg/pubsub/datasource/subscription_event_updater.go
```diff
@@ -0,0 +1,161 @@
...
+
+		wg.Wait()
+		close(done)
+	}()
+
+	select {
+	case <-done:
+		s.logger.Debug("All subscription updates completed")
+		// All subscriptions completed successfully
+	case <-time.After(s.timeout + timeoutGracePeriod):
```
StarpTech: Could we reuse `updaterCtx` here?

router/pkg/pubsub/kafka/adapter.go
dkorittki: Nope, fix incoming

router/pkg/pubsub/kafka/adapter.go
dkorittki: Its actually not very convinient to use an errgroup with `writeClient.Produce` because you need to deal with a callback function, which returns the result of the Produce operation (its async). It would look something like this:

```go
p.writeClient.Produce(ctx, &kgo.Record{
    Key:     evt.Key,
    Topic:   pubConf.Topic,
    Value:   evt.Data,
    Headers: headers,
}, func(record *kgo.Record, err error) {
    errCh <- err
})

// Spawn a goroutine that waits for the callback error
myErrGroup.Go(func() error {
    return <-errCh
})

[...]

// wait for errgroup to finish later in the code
```

Ideally I would like to pass g.Go directly as the callback but thats not possible.

Sticking to errgroups, another idea is to use `ProduceSync` inside `myErrGroup.Go()` but I would need to dig deep into `ProduceSync` to ensure it is thread-safe and still performant on concurrent calls. Unless I know more about this I would stick to `Produce`. We can always refine this afterwards.

But you brought up something. We should really handle partial failures here. Right now only the latest or first (depends on method implemetation between adapters) is the error we see, all others are hidden. I will implement partial failure handling for all three adapters Publish method.

router/pkg/pubsub/datasource/subscription_datasource.go
dkorittki: `r` can be anything, not just an error type. It wouldn't get displayed if I use `zap.Error()` for `error` field on the log, if `r` is not `error`. I borrowed this method using `zap.Any` from here https://github.com/wundergraph/cosmo/blob/main/router/internal/requestlogger/requestlogger.go#L163

router/pkg/pubsub/kafka/adapter.go
dkorittki: Fixed

router/pkg/pubsub/kafka/adapter.go
dkorittki: Fixed

router/core/graphql_handler.go
dkorittki: Fixed

router/pkg/pubsub/datasource/subscription_event_updater.go
dkorittki: Fixed

router/pkg/pubsub/datasource/subscription_event_updater.go
dkorittki: We can but we need to say goodbye to the timeout grace period. I think this is no problem in reality. We still process the events. Fixed.

router/pkg/pubsub/redis/adapter.go
```diff
@@ -94,19 +85,24 @@ func (p *ProviderAdapter) Shutdown(ctx context.Context) error {
 	return p.conn.Close()
 }
 
-func (p *ProviderAdapter) Subscribe(ctx context.Context, event SubscriptionEventConfiguration, updater resolve.SubscriptionUpdater) error {
+func (p *ProviderAdapter) Subscribe(ctx context.Context, conf datasource.SubscriptionEventConfiguration, updater datasource.SubscriptionEventUpdater) error {
+	subConf, ok := conf.(*SubscriptionEventConfiguration)
+	if !ok {
+		return datasource.NewError("subscription event not support by redis provider", nil)
+	}
```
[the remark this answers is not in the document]
dkorittki: Fixed

router/pkg/pubsub/redis/adapter.go
```diff
@@ -151,41 +151,88 @@ func (p *ProviderAdapter) Subscribe(ctx context.Context, event SubscriptionEvent
 	return nil
 }
 
-func (p *ProviderAdapter) Publish(ctx context.Context, event PublishEventConfiguration) error {
+func (p *ProviderAdapter) Publish(ctx context.Context, conf datasource.PublishEventConfiguration, events []datasource.StreamEvent) error {
+	pubConf, ok := conf.(*PublishEventConfiguration)
+	if !ok {
+		return datasource.NewError("publish event not support by redis provider", nil)
+	}
```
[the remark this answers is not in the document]
dkorittki: Fixed

adr/cosmo-streams-v1.md
dkorittki: Done

router/go.mod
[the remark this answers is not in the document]
dkorittki: go.mod got updated to use latest engine version, which includes the changes needed for Cosmo Streams. Closing comment.

dkorittki:
Query Planner Test keeps failing. This should be clarified before merging. Need to find out why it's failing, does not seem obvious to me.

dkorittki:
Waiting for confirmation that failing query planner checks are indeed not because of the changes in this pull request.

dkorittki:
Discussed internally why the check fails and if it is connected to this issue.
There is a query, which fails the planner check https://github.com/wundergraph/cosmo-celestial/pull/760#user-content-query-plan-check-failures. Usually there should be a reference to the failing check in the pull request comments of the query planner bot but at the time of testing this pull request, there was some work on the query planner automation which could have prevented this. We observed other pull requests having the same problem.

So skipping query planner checks is okay here.

PR #2276, opened 2025-10-14
feat(router): permit log service name to be configured via env var

brokenpip3:
This PR will make the "service" value in router logs configurable through an env var, while keeping the current hardcoded @wundergraph/router as the default.
I was not able to find any existing test for router/cmd/main.go, in case you prefer I can add a couple of test cases to cover this change.

@coderabbitai summary

## Checklist

close: #2277


router/cmd/main.go
```diff
@@ -90,9 +91,14 @@ func Main() {
 
 	logLevelAtomic := zap.NewAtomicLevelAt(result.Config.LogLevel)
 
+	serviceName := *logServiceName
```
StarpTech, line 94: Instead off having this as a flag, it would be more useful to support it in the config file. For this, you need to update https://github.com/wundergraph/cosmo/blob/d6a443e448470dac7391a4e68210d0fcdd95741d/router/pkg/config/config.go#L1006

There you can also enforce the default, make it available as ENV + make it required.

brokenpip3: yeah it make sense, ty, let me do it :)

PR #2288, opened 2025-10-20
fix: improve subscription latency by executing handlers async

dkorittki:
## Summary by CodeRabbit

* **New Features**
  * Added configuration option `max_concurrent_event_receive_handlers` to control concurrent subscription event receive processing.

* **Bug Fixes**
  * Implemented error deduplication and aggregated logging for subscription event handlers.

* **Tests**
  * Added test coverage for error deduplication scenarios and unique error message logging in subscription events.

## Checklist

**This is a change for Cosmo Streams.**
Changed the execution of `OnReceiveHandler` from sequentiel to asynchronous. The number of concurrent handlers running is limited by a semaphore, which is confurable via a new config parameter `events.subscription_hooks.max_concurrent_event_receive_handlers`. The default is max 100 concurrent handlers. 

I found a problem unrelated to this change during development: If all handlers return an error, all errors get logged. When having `x` active subscriptions and the hook errors for each, which I suppose could be quiet common, than we log `x` amount of error logs. I came with a error message deduplication logic: Based on the errors `Error()` we deduplicate logs while handlers are running and once all handlers have finished we log them.

The behaviour of using handlers hasn't changed. From a users POV he can use the handler as he has before.

### DIFF
--- a/router-tests/modules/stream_receive_test.go
+++ b/router-tests/modules/stream_receive_test.go
@@ -35,72 +35,6 @@ func TestReceiveHook(t *testing.T) {
 		errValue  error
 	}
 
-	subscribeAndProduce := func(t *testing.T, xEnv *testenv.Environment, amountSubscribers int) {
-		topics := []string{"employeeUpdated"}
-		events.KafkaEnsureTopicExists(t, xEnv, time.Second, topics...)
-
-		var subscriptionQuery struct {
-			employeeUpdatedMyKafka struct {
-				ID      float64 `graphql:"id"`
-				Details struct {
-					Forename string `graphql:"forename"`
-					Surname  string `graphql:"surname"`
-				} `graphql:"details"`
-			} `graphql:"employeeUpdatedMyKafka(employeeID: 3)"`
-		}
-
-		surl := xEnv.GraphQLWebSocketSubscriptionURL()
-
-		clients := make([]*graphql.SubscriptionClient, amountSubscribers)
-		clientRunChs := make([]chan error, amountSubscribers)
-		subscriptionArgsChs := make([]chan kafkaSubscriptionArgs, amountSubscribers)
-
-		for i := range amountSubscribers {
-			clients[i] = graphql.NewSubscriptionClient(surl)
-			clientRunChs[i] = make(chan error)
-			subscriptionArgsChs[i] = make(chan kafkaSubscriptionArgs, 1)
-
-			idx := i
-			subscriptionID, err := clients[i].Subscribe(&subscriptionQuery, nil, func(dataValue []byte, errValue error) error {
-				subscriptionArgsChs[idx] <- kafkaSubscriptionArgs{
-					dataValue: dataValue,
-					errValue:  errValue,
-				}
-				return nil
-			})
-			require.NoError(t, err)
-			require.NotEmpty(t, subscriptionID)
-
-			go func(i int) {
-				clientRunChs[i] <- clients[i].Run()
-			}(i)
-		}
-
-		xEnv.WaitForSubscriptionCount(uint64(amountSubscribers), Timeout)
-
-		events.ProduceKafkaMessage(t, xEnv, Timeout, topics[0], `{"__typename":"Employee","id": 1,"update":{"name":"foo"}}`)
-
-		// Collect events from all subscribers
-		for i := 0; i < amountSubscribers; i++ {
-			testenv.AwaitChannelWithT(t, Timeout, subscriptionArgsChs[i], func(t *testing.T, args kafkaSubscriptionArgs) {
-				require.NoError(t, args.errValue)
-				require.JSONEq(t, `{"employeeUpdatedMyKafka":{"id":1,"details":{"forename":"Jens","surname":"Neuse"}}}`, string(args.dataValue))
-			})
-		}
-
-		// Close all clients
-		for i := 0; i < amountSubscribers; i++ {
-			require.NoError(t, clients[i].Close())
-			testenv.AwaitChannelWithT(t, Timeout, clientRunChs[i], func(t *testing.T, err error) {
-				require.NoError(t, err)
-			}, "unable to close client before timeout")
-		}
-
-		for i := range subscriptionArgsChs {
-			close(subscriptionArgsChs[i])
-		}
-	}
-
 	t.Run("Test Receive hook is called", func(t *testing.T) {
 		t.Parallel()
 
@@ -797,22 +731,22 @@ func TestReceiveHook(t *testing.T) {
 			{
 				name:           "1 concurrent handler",
 				maxConcurrent:  1,
-				numSubscribers: 5, // 5 batches of 1 (sequential execution)
+				numSubscribers: 5,
 			},
 			{
 				name:           "2 concurrent handlers",
 				maxConcurrent:  2,
-				numSubscribers: 10, // 5 batches of 2
+				numSubscribers: 10,
 			},
 			{
 				name:           "10 concurrent handlers",
 				maxConcurrent:  10,
-				numSubscribers: 20, // 2 batches of 10
+				numSubscribers: 20,
 			},
 			{
 				name:           "20 concurrent handlers",
 				maxConcurrent:  20,
-				numSubscribers: 40, // 2 batches of 20
+				numSubscribers: 40,
 			},
 		}
 
@@ -843,6 +777,13 @@ func TestReceiveHook(t *testing.T) {
 									}
 
 									if current >= int32(tc.maxConcurrent) {
+										// wait to see if the updater spawns too many concurrent handlers
+										deadline := time.Now().Add(300 * time.Millisecond)
+										for time.Now().Before(deadline) {
+											if currentHandlers.Load() > int32(tc.maxConcurrent) {
+												break
+											}
+										}
 										break
 									}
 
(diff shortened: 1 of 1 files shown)


dkorittki:
@coderabbitai review

router/pkg/pubsub/datasource/subscription_event_updater.go
```diff
@@ -31,34 +32,32 @@ func (s *subscriptionEventUpdater) Update(events []StreamEvent) {
...
-		}
+
+	subscriptions := s.eventUpdater.Subscriptions()
+	maxConcurrentHandlers := max(s.hooks.MaxConcurrentOnReceiveHandlers, 1)
+
+	semaphore := make(chan struct{}, maxConcurrentHandlers)
+	for range maxConcurrentHandlers {
+		semaphore <- struct{}{}
+	}
+
```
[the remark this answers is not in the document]
dkorittki: In Go 1.25 max is defined for ints, as well as range over ints. I added a commit to change the semaphore aquire/release.

router-tests/modules/stream_publish_test.go
```diff
@@ -28,6 +28,11 @@ func TestPublishHook(t *testing.T) {
 
 		cfg := config.Config{
 			Graph: config.Graph{},
+			Events: config.EventsConfiguration{
```
alepane21: Why all these changes related to the event receive handler in the tests that are testing the publish hook?

router-tests/modules/stream_receive_test.go
```diff
@@ -41,6 +47,11 @@ func TestReceiveHook(t *testing.T) {
 			Modules: map[string]interface{}{
 				"streamReceiveModule": stream_receive.StreamReceiveModule{},
 			},
+			Events: config.EventsConfiguration{
+				SubscriptionHooks: config.SubscriptionHooksConfiguration{
+					MaxConcurrentEventReceiveHandlers: defaultMaxConcurrentEventReceiveHandlers,
```
alepane21: I think you should just test:
- the default when not specified
- that when you specify a value, it is used
Can't see any point in specify this value each time for every test, but maybe there is a reason, of course.

router/pkg/pubsub/datasource/subscription_event_updater.go
```diff
@@ -31,34 +32,28 @@ func (s *subscriptionEventUpdater) Update(events []StreamEvent) {
...
+	limit := max(s.hooks.MaxConcurrentOnReceiveHandlers, 1)
+	semaphore := make(chan struct{}, limit)
+
+	var (
+		wg    = sync.WaitGroup{}
+		errCh = make(chan error, len(subscriptions))
+	)
+
+	for ctx, subId := range subscriptions {
+		semaphore <- struct{}{} // aquire a slot
```
alepane21: typo: acquire

router/pkg/pubsub/datasource/subscription_event_updater.go
```diff
@@ -31,34 +32,28 @@ func (s *subscriptionEventUpdater) Update(events []StreamEvent) {
...
+		eventsCopy := copyEvents(events)
+		wg.Add(1)
```
alepane21, line 45: I think you can also use wg.Go

router/pkg/pubsub/datasource/subscription_event_updater.go
```diff
@@ -31,34 +32,28 @@ func (s *subscriptionEventUpdater) Update(events []StreamEvent) {
...
+		go s.updateSubscription(ctx, &wg, errCh, semaphore, subId, eventsCopy)
 	}
+
+	go s.deduplicateAndLogErrors(errCh, len(subscriptions))
```
alepane21: Small doubt: we are starting a go routine, but we are not waiting for it to finish. It will range over the errCh until the last line in this method is called (when you close the channel) and only then it will start logging out things. So it is possible that it will complete the execution after this method has completed everything. Is this the intended behaviour?

router/pkg/pubsub/datasource/subscription_event_updater.go
```diff
@@ -97,6 +92,74 @@ func applyStreamEventHooks(
...
+	for i := range in {
+		out[i] = in[i].Clone()
+	}
+	return out
+}
+
+func (s *subscriptionEventUpdater) updateSubscription(ctx context.Context, wg *sync.WaitGroup, errCh chan error, semaphore chan struct{}, subID resolve.SubscriptionIdentifier, events []StreamEvent) {
+	defer wg.Done()
+	defer func() {
+		<-semaphore // release the slot when done
```
alepane21: usually comments always start with a capitalized word, if it is not referring to a method name

router-tests/modules/streams_hooks_combined_test.go
```diff
@@ -34,6 +34,11 @@ func TestStreamsHooksCombined(t *testing.T) {
 
 		cfg := config.Config{
 			Graph: config.Graph{},
+			Events: config.EventsConfiguration{
```
alepane21: Also here, if we are not testing this parameter, we should not explicitly set it, it should just use the default.

router-tests/modules/stream_receive_test.go
```diff
@@ -518,4 +552,217 @@ func TestReceiveHook(t *testing.T) {
...
+					// Each unique error should have been thrown by exactly 1 handler
+					assert.Equal(t, int64(1), fields["amount_handlers"], "each unique error should have amount_handlers = 1")
+				}
+
+				// Verify we got exactly 3 unique error messages
+				assert.Len(t, errorMessages, 3, "should have exactly 3 unique error messages")
+			}
+		})
+	})
 }
```
alepane21, line 893: There are no tests testing that the parallelization works ... I think you should at least check that if set to 1  it process 1 by 1, if set to 2 process 2 by 2.

router-tests/modules/stream_publish_test.go
dkorittki: I did that for consistency but yeah right they don't need to be set there. Will remove them.

router-tests/modules/stream_receive_test.go
dkorittki: When using the router as a user the default is 100 which is enforced [by our config package](https://github.com/wundergraph/cosmo/pull/2288/files#diff-8740661ecded558061e109e5f2fffdade28695fc26eaa8078a3b01e261816770R647) but that is not applied here. I will see how I can still set a default without having to explicitely set them in tests.

router/pkg/pubsub/datasource/subscription_event_updater.go
dkorittki: It's not really intended but tolerated. It means log lines can appear on the console after `s.Update()` already finished updating all subscriptions. In that case the router, under normal circumstances, is still running and can send things to stderr/stdout. The worst thing I could image happening is we don't log those errors because the server has already shut down. Unlikely but hypothetically possible. I will improve it so we wait in `s.Update()` for `s.deduplicateAndLogErrors()` to finish.

router-tests/modules/stream_receive_test.go
dkorittki: Will do!

router/pkg/pubsub/datasource/subscription_event_updater.go
dkorittki: It works but the godoc of `wg.Go` states

> The function f must not panic.

Currently these functions can panic. I could add a recovery to prevent this but that is actually the next pull request on my agenda. If it's okay for you I will add using wg.Go in that pull request and keep it like this for now.

router-tests/modules/stream_receive_test.go
dkorittki: Fixed

router-tests/modules/streams_hooks_combined_test.go
dkorittki: Fixed

router/pkg/pubsub/datasource/subscription_event_updater.go
dkorittki: Fixed

router/pkg/pubsub/datasource/subscription_event_updater.go
dkorittki: Fixed

router/pkg/pubsub/datasource/subscription_event_updater.go
dkorittki: Fixed

router-tests/modules/stream_receive_test.go

router-tests/modules/stream_publish_test.go
dkorittki: Fixed

router/pkg/pubsub/datasource/subscription_event_updater.go
alepane21: Ah, I didn't know that. Yeah it's ok, it was just a suggestion to have the code a bit more clean!

router-tests/modules/stream_receive_test.go
```diff
@@ -518,4 +520,393 @@ func TestReceiveHook(t *testing.T) {
...
+					// Verify batches: within each batch, events should arrive nearly simultaneously
+					// Between batches, there should be ~20ms delay
+					for i := 0; i < tc.numSubscribers; i += tc.maxConcurrent {
+						batchEnd := i + tc.maxConcurrent
+						if batchEnd > tc.numSubscribers {
+							batchEnd = tc.numSubscribers
+						}
+
+						// Check time differences within the batch (should be small, < 10ms)
+						for j := i; j < batchEnd-1; j++ {
```
alepane21: Instead of doing this based on timing, couldn't you "lock" the handler to be sure that is called X different times? I think it is a bit dangerous to use the timing...

router/core/router.go
```diff
@@ -2122,6 +2127,17 @@ func WithDemoMode(demoMode bool) Option {
 	}
 }
 
+func WithSubscriptionHooks(cfg config.SubscriptionHooksConfiguration) Option {
+	return func(r *Router) {
+		// maxConcurrent := cfg.MaxConcurrentEventReceiveHandlers
```
alepane21: Do we need this commented code?

dkorittki: Whoops, its a left over! Will remove

dkorittki: Fixed

router-tests/modules/stream_receive_test.go
dkorittki: I switched to a different approach now. I count parrallel handlers in a variable and whenever I increase the counter, I check if it exceeded the maximum. I detect if we under- or overscaled parrallel running handlers reliably this way. There is only one spot left with a `time.Sleep()` inside the handler to simulate work and force a parrallel execution.

I tried out other sync mechanisms to force parallelism like channels and atomic variables to wait for all concurrent handlers to start, but it made the code very complex. For example I had to watch out not to block forever in case the last round of handlers might not be `maxConcurrentHandlers`, and never will be. That can happen if there is only one more subscriber to update, but we set maxConcurrency to 2, for example. While I also could check for that I scraped this idea in the end, because a simply `time.Sleep(50 * time.Millisecond)` does the trick as well. I think it is very unlikely this will ever not force parrallel handler execution, no matter on which machine we run this.

PR #2293, opened 2025-10-22
feat: support grpc plugins with ts and bun

SkArchon:
This PR implements typescript plugins with bun.

* There are no changes in the router, because we build a binary in the same name format it expects `{platform}-{arch}`

* This PR also removes the template TS constants into template files, which are appended into the previous constants dynamically. This makes it easier to manage a growing number of templates, instead of going through TS strings. Includes a CI check to verify that any template changes are generated and committed.

* To keep thing simple we didn't add a separate folder for the ts router plugin `router-plugin-ts`, this will be done in a follow up pr, after everything is approved.

* We used `bun patch` to override the internal `__dirname` because it is hardcoded to the directory the binary was compiled at on compile time (this especially breaks cosmo cloud plugins since the docker directory is temporary for building). Also we bun patch the `protobufjs/inquire` module as it has not been updated due to a build issue on their repository.

* Adds courses demo

## Summary by CodeRabbit

* **New Features**
  * TypeScript/Bun plugin support added alongside Go: scaffolding, example plugin, tests, and multi-platform build/test flows.
  * Generate/test/build commands are language-aware with clearer error messages and per-language dependency/tool installation.
  * New "courses" demo subgraph and expanded plugin templates for both Go and TypeScript.

* **Chores**
  * CI, install scripts and build tooling updated to support language-specific toolchains and template compilation.

## Checklist

### DIFF
--- a/admission-server/package.json
+++ b/admission-server/package.json
@@ -12,7 +12,7 @@
     "@hono/node-server": "1.13.7",
     "@wundergraph/cosmo-shared": "workspace:^",
     "dotenv": "^16.4.5",
-    "hono": "4.6.12"
+    "hono": "4.10.6"
   },
   "devDependencies": {
     "@types/node": "^20.11.17",
--- a/cdn-server/CHANGELOG.md
+++ b/cdn-server/CHANGELOG.md
@@ -4,6 +4,12 @@ Binaries are attached to the github release otherwise all images can be found [h
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+# [0.15.0](https://github.com/wundergraph/cosmo/compare/cdn@0.14.5...cdn@0.15.0) (2025-11-19)
+
+### Features
+
+* **admission-server,cdn-server:** upgrade hono to v4.10.6 ([#2340](https://github.com/wundergraph/cosmo/issues/2340)) ([9f9faa0](https://github.com/wundergraph/cosmo/commit/9f9faa00eed1027e8d3cbe08e594c94e08028553)) (@pepol)
+
 ## [0.14.5](https://github.com/wundergraph/cosmo/compare/cdn@0.14.4...cdn@0.14.5) (2025-10-03)
 
 **Note:** Version bump only for package cdn
--- a/cdn-server/cdn/CHANGELOG.md
+++ b/cdn-server/cdn/CHANGELOG.md
@@ -4,6 +4,12 @@ Binaries are attached to the github release otherwise all images can be found [h
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+# [0.10.0](https://github.com/wundergraph/cosmo/compare/@wundergraph/cosmo-cdn@0.9.5...@wundergraph/cosmo-cdn@0.10.0) (2025-11-19)
+
+### Features
+
+* **admission-server,cdn-server:** upgrade hono to v4.10.6 ([#2340](https://github.com/wundergraph/cosmo/issues/2340)) ([9f9faa0](https://github.com/wundergraph/cosmo/commit/9f9faa00eed1027e8d3cbe08e594c94e08028553)) (@pepol)
+
 ## [0.9.5](https://github.com/wundergraph/cosmo/compare/@wundergraph/cosmo-cdn@0.9.4...@wundergraph/cosmo-cdn@0.9.5) (2025-10-03)
 
 **Note:** Version bump only for package @wundergraph/cosmo-cdn
--- a/cdn-server/cdn/package.json
+++ b/cdn-server/cdn/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@wundergraph/cosmo-cdn",
-  "version": "0.9.5",
+  "version": "0.10.0",
   "author": {
     "name": "WunderGraph Maintainers",
     "email": "info@wundergraph.com"
@@ -26,7 +26,7 @@
     "dist"
   ],
   "dependencies": {
-    "hono": "4.6.12",
+    "hono": "4.10.6",
     "jose": "^5.2.4"
   },
   "devDependencies": {
--- a/cdn-server/package.json
+++ b/cdn-server/package.json
@@ -1,6 +1,6 @@
 {
   "name": "cdn",
-  "version": "0.14.5",
+  "version": "0.15.0",
   "private": true,
   "author": {
     "name": "WunderGraph Maintainers",
@@ -27,7 +27,7 @@
     "@hono/node-server": "1.13.7",
     "@wundergraph/cosmo-cdn": "workspace:*",
     "dotenv": "^16.4.5",
-    "hono": "4.6.12"
+    "hono": "4.10.6"
   },
   "devDependencies": {
     "@types/node": "^20.9.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -101,7 +101,7 @@ importers:
     dependencies:
       '@hono/node-server':
         specifier: 1.13.7
-        version: 1.13.7(hono@4.6.12)
+        version: 1.13.7(hono@4.10.6)
       '@wundergraph/cosmo-shared':
         specifier: workspace:^
         version: link:../shared
@@ -109,8 +109,8 @@ importers:
         specifier: ^16.4.5
         version: 16.4.5
       hono:
-        specifier: 4.6.12
-        version: 4.6.12
+        specifier: 4.10.6
+        version: 4.10.6
     devDependencies:
       '@types/node':
         specifier: ^20.11.17
@@ -129,7 +129,7 @@ importers:
         version: 3.529.1
       '@hono/node-server':
         specifier: 1.13.7
-        version: 1.13.7(hono@4.6.12)
+        version: 1.13.7(hono@4.10.6)
       '@wundergraph/cosmo-cdn':
         specifier: workspace:*
         version: link:cdn
@@ -137,8 +137,8 @@ importers:
         specifier: ^16.4.5
         version: 16.4.5
       hono:
-        specifier: 4.6.12
-        version: 4.6.12
+        specifier: 4.10.6
+        version: 4.10.6
     devDependencies:
       '@types/node':
         specifier: ^20.9.0
@@ -159,8 +159,8 @@ importers:
   cdn-server/cdn:
     dependencies:
       hono:
-        specifier: 4.6.12
-        version: 4.6.12
+        specifier: 4.10.6
+        version: 4.10.6
       jose:
         specifier: ^5.2.4
         version: 5.2.4
@@ -10542,8 +10542,8 @@ packages:
   hoist-non-react-statics@3.3.2:
     resolution: {integrity: sha512-/gGivxi8JPKWNm/W0jSmzcMPpfpPLc3dY/6GxhX2hQ9iGj3aDfklV4ET7NjKpSinLpJ5vafa9iiGIEZg10SfBw==}
 
-  hono@4.6.12:
-    resolution: {integrity: sha512-eHtf4kSDNw6VVrdbd5IQi16r22m3s7mWPLd7xOMhg1a/Yyb1A0qpUFq8xYMX4FMuDe1nTKeMX5rTx7Nmw+a+Ag==}
+  hono@4.10.6:
+    resolution: {integrity: sha512-BIdolzGpDO9MQ4nu3AUuDwHZZ+KViNm+EZ75Ae55eMXMqLVhDFqEMXxtUe9Qh8hjL+pIna/frs2j6Y2yD5Ua/g==}
     engines: {node: '>=16.9.0'}
 
   hosted-git-info@2.8.9:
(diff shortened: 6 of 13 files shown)


SkArchon:
@coderabbitai full review

cli/src/commands/router/commands/plugin/templates/typescript/plugin.ts.template
```diff
@@ -0,0 +1,50 @@
...
+  // Create the plugin server (health check automatically initialized)
+  const pluginServer = new PluginServer();
+  
+  // Add the {serviceName} service
+  pluginServer.addService({serviceName}Service, {serviceName}Implementation);
+
+  // Start the server
+  pluginServer.serve().catch((error) => {
+    process.exit(1);
+  });
```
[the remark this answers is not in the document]
SkArchon: This is fine since we use the console.log output for a handshake.

cli/src/commands/router/commands/plugin/templates/go/main.go.template
```diff
@@ -0,0 +1,43 @@
...
+func (s *{serviceName}) QueryHello(ctx context.Context, req *service.QueryHelloRequest) (*service.QueryHelloResponse, error) {
+  response := &service.QueryHelloResponse{
+    Hello: &service.World{
+      Id:   strconv.Itoa(s.nextID),
+      Name: req.Name,
+    },
+  }
+  s.nextID++
+  return response, nil
+}
```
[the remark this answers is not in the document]
SkArchon: This is existing, so its fine for now.

cli/src/commands/router/commands/plugin/templates/plugin/Makefile.template
```diff
@@ -0,0 +1,18 @@
+.PHONY: build test generate install-wgc
+
+install-wgc:
+\t@which wgc > /dev/null 2>&1 || npm install -g wgc@latest
+
+make: build
```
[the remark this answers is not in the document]
SkArchon: This is existing, so its fine for now.

SkArchon:
@coderabbitai review

cli/src/commands/router/commands/plugin/templates/typescript.ts
```diff
@@ -0,0 +1,38 @@
+// Templates for typescript (templating is done by pupa)
+// This file is auto-generated. Do not edit manually.
+/* eslint-disable no-template-curly-in-string */
+
+const dockerfileTs =
+  'FROM --platform=$BUILDPLATFORM oven/bun:1.3.0-alpine AS builder\n\n# Multi-platform build arguments\nARG TARGETOS\nARG TARGETARCH\n\nWORKDIR /build\n\n# Copy package files\nCOPY package.json bun.lock* ./\n\n# Install dependencies\nRUN bun install\n\n# Copy all source code\nCOPY src/ ./src/\nCOPY generated/ ./generated/\n\n# Set BUN_TARGET based on OS and architecture\nRUN BUN_TARGET="bun-$TARGETOS-$([ "$TARGETARCH" = "amd64" ] && echo "x64" || echo "$TARGETARCH")" && \\\n    echo "Building for $BUN_TARGET" && \\\n    bun build src/plugin.ts --compile --outfile bin/plugin --target=$BUN_TARGET\n\nFROM --platform=$BUILDPLATFORM scratch\n\nCOPY --from=builder /build/bin/plugin ./{originalPluginName}-plugin\n\nENTRYPOINT ["./{originalPluginName}-plugin"]\n\n';
```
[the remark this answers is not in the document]
SkArchon: this is existing, its fine

cli/src/commands/router/commands/plugin/toolchain.ts
```diff
@@ -17,10 +17,28 @@ import { dataDir } from '../../../../core/config.js';
...
+  'bun-darwin-x64': 'darwin-amd64',
+  'bun-darwin-arm64': 'darwin-arm64',
+  'bun-windows-x64': 'windows-amd64',
+  'bun-linux-x64-musl': 'linux-amd64',
+  'bun-linux-arm64-musl': 'linux-arm64',
+};
+
 const installScriptUrl =
-  'https://raw.githubusercontent.com/wundergraph/cosmo/refs/tags/wgc%400.80.0/scripts/install-proto-tools.sh';
+  'https://raw.githubusercontent.com/wundergraph/cosmo/11347a8300eb71e06348cef203716e560585a20f/scripts/install-proto-tools.sh';
```
SkArchon: To Update

cli/src/commands/router/commands/plugin/commands/build.ts
```diff
@@ -46,29 +51,45 @@ export default (opts: BaseCommandOptions) => {
...
         await checkAndInstallTools(options.forceToolsInstallation);
       }
 
-      // Normalize platform list
-      platforms = normalizePlatforms(options.platform, options.allPlatforms);
-
       // Start the main build process
       spinner.start('Building plugin...');
 
+      switch (language) {
```
JivusAyrus, line 72: why two switch statements? One here and one below

cli/src/commands/router/commands/plugin/commands/generate.ts
```diff
@@ -36,22 +38,35 @@ export default (opts: BaseCommandOptions) => {
...
 
     try {
+      const language = getLanguage(pluginDir);
+
       // Check and install tools if needed
       if (!options.skipToolsInstallation) {
         await checkAndInstallTools(options.forceToolsInstallation);
       }
 
+      switch (language) {
```
JivusAyrus, line 59: same here

cli/src/commands/router/commands/plugin/commands/build.ts
SkArchon: We need to install the typescript related `node_modules` BEFORE generating GRPC Code since it has some proto related binraries needed for proto js/ts generation. However for go, since we are running `go mod tidy` we need to do that after generating the code, since otherwise it will error out with
```
does not contain package github.com/wundergraph/cosmo/plugin/generated
```

cli/src/commands/router/commands/plugin/commands/build.ts
```diff
@@ -21,54 +25,71 @@ export default (opts: BaseCommandOptions) => {
...
+        case 'ts': {
+          await installTsDependencies(pluginDir, spinner);
+          break;
+        }
+      }
```
Noroth, line 81: Why is this a switch and not an `if`? No need for a single case and a `break`

cli/src/commands/router/commands/plugin/commands/build.ts
```diff
@@ -21,54 +25,71 @@ export default (opts: BaseCommandOptions) => {
...
     const pluginName = path.basename(pluginDir);
-    const goModulePath = options.goModulePath;
+
+    let goModulePath = options.goModulePath;
     let platforms: string[] = [];
 
     try {
+      const language = getLanguage(pluginDir);
+
+      goModulePath = validateAndGetGoModulePath(language, goModulePath);
```
Noroth: You could move this down in line 75 and just call 

```ts
await generateProtoAndMapping(pluginDir, validateAndGetGoModulePath(language, options.goModulePath), spinner);
```

cli/src/commands/router/commands/plugin/commands/generate.ts
```diff
@@ -22,36 +25,48 @@ export default (opts: BaseCommandOptions) => {
...
     const spinner = Spinner();
     const pluginName = path.basename(pluginDir);
-    const goModulePath = options.goModulePath;
+
+    let goModulePath = options.goModulePath;
 
     try {
+      const language = getLanguage(pluginDir);
+
+      goModulePath = validateAndGetGoModulePath(language, goModulePath);
```
Noroth: Same here

cli/src/commands/router/commands/plugin/commands/build.ts
```diff
@@ -21,54 +25,71 @@ export default (opts: BaseCommandOptions) => {
...
+          await installTsDependencies(pluginDir, spinner);
+          break;
+        }
+      }
+
+      // Normalize platform list
+      platforms = normalizePlatforms(options.platform, options.allPlatforms, language);
+
       // Generate proto and mapping files
       await generateProtoAndMapping(pluginDir, goModulePath, spinner);
```
Noroth: I personally believe this needs a bit of rework. When we created this function the only intention was to use `Go` but now we are going to provide more languages in the future.

Maybe instead of providing single arguments, we could have options.

Something like:

```ts
interface BaseGenOptions {
  pluginDir: string
}

interface TSProtoGenOptions extends BaseGenOptions  {
  readonly language: 'ts' 
  // Future fields for proto specific generation options
}

interface GoProtoGenOptions extends BaseGenOptions {
  readonly language: 'go'
  goModulePath: string
}

type ProtoGenOptions = TSProtoGenOptions | GoProtoGenOptions

async function generateProtoAndMapping(options: ProtoGenOptions)
```

cli/src/commands/router/commands/plugin/commands/test.ts
```diff
@@ -26,21 +34,44 @@ export default (opts: BaseCommandOptions) => {
...
+            const result = await runGoTests(srcDir, spinner, false);
+            failed = result.failed;
+            break;
+          }
+          case 'ts': {
+            const result = await runTsTests(pluginDir, spinner);
+            failed = result.failed;
+            break;
+          }
+        }
```
Noroth, line 80: Can we improve this now for future implementations? Maybe we could have types that define the language and then have individual functions for installing deps and running test etc.

Something like this, What do you think?
This could also be used in other files to have a more structured approach.

```ts
interface BasePlugin {
  pluginDir: string
}

interface GoPlugin extends BasePlugin {
  kind: PluginKind.Go
  ....
}

interface TSPlugin extends BasePlugin {
  kind: PluginKind.Typescript
  ...
}

type Plugin = GoPlugin | TSPlugin

let plugin = resolvePluginType(pluginDir);
await installDependencies(plugin, spinner);
const result = await runTests(plugin, spinner);

...

async function runTests(plugin: Plugin, spinner: Ora) {
  switch (plugin.kind) {
    ....
  }
}


```

demo/pkg/subgraphs/courses/patches/grpc-health-check@2.1.0.patch
```diff
@@ -0,0 +1,59 @@
+diff --git a/build/src/health.js b/build/src/health.js
```
Noroth, line 1: We need to somehow document this. Having an explanation only in the PR description will eventually cause us to forget why this needs to be patched. This is a very hacky workaround

demo/Makefile
```diff
@@ -10,15 +10,23 @@ compose-integration:
 bump-deps:
 	./bump-deps.sh
 
-PHONY: plugin-build plugin-generate plugin-build-binary plugin-build-integration
+PHONY: plugin-build plugin-generate plugin-build-bun-binary plugin-build-go-binary plugin-build-integration
```
Noroth: Please prefix `PHONY` with a `.`, this was a typo

Noroth:
I don't see any integration tests in the router that call the `Courses` subgraph.

cli/src/commands/router/commands/plugin/commands/build.ts
SkArchon: I kept all the language conditionals as `case` for consistency, under the assumption we would add something later, can change to `if` though for single cases.

Noroth: Yes that makes more sense and removes a few lines of code

cli/src/commands/router/commands/plugin/commands/build.ts
SkArchon: I modified this with protographic, we now pass in [option](https://protobuf.dev/reference/protobuf/proto3-spec/#option) to protographic.

cli/src/commands/router/commands/plugin/commands/build.ts
SkArchon: I ended up added a go case with the protographic / go module path refactroring

cli/src/commands/router/commands/plugin/commands/build.ts
SkArchon: Refactored to pass in proto options to protographic.

cli/src/commands/router/commands/plugin/commands/test.ts
SkArchon: I would say we keep things as is for now, and we handle this as part of refactoring and moving the language specific parts to hooks.

demo/pkg/subgraphs/courses/patches/grpc-health-check@2.1.0.patch
SkArchon: I added a comment in the cli in init.ts, will also add in documentation

I removed the fs-polyfill.ts (was not working properly when testing) and replaced it with a patch for [this](https://github.com/protobufjs/protobuf.js/blob/master/lib/inquire/index.js), found out that there build step is broken. Last update 9 years ago: https://www.npmjs.com/package/@protobufjs/inquire/v/1.1.0, related complaints: https://github.com/protobufjs/protobuf.js/issues/2099

SkArchon:
FYI: Discussed with Dustin that we only install tools specific per language (right now we install bun when using go plugins), will update and push soon

SkArchon:
@coderabbitai review

PR #2296, opened 2025-10-23
fix: omit introspection queries for complexity limits

ysmolski:
Introspection queries are not considered when calculating complexity limits.

## Summary by CodeRabbit

* **New Features**
  * Added a configuration option to exclude introspection queries from complexity checks (ignore_introspection / skip introspection).

* **Bug Fixes**
  * Introspection queries can be exempted from depth/complexity limits when the new option is enabled.

* **Tests**
  * Added tests verifying complexity-limit behavior for introspection (both enforced and skipped).

* **Chores**
  * Bumped GraphQL tooling dependency to a newer release.

### DIFF
--- a/router/pkg/config/testdata/config_full.json
+++ b/router/pkg/config/testdata/config_full.json
@@ -711,7 +711,7 @@
         "Limit": 4,
         "IgnorePersistedOperations": true
       },
-      "SkipIntrospection": false
+      "IgnoreIntrospection": false
     },
     "DepthLimit": null,
     "ParserLimits": {


router/pkg/config/testdata/config_full.json
```diff
@@ -710,7 +710,8 @@
         "Enabled": true,
         "Limit": 4,
         "IgnorePersistedOperations": true
-      }
+      },
+      "SkipIntrospection": false
```
StarpTech: I'd go with `ignore_introspection` since we use it already for the `authentication` section

```yaml
authentication:
  ignore_introspection: true
```

router/pkg/config/config.go
```diff
@@ -438,10 +438,13 @@ type ComplexityCalculationCache struct {
...
-	TotalFields      *ComplexityLimit `yaml:"total_fields"`
-	RootFields       *ComplexityLimit `yaml:"root_fields"`
-	RootFieldAliases *ComplexityLimit `yaml:"root_field_aliases"`
+	Depth             *ComplexityLimit `yaml:"depth"`
+	TotalFields       *ComplexityLimit `yaml:"total_fields"`
+	RootFields        *ComplexityLimit `yaml:"root_fields"`
+	RootFieldAliases  *ComplexityLimit `yaml:"root_field_aliases"`
+
+	// When true, introspection operations are excluded from all complexity checks.
+	SkipIntrospection bool             `yaml:"skip_introspection" envDefault:"false"`
```
StarpTech: Make it accessible via env

router/pkg/config/config.schema.json
```diff
@@ -2603,6 +2603,11 @@
                   "default": false
                 }
               }
+            },
+            "skip_introspection": {
+              "type": "boolean",
+              "default": false,
+              "description": "Set to true to skip introspection queries when calculating complexity limits."
```
StarpTech: > When set to true, complexity validation is ignored for all introspection queries.

router/pkg/config/testdata/config_full.json
ysmolski: done

router/pkg/config/config.go
ysmolski: done

router/pkg/config/config.schema.json
ysmolski: Thanks for the right wording.

PR #2300, opened 2025-10-24
feat: add getter for sha256Hash for modules

SkArchon:
This PR adds the following
* A way for modules (which are outside of the core package) to access the computed sha256Hash
* The Sha256Hash is computed only when it is required, for example when the access log has added the "operation_sha256" context field or an expression uses the sha256Hash, if none of these are true the getter alone would not be useful, thus we add a way for the user to force the computation of the hash from modules. This has to be called from the "RouterOnRequest" hook.

## Summary by CodeRabbit

* **New Features**
  * Added a SHA-256 verification module integrated into request handling
  * Added ability to force SHA-256 computation for operations
  * Exposed operation SHA-256 hashes for downstream use

* **Tests**
  * Added tests validating hash computation, consistency across runs, and force-on behavior

## Checklist

### DIFF
--- a/router/CHANGELOG.md
+++ b/router/CHANGELOG.md
@@ -4,6 +4,13 @@ Binaries are attached to the github release otherwise all images can be found [h
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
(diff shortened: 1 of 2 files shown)


SkArchon:
@coderabbitai generate docstrings

PR #2302, opened 2025-10-24
feat: generate proto from operations and sdl

asoorm:
This adds the ability to collect GraphQL operations (queries/mutations) and emit .proto files so we can auto-generate servers and clients (via gRPC/Buf Connect) to consume the graph.

* Added compileOperationsToProto function to map GraphQL operations -> protobuf AST -> proto text.
* Introduced a new CLI flag `--with-operations` that selects operations-based generation mode.
* field-number management (via a lock/manager) to avoid collisions and ensure deterministic ordering.
* fragment-denormalization support: named fragments and inline fragments are resolved
* Query operations now optionally support an --idempotent-queries flag to mark queries as having no side-effects (mapped to option idempotency_level = NO_SIDE_EFFECTS; in the generated proto).

Backwards compatibility
Existing schema-type-based generation unchanged. New mode opt-in.

## Summary by CodeRabbit

* **New Features**
  * Alpha: generate a single merged proto from GraphQL operation files (--with-operations) with idempotency control, maxDepth protection, operation-name prefixing, language-specific proto options, and custom scalar mappings.

* **CLI / UX**
  * New operation-related CLI options; improved validation, warnings, and results that include generation mode and a dynamic list of generated files.

* **Documentation**
  * Added detailed Operations→Proto guide and updated README examples.

* **Tests**
  * Extensive tests for operations, field-number stability, fragments, enums, builders, validation, recursion, and proto text output.

### DIFF
--- a/cli/src/commands/grpc-service/commands/generate.ts
+++ b/cli/src/commands/grpc-service/commands/generate.ts
@@ -24,8 +24,8 @@ type CLIOptions = {
   packageName?: string;
   protoLock?: string;
   withOperations?: string;
-  queryIdempotency?: string;
   customScalarMapping?: string;
+  customScalarMappingFile?: string;
   maxDepth?: string;
   prefixOperationType?: boolean;
 } & ProtoOptions;
@@ -59,13 +59,13 @@ export default (opts: BaseCommandOptions) => {
       'Subdirectories are traversed recursively. When provided, generates proto from operations instead of SDL types.',
   );
   command.option(
-    '--query-idempotency <level>',
-    'Set idempotency level for Query operations. Valid values: NO_SIDE_EFFECTS, DEFAULT. Only applies with --with-operations.',
+    '--custom-scalar-mapping <json>',
+    'Custom scalar type mappings as JSON string. ' +
+      'Example: \'{"DateTime":"google.protobuf.Timestamp","UUID":"string"}\'',
   );
   command.option(
-    '--custom-scalar-mapping <json-or-path>',
-    'Custom scalar type mappings as JSON string or path to JSON file (prefix file paths with @). ' +
-      'Example: \'{"DateTime":"google.protobuf.Timestamp","UUID":"string"}\' or \'@mappings.json\'',
+    '--custom-scalar-mapping-file <path>',
+    'Path to JSON file containing custom scalar type mappings. ' + 'Example: ./mappings.json',
   );
   command.option(
     '--max-depth <number>',
@@ -124,30 +124,31 @@ async function generateCommandAction(name: string, options: CLIOptions) {
       }
     }
 
-    // Validate and warn about query-idempotency usage
-    let queryIdempotency: string | undefined;
-    if (options.queryIdempotency) {
-      if (!options.withOperations) {
-        spinner.warn('--query-idempotency flag is ignored when not using --with-operations');
-      }
-
-      const validLevels = ['NO_SIDE_EFFECTS', 'DEFAULT'];
-      queryIdempotency = options.queryIdempotency.toUpperCase();
-      if (!validLevels.includes(queryIdempotency)) {
-        program.error(
-          `Invalid --query-idempotency value: ${options.queryIdempotency}. Valid values are: ${validLevels.join(', ')}`,
-        );
-      }
-    }
-
     // Parse custom scalar mappings if provided
     let customScalarMappings: Record<string, string> | undefined;
+    if (options.customScalarMapping && options.customScalarMappingFile) {
+      program.error('Cannot use both --custom-scalar-mapping and --custom-scalar-mapping-file. Please use only one.');
+    }
+
     if (options.customScalarMapping) {
       try {
-        customScalarMappings = await parseCustomScalarMapping(options.customScalarMapping);
+        customScalarMappings = JSON.parse(options.customScalarMapping);
       } catch (error) {
         program.error(
-          `Failed to parse custom scalar mapping: ${error instanceof Error ? error.message : String(error)}`,
+          `Failed to parse custom scalar mapping JSON: ${error instanceof Error ? error.message : String(error)}`,
+        );
+      }
+    } else if (options.customScalarMappingFile) {
+      try {
+        const filePath = resolve(options.customScalarMappingFile);
+        if (!(await exists(filePath))) {
+          program.error(`Custom scalar mapping file not found: ${options.customScalarMappingFile}`);
+        }
+        const fileContent = await readFile(filePath, 'utf8');
+        customScalarMappings = JSON.parse(fileContent);
+      } catch (error) {
+        program.error(
+          `Failed to read or parse custom scalar mapping file: ${error instanceof Error ? error.message : String(error)}`,
         );
       }
     }
@@ -189,7 +190,6 @@ async function generateCommandAction(name: string, options: CLIOptions) {
       languageOptions,
       lockFile: options.protoLock,
       operationsDir: options.withOperations,
-      queryIdempotency,
       customScalarMappings,
       maxDepth,
       prefixOperationType: options.prefixOperationType,
@@ -220,8 +220,8 @@ async function generateCommandAction(name: string, options: CLIOptions) {
       generated: generatedFiles.join(', '),
     };
 
-    if (result.isOperationsMode && queryIdempotency) {
-      resultInfo['query idempotency'] = queryIdempotency;
+    if (result.isOperationsMode) {
+      resultInfo['query idempotency'] = 'NO_SIDE_EFFECTS';
     }
 
     renderResultTree(spinner, 'Generated protobuf schema', true, name, resultInfo);
@@ -242,7 +242,6 @@ type GenerationOptions = {
   languageOptions: ProtoOptions;
   lockFile?: string;
   operationsDir?: string;
-  queryIdempotency?: string;
   customScalarMappings?: Record<string, string>;
   maxDepth?: number;
   prefixOperationType?: boolean;
@@ -256,10 +255,8 @@ type GenerationOptions = {
 async function readOperationFiles(operationsDir: string): Promise<Array<{ filename: string; content: string }>> {
   const files = await readdir(operationsDir, { recursive: true });
   const validExtensions = ['.graphql', '.gql', '.graphqls', '.gqls'];
-  const operationFiles = files.filter((file) => {
-    const ext = extname(file).toLowerCase();
-    return validExtensions.includes(ext);
-  });
+  // Sort files to ensure deterministic output and consistent RPC/method ordering across platforms
+  const operationFiles = files.filter((file) => validExtensions.includes(extname(file).toLowerCase())).sort();
 
   if (operationFiles.length === 0) {
     throw new Error(`No GraphQL operation files (${validExtensions.join(', ')}) found in ${operationsDir}`);
@@ -336,11 +333,11 @@ function mergeProtoRoots(roots: protobuf.Root[], serviceName: string): protobuf.
  * @param packageName - The proto package name
  * @param languageOptions - Language-specific proto options
  * @param lockFile - Path to the proto lock file
- * @param queryIdempotency - Query idempotency level
  * @param customScalarMappings - Custom scalar type mappings
  * @param maxDepth - Maximum recursion depth
  * @param prefixOperationType - Whether to prefix operation types
  * @returns Generation result with proto content and lock data
+ * @note All Query operations are automatically marked with NO_SIDE_EFFECTS idempotency level
  */
 async function generateFromOperations(
   schema: string,
@@ -350,7 +347,6 @@ async function generateFromOperations(
   packageName: string,
   languageOptions: ProtoOptions,
   lockFile: string,
-  queryIdempotency?: string,
   customScalarMappings?: Record<string, string>,
   maxDepth?: number,
   prefixOperationType?: boolean,
@@ -374,7 +370,9 @@ async function generateFromOperations(
         packageName: packageName || 'service.v1',
         ...languageOptions,
         includeComments: true,
-        queryIdempotency: queryIdempotency as 'NO_SIDE_EFFECTS' | 'DEFAULT' | undefined,
+        // All Query operations are automatically marked as NO_SIDE_EFFECTS (idempotent).
+        // This ensures consistent, safe retry behavior for all query operations.
+        queryIdempotency: 'NO_SIDE_EFFECTS',
         lockData: currentLockData,
         customScalarMappings,
         maxDepth,
@@ -495,7 +493,6 @@ async function generateProtoAndMapping({
   languageOptions,
   lockFile = resolve(outdir, 'service.proto.lock.json'),
   operationsDir,
-  queryIdempotency,
   customScalarMappings,
   maxDepth,
   prefixOperationType,
@@ -519,7 +516,6 @@ async function generateProtoAndMapping({
       packageName || 'service.v1',
       languageOptions,
       lockFile,
-      queryIdempotency,
       customScalarMappings,
       maxDepth,
       prefixOperationType,
@@ -536,24 +532,6 @@ async function fetchLockData(lockFile: string): Promise<ProtoLock | undefined> {
 
   const existingLockData = JSON.parse(await readFile(lockFile, 'utf8'));
   return existingLockData == null ? undefined : existingLockData;
-}
-
-/**
- * Parse custom scalar mapping from JSON string or file path
- */
-async function parseCustomScalarMapping(input: string): Promise<Record<string, string>> {
-  // Check if input starts with @ to indicate a file path
-  if (input.startsWith('@')) {
-    const filePath = resolve(input.slice(1));
-    if (!(await exists(filePath))) {
-      throw new Error(`Custom scalar mapping file not found: ${filePath}`);
-    }
-    const fileContent = await readFile(filePath, 'utf8');
-    return JSON.parse(fileContent);
-  }
-
-  // Otherwise, treat as inline JSON
-  return JSON.parse(input);
 }
 
 // Usage of exists from node:fs is not recommended. Use access instead.
--- a/protographic/OPERATIONS_TO_PROTO.md
+++ b/protographic/OPERATIONS_TO_PROTO.md
@@ -117,20 +117,9 @@ gRPC supports idempotency levels to indicate whether operations have side effect
 - **NO_SIDE_EFFECTS**: Safe to retry, doesn't modify state
 - **DEFAULT**: May have side effects, retry with caution
 
-The `queryIdempotency` option explicitly sets the idempotency level for all Query operations:
+When using operations-based generation, all Query operations are automatically marked with `NO_SIDE_EFFECTS` idempotency level, indicating they are safe to retry without side effects.
 
-```typescript
-compileOperationsToProto(operation, schema, {
-  queryIdempotency: 'NO_SIDE_EFFECTS'
-});
-```
-
-**Valid values:**
-- `NO_SIDE_EFFECTS` - Marks queries as safe to retry without side effects
-- `DEFAULT` - Explicitly marks queries with default behavior (may have side effects)
-- Omit the option - No idempotency level is set (default gRPC behavior)
-
-**Note:** Mutations and Subscriptions are never marked with idempotency levels, regardless of this option.
+**Note:** Mutations and Subscriptions are never marked with idempotency levels.
 
 ---
 
@@ -167,7 +156,8 @@ wgc grpc-service generate [name] [options]
 |--------|-------------|
 | `-w, --with-operations <path>` | Path to directory containing `.graphql` or `.gql` operation files. Subdirectories are traversed recursively. Enables operations-based generation. |
 | `--prefix-operation-type` | Prefix RPC method names with operation type (Query/Mutation/Subscription) |
-| `--query-idempotency <level>` | Set idempotency level for Query operations. Valid values: `NO_SIDE_EFFECTS`, `DEFAULT`. Only applies with `--with-operations`. |
+| `--custom-scalar-mapping <json>` | Custom scalar type mappings as inline JSON string. Example: `'{"DateTime":"google.protobuf.Timestamp","UUID":"string"}'` |
+| `--custom-scalar-mapping-file <path>` | Path to JSON file containing custom scalar type mappings. Example: `./mappings.json` |
 
 #### Language-Specific Options
 
@@ -196,14 +186,24 @@ wgc grpc-service generate UserService \
   --prefix-operation-type
 ```
 
-#### With Idempotent Queries
+#### With Custom Scalar Mappings (Inline)
 
 ```bash
 wgc grpc-service generate UserService \
   --input schema.graphql \
   --output ./proto \
   --with-operations ./operations \
-  --query-idempotency NO_SIDE_EFFECTS
+  --custom-scalar-mapping '{"DateTime":"google.protobuf.Timestamp","UUID":"string"}'
+```
+
+#### With Custom Scalar Mappings (File)
+
+```bash
+wgc grpc-service generate UserService \
+  --input schema.graphql \
+  --output ./proto \
+  --with-operations ./operations \
+  --custom-scalar-mapping-file ./scalar-mappings.json
 ```
 
 #### With Go Package
@@ -296,8 +296,12 @@ const result = compileOperationsToProto(operations, schema, {
   packageName: 'myorg.user.v1',
   goPackage: 'github.com/myorg/myapp/proto/user/v1',
   prefixOperationType: true,
-  queryIdempotency: 'NO_SIDE_EFFECTS',
+  queryIdempotency: 'NO_SIDE_EFFECTS',  // All queries are marked as idempotent
   includeComments: true,
+  customScalarMappings: {
+    'DateTime': 'google.protobuf.Timestamp',
+    'UUID': 'string'
+  },
 });
 
 console.log(result.proto);
@@ -549,7 +553,9 @@ Regenerate - the lock file preserves existing field numbers and assigns the next
 
 ### Custom Scalar Mappings
 
-GraphQL custom scalars are mapped to proto types. Common mappings:
+GraphQL custom scalars can be mapped to proto types using either inline JSON or a separate configuration file.
+
+#### Common Scalar Mappings
 
 | GraphQL Scalar | Recommended Proto Type |
 |----------------|----------------------|
@@ -558,6 +564,59 @@ GraphQL custom scalars are mapped to proto types. Common mappings:
 | `JSON` | `google.protobuf.Struct` |
 | `UUID` | `string` |
 | `BigInt` | `int64` |
+
+#### Using Inline JSON
+
+Pass custom scalar mappings directly as a JSON string:
+
+```bash
+wgc grpc-service generate UserService \
+  --input schema.graphql \
+  --output ./proto \
+  --with-operations ./operations \
+  --custom-scalar-mapping '{"DateTime":"google.protobuf.Timestamp","UUID":"string"}'
+```
+
+#### Using a Configuration File
+
+Create a JSON file with your scalar mappings:
+
+**scalar-mappings.json:**
+```json
+{
+  "DateTime": "google.protobuf.Timestamp",
+  "Date": "google.protobuf.Timestamp",
+  "UUID": "string",
+  "JSON": "google.protobuf.Struct",
+  "BigInt": "int64"
+}
+```
+
+Then reference it in your command:
+
+```bash
+wgc grpc-service generate UserService \
+  --input schema.graphql \
+  --output ./proto \
+  --with-operations ./operations \
+  --custom-scalar-mapping-file ./scalar-mappings.json
+```
+
+**Note:** You cannot use both `--custom-scalar-mapping` and `--custom-scalar-mapping-file` simultaneously. Choose one approach based on your needs.
+
+#### In Code
+
+When using the API directly, pass the mappings as an object:
+
+```typescript
+const result = compileOperationsToProto(operations, schema, {
+  customScalarMappings: {
+    'DateTime': 'google.protobuf.Timestamp',
+    'UUID': 'string',
+    'JSON': 'google.protobuf.Struct'
+  }
+});
+```
 
 ### Proto Lock Files
 


protographic/tests/operations/enum-support.test.ts
```diff
@@ -0,0 +1,715 @@
...
+          }
+        }
+      `;
+
+      const { proto, root } = compileOperationsToProto(operation, schema);
+
+      expectValidProto(proto);
+
+      // Should contain enum definition
+      expect(proto).toContain('enum UserStatus {');
```
StarpTech: Using inline snapshots here would be easier read, especially if indentation is also correct.

protographic/src/operations/field-numbering.ts
```diff
@@ -0,0 +1,204 @@
...
+  getLockManager?(): ProtoLockManager | undefined;
+}
+
+/**
+ * Creates a new field number manager instance
+ *
+ * @param lockManager - Optional ProtoLockManager for field number stability
+ * @returns A new field number manager
+ */
+export function createFieldNumberManager(lockManager?: ProtoLockManager): FieldNumberManager {
```
StarpTech, line 76: How will this integrate with the current connect approach? Will you create a separate lock file or will you touch an existing one?

protographic/src/operations/field-numbering.ts
```diff
@@ -0,0 +1,204 @@
...
+  reconcileFieldOrder(messageName: string, fieldNames: string[]): string[];
+
+  /**
+   * Gets the lock manager if available
+   */
+  getLockManager?(): ProtoLockManager | undefined;
+}
+
+/**
+ * Creates a new field number manager instance
```
StarpTech, line 71: How do you handle argument order?

protographic/src/operations/message-builder.ts
```diff
@@ -0,0 +1,449 @@
...
+  options?: MessageBuilderOptions,
+): protobuf.Type {
+  const message = new protobuf.Type(messageName);
+  const fieldNumberManager = options?.fieldNumberManager;
+
+  // First pass: collect all field names that will be in this message
+  const fieldNames: string[] = [];
+  const fieldSelections = new Map<string, { selection: FieldNode; type: GraphQLObjectType }>();
+
+  const collectFields = (selections: readonly any[], currentType: any) => {
```
StarpTech: Can we avoid any here?

protographic/src/operations/message-builder.ts
```diff
@@ -0,0 +1,449 @@
...
+    for (const selection of selections) {
+      if (selection.kind === 'Field') {
+        // Check if type has getFields method (objects and interfaces)
```
StarpTech: Could elaborate on this?

protographic/src/operations/message-builder.ts
```diff
@@ -0,0 +1,449 @@
...
+          const type = options.schema.getType(typeName);
+          if (type && (isObjectType(type) || isInterfaceType(type))) {
+            collectFields(fragmentDef.selectionSet.selections, type);
+          }
+        }
+      }
+    }
+  };
+
+  collectFields(selectionSet.selections, parentType);
```
StarpTech: Do we have a stop condition for recursion?

protographic/src/operations/proto-text-generator.ts
```diff
@@ -0,0 +1,355 @@
...
+    lines.push('');
+  }
+
+  return lines;
+}
+
+/**
+ * Formats a protobuf field as proto text
+ */
+export function formatField(field: protobuf.Field, options?: ProtoTextOptions, indent: number = 1): string[] {
```
StarpTech, line 278: Do we really need to duplicate this logic? For connect we have implemented this already.

cli/src/commands/grpc-service/commands/generate.ts
```diff
@@ -28,20 +41,44 @@ export default (opts: BaseCommandOptions) => {
   command.option('-o, --output <path-to-output>', 'The output directory for the protobuf schema. (default ".").', '.');
   command.option('-p, --package-name <name>', 'The name of the proto package. (default "service.v1")', 'service.v1');
   command.option('-g, --go-package <name>', 'Adds an `option go_package` to the proto file.');
+  command.option('--java-package <name>', 'Adds an `option java_package` to the proto file.');
```
StarpTech: I'd not add this to the CLI before it is not supported.

protographic/tests/operations/field-ordering-stability.test.ts
```diff
@@ -0,0 +1,1262 @@
...
+      expect(requestFields2['id']).toBe(idNumber);
+      expect(requestFields2['name']).toBe(nameNumber);
+      expect(requestFields2['email']).toBe(emailNumber);
+      expect(requestFields2['age']).toBe(ageNumber);
+    });
+
+    test('should handle adding and removing variables', () => {
+      const schema = `
+        type Query {
+          filterUsers(id: ID, name: String, age: Int, email: String, active: Boolean): [User]
```
StarpTech, line 391: How about complex Input Objects and nesting?

protographic/tests/util.ts
```diff
@@ -46,11 +46,110 @@ export function loadProtoFromText(protoText: string): protobufjs.Root {
...
+
+  try {
+    // Try direct lookup first (for flat names or already-correct paths)
+    message = root.lookupType(messageName);
+  } catch (error) {
+    // If that fails, try to convert verbose name to nested path
+    // Strategy: Try different ways to split the name until one works
+
+    // Find the FIRST occurrence of Response/Request suffix
+    const responseIndex = messageName.indexOf('Response');
```
StarpTech: The logic makes many assumptions about unstructured data and has tight coupling to field name conventions like `Response` or `Request`, which are not connected here.

protographic/tests/util.ts
```diff
@@ -46,11 +46,110 @@ export function loadProtoFromText(protoText: string): protobufjs.Root {
...
+        // For 'SearchUsersUsers', try:
+        // 1. 'SearchUsersResponse.SearchUsersUsers' (no split)
+        // 2. 'SearchUsersResponse.SearchUsers.Users' (split after SearchUsers)
+        // 3. 'SearchUsersResponse.Search.Users.Users' (split after Search)
+
+        // Try different ways to split the nested part
+        // Strategy: Split at each capital letter and try all combinations
+
+        // Find all capital letter positions
+        const capitals: number[] = [];
```
StarpTech: You have to explain this to me. The method is super hard to reason about.

protographic/tests/util.ts
```diff
@@ -46,11 +46,110 @@ export function loadProtoFromText(protoText: string): protobufjs.Root {
...
 export function getFieldNumbersFromMessage(root: protobufjs.Root, messageName: string): Record<string, number> {
-  const message = root.lookupType(messageName);
+  let message: protobufjs.Type;
+
+  try {
+    // Try direct lookup first (for flat names or already-correct paths)
+    message = root.lookupType(messageName);
+  } catch (error) {
+    // If that fails, try to convert verbose name to nested path
+    // Strategy: Try different ways to split the name until one works
```
StarpTech: Why do we need to brute force here?

protographic/tests/operations/fragments.test.ts
```diff
@@ -0,0 +1,823 @@
...
+        
+        type User {
+          id: ID!
+          name: String
+          email: String
+        }
+      `;
+
+      const operation = `
+        fragment BasicFields on User {
```
StarpTech, line 222: I do miss tests for interface and union usage, including nested field resolver.

protographic/tests/operations/message-builder.test.ts
```diff
@@ -0,0 +1,386 @@
...
+      const userType = schema.getType('User') as GraphQLObjectType;
+
+      const message = buildMessageFromSelectionSet('UsersResponse', usersSelection.selectionSet, userType, typeInfo);
+
+      expect(message.name).toBe('UsersResponse');
+      expect(message.fields.id).toBeDefined();
+      expect(message.fields.name).toBeDefined();
+    });
+
+    test('should handle field aliases', () => {
```
StarpTech, line 284: Do we also have tests for root aliases or is this forbidden? Do we ensure that only a single operation can be inside a document?

protographic/tests/operations/message-builder.test.ts
```diff
@@ -0,0 +1,386 @@
...
+          user: User
+        }
+        
+        type User {
+          id: ID!
+        }
+      `);
+
+      // This query references a field that doesn't exist
+      const query = parse(`
```
StarpTech, line 360: Tbh this should return an error. These errors will be caught by a GraphQL parser using its validation rules. We should not swallow them and make it look like it will succeed.

protographic/tests/operations/message-builder.test.ts
```diff
@@ -0,0 +1,386 @@
...
+        type Query {
+          user: User
+        }
+        
+        type User {
+          id: ID!
+        }
+      `);
+
+      // This query references a field that doesn't exist
```
StarpTech: What field doesn't exist?

protographic/tests/operations/fragments.test.ts
```diff
@@ -0,0 +1,823 @@
...
+        }
+      `;
+
+      const operation = `
+        fragment UserContact on User {
+          email
+        }
+        
+        query GetUser {
+          user {
```
StarpTech, line 289: Will `__typename` work as well?

protographic/tests/operations/fragments.test.ts
```diff
@@ -0,0 +1,823 @@
...
+            }
+          }
+        }
+      `;
+
+      const { proto, root } = compileOperationsToProto(operation, schema);
+
+      expectValidProto(proto);
+
+      expect(proto).toMatch(/id/);
```
StarpTech: This doesn't test the logic well. There are a lot things that can make this a false positive. I suggest practicing inline snapshots or adopting a more structured testing approach.

cli/src/commands/grpc-service/commands/generate.ts
```diff
@@ -104,9 +211,45 @@ type GenerationOptions = {
...
 };
 
+/**
+ * Read all GraphQL operation files from a directory
+ */
+async function readOperationFiles(operationsDir: string): Promise<string> {
+  const files = await readdir(operationsDir);
+  const operationFiles = files.filter((file) => {
+    const ext = extname(file).toLowerCase();
+    return ext === '.graphql' || ext === '.gql';
```
StarpTech: There is also `.graphqls`, `.gqls`

StarpTech (changes requested):

cli/src/commands/grpc-service/commands/generate.ts
asoorm: OK - I thought that these were for schema files - not operation documents - will update.

protographic/tests/operations/fragments.test.ts
asoorm: Yes - I've added a test for it.

protographic/tests/operations/message-builder.test.ts
asoorm: Going to add validation to ensure single operation per document.
root aliases will mean we cannot reverse when converting the proto msg back to GraphQL.

cli/src/commands/grpc-service/commands/generate.ts
asoorm: This is mainly for buf connect so that customers can generate their clients

Do you mean - just omit Java till it is tested? or should I remove all the languages that I added?

I guess we could tell the customer to add their own package name into the proto.

protographic/src/operations/proto-text-generator.ts
asoorm: @StarpTech Help me understand what you are referring to before I go ahead and make lots of changes...

When you say "For connect we have implemented this already"... 

* formatField function (that you highlighted) is just 10LOC formatting a single protobuf field.
* The entire proto text printer/generator is about ~350LoC that converts protobufjs AST to proto text strings

Connect package seems to only have generated code - not proto printing logic

And sdl-to-proto-visitor builds text strings directly so it can't really be reused because I am converting protobuf AST to proto text.

If you could point me in the right direction before I go down a rabbit hole? 🙏

protographic/src/operations/message-builder.ts
asoorm: Only object and interface types have fields that can be selected - I've updated to include clearer comments and changed approach to:

```
isObjectType(currentType) || isInterfaceType(currentType)
```

protographic/src/operations/message-builder.ts
asoorm: Swapped for SelectionNode

protographic/src/operations/field-numbering.ts
asoorm: My thoughts were that they are two separate APIs / servers with separate purposes. So therefore they should have separate lock files.

SDL to Proto - Conenct API converts GraphQL SDL (federation schema) to proto.
Operations to proto / Persisted Operations converts individual named operations to proto - the proto depends on each operation's shape.

I didn't think too much further - but my approach was:

One lock file per service / proto file.
Each proto generation maintains its own lock
The two systems are independent - SDL to proto lock shouldn't interfere with operations to proto lock - as they are different APIs.

Both use the same ProtoLockManager but maintain separate lock data.

protographic/src/operations/field-numbering.ts
asoorm: In same way as field order - by wrapping it:

```
  /**
   * Reconcile and get the ordered argument names for a field
   *
   * @param fieldPath - Path to the field in the format "TypeName.fieldName"
   * @param availableArgs - Available argument names (used when no lock exists)
   * @returns Ordered array of argument names
   */
  public reconcileArgumentOrder(fieldPath: string, availableArgs: string[]): string[] {
    // Use the regular message field ordering for arguments
    return this.reconcileMessageFieldOrder(fieldPath, availableArgs);
  }
```

```
describe('Request Message Variable Ordering', () => {
    test('should maintain field numbers when variables are reordered', () => {
      const schema = `
        type Query {
          searchUsers(id: ID, name: String, email: String, age: Int): [User]
        }
        
        type User {
          id: ID!
          name: String!
        }
      `;

      // First operation with specific variable order
      const operation1 = `
        query SearchUsers($id: ID, $name: String, $email: String, $age: Int) {
          searchUsers(id: $id, name: $name, email: $email, age: $age) {
            id
            name
          }
        }
      `;

      const result1 = compileOperationsToProto(operation1, schema);
      expectValidProto(result1.proto);

      const root1 = loadProtoFromText(result1.proto);
      const requestFields1 = getFieldNumbersFromMessage(root1, 'SearchUsersRequest');

      const idNumber = requestFields1['id'];
      const nameNumber = requestFields1['name'];
      const emailNumber = requestFields1['email'];
      const ageNumber = requestFields1['age'];

      // Second operation with completely different variable order
      const operation2 = `
        query SearchUsers($age: Int, $email: String, $id: ID, $name: String) {
          searchUsers(id: $id, name: $name, email: $email, age: $age) {
            id
            name
          }
        }
      `;

      const result2 = compileOperationsToProto(operation2, schema, {
        lockData: result1.lockData,
      });
      expectValidProto(result2.proto);

      const root2 = loadProtoFromText(result2.proto);
      const requestFields2 = getFieldNumbersFromMessage(root2, 'SearchUsersRequest');

      // Verify field numbers are preserved
      expect(requestFields2['id']).toBe(idNumber);
      expect(requestFields2['name']).toBe(nameNumber);
      expect(requestFields2['email']).toBe(emailNumber);
      expect(requestFields2['age']).toBe(ageNumber);
    });
```

protographic/src/operations/message-builder.ts
asoorm: introduced stoop condition and also operation validation.

cli/src/commands/grpc-service/commands/generate.ts
StarpTech: Clarified.

asoorm: I've disabled this in the CLI.

cli/src/commands/grpc-service/commands/generate.ts
```diff
@@ -28,20 +43,54 @@ export default (opts: BaseCommandOptions) => {
...
   command.option('-g, --go-package <name>', 'Adds an `option go_package` to the proto file.');
+  // command.option('--java-package <name>', 'Adds an `option java_package` to the proto file.');
+  // command.option('--java-outer-classname <name>', 'Adds an `option java_outer_classname` to the proto file.');
+  // command.option('--java-multiple-files', 'Adds `option java_multiple_files = true` to the proto file.');
+  // command.option('--csharp-namespace <name>', 'Adds an `option csharp_namespace` to the proto file.');
+  // command.option('--ruby-package <name>', 'Adds an `option ruby_package` to the proto file.');
+  // command.option('--php-namespace <name>', 'Adds an `option php_namespace` to the proto file.');
+  // command.option('--php-metadata-namespace <name>', 'Adds an `option php_metadata_namespace` to the proto file.');
+  // command.option('--objc-class-prefix <name>', 'Adds an `option objc_class_prefix` to the proto file.');
+  // command.option('--swift-prefix <name>', 'Adds an `option swift_prefix` to the proto file.');
```
Noroth, line 50: Just add a comment to indicate that this is going to be supported but not part of the alpha release

cli/src/commands/grpc-service/commands/generate.ts
```diff
@@ -28,20 +43,54 @@ export default (opts: BaseCommandOptions) => {
...
+    'Path to directory containing GraphQL operation files (.graphql, .gql, .graphqls, .gqls). ' +
+      'When provided, generates proto from operations instead of SDL types.',
+  );
+  command.option(
+    '--query-idempotency <level>',
+    'Set idempotency level for Query operations. Valid values: NO_SIDE_EFFECTS, DEFAULT. Only applies with --with-operations.',
+  );
+  command.option(
+    '--custom-scalar-mapping <json-or-path>',
+    'Custom scalar type mappings as JSON string or path to JSON file. ' +
```
Noroth: Add a hint that you need to specify `@` when providing a path `@path/to/file.json`

cli/src/commands/grpc-service/commands/generate.ts
```diff
@@ -68,27 +117,114 @@ async function generateCommandAction(name: string, options: CLIOptions) {
...
       goPackage: options.goPackage,
+      javaPackage: options.javaPackage,
+      javaOuterClassname: options.javaOuterClassname,
+      javaMultipleFiles: options.javaMultipleFiles,
+      csharpNamespace: options.csharpNamespace,
+      rubyPackage: options.rubyPackage,
+      phpNamespace: options.phpNamespace,
+      phpMetadataNamespace: options.phpMetadataNamespace,
+      objcClassPrefix: options.objcClassPrefix,
+      swiftPrefix: options.swiftPrefix,
```
Noroth, line 181: We should make this an object or class and pass this as a single element instead of having an infinite amount of parameters

e.g.:

```ts
type JavaOptions = {
  javaPackage?: string;
  javaOuterClassName: string;
}

type GoOptions = {
  goPackage?: string;
}

type PythonOptions = {
  pythonPackage?: string;
}

// and so on

type LanguageOptions = {
  packageName: string; 
  javaOptions?: JavaOptions;
  goOptions?: GoOptions;
  pythonOptions?: PythonOptions;
  // and so on
}

class ProtoOptions {
  constructor(options: LanguageOptions) {
    // TODO validation for all options
  }
}
```

cli/src/commands/grpc-service/commands/generate.ts
```diff
@@ -68,27 +117,114 @@ async function generateCommandAction(name: string, options: CLIOptions) {
...
       lockFile: options.protoLock,
+      operationsDir: options.withOperations,
+      queryIdempotency: options.queryIdempotency?.toUpperCase(),
```
Noroth: Please create a variable that is either undefined or already contains the uppercase values instead of calling upperCase again as it was done in line 138.

cli/src/commands/grpc-service/commands/generate.ts
```diff
@@ -104,9 +240,267 @@ type GenerationOptions = {
...
   goPackage?: string;
+  javaPackage?: string;
+  javaOuterClassname?: string;
+  javaMultipleFiles?: boolean;
+  csharpNamespace?: string;
+  rubyPackage?: string;
+  phpNamespace?: string;
+  phpMetadataNamespace?: string;
+  objcClassPrefix?: string;
+  swiftPrefix?: string;
```
Noroth: put all the options into another type/class

cli/src/commands/grpc-service/commands/generate.ts
```diff
@@ -68,27 +117,114 @@ async function generateCommandAction(name: string, options: CLIOptions) {
...
+    }
+
+    const generatedFiles = [];
+    if (result.mapping) {
+      generatedFiles.push('mapping.json');
+    }
+    generatedFiles.push('service.proto');
+    if (result.lockData) {
+      generatedFiles.push('service.proto.lock.json');
+    }
```
Noroth: You always add service.proto. No need to have an empty array
```suggestion
    const generatedFiles = ['service.proto'];
    if (result.mapping) {
      generatedFiles.push('mapping.json');
    }
    if (result.lockData) {
      generatedFiles.push('service.proto.lock.json');
    }
```

cli/src/commands/grpc-service/commands/generate.ts
```diff
@@ -104,9 +240,267 @@ type GenerationOptions = {
...
+  customScalarMappings?: Record<string, string>;
+  maxDepth?: number;
+  prefixOperationType?: boolean;
 };
 
+/**
+ * Read all GraphQL operation files from a directory
+ * Returns an array of {filename, content} objects
+ */
+async function readOperationFiles(operationsDir: string): Promise<Array<{ filename: string; content: string }>> {
```
Noroth, line 255: nit: In Typescript you have hints in the comments to describe parameters and return types.

example:

```ts
  /**
   * Validate the context provided for the @resolved directive
   * @param ctx - The VisitContext containing the field definition node to check for @resolved directive
   * @param parent - The parent object type definition node
   * @param node - The argument node for the @resolved directive
   * @returns true if we need to continue with the validation, false otherwise
   * @private
   */
```

cli/src/commands/grpc-service/commands/generate.ts
```diff
@@ -104,9 +240,267 @@ type GenerationOptions = {
...
+  const files = await readdir(operationsDir);
+  const operationFiles = files.filter((file) => {
+    const ext = extname(file).toLowerCase();
+    return ext === '.graphql' || ext === '.gql' || ext === '.graphqls' || ext === '.gqls';
```
Noroth: Recommendation: 

initialize and array with valid elements and then use `includes()`
This way you save some pain with `||`
```ts
const validExtensions = ['.graphql', '.gql', .... ]
validExtensions.includes(ext)
```

cli/src/commands/grpc-service/commands/generate.ts
```diff
@@ -104,9 +240,267 @@ type GenerationOptions = {
...
+  });
+
+  if (operationFiles.length === 0) {
+    throw new Error(`No GraphQL operation files (.graphql, .gql, .graphqls, .gqls) found in ${operationsDir}`);
```
Noroth: You can then spread the elements from the array in this error message ;)

```ts
validExtensions.join(' ')
```

cli/src/commands/grpc-service/commands/generate.ts
```diff
@@ -117,36 +511,131 @@ async function generateProtoAndMapping({
...
+      javaPackage,
+      javaOuterClassname,
+      javaMultipleFiles,
+      csharpNamespace,
+      rubyPackage,
+      phpNamespace,
+      phpMetadataNamespace,
+      objcClassPrefix,
+      swiftPrefix,
+    });
```
Noroth: This is probably fine, but

Instead of generating the proto strings we could have a new function that creates a proto AST (containing imports, services, messages and enums) and then we merge everything into one AST and print the AST as a string.

cli/src/commands/grpc-service/commands/generate.ts
```diff
@@ -117,36 +511,131 @@ async function generateProtoAndMapping({
...
-  spinner.text = 'Generating mapping and proto files...';
-  const mapping = compileGraphQLToMapping(schema, serviceName);
-  const proto = compileGraphQLToProto(schema, {
-    serviceName,
-    packageName,
-    goPackage,
-    lockData,
-  });
+  // Determine generation mode
+  if (operationsDir) {
```
Noroth, line 509: Maybe we can extract the logic into two different functions to not have a very big if-else?

protographic/src/operation-to-proto.ts
```diff
@@ -0,0 +1,554 @@
...
+   * Callback for builders to create nested list wrappers
+   * This method is called by request-builder and message-builder when they encounter
+   * a GraphQL type that requires a nested list wrapper
+   *
+   * @param graphqlType - The GraphQL type that needs a wrapper
+   * @returns The wrapper message name
+   */
+  private createNestedListWrapperCallback(graphqlType: any): string {
+    // Import mapGraphQLTypeToProto to get type info
+    const { mapGraphQLTypeToProto } = require('./operations/type-mapper.js');
```
Noroth: We shouldn't have inline requires, this should be in the imports

protographic/src/operations/message-builder.ts
```diff
@@ -0,0 +1,539 @@
...
+  const message = new protobuf.Type(messageName);
+  const fieldNumberManager = options?.fieldNumberManager;
+
+  // First pass: collect all field names that will be in this message
+  const fieldNames: string[] = [];
+  const fieldSelections = new Map<string, { selection: FieldNode; type: GraphQLObjectType | GraphQLInterfaceType }>();
+
+  // Maximum recursion depth to prevent stack overflow
+  const maxDepth = options?.maxDepth ?? 50;
+  const currentDepth = options?._depth ?? 0;
```
Noroth: Magic numbers? Add a comment or put them in a const. We should provide some info why we have a maxDepth of 50 by default.

protographic/src/operations/message-builder.ts
```diff
@@ -0,0 +1,539 @@
...
+        }
+      } else if (selection.kind === 'FragmentSpread' && options?.fragments) {
+        const fragmentDef = options.fragments.get(selection.name.value);
+        if (fragmentDef && options?.schema) {
+          const typeName = fragmentDef.typeCondition.name.value;
+          const type = options.schema.getType(typeName);
+          if (type && (isObjectType(type) || isInterfaceType(type))) {
+            collectFields(fragmentDef.selectionSet.selections, type, depth + 1);
+          }
+        }
```
Noroth: Great moment for a `switch` instead of `if else if else`. Especially with the inner `ifs` it will improve the readability a bit

protographic/src/operations/message-builder.ts
```diff
@@ -0,0 +1,539 @@
...
+
+  // Reconcile field order using lock manager if available
+  let orderedFieldNames = fieldNames;
+  if (fieldNumberManager && 'reconcileFieldOrder' in fieldNumberManager) {
+    orderedFieldNames = fieldNumberManager.reconcileFieldOrder(messageName, fieldNames);
+  }
+
+  // Second pass: process fields in reconciled order
+  // Pre-assign field numbers from lock data if available
+  if (fieldNumberManager?.getLockManager) {
```
Noroth: is this useful or can we do something like

```ts
const lockManager = fieldNumberManager?.getLockManager();
```

and remove the first `if` entirely

protographic/src/operations/message-builder.ts
```diff
@@ -0,0 +1,539 @@
...
+    if (isRepeated) {
+      protoField.repeated = true;
+    }
+
+    if (options?.includeComments && fieldDef.description) {
+      protoField.comment = fieldDef.description;
+    }
+
+    message.add(protoField);
+  }
```
Noroth: I know we are checking field which are object types or scalar types, but it looks like there is a lot of duplicated code here which makes the function quite long. Maybe there is a better way to run the same logic for both and just run additional logic for nested types.

protographic/src/operations/message-builder.ts
```diff
@@ -0,0 +1,539 @@
...
+
+  if (!schema) {
+    // Without schema, we can't resolve the type - skip
+    return;
+  }
+
+  const type = schema.getType(typeName);
+  if (!type || !isObjectType(type)) {
+    // Type not found or not an object type - skip
+    return;
```
Noroth: Why only object type here?

```graphql
query { 
  employees {    
    details {
      forename
      surname
      # I can select on an interface in a fragment spread
      pets {
      	... A
      }
    }
  }
}

# Animal is an interface
fragment A on Animal {
  class
  gender
}
```

protographic/src/operations/proto-text-generator.ts
```diff
@@ -0,0 +1,363 @@
+import protobuf from 'protobufjs';
+
+/**
+ * Extended Method interface that includes custom properties
+ */
+interface MethodWithIdempotency extends protobuf.Method {
+  idempotencyLevel?: 'NO_SIDE_EFFECTS' | 'DEFAULT';
+}
```
Noroth: Declared in multiple files. Is this intended?

protographic/src/operations/proto-text-generator.ts
```diff
@@ -0,0 +1,363 @@
...
+  objcClassPrefix?: string;
+  /** Swift prefix option */
+  swiftPrefix?: string;
+  /** Additional imports to include */
+  imports?: string[];
+  /** Additional options to include */
+  options?: string[];
+  /** Whether to include comments */
+  includeComments?: boolean;
+}
```
Noroth, line 44: This structure or something similar has been declared multiple times already. This is a bit confusing.

protographic/src/operations/proto-text-generator.ts
```diff
@@ -0,0 +1,363 @@
...
+ *
+ * @param root - The protobufjs Root object containing all definitions
+ * @param options - Optional configuration for text generation
+ * @returns The proto text as a string
+ */
+export function rootToProtoText(root: protobuf.Root, options?: ProtoTextOptions): string {
+  const lines: string[] = [];
+
+  // Generate header
+  lines.push(...generateHeader(options));
```
Noroth:
```suggestion
  const lines: string[] = generateHeader(options);
```

protographic/src/operations/proto-text-generator.ts
```diff
@@ -0,0 +1,363 @@
...
+  // Package declaration
+  const packageName = options?.packageName || 'service.v1';
+  lines.push(`package ${packageName};`);
+  lines.push('');
+
+  // Imports
+  const imports = new Set<string>();
+
+  // Add default imports if using wrapper types
+  imports.add('google/protobuf/wrappers.proto');
```
Noroth: Should we always add the import for wrapper types even when we don't use them?

protographic/src/operations/proto-text-generator.ts
```diff
@@ -0,0 +1,363 @@
...
+      methodLine += ' {}';
+      lines.push(methodLine);
+    }
+  }
+
+  lines.push('}');
+  lines.push('');
+
+  return lines;
+}
```
Noroth, line 194: Can this be done with simple string formatting?

Something like see sdl-to-proto-visitor `function buildMessage`
```ts
      messageLines.push(
        this.formatIndent(1, `${repeated}${field.typeName} ${field.fieldName} = ${field.fieldNumber};`),
      );
```

protographic/src/operations/request-builder.ts
```diff
@@ -0,0 +1,307 @@
...
+
+  // Create a map for quick lookup
+  const variableMap = new Map<string, VariableDefinitionNode>();
+  for (const variable of variables) {
+    const protoName = graphqlArgumentToProtoField(variable.variable.name.value);
+    variableMap.set(protoName, variable);
+  }
+
+  // Pre-assign field numbers from lock data if available
+  if (fieldNumberManager?.getLockManager) {
```
Noroth: Can be removed as well

protographic/src/operations/request-builder.ts
```diff
@@ -0,0 +1,307 @@
...
+
+  // Create a map for quick lookup
+  const fieldMap = new Map<string, (typeof fields)[string]>();
+  for (const [fieldName, inputField] of Object.entries(fields)) {
+    const protoFieldName = graphqlFieldToProtoField(fieldName);
+    fieldMap.set(protoFieldName, inputField);
+  }
+
+  // Pre-assign field numbers from lock data if available
+  if (fieldNumberManager?.getLockManager) {
```
Noroth: Can be removed

protographic/src/operations/request-builder.ts
```diff
@@ -0,0 +1,307 @@
...
+  const enumValues = enumType.getValues();
+  for (const enumValue of enumValues) {
+    // Prefix enum values with the enum type name to avoid collisions
+    const protoEnumValue = graphqlEnumValueToProtoEnumValue(enumType.name, enumValue.name);
+    protoEnum.add(protoEnumValue, enumNumber);
+
+    if (options?.includeComments && enumValue.description) {
+      // Note: protobufjs doesn't have direct comment support for enum values
+      // In a full implementation, you'd track these separately for text generation
+    }
```
Noroth: Apart from the comment, do we have any need to keep the `if`?

protographic/src/sdl-to-proto-visitor.ts
```diff
@@ -223,6 +242,42 @@ export class GraphQLToProtoTextVisitor {
...
+      this.options.push(`option php_metadata_namespace = "${phpMetadataNamespace}";`);
+    }
+
+    if (objcClassPrefix) {
+      this.options.push(`option objc_class_prefix = "${objcClassPrefix}";`);
+    }
+
+    if (swiftPrefix) {
+      this.options.push(`option swift_prefix = "${swiftPrefix}";`);
+    }
```
Noroth: This logic already exists somewhere. We could have that in a single place

protographic/tests/operations/type-mapper.test.ts
```diff
@@ -0,0 +1,418 @@
...
+
+        expect(result).toMatchInlineSnapshot(`
+          {
+            "isRepeated": true,
+            "isScalar": true,
+            "isWrapper": false,
+            "typeName": "string",
+          }
+        `);
+      });
```
Noroth, line 168: We should take a look at this, because repeated fields in protobuf are non nullable. We are using a ListOf wrapper type in the grpc datasource

protographic/src/operations/proto-text-generator.ts
```diff
@@ -0,0 +1,363 @@
...
+  }
+
+  // Options
+  const protoOptions: string[] = [];
+
+  if (options?.goPackage) {
+    protoOptions.push(`option go_package = "${options.goPackage}";`);
+  }
+
+  if (options?.javaPackage) {
```
SkArchon: I don't have full context, but I did a similar change to protographic though instead of adding more language specific options, I extracted it out, allowing the user to specify any custom proto options

https://github.com/wundergraph/cosmo/pull/2293/files#diff-726735973c8bf593f77055d5fba589b59612fa051e39ab02bcb8aea4cc8332a3L219-R224

If it makes sense we could align these changes. I extracted it out as we want to allow users to create their own "hooks" to generate router plugins for whichever language.

asoorm:
@coderabbitai generate docstrings

protographic/src/operations/proto-text-generator.ts
asoorm: Hey @SkArchon i've refactored based on previous comments from @Noroth ...

I took similar approach, but kept it simpler - just `options?: string[]`.

This gives the user most flexibility to do things like:

```
compileOperationsToProto(ops, schema, {
  options: [
    'option java_package = "com.example";',
    'option (google.api.http) = { get: "/v1" };'
  ]
});
```

What do you think if we align with this approach?

cli/src/commands/grpc-service/commands/generate.ts
```diff
@@ -15,10 +17,7 @@ import { BaseCommandOptions } from '../../../core/types/types.js';
 import { renderResultTree, renderValidationResults } from '../../router/commands/plugin/helper.js';
 import { getGoModulePathProtoOption } from '../../router/commands/plugin/toolchain.js';
 
-type CLIOptions = {
-  input: string;
-  output: string;
-  packageName?: string;
+type LanguageOptions = {
```
Noroth, line 17: nit:
Just a question. Why not using the `ProtoOptions` interface from the protographic library? Then you don't need to manage two types.

cli/src/commands/grpc-service/commands/generate.ts
```diff
@@ -30,20 +50,55 @@ export default (opts: BaseCommandOptions) => {
...
+  // command.option('--swift-prefix <name>', 'Adds an `option swift_prefix` to the proto file.');
   command.option(
     '-l, --proto-lock <path-to-proto-lock>',
     'The path to the existing proto lock file to use as the starting point for the updated proto lock file. ' +
       'Default is to use and overwrite the output file "<outdir>/service.proto.lock.json".',
   );
+  command.option(
+    '-w, --with-operations <path-to-operations>',
+    'Path to directory containing GraphQL operation files (.graphql, .gql, .graphqls, .gqls). ' +
+      'When provided, generates proto from operations instead of SDL types.',
```
StarpTech: Does it accept patterns or work recursively?

cli/src/commands/grpc-service/commands/generate.ts
```diff
@@ -30,20 +50,55 @@ export default (opts: BaseCommandOptions) => {
...
+  );
+  command.option(
+    '--query-idempotency <level>',
```
StarpTech: Is `NO_SIDE_EFFECTS` the default?

cli/src/commands/grpc-service/commands/generate.ts
```diff
@@ -166,6 +550,24 @@ async function fetchLockData(lockFile: string): Promise<ProtoLock | undefined> {
   return existingLockData == null ? undefined : existingLockData;
 }
 
+/**
+ * Parse custom scalar mapping from JSON string or file path
+ */
+async function parseCustomScalarMapping(input: string): Promise<Record<string, string>> {
+  // Check if input starts with @ to indicate a file path
+  if (input.startsWith('@')) {
```
StarpTech: Is that a common convention or why did we not introduce two cli flags?

StarpTech (changes requested):

cli/src/commands/grpc-service/commands/generate.ts
asoorm: Updated

cli/src/commands/grpc-service/commands/generate.ts
asoorm: Omitted: Same behavior as `DEFAULT`, but implicit.

DEFAULT: Explicitly adds

`option idempotency_level = DEFAULT;`

NO_SIDE_EFFECTS: Explicitly adds

`option idempotency_level = NO_SIDE_EFFECTS;`

So omission is equivalent to DEFAULT.

cli/src/commands/grpc-service/commands/generate.ts
asoorm: I borrowed the convention from cURL - where you use `-d '@filename'`.

I chose this approach because when you pass JSON directly in CLI args, it can get painful - with the need to escape quotes.

also, single flag is simpler than having separate flags `--custom-scalars` + `--custom-scalrars-file`

And the `@` prefix makes it clear that you are referencing a file rather than inline json.

examples:

```
--custom-scalars "{\"google.protobuf.Timestamp\":\"string\",\"custom.UUID\":\"string\"}"
```

```
--custom-scalars '{"google.protobuf.Timestamp":"string","custom.UUID":"string"}'
```

```
--custom-scalars @scalars.json
```

StarpTech: I would agree, but people need to understand the syntax first; it is not something everybody is used to or smt we already practice in the CLI.

cli/src/commands/grpc-service/commands/generate.ts
StarpTech: ok

PR #2304, opened 2025-10-26
feat: distinguish between read / write events (Cosmo Streams)

dkorittki:
## Summary by CodeRabbit

* **Refactor**
  * Internal event handling system refactored to support mutable event operations throughout the pub/sub pipeline, enabling in-flight event transformation in subscription hooks and adapters while maintaining consistency across all supported transports.

## Checklist

# Motivation

This pull is a draft to discuss a way of avoiding lots of allocations being done whenever we call the `OnReceiveEvents` hook. We do this to isolate event changes between subscribers on that hook, so a change in a hook from one subscriber does not affect the other. The downside is we need to allocate memory for every event for every subscriber. The question is: How many hooks will actually modify the event slice items? Because only then do we need a deep copy of them. If the hooks don't do that we do lots of allocation for nothing.

This PR for now provides the code changes, without touching tests yet. This is only to show the idea and once thats settling, I will do the corrections and adjust tests.

# The idea

Don't copy events before calling OnReceiveEvents hook. Instead make the user aware via type/method names, godoc and cosmo docs he needs to clone events when he wants to modify them to prevent data races. He can use the original events object to read events (and avoid allocations along the way) as long as he doesn't touch the actual event items themself.

# The design

Thx to @alepane21 for this idea!

Instead of providing events as a slice to the hook, we provide a new type `datasource.StreamEvents` to hooks:

```go
OnPublishEvents(
    ctx core.StreamPublishEventHandlerContext,
    events datasource.StreamEvents)
(datasource.StreamEvents, error)
```

You cannot access the slice directly, because we hide the slice in a private struct field.

```go
func (m *CosmoStreamsModule) OnPublishEvents(ctx core.StreamPublishEventHandlerContext, events datasource.StreamEvents) (datasource.StreamEvents, error) {
    // this throws a compile time error
    for i = range events { // cant range over 'events'
        events[i] = myNewEvent // 'events' not indexable
    }
}
```

To access it you have two options:

## `events.All()`

allocation-free iterator without the option to override slice elements.  
Good for getting read access to events.

```go
func (m *CosmoStreamsModule) OnPublishEvents(ctx core.StreamPublishEventHandlerContext, events datasource.StreamEvents) (datasource.StreamEvents, error) {
    for i, evt := range events.All() {
        evt = &kafka.Event{} // ineff assignment
        events.All()[i] = &kafka.Event{} // compile-time error, not indexable
        data := evt.GetData() // returns a copy of data --> thread safe
    }
}
```

To modify the events in a thread safe way, a hook user has to do this

> [!NOTE]  
> Working on getting this easier and more straight forward

```go
func (m *CosmoStreamsModule) OnPublishEvents(ctx core.StreamPublishEventHandlerContext, events datasource.StreamEvents) (datasource.StreamEvents, error) {
    newEvents := make([]datasource.StreamEvent, 0, events.Len())

    for _, evt := range events.All() {
            newEvt := evt.GetUnsafeEvent().Clone()
	    newEvt.SetData([]byte("new event data"))
	    newEvents = append(newEvents, kafka.NewEvent(newEvt.(*kafka.UnsafeEvent)))
    }

    return datasource.NewStreamEvents(newEvents), nil
}
```

## `events.UnsafeStreamEvents()`:

There is a way to access the underlying slice directly for those who really want to.  
For example they need indexable read access to the slice 

```go
func (m *CosmoStreamsModule) OnPublishEvents(ctx core.StreamPublishEventHandlerContext, events datasource.StreamEvents) (datasource.StreamEvents, error) {
    unsafeEvents := events.UnsafeStreamEvents()
    iReallyNeedTheSecondElement := unsafeEvents[2]
}
```

The name implies that doing this is dangerous. It should lead to the user checking godocs and see that this can cause race conditions, if data is modified.

### DIFF
--- a/router/core/subscriptions_modules.go
+++ b/router/core/subscriptions_modules.go
@@ -118,6 +118,10 @@ func (e *EngineEvent) WriteCopy() datasource.MutableStreamEvent {
 	return e.data.Clone()
 }
 
+func (e *EngineEvent) Clone() datasource.MutableStreamEvent {
+	return slices.Clone(e.data)
+}
+
 type engineSubscriptionOnStartHookContext struct {
 	request        *http.Request
 	logger         *zap.Logger


router/pkg/pubsub/kafka/engine_datasource.go
```diff
@@ -15,18 +15,69 @@ import (
...
+	return cloneHeaders(e.evt.Headers)
+}
+
+func (e Event) GetUnsafeEvent() datasource.UnsafeStreamEvent {
+	return e.evt
+}
+
+func NewEvent(evt *UnsafeEvent) datasource.StreamEvent {
+	return &Event{evt: evt}
+}
```
dkorittki: Using this feels a bit weird, because you have pass an unsafe event but creating new events is not unsafe per se:

```go
newEvt := kafka.NewEvent(&kafka.UnsafeEvent{})
```

What you think about using this implementation instead?
```go
func NewEvent(key []byte, data []byte, headers map[string][]byte) datasource.StreamEvent {
	return &Event{
		evt: &UnsafeEvent{
			Key: key,
			Data: data,
			Headers: headers,
		},
	}
}
```

A call would then look like:

```go
// suppose myKey, myData, myHeaders already exist
newEvt := kafka.NewEvent(myKey, myData, myHeaders)
```

This way you dont have to type something with "unsafe" in the name.

router/pkg/pubsub/kafka/engine_datasource.go
```diff
@@ -15,18 +15,69 @@ import (
...
-func (e *Event) GetData() []byte {
+func (e *UnsafeEvent) GetData() []byte {
 	return e.Data
 }
 
-func (e *Event) Clone() datasource.StreamEvent {
+func (e *UnsafeEvent) SetData(data []byte) {
+	if e == nil {
+		return
+	}
```
dkorittki, line 73: I think it should be okay to set `e.Data` to `nil`. If a user really wants to, they should be able to. It could be confusing if you do `myUnsafeEvent.SetData(nil)` and the data isn't actually nil.

router/pkg/pubsub/kafka/engine_datasource.go
```diff
@@ -15,18 +15,69 @@ import (
...
+}
+
+func (e Event) GetHeaders() map[string][]byte {
+	if e.evt == nil {
+		return nil
+	}
+	return cloneHeaders(e.evt.Headers)
+}
+
+func (e Event) GetUnsafeEvent() datasource.UnsafeStreamEvent {
```
dkorittki: I think we can shorten the method name without loosing semantics.

```go
myEvent.GetData() // return copy of data
myEvent.Unsafe().GetData() // get original data
myEvent.Unsafe().SetData(data) // overwrite data
```

Same for `datasource.StreamEvents`´s `UnsafeStreamEvents()` .

router/pkg/pubsub/kafka/engine_datasource.go
```diff
@@ -15,18 +15,69 @@ import (
...
 
-func (e *Event) Clone() datasource.StreamEvent {
+func (e *UnsafeEvent) SetData(data []byte) {
+	if e == nil {
+		return
+	}
+	e.Data = data
+}
+
+func (e *UnsafeEvent) Clone() datasource.UnsafeStreamEvent {
```
dkorittki: Right now this method only exists for `UnsafeEvent`, so whenever a user wants to clone an event they have to access the unsafe event

```go
clonedEvent := myEvent.GetUnsafeEvent().Clone()
```

Maybe we can move away this method from `UnsafeEvent` to `Event`? So a user can do

```go
clonedEvent := myEvent.Clone() // should be a safe operation because the object gets copied
```

The Clone method could return an `datasource.StreamEvent`. Then duplicating and appending a new event would look like this

```go
newEvents := make([]datasource.StreamEvent, 0, events.Len()) // Len() does not exist atm, but easy to add

for _, evt := range events.All() {
	evtCopy := evt.Clone() // returns datasource.StreamEvent
	evtCopy.Unsafe().SetData([]byte("new data")) // slightly ugly we need to resort to call Unsafe(), but I think its still okay!
	newEvents = append(newEvents, evtCopy) // works because evtCopy is a datasource.StreamEvent
}

return datasource.NewStreamEvents(newEvents), nil
```

router/core/subscriptions_modules.go
```diff
@@ -91,19 +91,31 @@ func (c *pubSubSubscriptionOnStartHookContext) WriteEvent(event datasource.Strea
...
 
+type UnsafeEngineEvent []byte
+
+func (e UnsafeEngineEvent) GetData() []byte {
+	return e
+}
+
+func (e UnsafeEngineEvent) SetData(data []byte) {
+	copy(e, data)
+}
```
dkorittki: ℹ This is not about the interface design, just for us to not forget about it later on

This will cause a bug because while copy swaps out the underlying byte data on the existing memory location, it only copies whatever is shorter (`len(e)` vs `len(data)`): https://go.dev/play/p/umbTn0a-NFa

I think the only safe way to do it is pointer receivers.

router/pkg/pubsub/kafka/engine_datasource.go
alepane21: I like it!

router/pkg/pubsub/kafka/engine_datasource.go
alepane21: The point of the check was that if e is nil, setting e.Data would cause a panic.

router/pkg/pubsub/kafka/engine_datasource.go
alepane21: So right now there is no need to get a Clone of a read-only event. Maybe we can make it so that when calling GetUnsafeEvent it is always cloning it? But than if we are cloning it, it is not safe anymore!
Maybe instead of UnsafeEvent we can call it MutableEvent or ChangeableEvent, so the method directly in the read-only event can be .ChangeableEvent() or something like this.
We can than keep the Unsafe name only in the unsafe slice version of the StreamEvents!

router/pkg/pubsub/kafka/engine_datasource.go
alepane21: Make sense, thanks! Just to decide if in the Event we want to keep calling it Unsafe or Changeable (or something like this) as better explained in the other comment.

router/core/subscriptions_modules.go
alepane21: nice catch, thanks!

router/pkg/pubsub/kafka/engine_datasource.go
alepane21: We fixed it by adding the ToStreamEvent to the ChangeableEvents.

router/pkg/pubsub/kafka/engine_datasource.go
alepane21: We found a different (hopefully better) way to improve this part!

router/pkg/pubsub/kafka/engine_datasource.go
alepane21: We found a different (hopefully better) way to improve this part!

router/core/subscriptions_modules.go
alepane21: fixed together!

router/pkg/pubsub/kafka/engine_datasource.go
dkorittki: Sorry my bad, I misread that

dkorittki:
@coderabbitai help

dkorittki:
@coderabbitai summary

router/core/subscriptions_modules.go
```diff
@@ -202,12 +214,15 @@ type StreamReceiveEventHandlerContext interface {
 
 type StreamReceiveEventHandler interface {
 	// OnReceiveEvents is called each time a batch of events is received from the provider before delivering them to the
-	// client. So for a single batch of events received from the provider, this hook will be called one time for each
-	// active subscription.
+	// client. So for a single batch of events received from the provider, this hook will be called once for each
+	// active subscribers on the subscription this events belongs to.
+	// Be careful when modifying events, as it's elements are shared between all subscribers of that event.
+	// You can add, remove or sort the slice but modifying the elements itself will likely cause side effects or
+	// race conditions in other handler calls. Use events.Clone() when you need to modify slice elements.
```
alepane21: I think events.Clone() doesn't exist anymore! We had the ChangableEvents() before, but how there is nothing (you have to loop through events and call clone on each instance)

router/pkg/pubsub/datasource/provider.go
```diff
@@ -55,12 +57,44 @@ const (
...
+	return StreamEvents{evts: evts}
+}
+
+// A StreamEvent is a single event coming from or going to an event provider.
 type StreamEvent interface {
+	// GetData returns the data of the event.
 	GetData() []byte
-	Clone() StreamEvent
+	// Clone returns a mutable copy of the event.
+	Clone() MutableStreamEvent
```
alepane21, line 90: I think that is not semantically correct to call Clone a method that is returning a different kind. I would never expect to get something different.
Maybe we should focus on way a customer should call that method? It should call that to get a copy of the event that is editable, right? So maybe we can call it `CopyToMutable()`, `ToMutable()` or something like that.

router/pkg/pubsub/kafka/engine_datasource.go
```diff
@@ -36,6 +84,10 @@ func (e *Event) Clone() datasource.StreamEvent {
 	return &e2
 }
 
+// func (e *Event) ToStreamEvent() datasource.StreamEvent {
```
alepane21: I think this can be removed or uncomment if we need it

router/pkg/pubsub/redis/engine_datasource.go
```diff
@@ -15,21 +15,53 @@ import (
...
+	if e == nil {
+		return nil
+	}
+
+	return &MutableEvent{
 		Data: slices.Clone(e.Data),
 	}
 }
 
+func (e *MutableEvent) ToStreamEvent() datasource.StreamEvent {
```
alepane21: Why we left that here but not in kafka?

router/core/subscriptions_modules.go
```diff
@@ -202,12 +214,15 @@ type StreamReceiveEventHandlerContext interface {
 
 type StreamReceiveEventHandler interface {
 	// OnReceiveEvents is called each time a batch of events is received from the provider before delivering them to the
-	// client. So for a single batch of events received from the provider, this hook will be called one time for each
-	// active subscription.
+	// client. So for a single batch of events received from the provider, this hook will be called once for each
+	// active subscribers on the subscription this events belongs to.
```
alepane21: this hook will be called once for each active subscribers on the subscription this events belongs to -> I preferred the shorter description that we had before. Also I think `subscribers` is wrong, it should be singular.

router/core/subscriptions_modules.go
```diff
@@ -227,10 +242,10 @@ type StreamPublishEventHandler interface {
 	// OnPublishEvents is called each time a batch of events is going to be sent to the provider
 	// Returning an error will result in a GraphQL error being returned to the client, could be customized returning a
 	// StreamHookError.
```
alepane21, line 249: I think we should add also here some instruction about how to change the events how we did in the OnReceiveEvents hook!

router/core/subscriptions_modules.go
dkorittki: Yes I removed it because I think looping through events with `events.All()` and having to clone the event is simple enough. We can reduce the api surface this way, and to me this feels very intuitiv. If users complain about this during user-tests we can still add it back without breaking the api :)

router/pkg/pubsub/datasource/provider.go
dkorittki: Yeah had the same thoughts. Talked with @StarpTech about it but we came to the conclusion that `Clone()` still feels like the best name. It's true that usually a clone function returns its own type and we don't do this here. But from a conceptional POV a `StreamEvent` and a `MutableStreamEvent` are both stream events but for different contexts. So in that sense `Clone()` does return itself even if they are technically different types.. kinda

Anyway from all bad names to choose from we found that this is probably the most intuitiv one :D

router/core/subscriptions_modules.go
dkorittki: Added

router/core/subscriptions_modules.go
dkorittki: Fixed, can you take a look at it again?

router/pkg/pubsub/redis/engine_datasource.go
dkorittki: Whoops, forgot this one, thanks

router/pkg/pubsub/kafka/engine_datasource.go
dkorittki: Removed it

router/core/subscriptions_modules.go
dkorittki: Can you think of a use case where a user would benefit from having such a method on `events`?

router/pkg/pubsub/datasource/provider.go
dkorittki: Would like to keep it as `Clone()` for now. We can reevaluate during testing phase if it confuses users and rename it then.

router/core/subscriptions_modules.go
alepane21: Nah, I was just pointing out that the comment was outdated.

router/pkg/pubsub/datasource/provider.go
alepane21: Ok, I guess it is just about taste, at the end!

alepane21 (approved):

PR #2305, opened 2025-10-27
feat(mcp): 8245 header forwarding

asoorm:
Adds header forwarding from MCP client through MCP server to the GraphQL router, enabling authentication tokens, tracing headers, and custom metadata to be propagated through the request chain.

## Summary by CodeRabbit

* **New Features**
  * MCP server now forwards all incoming HTTP request headers to downstream GraphQL services, preserving tracing, custom and non-standard headers while ensuring required GraphQL headers are set.
  * Added ability to configure a custom HTTP client for GraphQL requests.

* **Tests**
  * End-to-end tests validate header forwarding in stateless mode, confirming custom/non-standard headers reach subgraphs and return 200 OK.

### DIFF
--- a/router-tests/mcp_test.go
+++ b/router-tests/mcp_test.go
@@ -5,12 +5,14 @@ import (
 	"fmt"
 	"net/http"
 	"strings"
+	"sync"
 	"testing"
 
 	"github.com/mark3labs/mcp-go/mcp"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"github.com/wundergraph/cosmo/router-tests/testenv"
+	"github.com/wundergraph/cosmo/router/core"
 	"github.com/wundergraph/cosmo/router/pkg/config"
 )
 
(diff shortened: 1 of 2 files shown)


router-tests/mcp_test.go
```diff
@@ -553,4 +555,126 @@ func TestMCP(t *testing.T) {
...
+			// Strategy:
+			// 1. Enable stateless mode (no session ID required)
+			// 2. Make direct HTTP POST requests with custom headers
+			// 3. Verify custom headers reach the subgraphs
+			
+			var capturedSubgraphRequest *http.Request
+			var subgraphMutex sync.Mutex
+			
+			testenv.Run(t, &testenv.Config{
+				EnableNats: true,
```
StarpTech: Why is NATS required for the test?

router-tests/mcp_test.go
```diff
@@ -553,4 +555,126 @@ func TestMCP(t *testing.T) {
 			})
 		})
 	})
+
+	t.Run("Header Forwarding", func(t *testing.T) {
+		t.Run("MCP server forwards all headers including custom headers like foo:bar", func(t *testing.T) {
```
StarpTech: The title sounds weird. Why is foo:bar different from other headers when we use wildcard?

router-tests/mcp_test.go
asoorm: Artefact from copy paste other test - will remove. 🙏

PR #2306, opened 2025-10-27
feat(router): provide an agnostic way of creating events in hooks (Cosmo Streams)

dkorittki:
## Checklist

> [!NOTE]  
> First code design idea as a basis for discussion. Once the design is settled I will edit and extend tests.

# Motivation

Currently it is relatively "uncomfortable" to create new events in Cosmo Streams hooks.
You have to know what type of message broker is being at play when a hook is called on order to then create a broker-specific event like `kafka.Event`, `nats.Event` or `redis.Event`.
While it's certainly possible to do this, it is not very user-friendly.
It would be great to have a way for hook users to create new events without needing to know what provider the hook is associated with.

# Implementation Design

This PR extends the handler contexts with a new method: `NewEvent(data []byte) datasource.StreamEvent`.
It allows a hook user to create new events independently of providers:

```go
func (m *CosmoStreamsModule) OnPublishEvents(ctx core.StreamPublishEventHandlerContext, events []datasource.StreamEvent) ([]datasource.StreamEvent, error) {
	newEvent := ctx.NewEvent([]byte(`{"__typename": "Employee", "id": 3}`))
	events = append(events, newEvent)
	return events, nil
}

func (m *CosmoStreamsModule) OnReceiveEvents(ctx core.StreamReceiveEventHandlerContext, events []datasource.StreamEvent) ([]datasource.StreamEvent, error) {
	newEvent := ctx.NewEvent([]byte(`{"__typename": "Employee", "id": 3}`))
	events = append(events, newEvent)
	return events, nil
}

func (m *CosmoStreamsModule) SubscriptionOnStart(ctx core.SubscriptionOnStartHandlerContext) error {
	firstEvent := ctx.NewEvent([]byte(`{"__typename": "Employee", "id": 3}`))
	ctx.WriteEvent(firstEvent)
	return nil
}
```

The above code works for all providers. Depending on the provider being at play when the hook is called, `ctx.NewEvent` returns either a concrete `kafka.Event`, `nats.Event` or `redis.Event`, hidden behind the `datasource.StreamEvent` interface.

## Release Notes

* **Refactor**
  * Updated event handling pipeline in subscription and publishing workflows with improved event construction capabilities.
  * Refined hook function signatures across multiple pub/sub implementations to support enhanced event creation.
  * Extended context interfaces with new methods for raw event construction.

### DIFF
--- a/router/core/subscriptions_modules.go
+++ b/router/core/subscriptions_modules.go
@@ -26,8 +26,8 @@ type SubscriptionOnStartHandlerContext interface {
 	// WriteEvent writes an event to the stream of the current subscription
 	// It returns true if the event was written to the stream, false if the event was dropped
 	WriteEvent(event datasource.StreamEvent) bool
-	// NewRawEvent creates a new event that can be used in the subscription.
-	NewRawEvent(data []byte) datasource.MutableStreamEvent
+	// NewEvent creates a new event that can be used in the subscription.
+	NewEvent(data []byte) datasource.MutableStreamEvent
 }
 
 type pubSubPublishEventHookContext struct {
@@ -59,7 +59,7 @@ func (c *pubSubPublishEventHookContext) PublishEventConfiguration() datasource.P
 	return c.publishEventConfiguration
 }
 
-func (c *pubSubPublishEventHookContext) NewRawEvent(data []byte) datasource.MutableStreamEvent {
+func (c *pubSubPublishEventHookContext) NewEvent(data []byte) datasource.MutableStreamEvent {
 	return c.eventBuilder(data)
 }
 
@@ -99,7 +99,7 @@ func (c *pubSubSubscriptionOnStartHookContext) WriteEvent(event datasource.Strea
 	return true
 }
 
-func (c *pubSubSubscriptionOnStartHookContext) NewRawEvent(data []byte) datasource.MutableStreamEvent {
+func (c *pubSubSubscriptionOnStartHookContext) NewEvent(data []byte) datasource.MutableStreamEvent {
 	return c.eventBuilder(data)
 }
 
@@ -169,7 +169,7 @@ func (c *engineSubscriptionOnStartHookContext) WriteEvent(event datasource.Strea
 	return true
 }
 
-func (c *engineSubscriptionOnStartHookContext) NewRawEvent(data []byte) datasource.MutableStreamEvent {
+func (c *engineSubscriptionOnStartHookContext) NewEvent(data []byte) datasource.MutableStreamEvent {
 	return &MutableEngineEvent{data: data}
 }
 
@@ -236,8 +236,8 @@ type StreamReceiveEventHandlerContext interface {
 	Authentication() authentication.Authentication
 	// SubscriptionEventConfiguration the subscription event configuration
 	SubscriptionEventConfiguration() datasource.SubscriptionEventConfiguration
-	// NewRawEvent creates a new event that can be used in the subscription.
-	NewRawEvent(data []byte) datasource.MutableStreamEvent
+	// NewEvent creates a new event that can be used in the subscription.
+	NewEvent(data []byte) datasource.MutableStreamEvent
 }
 
 type StreamReceiveEventHandler interface {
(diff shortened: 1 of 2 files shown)


router/core/subscriptions_modules.go
```diff
@@ -26,6 +26,8 @@ type SubscriptionOnStartHandlerContext interface {
 	// WriteEvent writes an event to the stream of the current subscription
 	// It returns true if the event was written to the stream, false if the event was dropped
 	WriteEvent(event datasource.StreamEvent) bool
+	// NewEvent creates a new event from the given data
```
alepane21: `NewEvent creates a new event from the given data` -> `NewEvent creates a new event that can be used in the subscription` (or something like this)

router/pkg/pubsub/datasource/subscription_datasource.go
```diff
@@ -30,6 +33,7 @@ func (s *PubSubSubscriptionDataSource[C]) UniqueRequestID(ctx *resolve.Context,
 	return s.uniqueRequestID(ctx, input, xxh)
 }
 
+// s = kafka.SubscriptionEventConfiguration
```
alepane21: I guess it's an error!

dkorittki: Yes! Fixed

router/core/subscriptions_modules.go
dkorittki: Thanks, fixed

alepane21 (approved):

dkorittki:
fyi merged #2304 into this and after that one got merged into the topic branch, I also merged that in here as well. This allowed me to work on this branch while the other was still in review. The merge into this branch causes some new commits to appear here because #2304 got squash-merged into the topic branch but when comparing the code diff from this branch against the topic base branch, it only contains the new additions from this PR.

dkorittki:
Based on @StarpTech ´s suggestion I renamed `NewEvent` to `NewRawEvent`. The idea is to have a `NewEvent` method in the future, which accepts some sort of type instead of raw byte data to create an event.

dkorittki:
@coderabbitai review

router/pkg/pubsub/datasource/subscription_event_updater_test.go
```diff
@@ -37,6 +37,11 @@ type receivedHooksArgs struct {
 	cfg    SubscriptionEventConfiguration
 }
 
+// testEventBuilder is a reusable event builder for tests
+func testEventBuilder(data []byte) MutableStreamEvent {
+	return mutableTestEvent(data)
+}
```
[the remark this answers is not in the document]
dkorittki: Please reevaluate this. The type `mutableTestEvent` is defined as `type mutableTestEvent []byte`, so why should I convert `data` to string here?

dkorittki:
Adressed coderrabbits points, which all are valid with the exception of his hallucination.

router/core/subscriptions_modules.go
```diff
@@ -26,6 +26,8 @@ type SubscriptionOnStartHandlerContext interface {
 	// WriteEvent writes an event to the stream of the current subscription
 	// It returns true if the event was written to the stream, false if the event was dropped
 	WriteEvent(event datasource.StreamEvent) bool
+	// NewRawEvent creates a new event that can be used in the subscription.
+	NewRawEvent(data []byte) datasource.MutableStreamEvent
```
alepane21: Not a fan of the name. I think that a customer using this system would just have a question in his mind: What is a raw event? I would keep `NewEvent` or use `NewMutableEvent` if we want to be more correct.

alepane21 (changes requested):

router/core/subscriptions_modules.go
dkorittki: After discussion with @alepane21 we decided to keep the `NewEvent` method name for now.

The idea behind this was to keep the name `NewEvent` reserved for a future implementation, where a hook user can pass a high level object instead of bytes. However the problem with that is that the pubsub package has to deal with these object and according to @alepane21 this is not the right place. We might alternatively provide a marshaller, which can generate the bytes out of a high level type, which you then feed the `ctx.CreateEvent([]byte)` with. I for myself lack the historical knowledge of this package atm and we both want to look into this deeper. For now we need a name, and we have the option to rename this during user tests but before final release of Cosmo Streams v1.

PR #2314, opened 2025-10-31
chore(router): update Cosmo Streams ADR

dkorittki:
## Checklist

This updates the ADR to reflect the latest changes we made on Cosmo Streams. Besides updating the existing code examples I also added two for the OnReceiveEvents and OnPublishEvents hooks.

### DIFF
--- a/adr/cosmo-streams-v1.md
+++ b/adr/cosmo-streams-v1.md
@@ -158,6 +158,20 @@ type StreamPublishEventHandler interface {
 }
 ```
 
+## Immutable vs Mutable events
+
+The design of `StreamEvent` and `MutableStreamEvent` interfaces addresses a critical performance and safety trade-off in the event handling system. When events are received from a provider, they are typically delivered to multiple active subscriptions simultaneously. The `OnReceiveEvents` handler is called once for each active subscription, meaning the same batch of events needs to be processed by multiple handlers concurrently.
+
+The primary design challenge was avoiding unnecessary memory allocations and data copying while maintaining safety guarantees. If we automatically created a deep copy of all events before each handler invocation, the performance cost would be significant, especially under high load with many active subscriptions. However, if we simply passed mutable references to all handlers, we would risk handlers inadvertently modifying shared event data, causing unexpected behavior for other subscribers processing the same events.
+
+The current solution leverages immutability as the default behavior with explicit opt-in mutability. The `StreamEvent` interface is designed to be immutable: the `GetData()` method returns a copy of the payload data, ensuring that read operations are safe by default. When a handler needs to modify an event, it must explicitly call the `Clone()` method to obtain a `MutableStreamEvent`. This creates a conscious decision point where developers understand they are creating a new copy that can be safely modified without affecting other subscriptions.
+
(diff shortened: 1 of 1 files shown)


adr/cosmo-streams-v1.md
```diff
@@ -17,47 +17,64 @@ This ADR describes new hooks that will be added to the router to support more cu
...
+    evts []StreamEvent
+}
+
+func (e StreamEvents) All() iter.Seq2[int, StreamEvent]  // iterator for all events
+func (e StreamEvents) Len() int                          // returns the number of events
+func (e StreamEvents) Unsafe() []StreamEvent             // returns the underlying slice
+
+func NewStreamEvents(evts []StreamEvent) StreamEvents
+
 // each provider will have its own event type with custom fields
```
alepane21: I would change the comment to be more "standard":
`StreamEvent is a read-only event. Every provider will have it's specific implementation with additionals fields`

adr/cosmo-streams-v1.md
```diff
@@ -17,47 +17,64 @@ This ADR describes new hooks that will be added to the router to support more cu
...
 // the StreamEvent interface is used to allow the hooks system to be provider-agnostic
 type StreamEvent interface {
     GetData() []byte
+    Clone() MutableStreamEvent  // returns a mutable copy of the event
```
alepane21: Why not adding a comment over the definition?
`// Clone returns a mutable copy of the event`

adr/cosmo-streams-v1.md
```diff
@@ -17,47 +17,64 @@ This ADR describes new hooks that will be added to the router to support more cu
...
+}
+
+// MutableStreamEvent is a stream event that can be modified
+type MutableStreamEvent interface {
+    StreamEvent
+    SetData([]byte)  // sets the data of the event
```
alepane21: Also here I would add over the method:
`SetData sets the data of the event`

adr/cosmo-streams-v1.md
```diff
@@ -278,6 +309,162 @@ var (
...
+
+// Interface guards
+var (
+    _ core.StreamReceiveEventHandler = (*HeaderFilterModule)(nil)
+)
+```
+
+## Add metadata to published events based on headers
+
+This example shows how to use HTTP headers to add metadata to events before publishing them using the `OnPublishEvents` handler.
```
alepane21: What do we want to do here? Metadata is also a construct of nats, for example.
So if we want to format the content of the event by adding a metadata field this example make sense, but if we want to set the metadata of nats, it does not.
If we do not want to change the nats metadata, isn't this just another example of data mapping? Why add a new example? Maybe you can change the data mapping one!

adr/cosmo-streams-v1.md
dkorittki: Fixed

adr/cosmo-streams-v1.md
dkorittki: Fixed

adr/cosmo-streams-v1.md
dkorittki: It was meant to provide usage examples to OnStreamEvents and OnPublishEvents hooks. I scraped them. It's more something for the Cosmo Streams docs and not the ADR.

PR #2317, opened 2025-10-31
Update otel collector config to send delta values between batches

SGudbrandsson:
The current configuration causes metrics to be a cumulative sum, causing all metrics in WG studio to be inflated beyond belief.

Using a cumulative delta processor will fix this by removing reducing the sum delta for each batch.

Before:
```
batch 1 - sum 10
batch 2 - sum 20
batch 3 - sum 30
```

After:
```
batch 1 - sum 10
batch 2 - sum 10
batch 3 - sum 10
```

This will cause the sum to be correct in the WG studio.

The alternative is to modify the SQL query to do delta calculations.

## Summary by CodeRabbit

* **Chores**
  * Updated OpenTelemetry configuration to enhance telemetry data processing by adding cumulative-to-delta metric conversion in the traces pipeline.

## Checklist

### DIFF
--- a/otelcollector/otel-config.yaml
+++ b/otelcollector/otel-config.yaml
@@ -29,6 +29,7 @@ processors:
     send_batch_size: 10000
     send_batch_max_size: 20000
     timeout: 15s
+  cumulativetodelta: {}
 
 exporters:
   clickhouse:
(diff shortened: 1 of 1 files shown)


SGudbrandsson:
Turns out this is just bad config on our side.
To fix this, set the router exporter temporality to delta.
The default seems to be cumulative.

PR #2318, opened 2025-11-05
fix: cache vars and remapping normalization

devsergiy:
Use cache for variables normalization and variables remapping.

Closes ENG-8477

## Summary by CodeRabbit

* **Performance**
  * Added caching for variable normalization and variable remapping to improve request throughput.

* **Observability**
  * Cache-hit/miss info for normalization and remapping now recorded in telemetry, spans, and response headers.

* **New Features / Debug**
  * Debug response headers consolidated under a single flag to report normalization, remapping, persisted-operation, and plan cache hits.

* **Tests**
  * Added comprehensive tests covering normalization/remapping cache behavior, headers, and telemetry.

* **Chores**
  * Deprecated old per-cache header flags and updated config/schema/defaults to use the consolidated flag.


router/core/graph_server.go
```diff
@@ -723,6 +750,14 @@ func (s *graphMux) Shutdown(ctx context.Context) error {
...
+		s.variablesNormalizationCache.Close()
+	}
+
+	if s.remapVariablesCache != nil {
+		s.remapVariablesCache.Close()
+	}
+
 	if s.complexityCalculationCache != nil {
 		s.complexityCalculationCache.Close()
 	}
```
Noroth: I suggest to move this to a function to reduce the size a little bit. Something like

```go
// Or define a unexported interface with a close function
func closeResources[T io.Closer](caches ...T) {
  for _, cache := range caches {
    if cache != nil {
      cache.Close()
    }
  }
}
```

router/core/operation_processor.go
```diff
@@ -840,7 +857,43 @@ func (o *OperationKit) setAndParseOperationDoc() error {
 	return nil
 }
 
+func (o *OperationKit) normalizeVariablesCacheKey() uint64 {
+	_, _ = o.kit.keyGen.Write(o.kit.doc.Input.Variables)
+
+	// fmt.Println("####### NormalizeVariables: variables len:", len(o.kit.doc.Input.Variables))
+	// fmt.Println("####### NormalizeVariables: variables:", string(o.kit.doc.Input.Variables))
```
Noroth: Should be removed?

There are multiple comments with fmt statements in this file. Should they be removed?

router/core/operation_processor.go
```diff
@@ -749,12 +753,25 @@ func (o *OperationKit) normalizePersistedOperation(clientName string, isApq bool
...
+	variables                json.RawMessage
+	reparse                  bool
+}
+
+type RemapVariablesCacheEntry struct {
+	internalID               uint64
+	normalizedRepresentation string
+	remapVariables           map[string]string
+}
+
```
Noroth, line 797: Can we get a few comments and descriptions here to understand the purpose and the fields?

router/core/graph_server.go
ysmolski: Cache does not satisfy the "io.Closer" interface. I will simplify it by removing redundant checks for nils. Method "Close" has that check.

router/core/operation_processor.go
ysmolski: I have added a couple of comments there

router/core/operation_processor.go
ysmolski: Removed

router-tests/normalization_cache_test.go
```diff
@@ -1,14 +1,161 @@
...
+				})
+				assertCacheHeaders(t, res, expected)
+				require.Equal(t, `{"data":{"employee":{"details":{"pets":[{"name":"Abby","__typename":"Dog","breed":"GOLDEN_RETRIEVER","class":"MAMMAL","gender":"FEMALE"},{"name":"Survivor","__typename":"Pony"}]}}}}`, res.Body)
+			}
+
+			f(cacheHit{false, false, false}, true)
+			f(cacheHit{true, true, true}, true)
+			f(cacheHit{true, true, true}, true)
+			f(cacheHit{false, false, false}, false)
+			f(cacheHit{true, true, true}, true)
```
jensneuse: can you add one more skipMouse: false at the end?

router-tests/normalization_cache_test.go
```diff
@@ -1,14 +1,161 @@
...
+	})
+
+	t.Run("Variables normalization cache - inline value extraction", func(t *testing.T) {
+		t.Parallel()
+		testenv.Run(t, &testenv.Config{}, func(t *testing.T, xEnv *testenv.Environment) {
+			// Inline value gets extracted to variable
+			res := xEnv.MakeGraphQLRequestOK(testenv.GraphQLRequest{
+				Query: `query { employee(id: 1) { id details { forename } } }`,
+			})
+			assertCacheHeaders(t, res, cacheHit{false, false, false})
```
jensneuse: I would prefer if you always write the names of each field, not just "false, false, false". I keep looking back to the definition to evaluate each case.

router/core/graph_server.go
```diff
@@ -506,18 +506,22 @@ func (s *graphServer) setupEngineStatistics(baseAttributes []attribute.KeyValue)
...
+	normalizationCache          *ristretto.Cache[uint64, NormalizationCacheEntry]
+	complexityCalculationCache  *ristretto.Cache[uint64, ComplexityCacheEntry]
+	variablesNormalizationCache *ristretto.Cache[uint64, VariablesNormalizationCacheEntry]
+	remapVariablesCache         *ristretto.Cache[uint64, RemapVariablesCacheEntry]
+	validationCache             *ristretto.Cache[uint64, bool]
+	operationHashCache          *ristretto.Cache[uint64, string]
+
+	accessLogsFileLogger   *logging.BufferedLogger
+	metricStore            rmetric.Store
+	prometheusCacheMetrics *rmetric.CacheMetrics
```
jensneuse, line 524: does it make sense that we have cache metrics twice, once for prom, once for otel?

router/core/graph_server.go
```diff
@@ -709,31 +745,16 @@ func (s *graphMux) configureCacheMetrics(srv *graphServer, baseOtelAttributes []
 }
 
 func (s *graphMux) Shutdown(ctx context.Context) error {
-	var err error
+	s.planCache.Close()
```
jensneuse, line 751: nil check unnecessary?

router/core/graphql_handler.go
```diff
@@ -428,25 +430,22 @@ func (h *GraphQLHandler) WriteError(ctx *resolve.Context, err error, res *resolv
...
-		} else {
-			w.Header().Set(NormalizationCacheHeader, "MISS")
+	s := func(hit bool) string {
+		if hit {
+			return "HIT"
 		}
+		return "MISS"
+	}
+
+	if h.enableNormalizationCacheResponseHeader {
```
jensneuse: Should we make this more generic? sth like enable cache response headers or similar? it's no longer focused solely on normalization

router/core/operation_processor.go
```diff
@@ -772,6 +809,13 @@ func (o *OperationKit) normalizeNonPersistedOperation() (cached bool, err error)
 			o.parsedOperation.NormalizedRepresentation = entry.normalizedRepresentation
 			o.parsedOperation.Type = entry.operationType
 			o.parsedOperation.NormalizationCacheHit = true
+
+			// remove skip include variables from variables
+			// as they were removed during normalization, but still present when we get operation from cache
+			for _, varName := range skipIncludeVariableNames {
```
jensneuse, line 818: why can we not do this processing before storing in the cache?

router/core/operation_processor.go
```diff
@@ -840,7 +877,34 @@ func (o *OperationKit) setAndParseOperationDoc() error {
 	return nil
 }
 
-func (o *OperationKit) NormalizeVariables() ([]uploads.UploadPathMapping, error) {
+func (o *OperationKit) normalizeVariablesCacheKey() uint64 {
+	_, _ = o.kit.keyGen.Write(o.kit.doc.Input.Variables)
+	_, _ = o.kit.keyGen.WriteString(o.parsedOperation.NormalizedRepresentation)
+
+	sum := o.kit.keyGen.Sum64()
+	o.kit.keyGen.Reset()
```
jensneuse, line 905: are we generally using Reset before or after using keyGen?
Could we settle on one pattern and use it everywhere?
maybe we could make keyGen private and add a wrapper API around it to enforce correct behaviour

I know from using this code a bit it's not always clear and our mix of usages could lead to bugs

router/core/operation_processor.go
```diff
@@ -840,7 +877,34 @@ func (o *OperationKit) setAndParseOperationDoc() error {
...
+	return sum
+}
+
+// NormalizeVariables returns a slice of upload mappings if there are any of them present in a query.
+func (o *OperationKit) NormalizeVariables() (cached bool, mapping []uploads.UploadPathMapping, err error) {
```
jensneuse, line 913: I think you could make the description clearer. It reads as if it just returns upload mappings, but in fact it does much more than that.

router/core/operation_processor.go
```diff
@@ -872,50 +934,93 @@ func (o *OperationKit) NormalizeVariables() ([]uploads.UploadPathMapping, error)
...
 	o.parsedOperation.Request.Variables = o.kit.doc.Input.Variables
 
-	return uploadsMapping, nil
+	if o.cache != nil && o.cache.variablesNormalizationCache != nil {
+		entry := VariablesNormalizationCacheEntry{
+			uploadsMapping:           uploadsMapping,
+			id:                       o.parsedOperation.ID,
+			normalizedRepresentation: o.parsedOperation.NormalizedRepresentation,
+			variables:                o.parsedOperation.Request.Variables,
+			reparse:                  true,
```
jensneuse, line 1006: if we need to reparse a query, should we do it after retrieving it from the cache (for every query) or should we do it before putting it into the cache? (once)

router/core/graph_server.go
ysmolski: Yes, Close method has the check:
```go
// Close stops all goroutines and closes all channels.
func (c *Cache[K, V]) Close() {
	if c == nil || c.isClosed.Load() {
		return
	}
	c.Clear()
        // ....
```

router-tests/normalization_cache_test.go
ysmolski: done

router-tests/normalization_cache_test.go
ysmolski: I guess, I was spoiled by the GoLand. Fixing that

router/core/graph_server.go
ysmolski: I have not introduced this metrics. I have only added an empty line and it changed the alignment (formatting) of fields. I asked Ludwig about this...

router/core/graphql_handler.go
ysmolski: I do not like to rename the config option nor introduce two new options for newly addded caching. Do you think that renaming config option is okay? In that case I think it makes sense to have just one option to enable disable headers for all the caches we have, instead of having 3 or 5 config options.

router/core/graph_server.go
Noroth: I remember that I initially tried to have one function that accepts multiple providers. For some reason it didn't work and I decided to split it. 

We could take a look into combining it.

router/core/operation_processor.go
ysmolski: We cannot do that because this particular cache does not include variables. We allow different values of variables for the same normalized entry cache.

ysmolski: I have clarified comments around variables usage in caching

router/core/operation_processor.go
ysmolski: In 9 cases (`o.kit.keyGen`) out of 10 we do `sum` and then `reset`. Only in 1 case (in the loader `batchEntityFetchBuffer.keyGen`) we do reset, small write, and sum. I agree that current behaviour is somewhat error prone. But if we reset (clean) after each Sum, we should be fine. There are not so many usage in general and it is easy to track them all.

ysmolski: Added a comment about that

router/core/operation_processor.go
ysmolski: done

router/core/operation_processor.go
ysmolski: Hmm. I do not see how we can do this before putting it into the cache, since we do not store parsed query in the cache at all.

router/core/operation_processor.go
```diff
@@ -772,6 +811,14 @@ func (o *OperationKit) normalizeNonPersistedOperation() (cached bool, err error)
...
 			o.parsedOperation.NormalizationCacheHit = true
+
+			// Remove skip/include variables because they come directly from the user.
+			// They were removed during normalization, but we did not cache variables,
+			// thus we have to do it every time.
+			for _, varName := range skipIncludeVariableNames {
+				o.parsedOperation.Request.Variables = jsonparser.Delete(o.parsedOperation.Request.Variables, varName)
+			}
+
 			err = o.setAndParseOperationDoc()
```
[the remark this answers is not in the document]
ysmolski: setting `o.parsedOperation.Variables` is not needed because it is not being used anywhere after this point. it is being read before only, in `writeSkipIncludeCacheKeyToKeyGen`.

router/core/graphql_handler.go
jensneuse: I'd rename for clarity

ysmolski: In other parts of the code I found mentions that VariableNormalizations and Remapping are all part of the Operation Normalization.

ysmolski: I have replaced those 3 options with one: `enableCacheResponseHeaders`

jensneuse (approved):

router/core/operation_processor.go
devsergiy: Actually, I was also thinking about a wrapper, which will be reset after calling callback


smth like
```
o.kit.keygen.Generate(func(h *xxhash.Digest) {
   h.WriteString("something")
})
```

Because one of the fixes in this PR was to properly reset the hasher after we had missed resetting it in one of the cases.
There are not so many places where it is used, but it is rather easy to miss reset

ysmolski: I agree with you. I could work on it as next and sepate PR.

router/pkg/config/testdata/config_defaults.json
```diff
@@ -360,14 +360,12 @@
...
       "EnableResolverDebugging": false,
-      "EnablePersistedOperationsCacheResponseHeader": false,
-      "EnableNormalizationCacheResponseHeader": false,
+      "EnableCacheResponseHeaders": false,
       "AlwaysIncludeQueryPlan": false,
       "AlwaysSkipLoader": false
     },
     "EnableSingleFlight": true,
     "EnableRequestTracing": true,
-    "EnableExecutionPlanCacheResponseHeader": false,
```
StarpTech, line 375: Why was it removed? We need to deprecate it

router/pkg/otel/attributes.go
```diff
@@ -36,6 +36,8 @@ const (
 	WgFeatureFlag                      = attribute.Key("wg.feature_flag")
 	WgAcquireResolverWaitTimeMs        = attribute.Key("wg.engine.resolver.wait_time_ms")
 	WgNormalizationCacheHit            = attribute.Key("wg.engine.normalization_cache_hit")
+	WgVariablesNormalizationCacheHit   = attribute.Key("wg.engine.variables_normalization_cache_hit")
+	WgVariablesRemappingCacheHit       = attribute.Key("wg.engine.variables_remapping_cache_hit")
```
StarpTech, line 40: We should show this in the trace UI in the studio too. Please create an issue.

jensneuse (approved):

router/core/graph_server.go
```diff
@@ -1318,26 +1341,26 @@ func (s *graphServer) buildGraphMux(
...
 
+	loaderHooks := NewEngineRequestHooks(
+		gm.metricStore,
+		subgraphAccessLogger,
+		s.tracerProvider,
+		tracingAttExpressions,
+		telemetryAttExpressions,
+		metricAttExpressions,
+		exprManager.VisitorManager.IsSubgraphResponseBodyUsedInExpressions(),
+	)
```
Noroth, line 1385: Why did you extract the hooks into a variable here? It's only used once anyway

router/core/graphql_handler.go
```diff
@@ -441,21 +440,27 @@ func (h *GraphQLHandler) WriteError(ctx *resolve.Context, err error, res *resolv
...
+		if opCtx.variablesNormalizationCacheHit {
+			w.Header().Set(VariablesNormalizationCacheHeader, "HIT")
+		} else {
+			w.Header().Set(VariablesNormalizationCacheHeader, "MISS")
+		}
+		if opCtx.variablesRemappingCacheHit {
+			w.Header().Set(VariablesRemappingCacheHeader, "HIT")
+		} else {
+			w.Header().Set(VariablesRemappingCacheHeader, "MISS")
+		}
```
Noroth, line 458: I suggest a bit of cleanup here:

```go
func (h *GraphQLHandler) setDebugCacheHeader(w http.ResponseWriter, headerName string, hit bool) {
  headerValue := "MISS"
  if hit {
    headerValue = "HIT"
  }

  w.Header().Set(headerName, headerValue)
}

func (h *GraphQLHandler) setDebugCacheHeaders(w http.ResponseWriter, opCtx *operationContext) {
	if h.enableCacheResponseHeaders {
        h.setDebugCacheHeader(w, NormalizationCacheHeader, opCtx.normalizationCacheHit)
        h.setDebugCacheHeader(w, VariablesNormalizationCacheHeader, opCtx.variablesNormalizationCacheHit)
        h.setDebugCacheHeader(w, VariablesRemappingCacheHeader, opCtx.variablesRemappingCacheHit)
        h.setDebugCacheHeader(w, PersistedOperationCacheHeader, opCtx.persistedOperationCacheHit)
        h.setDebugCacheHeader(w, ExecutionPlanCacheHeader, opCtx.planCacheHit)
	}
}

```

router/core/graphql_prehandler.go
```diff
@@ -776,22 +776,14 @@ func (h *PreHandler) handleOperation(w http.ResponseWriter, req *http.Request, v
...
 	 */
 
-	// Normalize the variables returns list of uploads mapping if there are any of them present in a query
-	// type UploadPathMapping struct {
-	// 	VariableName       string - is a variable name holding the direct or nested value of type Upload, example "f"
-	// 	OriginalUploadPath string - is a path relative to variables which have an Upload type, example "variables.f"
-	// 	NewUploadPath      string - if variable was used in the inline object like this `arg: {f: $f}` this field will hold the new extracted path, example "variables.a.f", if it is an empty, there was no change in the path
-	// }
-	uploadsMapping, err := operationKit.NormalizeVariables()
+	cached, uploadsMapping, err := operationKit.NormalizeVariables()
```
Noroth, line 786: Why was the comment removed? Should it be updated to explain how the `NormalizeVariables` function has changed?

router/core/graph_server.go
ysmolski: To make it more readable. Is the new version worse for reading?

router/core/graphql_handler.go
ysmolski: It was almost like you suggest here. But Jens asked me to revert to the verbose version you see in the PR right now. So I will leave it as it is.

router/core/graphql_prehandler.go
ysmolski: These comments did not belong here. I moved the first file of that comment to the corresponding function. And the rest lives with the structure itself.

router/pkg/config/testdata/config_defaults.json
ysmolski: You are right. We cannot remove options like I did. Do you think we should support both, the old and new options at the same time or can we just ignore old debug settings while allowing them in the config?

router/pkg/otel/attributes.go
ysmolski: Done: ENG-8615

router/pkg/config/testdata/config_defaults.json
ysmolski: I made them properly deprecated.

Noroth (approved):

PR #2326, opened 2025-11-12
feat(controlplane): run federated graph composition in worker pool to unblock main event loop

pepol:
- **feat(controlplane): run federated composition in worker pool to unblock main event loop**

## Summary by CodeRabbit

* **New Features**
  * Composition now runs on a dedicated worker pool, improving concurrency and reliability across graph, subgraph, contract, monograph, and feature-flag operations.

* **Chores**
  * Added configurable composeWorkers (maxCount) and integrated worker-pool lifecycle into server startup and graceful shutdown.
  * Composition error/warning payloads now use string/raw-object messages (format change).

* **Tests**
  * Test setup exposes the new compose worker pool for integration testing.

## Checklist

### DIFF
--- a/controlplane/src/workers/compose.ts
+++ b/controlplane/src/workers/compose.ts
@@ -5,6 +5,7 @@
 import { randomUUID } from 'node:crypto';
 import { MessagePort } from 'node:worker_threads';
 import { PlainMessage } from '@bufbuild/protobuf';
+import { RouterConfig } from '@wundergraph/cosmo-connect/dist/node/v1/node_pb';
 import { CompositionError, CompositionWarning } from '@wundergraph/cosmo-connect/dist/platform/v1/platform_pb';
 import { ContractTagOptions, FederationResult, FederationResultWithContracts } from '@wundergraph/composition';
 import { getFederationResultWithPotentialContracts } from '../core/util.js';
@@ -31,7 +32,7 @@ export interface ContractGraph {
   contractResultSuccess: boolean;
   contractGraphId: string;
   contractSchemaVersionId: `${string}-${string}-${string}-${string}-${string}`;
-  // contractRouterExecutionConfig?: RouterConfig;
+  contractRouterConfig: string;
 }
 
 export interface Outputs {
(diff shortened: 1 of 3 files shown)


controlplane/src/core/composition/composition.ts
```diff
@@ -11,6 +11,7 @@ import {
   SupportedRouterCompatibilityVersion,
 } from '@wundergraph/composition';
 import { CompositionOptions } from 'src/types/index.js';
+import child_process from 'node:child_process';
```
Aenimus: You might to do an ignore on the linter for this

controlplane/package.json
```diff
@@ -10,7 +10,7 @@
     "email": "info@wundergraph.com"
   },
   "scripts": {
-    "dev": "tsx watch src/index.ts",
+    "dev": "tsx watch dist/index.js",
```
Aenimus: I don't understand this at all I'm afraid

pepol: This was a testing script. It's reverted now.

controlplane/src/core/composition/composition.ts
pepol: This was part of testing, reverted now.

controlplane/src/workers/compose.ts
```diff
@@ -0,0 +1,60 @@
...
+
+interface Inputs {
+  federatedGraph: FederatedGraphDTO;
+  subgraphsToCompose: SubgraphsToCompose;
+  tagOptionsByContractName: Map<string, ContractTagOptions>;
+  compositionOptions?: CompositionOptions;
+}
+
+export function getWorkerPool(maxCount?: number): Tinypool | undefined {
+  if (isMainThread) {
```
StarpTech: This shouldn't be necessary.

controlplane/src/workers/compose.ts
```diff
@@ -0,0 +1,60 @@
...
+    const filename = import.meta.url.endsWith('.ts') ?
```
StarpTech: What's the purpose of this? In all cases, we need to point to `dist/workers/compose.js` right?

pepol: Yes, however in "production" case, `import.meta.url` will point to `dist/workers/compose.js`. This is just to ensure that in "dev/watch" case, we use the correct module. I'm happy to unify it, however.

pepol: Fixed.

PR #2329, opened 2025-11-12
feat(router): add a timeout for on_receive_event hooks

dkorittki:
## Checklist

# Motivation and Context

I use an synonym example here to demonstrate what this change is all about, which hopefully accurately describes everything.

**The Great Ice Cream Delivery Crisis: A Timeout Tale**

Imagine an ice cream factory (Kafka) that's churning out delicious frozen treats (events) faster than you can say "brain freeze." These ice creams need to get to hungry customers (subscription clients) via our fleet of deliveryman (the router's pubsub system).

Each deliveryman has to pass through a traffic light (a users potentially long running hook) on their way to the customer. Usually it's green and they zoom right through. But sometimes? That light turns red. Really, REALLY red. Like, "stuck-for-5-minutes-questioning-your-life-choices" red.

Let's assume you have 3 deliveryman and 3 customers. The first two deliveryman make it over the green light, the third one has to wait in front of the red light (hook is blocking for whatever reason). In the old implementation the two deliveryman will wait for the third deliveryman before picking up new ice-cream (events) from the ice cream factory (Kafka). This makes the ice cream delivery as slow as the weakest of the three.

## Solution idea

After this change the first two deliveryman will not wait for long (until timeout exceeds) for their third colleague before they will pick up new ice cream (new events from broker), antisocially but efficiently so.

### Why even use a timeout

I need to break out of the ice cream example for a bit here.

The reason to use a timeout instead of simply moving on immediately when a semaphore slot is free is because I try to honor event ordering and only break it if we exceed a timeout. We cannot stop a running goroutine, we can only abandon it. But it will still eventually progress and update a clients subscription. This could mean an event #2 could make it fast to a subscriber than event #1, if the hook processing it takes too long. With a timeout waiting for long running hooks we can somewhat mitigate this to a degree.

The timeout is configurable via config: `events.subscription_hooks.on_receive_events.handler_timeout`

### Always honor semaphore limit

If all deliveryman are stuck at the traffic light then we will not pick up new ice cream from the factory. That was a problem before this change and is still one after this change. But the only way to make it work is to spawn new deliveryman even if the limit (semaphore limit) is already exceeded. I did not opt for this and instead always honor the semaphore limit. There will never be more hook goroutines, ÄH deliveryman, than the configured semaphore limit. Else we would potentially run into a situation, where we have no control over the amount of routines running at a time on the router. If the limit is too low, a user can set it higher via configuration `events.subscription_hooks.on_receive_events.max_concurrent_handlers`.

### Let the hook developers know their time is over

Okay that sounds more dramatic than it really is. What I mean by that is we need to let hook developers know that their hook execution must stop now. We as the pubsub system cannot stop the routine from the outside, and have no control of the inside of the hook. But we can let the hook know to stop voluntarily via timeout contexts. The hooks context now has `ctx.Context()` which is timed with the routers timeout mechanism. Once we stop waiting for the routine, the `ctx.Context()` is also cancelled. This gives hook developers the option to stop their routines and avoid running abandoned in the background.

## Technical changes

- Semaphore is now global for each call to Update() on the event subscription updater, allowing free slots to be used for the next batch of events
- Hook configuration types have been streamlined, which allows us to easily add config parameters to hook behaviour in the future
- I also removed error logging when a hook returns an error, leaving this up to the hook developer (unrelated, but a small change)
- I also added some context to loggers like a `component` key for better logging (unrelated, but small change)
- Now we either wait for all updater routines to finish or a configurable timeout, whatever happens first. Then we will close the Update() function and receive the next batch of events.
- Pass a timeout context to the hooks via `ctx.Context()`
- Config parameter change: `events.subscription_hooks.on_receive_events.max_concurrent_handlers` (new path)
- New Config parameter: `events.subscription_hooks.on_receive_events.handler_timeout`

## Release Notes

* **New Features**
  * Introduced handler timeout configuration for subscription event processing (5-second default).
  * Refined concurrent event handler configuration structure for improved management.

* **Bug Fixes**
  * Enhanced out-of-order event delivery handling with timeout protection during processing delays.

* **Tests**
  * Expanded test coverage for timeout-based event delivery scenarios.

### DIFF
--- a/router/pkg/pubsub/datasource/subscription_event_updater_test.go
+++ b/router/pkg/pubsub/datasource/subscription_event_updater_test.go
@@ -87,7 +87,7 @@ func TestSubscriptionEventUpdater_UpdateSubscription_WithHooks_Success(t *testin
 
 	// Create wrapper function for the mock
 	receivedArgs := make(chan receivedHooksArgs, 1)
-	testHook := func(ctx context.Context, cfg SubscriptionEventConfiguration, eventBuilder EventBuilderFn, events []StreamEvent) ([]StreamEvent, error) {
+	testHook := func(subCtx context.Context, updaterCtx context.Context, cfg SubscriptionEventConfiguration, eventBuilder EventBuilderFn, events []StreamEvent) ([]StreamEvent, error) {
 		receivedArgs <- receivedHooksArgs{events: events, cfg: cfg}
 		return modifiedEvents, nil
 	}
@@ -135,7 +135,7 @@ func TestSubscriptionEventUpdater_UpdateSubscriptions_WithHooks_Error(t *testing
 	hookError := errors.New("hook processing error")
 
 	// Define hook that returns an error
-	testHook := func(ctx context.Context, cfg SubscriptionEventConfiguration, eventBuilder EventBuilderFn, events []StreamEvent) ([]StreamEvent, error) {
+	testHook := func(subCtx context.Context, updaterCtx context.Context, cfg SubscriptionEventConfiguration, eventBuilder EventBuilderFn, events []StreamEvent) ([]StreamEvent, error) {
 		return nil, hookError
 	}
 
@@ -180,13 +180,13 @@ func TestSubscriptionEventUpdater_Update_WithMultipleHooks_Success(t *testing.T)
 
 	// Chain of hooks that modify the data
 	receivedArgs1 := make(chan receivedHooksArgs, 1)
-	hook1 := func(ctx context.Context, cfg SubscriptionEventConfiguration, eventBuilder EventBuilderFn, events []StreamEvent) ([]StreamEvent, error) {
+	hook1 := func(subCtx context.Context, updaterCtx context.Context, cfg SubscriptionEventConfiguration, eventBuilder EventBuilderFn, events []StreamEvent) ([]StreamEvent, error) {
 		receivedArgs1 <- receivedHooksArgs{events: events, cfg: cfg}
 		return []StreamEvent{&testEvent{mutableTestEvent("modified by hook1")}}, nil
 	}
 
 	receivedArgs2 := make(chan receivedHooksArgs, 1)
-	hook2 := func(ctx context.Context, cfg SubscriptionEventConfiguration, eventBuilder EventBuilderFn, events []StreamEvent) ([]StreamEvent, error) {
+	hook2 := func(subCtx context.Context, updaterCtx context.Context, cfg SubscriptionEventConfiguration, eventBuilder EventBuilderFn, events []StreamEvent) ([]StreamEvent, error) {
 		receivedArgs2 <- receivedHooksArgs{events: events, cfg: cfg}
 		return []StreamEvent{&testEvent{mutableTestEvent("modified by hook2")}}, nil
 	}
@@ -280,7 +280,7 @@ func TestSubscriptionEventUpdater_SetHooks(t *testing.T) {
 		fieldName:    "testField",
 	}
 
-	testHook := func(ctx context.Context, cfg SubscriptionEventConfiguration, eventBuilder EventBuilderFn, events []StreamEvent) ([]StreamEvent, error) {
+	testHook := func(subCtx context.Context, updaterCtx context.Context, cfg SubscriptionEventConfiguration, eventBuilder EventBuilderFn, events []StreamEvent) ([]StreamEvent, error) {
 		return events, nil
 	}
 
@@ -314,7 +314,7 @@ func TestNewSubscriptionEventUpdater(t *testing.T) {
 		fieldName:    "testField",
 	}
 
-	testHook := func(ctx context.Context, cfg SubscriptionEventConfiguration, eventBuilder EventBuilderFn, events []StreamEvent) ([]StreamEvent, error) {
+	testHook := func(subCtx context.Context, updaterCtx context.Context, cfg SubscriptionEventConfiguration, eventBuilder EventBuilderFn, events []StreamEvent) ([]StreamEvent, error) {
 		return events, nil
 	}
 
@@ -385,7 +385,7 @@ func TestSubscriptionEventUpdater_Update_WithSingleHookModification(t *testing.T
 	}
 
 	// Hook that modifies events by adding a prefix
-	hook := func(ctx context.Context, cfg SubscriptionEventConfiguration, eventBuilder EventBuilderFn, events []StreamEvent) ([]StreamEvent, error) {
+	hook := func(subCtx context.Context, updaterCtx context.Context, cfg SubscriptionEventConfiguration, eventBuilder EventBuilderFn, events []StreamEvent) ([]StreamEvent, error) {
 		modifiedEvents := make([]StreamEvent, len(events))
 		for i, event := range events {
 			modifiedData := "modified: " + string(event.GetData())
@@ -438,7 +438,7 @@ func TestSubscriptionEventUpdater_Update_WithSingleHookError_ClosesSubscription(
 	hookError := errors.New("hook processing failed")
 
 	// Hook that returns an error
-	hook := func(ctx context.Context, cfg SubscriptionEventConfiguration, eventBuilder EventBuilderFn, events []StreamEvent) ([]StreamEvent, error) {
+	hook := func(subCtx context.Context, updaterCtx context.Context, cfg SubscriptionEventConfiguration, eventBuilder EventBuilderFn, events []StreamEvent) ([]StreamEvent, error) {
 		// Return the events but also return an error
 		return events, hookError
 	}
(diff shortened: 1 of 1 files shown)


dkorittki:
@coderabbitai review

router/core/subscriptions_modules.go
```diff
@@ -213,9 +226,15 @@ func NewEngineSubscriptionOnStartHook(fn func(ctx SubscriptionOnStartHandlerCont
 
 	return func(resolveCtx resolve.StartupHookContext, input []byte) error {
 		requestContext := getRequestContext(resolveCtx.Context)
+
+		logger := requestContext.Logger()
+		if logger != nil {
+			logger = logger.With(zap.String("component", "on_publish_events_hook"))
```
alepane21: mmm isn't this the on start hook?

router/core/subscriptions_modules.go
```diff
@@ -191,9 +191,22 @@ func NewPubSubSubscriptionOnStartHook(fn func(ctx SubscriptionOnStartHandlerCont
 
 	return func(resolveCtx resolve.StartupHookContext, subConf datasource.SubscriptionEventConfiguration, eventBuilder datasource.EventBuilderFn) error {
 		requestContext := getRequestContext(resolveCtx.Context)
+
+		logger := requestContext.Logger()
+		if logger != nil {
+			logger = logger.With(zap.String("component", "subscription_on_start_hook"))
```
alepane21: Maybe we should differentiate between pubsub_subscription_on_start and engind_subscription_on_start?

router/core/subscriptions_modules.go
```diff
@@ -226,6 +245,9 @@ func NewEngineSubscriptionOnStartHook(fn func(ctx SubscriptionOnStartHandlerCont
 }
 
 type StreamReceiveEventHandlerContext interface {
+	// Context is a timeout based context.
+	// It signals a timeout when the handler should stop processing.
+	Context() context.Context
```
alepane21, line 250: I think that here we can avoid to get into the details of it being a timeout based context.
Also it is not correct, because it should also be cancelled when it is cancelled the main resolver context, and probably in the future we could change it further. Probably the user of this interface just have to know that when the context is done, they are done.

router/pkg/config/config.go
```diff
@@ -646,7 +646,12 @@ type EventsConfiguration struct {
 }
 
 type SubscriptionHooksConfiguration struct {
-	MaxConcurrentEventReceiveHandlers int `yaml:"max_concurrent_event_receive_handlers" envDefault:"100"`
+	OnReceiveEvents OnReceiveEventsConfiguration `yaml:"on_receive_events"`
```
alepane21, line 649: We should also change the doc. I don't see any PR on the doc repo or change in the current PR about this.

router/pkg/pubsub/datasource/subscription_event_updater.go
```diff
@@ -26,38 +27,51 @@ type subscriptionEventUpdater struct {
...
+	done := make(chan struct{})
 
-	doneLogging := make(chan struct{})
 	go func() {
-		s.deduplicateAndLogErrors(errCh, len(subscriptions))
-		doneLogging <- struct{}{}
+		for ctx, subId := range subscriptions {
+			s.semaphore <- struct{}{} // Acquire slot, blocks if all slots are taken
+			wg.Add(1)
+			ctx, cancel := context.WithDeadline(ctx, deadline)
```
alepane21: Do we need different ctx for each subscription? I would create a single ctx, the cancel is already running at the end of the go routine out of the loop.

router/pkg/pubsub/datasource/subscription_event_updater.go
```diff
@@ -26,38 +27,51 @@ type subscriptionEventUpdater struct {
...
 
-	wg.Wait()
-	close(semaphore)
-	close(errCh)
-	<-doneLogging
+	select {
+	case <-done:
+		s.logger.Debug("All subscription updates completed")
+		// All subscriptions completed successfully
+	case <-time.After(s.timeout + time.Millisecond*50):
```
alepane21: I would make the "exceedable time" configurable or at least a constant with a name. Also I think this behaviour should be made explicit in the documentation. If someone stays in the hook after this limit there is no guarantee that the client is still connected and the message sent.

router/pkg/pubsub/datasource/subscription_event_updater.go
```diff
@@ -26,38 +27,51 @@ type subscriptionEventUpdater struct {
...
+		// Timeout exceeded, some subscription updates may still be running.
+		// We can't stop them but we will also not wait for them, basically abandoning them.
+		// They will continue to hold their semaphore slots until they complete,
+		// which means the next Update() call will have fewer available slots.
+		// Also since we will process the next batch of events while having abandoned updaters,
+		// those updaters might eventually push their events to the subscription late,
+		// which means events might arrive out of order.
+		s.logger.Warn("Timeout exceeded during subscription updates, events may arrive out of order")
```
alepane21, line 77: Can we signal which hook has been exceeding the timeout? I think it would be really important for our users to debug where the problem is.

router/pkg/pubsub/datasource/subscription_event_updater.go
alepane21: If done is received before this time.After, aren't we leaking memory?

router/pkg/pubsub/datasource/subscription_event_updater.go
```diff
@@ -120,48 +131,26 @@ func (s *subscriptionEventUpdater) recoverPanic(subID resolve.SubscriptionIdenti
...
 	cfg SubscriptionEventConfiguration,
 	hooks Hooks,
 	eventUpdater resolve.SubscriptionUpdater,
 	logger *zap.Logger,
 	eventBuilder EventBuilderFn,
 ) SubscriptionEventUpdater {
+	limit := max(hooks.OnReceiveEvents.MaxConcurrentHandlers, 1)
+	timeout := hooks.OnReceiveEvents.Timeout
+	if timeout == 0 {
+		timeout = 5 * time.Second
```
alepane21: I think it should be a constant or just come down from the default configuration.

alepane21 (changes requested):
I like what you did here, good work! I found some small issues.

router/pkg/pubsub/datasource/subscription_event_updater.go
```diff
@@ -66,22 +80,23 @@ func (s *subscriptionEventUpdater) Complete() {
 
 func (s *subscriptionEventUpdater) Close(kind resolve.SubscriptionCloseKind) {
 	s.eventUpdater.Close(kind)
+	close(s.semaphore)
 }
 
 func (s *subscriptionEventUpdater) SetHooks(hooks Hooks) {
 	s.hooks = hooks
 }
```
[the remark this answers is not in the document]
dkorittki: The `SetHooks` method is only there to set new hooks, not reset hook settings. `s.timeout` and `s.semaphore` are meant to be set once at subscription start and that has to be done by creating a new updater. Also right now `SetHooks` is only used inside a router-test, not production code.

router/pkg/pubsub/datasource/subscription_event_updater.go
dkorittki: The information available at that logger currently is:
provider_id, root_field_name, provider_type, request_id, component (=subscription_updater)

I could maybe add the subscriptoin id to it but don't know if it would really help. What do you think about that?

router/pkg/pubsub/datasource/subscription_event_updater.go
dkorittki: Can you explain what parts could leak memory?

router/pkg/pubsub/datasource/subscription_event_updater.go
dkorittki: I will make it a constant. The configuration settings higher up in the router code already ensures it's 5s default. This is just a safety guard in case someone messes up the defaulting logic in the future. This code right now is actually not hit.

router/pkg/pubsub/datasource/subscription_event_updater.go
alepane21: Ok, fixed in go 1.23, my bad!

dkorittki: I think making it configurable isn't a great idea, it would confuse hook developers rather than helping them. It's just a small buffer for allowing all hooks to react to the context cancellation. I mean we don't kill their routines if their cleanup takes longer than that, so I think they don't need to worry about this. But I will make it a constant for sure

router/pkg/pubsub/datasource/subscription_event_updater.go
dkorittki: After discussion: It would be great to provide the hook developers custom module name in the logger as well. But thats something more sophisticated than a small fix and out of scope here.

router/demo.config.yaml
```diff
@@ -20,3 +20,7 @@ events:
       - id: my-redis
         urls:
         - "redis://localhost:6379/2"
+  subscription_hooks:
+    on_receive_events:
+      max_concurrent_handlers: 97
+      handler_timeout: 8800
```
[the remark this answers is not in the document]
dkorittki: demo.config.yaml was committed by me by accident. It should not change. I undid the changes.

router-tests/modules/stream_receive_test.go
```diff
@@ -893,4 +694,125 @@ func TestReceiveHook(t *testing.T) {
...
+			}
+
+			require.NoError(t, client.Close())
+			testenv.AwaitChannelWithT(t, Timeout, clientRunCh, func(t *testing.T, err error) {
+				require.NoError(t, err)
+			}, "unable to close client before timeout")
+
+			// Verify events arrived out of order: event 2 and 3 before event 1
+			assert.Equal(t, []float64{2, 3, 1}, receivedIDs, "expected events to arrive out of order due to timeout")
+
```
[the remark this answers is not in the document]

PR #2335, opened 2025-11-16
feat: add user/app context in controlplane sentry

miklosbarabas:
## Summary by CodeRabbit

* **New Features**
  * Enhanced error monitoring with Sentry integration enabled, including release version tracking for improved incident diagnostics.
  * User context enrichment in error reports for better support and debugging visibility.

* **Chores**
  * Updated configuration infrastructure to support monitoring setup.

## Checklist

### DIFF
--- a/.github/workflows/controlplane-ci.yaml
+++ b/.github/workflows/controlplane-ci.yaml
@@ -80,9 +80,6 @@ jobs:
       - name: Setup Keycloak
         run: nohup .github/scripts/setup-keycloak.sh &
 
-      - name: Copy dummy env file
-        run: cp ./controlplane/.env.example ./controlplane/.env
-
       - name: Test
         run: pnpm run --filter controlplane test
         env:
--- a/controlplane/src/core/sentry.config.ts
+++ b/controlplane/src/core/sentry.config.ts
@@ -1,8 +1,8 @@
+import { createRequire } from 'node:module';
 import * as Sentry from '@sentry/node';
 import { nodeProfilingIntegration } from '@sentry/profiling-node';
 import { eventLoopBlockIntegration } from '@sentry/node-native';
 import { fastifyIntegration, pinoIntegration } from '@sentry/node';
-import { version } from '../../package.json';
 import { envVariables } from './env.schema.js';
 
 const {
@@ -17,9 +17,18 @@ const {
 } = envVariables.parse(process.env);
 
 if (SENTRY_ENABLED && SENTRY_DSN) {
+  const require = createRequire(import.meta.url);
+  let release: string | undefined;
+  try {
+    const pkg = require('../../package.json') as { version?: string };
+    release = pkg.version;
+  } catch (error) {
+    console.debug('Sentry: failed to read package.json version for release', error);
+  }
+
   Sentry.init({
     dsn: SENTRY_DSN,
-    release: version,
+    release,
     integrations: [
       fastifyIntegration(),
       eventLoopBlockIntegration({ threshold: SENTRY_EVENT_LOOP_BLOCK_THRESHOLD_MS }),
--- a/controlplane/src/core/util.ts
+++ b/controlplane/src/core/util.ts
@@ -1,5 +1,4 @@
 import { randomFill } from 'node:crypto';
-import * as Sentry from '@sentry/node';
 import { S3ClientConfig } from '@aws-sdk/client-s3';
 import { HandlerContext } from '@connectrpc/connect';
 import {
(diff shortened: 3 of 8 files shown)


controlplane/src/core/util.ts
```diff
@@ -34,13 +35,15 @@ import { isAuthenticationError, isAuthorizationError, isPublicError } from './er
...
 import { SubgraphsToCompose } from './repositories/FeatureFlagRepository.js';
+import { envVariables } from './env.schema.js';
 
 const labelRegex = /^[\dA-Za-z](?:[\w.-]{0,61}[\dA-Za-z])?$/;
 const organizationSlugRegex = /^[\da-z]+(?:-[\da-z]+)*$/;
 const namespaceRegex = /^[\da-z]+(?:[_-][\da-z]+)*$/;
 const schemaTagRegex = /^(?![/-])[\d/A-Za-z-]+(?<![/-])$/;
 const graphNameRegex = /^[\dA-Za-z]+(?:[./@_-][\dA-Za-z]+)*$/;
 const pluginVersionRegex = /^v\d+$/;
+const { SENTRY_ENABLED, SENTRY_DSN } = envVariables.parse(process.env);
```
StarpTech: There should be only the main entry file that interacts with globals directly.

controlplane/src/core/util.ts
```diff
@@ -107,6 +110,40 @@ export const enrichLogger = (
     },
   });
 
+  if (SENTRY_ENABLED && SENTRY_DSN) {
+    try {
+      // NB: Fastify integration automatically creates request-specific scopes
+      if (authContext.userId) {
+        Sentry.setUser({
```
StarpTech: Isn't sentry a noop if disabled?

.github/workflows/controlplane-ci.yaml
```diff
@@ -80,6 +80,9 @@ jobs:
       - name: Setup Keycloak
         run: nohup .github/scripts/setup-keycloak.sh &
 
+      - name: Copy dummy env file
```
StarpTech: Why is it necessary? To me it looks like it would be better to set necessary env in the test setup file.

controlplane/src/core/sentry.config.ts
```diff
@@ -16,8 +17,18 @@ const {
 } = envVariables.parse(process.env);
 
 if (SENTRY_ENABLED && SENTRY_DSN) {
+  const require = createRequire(import.meta.url);
+  let release: string | undefined;
+  try {
+    const pkg = require('../../package.json') as { version?: string };
```
StarpTech: You can just use the file api and don't have to deal with require OR

```ts
import { version } from '../package.json';

console.log(`Starting Cosmo ${version}`);
```

controlplane/src/core/util.ts
miklosbarabas: so how would you use those envvars in here?

controlplane/src/core/util.ts
miklosbarabas: Not sure, but would be better not to run on this code path at all if Sentry is not enabled.

.github/workflows/controlplane-ci.yaml
miklosbarabas: This was required for Vitest once I added this, it was requiring all envvars not just the imported/parsed two:
```
const { SENTRY_ENABLED, SENTRY_DSN } = envVariables.parse(process.env);
```
Do you have any suggestion instead?

controlplane/src/core/sentry.config.ts
miklosbarabas: Sounds good, will give it a try.

controlplane/src/core/util.ts
```diff
@@ -107,6 +110,40 @@ export const enrichLogger = (
     },
   });
 
+  if (SENTRY_ENABLED && SENTRY_DSN) {
```
StarpTech: What's the reason why we initialize Sentry within `enrichLogger`? How are they related?

StarpTech (changes requested):

controlplane/src/core/util.ts
miklosbarabas: There're couple of reasons why it was chosen:
 - `enrichLogger(ctx, logger, authContext)` is the first point in each Connect handler where you have:
      - `authContext` (user/org/api-key/graph info), and
      - ctx (request-scoped context tied to the Fastify request via contextValues in `src/core/build-server.ts:493`).
  - That's also the only shared helper called from all the bufservice handlers immediately after opts.authenticator.authenticate(...).
 
 
From a design/concerns perspective, conceptually, Sentry context would be better handled by a Sentry-specific helper, not a logging helper. Moving Sentry into Authentication or a Fastify hook in `build-server.ts` sounds attractive, but:
- Authentication doesn't have access to HandlerContext/Fastify logger.
- The Fastify contextValues hook in `build-server.ts` doesn’t have authContext (auth hasn’t run yet).
- To do that properly we'd need a custom Connect interceptor/middleware layer, which is a more invasive change.

PR #2337, opened 2025-11-17
feat: use a static version id for the router configs generated by `wgc router compose`

JivusAyrus:
## Summary by CodeRabbit

* **Improvements**
  * Router configuration now uses stable, deterministic schema version identifiers instead of per-build random IDs, producing consistent configs across builds and deployments.
  * Federated client schema generation now preserves and includes schema directives in the output, improving accuracy for schema-driven behavior and tooling.
  * Feature-flag versioning aligned with deterministic schema IDs.

## Checklist


cli/src/commands/router/commands/compose.ts
```diff
@@ -158,6 +158,27 @@ function constructRouterSubgraph(result: FederationSuccess, s: SubgraphMetadata,
   return composedSubgraphGRPC;
 }
 
+/**
+ * Converts a hash to a deterministic UUID format.
+ * Takes a SHA-256 hash and formats it as a UUID v4-like string.
+ * The same input will always produce the same UUID.
+ */
+function hashToUUID(input: string): string {
```
StarpTech: Why do we care if it is a UUID?

Aenimus: What are you suggesting otherwise? It shouldn't be likely to clash. And it's inexpensive/not meaningful because it's run by the end user rather than Cosmo.

JivusAyrus: The reason i made it a uuid is, I dont know where the version is used and wanted to make sure that we dont break anything

cli/src/commands/router/commands/compose.ts
```diff
@@ -241,19 +240,20 @@ export default (opts: BaseCommandOptions) => {
...
+    const federatedSDL = printSchemaWithDirectives(result.federatedGraphSchema);
     const routerConfig = buildRouterConfig({
       federatedClientSDL,
-      federatedSDL: printSchemaWithDirectives(result.federatedGraphSchema),
+      federatedSDL,
       fieldConfigurations: result.fieldConfigurations,
       // @TODO get router compatibility version programmatically
       routerCompatibilityVersion: ROUTER_COMPATIBILITY_VERSION_ONE,
-      schemaVersionId: 'static',
+      schemaVersionId: '00000000-0000-0000-0000-000000000000',
```
Aenimus: set it as a const above and pass it and set the feature flag version to the const?

JivusAyrus: Done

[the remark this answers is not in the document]
JivusAyrus: As this is for local router config, we dont care about versioning. The version attribute exists for the schema registry to use

PR #2365, opened 2025-11-26
feat: improve memory management

jensneuse:
## How to use this version

You have two ways:

### a) Users with custom modules

Follow the router upgrade instructions here:
[https://github.com/wundergraph/router-examples?tab=readme-ov-file#-upgrade-router](https://github.com/wundergraph/router-examples?tab=readme-ov-file#-upgrade-router)

When referencing this PR from your branch, use the following commit hash:

```
e67bc7307d3fc15a483b026a02ec1530f4f0ff5e
```

---

### b) Regular users

Use one of the following Docker images:

* **Root image**
  `ghcr.io/wundergraph/cosmo/router:pr-2365`

* **Non-root image**
  `ghcr.io/wundergraph/cosmo/router:pr-2365-nonroot`

## Summary by CodeRabbit

* **New Features**
  * Configurable header propagation across subgraphs; header-aware subscriptions and new inbound/request deduplication toggles (including force flags).

* **Performance**
  * Increased default HTTP connection pooling and timeouts for higher concurrency.

* **Tests**
  * Vastly expanded deduplication and subscription test coverage; added header-propagation unit tests and updated telemetry/metrics expectations.

* **Chores**
  * Dependency upgrades and added telemetry attribute to mark resolver deduplicated requests; various test and cleanup refinements.

### DIFF
--- a/router/pkg/pubsub/nats/engine_datasource_factory_test.go
+++ b/router/pkg/pubsub/nats/engine_datasource_factory_test.go
@@ -180,7 +180,7 @@ func TestEngineDataSourceFactory_RequestDataSource(t *testing.T) {
 		return event.ProviderID() == "test-provider" && event.Subject == "test-subject"
 	}), mock.MatchedBy(func(event datasource.StreamEvent) bool {
 		return event != nil && strings.EqualFold(string(event.GetData()), `{"test":"data"}`)
-	}), mock.Anything).Return([]byte(`{"response": "test"}`), nil)
+	})).Return([]byte(`{"response": "test"}`), nil)
 
 	// Create the data source with mock adapter
 	pubsub := &EngineDataSourceFactory{


router/core/context.go
```diff
@@ -272,6 +279,73 @@ type requestContext struct {
...
+	allHash uint64
+}
+
+func (c *headerBuilder) HashAll() (out uint64) {
+	return c.allHash
+}
+
+func (c *headerBuilder) HeadersForSubgraph(subgraphName string) (http.Header, uint64) {
+	if header, ok := c.headers[subgraphName]; ok {
+		return header.Header.Clone(), header.Hash
```
dkorittki, line 296: One thing I found here, which might or might not be intended is:
It seems `header.Header` is nil for EDG's. I tested this and found this to be true at least for Kafka.
The reason seems to be that for EDG´s we dont have any headers (because it's not http connections to subgraphs but broker-specific connections we do right?). The function returns nil, 0 as the consequence. 

The `prepareTrigger` function in the engine does find a header builder and then adds `nil` to the hash input. It means the triggerID for EDFS subscriptions in the engine is based solely on the input, not headers. Again, could be by design. Just wanted to raise awereness in case this is unexpected behaviour.

router-tests/go.mod
```diff
@@ -209,7 +210,7 @@ replace (
 	github.com/wundergraph/cosmo/demo/pkg/subgraphs/projects => ../demo/pkg/subgraphs/projects
 	github.com/wundergraph/cosmo/router => ../router
 	github.com/wundergraph/cosmo/router-plugin => ../router-plugin
-//github.com/wundergraph/graphql-go-tools/v2 => ../../graphql-go-tools/v2
+	github.com/wundergraph/graphql-go-tools/v2 => ../../../graphql-go-tools/v2
```
StarpTech: This should not be commited.

router-tests/prometheus_test.go
```diff
@@ -4689,7 +4689,7 @@ func TestFlakyPrometheusRouterConnectionMetrics(t *testing.T) {
 				require.Len(t, metrics, 2)
 
 				metricDataPoint1 := metrics[0]
-				require.Equal(t, 100.0, *metricDataPoint1.Gauge.Value)
+				require.Equal(t, 0.0, *metricDataPoint1.Gauge.Value)
```
StarpTech: Those values don't match the previous ones; they are all zero. What's going on here?

router/core/graphql_prehandler.go
```diff
@@ -1194,6 +1205,34 @@ func (h *PreHandler) parseRequestOptions(r *http.Request, clientInfo *ClientInfo
...
+	// Disable inbound request deduplication if request deduplication is disabled
+	ex.DisableInboundRequestDeduplication = !h.enableInboundRequestDeduplication
+	if !h.enableRequestDeduplication {
+		// DisableInboundRequestDeduplication is disabled when RequestDeduplication in general is disabled
+		ex.DisableInboundRequestDeduplication = true
+	}
+	if h.hasPreOriginHandlers {
+		// if we have pre origin handlers, we cannot guarantee that these handlers won't modify headers
+		// as such, we're automatically disabling request deduplication
+		ex.DisableSubgraphRequestDeduplication = true
```
StarpTech, line 1229: What's the difference between Inbound and Subgraph?

router/core/router.go
```diff
@@ -307,19 +308,18 @@ func NewRouter(opts ...Option) (*Router, error) {
...
-
-	if hr.HasRequestRules() {
-		r.preOriginHandlers = append(r.preOriginHandlers, hr.OnOriginRequest)
-	}
-	if hr.HasResponseRules() {
-		r.postOriginHandlers = append(r.postOriginHandlers, hr.OnOriginResponse)
+	// we only add post origin handler for header rules
+	// pre handlers (header propagation rules) are handled via the engine
+	if r.headerPropagation.HasResponseRules() {
+		r.postOriginHandlers = append(r.postOriginHandlers, r.headerPropagation.OnOriginResponse)
```
StarpTech, line 320: That's a great catch by coderabbit. Ideally, an interface never returns (nil,nil) because the implementation is essentially non-working. It either success or fails. You can fix it by returning an error when rules are empty in `NewHeaderPropagation`

router/core/router.go
```diff
@@ -1911,9 +1911,9 @@ func DefaultTransportRequestOptions() *TransportRequestOptions {
 		KeepAliveIdleTimeout:   90 * time.Second,
 		DialTimeout:            30 * time.Second,
 
-		MaxConnsPerHost:     100,
-		MaxIdleConns:        1024,
-		MaxIdleConnsPerHost: 20,
+		MaxConnsPerHost:     0,
```
StarpTech: Why did you change the defaults? Those are aligned with the config defaults.

router/pkg/config/config.go
```diff
@@ -372,35 +372,43 @@ type EngineDebugConfiguration struct {
...
-	ValidateRequiredExternalFields                   bool                     `envDefault:"false" env:"ENGINE_VALIDATE_REQUIRED_EXTERNAL_FIELDS" yaml:"validate_required_external_fields"`
+	Debug              EngineDebugConfiguration `yaml:"debug"`
+	EnableSingleFlight bool                     `envDefault:"true" env:"ENGINE_ENABLE_SINGLE_FLIGHT" yaml:"enable_single_flight"`
+	// ForceEnableSingleFlight always enables single flight, except for mutations
+	// By default, SingleFlight / Request Deduplication is disabled when PreOriginHandlers are configured
+	// This is because PreOriginHandlers can modify request headers, which has influence on the request deduplication key
+	// If you're sure that your PreOriginHandlers won't interfere with the request deduplication key, you can enable it with this flag
+	ForceEnableSingleFlight           bool `envDefault:"false" env:"ENGINE_FORCE_ENABLE_SINGLE_FLIGHT" yaml:"force_enable_single_flight"`
+	EnableInboundRequestDeduplication bool `envDefault:"true" env:"ENGINE_ENABLE_INBOUND_REQUEST_DEDUPLICATION" yaml:"enable_inbound_request_deduplication"`
```
StarpTech, line 404: Why is the diff so huge? Please check if nothing broke.

router/core/graphql_prehandler.go
jensneuse: inbound (client->router), subgraph (router-subgraph)

router/core/context.go
jensneuse: Headers will be nil if there are not header forwarding rules. This totally depends on the config. In general, even an EDG is just a regular data source. That said, even if headers are nil, we still hash the input bytes for the data source, so it's working as intended.

PR #2377, opened 2025-11-28
fix(router): flush errors on sse streams

dkorittki:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error propagation for GraphQL subscriptions over Server-Sent Events (SSE).
  * Enhanced HTTP response flushing and formatting to ensure error responses are reliably transmitted.

* **Tests**
  * Added test coverage verifying SSE error handling when downstream requests fail.

## Checklist

# Context

When we write errors to a connection backed by `HttpFlushWriter` (used on subscriptions via SSE/Multipart) we do not flush the connection as oposed to a websocket connection, where this actually happens. I fixed it by simply checking if we are dealing with a subscription response writer (which needs to be flushed and covers both websockets and sse) and flush if so.


dkorittki:
Does Codecov recognize that tests have been added to the router-tests package? Seems like not.

alepane21 (approved):
LGTM!

dkorittki:
Note: Did some local tests to see if it works, but only with one simultaneous client. Gonna be sure and test with the Apollo client with some simultaneous connections.

dkorittki:
I did some tests with a custom script using apollo client with both [graphql-ws](https://www.npmjs.com/package/graphql-ws) and [graphql-sse](https://www.npmjs.com/package/graphql-sse) as connection clients.

The error seems to be propagated correctly to clients on sse, now matching what ws returns. Here are log outputs:

ws-client
```
Starting subscription...
Received: {
  data: undefined,
  error: CombinedGraphQLErrors: My Test Error
      at Object.error (file:///[...]/apollo-client-demo/node_modules/@apollo/client/link/subscriptions/index.js:88:43)
      at Object.e12e92bd-23ca-497c-a384-e19a5651940e (file:///[...]/apollo-client-demo/node_modules/graphql-ws/dist/client.js:338:22)
      at emit (file:///[...]/apollo-client-demo/node_modules/graphql-ws/dist/client.js:75:58)
      at Object.emit (file:///[...]/apollo-client-demo/node_modules/graphql-ws/dist/client.js:100:11)
      at socket2.onmessage (file:///[...]/apollo-client-demo/node_modules/graphql-ws/dist/client.js:203:21)
      at [nodejs.internal.kHybridDispatch] (node:internal/event_target:827:20)
      at WebSocket.dispatchEvent (node:internal/event_target:762:26)
      at fireEvent (node:internal/deps/undici/undici:11663:14)
      at websocketMessageReceived (node:internal/deps/undici/undici:11685:7)
      at ByteParser.run (node:internal/deps/undici/undici:12307:19) {
    errors: [ [Object] ],
    data: undefined,
    extensions: undefined
  }
}
```

sse-client
```
Starting subscription...
Received: {
  data: undefined,
  error: CombinedGraphQLErrors: My Test Error
      at file:///[...]/apollo-client-demo/node_modules/@apollo/client/core/QueryManager.js:475:36
      at /[...]/apollo-client-demo/node_modules/rxjs/dist/cjs/internal/operators/map.js:10:37
      at OperatorSubscriber._this._next (/[...]/apollo-client-demo/node_modules/rxjs/dist/cjs/internal/operators/OperatorSubscriber.js:33:21)
      at Subscriber.next (/[...]/apollo-client-demo/node_modules/rxjs/dist/cjs/internal/Subscriber.js:51:18)
      at Subscriber._next (/[...]/apollo-client-demo/node_modules/rxjs/dist/cjs/internal/Subscriber.js:80:26)
      at Subscriber.next (/[...]/apollo-client-demo/node_modules/rxjs/dist/cjs/internal/Subscriber.js:51:18)
      at /[...]/apollo-client-demo/node_modules/rxjs/dist/cjs/internal/Subject.js:69:34
      at Object.errorContext (/[...]/apollo-client-demo/node_modules/rxjs/dist/cjs/internal/util/errorContext.js:22:9)
      at Subject.next (/[...]/apollo-client-demo/node_modules/rxjs/dist/cjs/internal/Subject.js:59:24)
      at Object.next (/[...]/apollo-client-demo/node_modules/rxjs/dist/cjs/internal/operators/share.js:69:58) {
    errors: [ [Object] ],
    data: null,
    extensions: undefined
  }
}
```

The only difference is that sse-client keeps reconnecting after the error. This seems to stem from the sse-client itself and has been the case before this change. It was just silent, no log output in that case, because the router never returned data on the connection prior to closing it.

I suspect the sse-client retries because we never sent a `complete` event after the error. According to the SSE specification, a complete event needs to be sent whenever we want to close the connection, which in this case is true. I'll see how we can add this to the code and if it really makes sense.

dkorittki:
~~Added commits to make the router send a complete event after sending an error event. I also updated the pull request description with details about that. Please have a look at that for more context.~~

EDIT: Removed it again. Sending complete events is not the task of the `WriteError` method. That method is responsible for sending errors on the stream, not closing/completing them. I believe the router should sent them at various places where we also write an error to the stream, but we are better off fixing this in a separate pull request. It might involve a bit more work and it's an unrelated problem, really.

router/core/graphql_handler.go
```diff
@@ -433,6 +433,10 @@ func (h *GraphQLHandler) WriteError(ctx *resolve.Context, err error, res *resolv
 	if wsRw, ok := w.(*websocketResponseWriter); ok {
 		_ = wsRw.Flush()
 	}
+
+	if httpRw, ok := w.(*HttpFlushWriter); ok {
+		_ = httpRw.Flush()
+	}
```
endigma: Can this be checked against the subscription writer interface or like an inline "Flusher" interface to avoid the duplication?

router/core/flushwriter.go
```diff
@@ -143,6 +143,10 @@ func (f *HttpFlushWriter) Flush() (err error) {
 		separation = ""
 	}
 
+	if bytes.HasSuffix(resp, []byte{'\n'}) {
+		resp = bytes.TrimRight(resp, "\n")
+	}
```
endigma, line 150: Can you add a comment justifying this behaviour?

endigma (changes requested):

router/core/graphql_handler.go
dkorittki: Fixed

router/core/flushwriter.go
dkorittki: Fixed

PR #2379, opened 2025-11-28
feat(router): add connectrpc handler for graphql-to-grpc bridge

asoorm:
## Overview

This PR introduces a ConnectRPC handler that enables gRPC/Connect/gRPC-Web clients to interact with GraphQL APIs through protocol translation. The implementation provides a bridge between protobuf-based RPC protocols and GraphQL operations.

⚠️ Ignore changes to cli/ and protographic/ directories - these improvements have been extracted to PR #2436 for independent review and will be merged first.

## Key Features

- **Multi-Protocol Support**: Handles gRPC, Connect, and gRPC-Web protocols via Vanguard transcoder
- **Service Discovery**: Convention-based service discovery from directory structure
- **Operation Registry**: Service-scoped GraphQL operation management
- **Error Handling**: Relay-inspired error classification (CRITICAL vs NON-CRITICAL)
- **Header Forwarding**: Transparent header propagation to GraphQL endpoint

### File Responsibilities

- **`handler.go`**: Core RPC request handler, GraphQL execution, error classification
- **`operation_registry.go`**: Thread-safe registry for service-scoped GraphQL operations
- **`proto_loader.go`**: Protobuf file parsing and service definition extraction
- **`server.go`**: HTTP/2 server with lifecycle management (start/stop/reload)
- **`service_discovery.go`**: Convention-based service discovery from filesystem
- **`vanguard_service.go`**: Vanguard integration for protocol transcoding

### Supporting Files

- **`handler_error_test.go`**: Error handling and classification tests
- **`handler_test.go`**: Core handler functionality tests
- **`operation_registry_test.go`**: Registry thread-safety and service-scoping tests
- **`proto_loader_test.go`**: Proto parsing tests
- **`server_test.go`**: Server configuration tests
- **`server_lifecycle_test.go`**: Server lifecycle and reload tests
- **`service_discovery_test.go`**: Service discovery convention tests
- **`vanguard_service_test.go`**: Protocol transcoding tests

### Prerequisites
Start the router with appropriate config

```bash
listen_addr: "localhost:3002"
graphql_path: "/graphql"

# ConnectRPC configuration
connect_rpc:
  enabled: true
  server:
    listen_addr: "localhost:5026"
  services_provider_id: "fs-services"
  graphql_endpoint: "http://localhost:3002/graphql"

# Storage providers for services directory
storage_providers:
  file_system:
    - id: "fs-services"
      path: "./pkg/connectrpc/samples/services"
```

The services directory should contain `proto` and named GraphQL operations. See `./pkg/connectrpc/samples/services` as an example:

```bash
tree ./pkg/connectrpc/samples/services 
./pkg/connectrpc/samples/services
└── employee.v1
    ├── MutationUpdateEmployeeMood.graphql
    ├── QueryGetEmployeeById.graphql
    ├── QueryGetEmployeeByPets.graphql
    ├── QueryGetEmployeeWithMood.graphql
    ├── QueryGetEmployees.graphql
    ├── QueryGetEmployeesByPetsInlineFragment.graphql
    ├── QueryGetEmployeesByPetsNamedFragment.graphql
    ├── service.proto
    └── service.proto.lock.json
```

### Testing with curl (Connect Protocol)

Client Validation Error
```
curl -s -X POST http://localhost:5026/employee.v1.EmployeeService/GetEmployeeById \
  -H "Content-Type: application/json" \
  -H "Connect-Protocol-Version: 1" \
  -d '{"employee_id": "a"}' | jq
{
  "code": "invalid_argument",
  "message": "invalid request: field 'employee_id': Int32 cannot represent non-integer value: a"
}
```

OK

```
curl -s -X POST http://localhost:5026/employee.v1.EmployeeService/GetEmployeeById \
  -H "Content-Type: application/json" \
  -H "Connect-Protocol-Version: 1" \
  -d '{"employee_id": 1}' | jq  
{
  "employee": {
    "id": 1,
    "tag": "",
    "details": {
      "pets": null,
      "location": {
        "key": {
          "name": "Germany"
        }
      },
      "forename": "Jens",
      "surname": "Neuse"
    }
  }
}
```

### Testing with grpcurl (gRPC Protocol)

**Note**: Requires proto files as reflection protocol is not supported.

```bash
grpcurl -plaintext \
  -proto ./pkg/connectrpc/samples/services/employee.v1/service.proto \
  -d '{"employee_id": 1}' \
  localhost:5026 \
  employee.v1.EmployeeService/GetEmployeeById | jq
{
  "employee": {
    "id": 1,
    "details": {
      "forename": "Jens",
      "surname": "Neuse",
      "location": {
        "key": {
          "name": "Germany"
        }
      }
    }
  }
}
```

### Testing with buf CLI (Connect Protocol)

```bash
buf curl --http2-prior-knowledge --protocol connect \
  --schema ./pkg/connectrpc/samples/services/employee.v1/service.proto \
  --data '{"employee_id": 1}' \
  http://localhost:5026/employee.v1.EmployeeService/GetEmployeeById | jq
{
  "employee": {
    "id": 1,
    "details": {
      "forename": "Jens",
      "surname": "Neuse",
      "location": {
        "key": {
          "name": "Germany"
        }
      }
    }
  }
}
```

### grpc-web
```bash
buf curl --protocol grpcweb \
  --schema ./pkg/connectrpc/samples/services/employee.v1/service.proto \
  --data '{"employee_id": 1}' \
  http://localhost:5026/employee.v1.EmployeeService/GetEmployeeById
```

## Service Discovery Convention

Services are discovered using this directory structure:
```
services/
├── employee.v1/
│   ├── employee.proto          # Service definition
│   └── operations/
│       ├── GetEmployee.graphql
│       └── UpdateEmployee.graphql
└── product.v1/
    ├── product.proto
    └── operations/
        └── GetProduct.graphql
```

## Related Issues

Closes #ENG-8277

## Release Notes

* **New Features**
  * Added ConnectRPC server support with gRPC, gRPC-Web, and Connect protocol transcoding.
  * Implemented automatic service discovery and GraphQL operation execution for protobuf services.
  * Added dynamic service reloading and graceful shutdown lifecycle.

* **Configuration**
  * New ConnectRPC configuration options for server listen address, storage provider, and GraphQL endpoint.
  * Updated schema and validation for ConnectRPC settings.

* **Tests**
  * Comprehensive test coverage for service discovery, protocol handling, error scenarios, and server lifecycle.

### DIFF
--- a/router/pkg/connectrpc/handler.go
+++ b/router/pkg/connectrpc/handler.go
@@ -416,7 +416,11 @@ func toUpperSnakeCase(s string) string {
 // This follows Relay's error classification pattern for critical errors.
 func (h *RPCHandler) makeCriticalGraphQLError(errors []GraphQLError, httpStatus int) error {
 	// Serialize GraphQL errors to JSON for metadata
-	errorsJSON, _ := json.Marshal(errors)
+	errorsJSON, err := json.Marshal(errors)
+	if err != nil {
+		h.logger.Error("failed to marshal GraphQL errors", zap.Error(err))
+		return connect.NewError(connect.CodeInternal, fmt.Errorf("failed to marshal GraphQL errors: %w", err))
+	}
 
 	// Create Connect error with CRITICAL classification
 	// Use CodeUnknown for GraphQL errors (not CodeInternal which implies server defects)
@@ -444,7 +448,11 @@ func (h *RPCHandler) makeCriticalGraphQLError(errors []GraphQLError, httpStatus 
 // This follows Relay's pattern for field-level errors where some data was successfully retrieved.
 func (h *RPCHandler) makePartialGraphQLError(errors []GraphQLError, data json.RawMessage, httpStatus int) error {
 	// Serialize errors to JSON for metadata
-	errorsJSON, _ := json.Marshal(errors)
+	errorsJSON, err := json.Marshal(errors)
+	if err != nil {
+		h.logger.Error("failed to marshal GraphQL errors", zap.Error(err))
+		return connect.NewError(connect.CodeInternal, fmt.Errorf("failed to marshal GraphQL errors: %w", err))
+	}
 
 	// Compact the partial data JSON to remove whitespace
 	var compactData bytes.Buffer


protographic/src/operation-to-proto.ts
```diff
@@ -264,7 +264,19 @@ class OperationsToProtoVisitor {
       return;
     }
 
-    // 2. Validate no root-level field aliases (breaks reversibility)
+    // 2. Validate operation name is PascalCase
+    // This ensures exact matching between GraphQL operation names and RPC method names
+    // PascalCase: starts with uppercase, contains at least one lowercase letter
+    if (!/^[A-Z](?=.*[a-z])[a-zA-Z0-9]*$/.test(operationName)) {
+      throw new Error(
+        `Operation name "${operationName}" must be in PascalCase ` +
```
endigma: It should be possible to print some approximation of the correct name using `upperFirst(camelCase(operationName))` like we did before inside the error.

router-tests/testdata/connectrpc/services/employee.v1/service.proto
```diff
@@ -0,0 +1,191 @@
...
+    message Details {
+      string forename = 5;
+      string surname = 6;
+    }
+    int32 id = 1;
+    Details details = 2;
+    Mood current_mood = 3;
+    Mood derived_mood = 4;
+  }
+  // This mutation update the mood of an employee.
```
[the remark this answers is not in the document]
endigma: @asoorm where is generator getting this comment from?

router-tests/connectrpc/README.md
endigma, line 1: Maybe just godoc, we don't need to fill `router-tests` with README's describing functions, this is built into Go. Directory structure is also readily available in the editor sidebar.

router-tests/testdata/connectrpc/README.md
endigma, line 1: Can we cut this down to just explaining how to write new tests?

router-tests/testdata/connectrpc/buf.gen.yaml
```diff
@@ -0,0 +1,15 @@
+version: v2
+managed:
+  enabled: true
+  override:
+    - file_option: go_package_prefix
+      value: github.com/wundergraph/cosmo/router-tests/testdata/connectrpc/client
+plugins:
+  - remote: buf.build/protocolbuffers/go
```
endigma: I believe remote plugins have generation rate limits? Can we avoid this?

endigma: https://buf.build/docs/configuration/v2/buf-gen-yaml/#type-of-plugin

router/core/router.go
```diff
@@ -955,6 +956,88 @@ func (r *Router) bootstrap(ctx context.Context) error {
 		r.mcpServer = mcpss
 	}
 
+	if r.connectRPC.Enabled {
+		if r.connectRPC.ServicesProviderID == "" {
```
endigma: This can and should be enforced in schema

router/core/router.go
```diff
@@ -955,6 +956,88 @@ func (r *Router) bootstrap(ctx context.Context) error {
...
+			if provider.ID == r.connectRPC.ServicesProviderID {
+				servicesDir = provider.Path
+				found = true
+				r.logger.Debug("Resolved services provider",
+					zap.String("provider_id", provider.ID),
+					zap.String("path", provider.Path))
+				break
+			}
+		}
+		if !found {
```
endigma: Could just check the `servicesDir` variable for contents

router/core/router.go
```diff
@@ -955,6 +956,88 @@ func (r *Router) bootstrap(ctx context.Context) error {
...
+		})
+		if err != nil {
+			return fmt.Errorf("failed to discover ConnectRPC services: %w", err)
+		}
+
+		r.logger.Info("Discovered ConnectRPC services",
+			zap.Int("service_count", len(discoveredServices)))
+
+		// Determine the router GraphQL endpoint
+		var routerGraphQLEndpoint string
```
endigma: I see this is also how MCP works, but have you tried directly calling the router's HTTP handler with your request instead of using an http client?

router/core/router.go
```diff
@@ -955,6 +956,88 @@ func (r *Router) bootstrap(ctx context.Context) error {
...
+
+		// Initialize the ConnectRPC server with the services directory
+		serverConfig := connectrpc.ServerConfig{
+			ServicesDir:     servicesDir,
+			ListenAddr:      r.connectRPC.Server.ListenAddr,
+			GraphQLEndpoint: routerGraphQLEndpoint,
+			Logger:          r.logger,
+		}
+
+		r.logger.Debug("Creating ConnectRPC server",
```
endigma: Should be `info` level I think

router/core/router.go
```diff
@@ -1474,6 +1557,19 @@ func (r *Router) Shutdown(ctx context.Context) error {
 		}()
 	}
 
+	if r.connectRPCServer != nil {
+		wg.Add(1)
```
endigma: We have [`waitgroup.Go`](https://pkg.go.dev/sync#WaitGroup.Go) since Go1.25

router/core/router_config.go
```diff
@@ -107,6 +107,7 @@ type Config struct {
 	retryOptions                    retrytransport.RetryOptions
 	redisClient                     rd.RDCloser
 	mcpServer                       *mcpserver.GraphQLSchemaServer
+	connectRPCServer                interface{} // Will be *connectrpc.Server
```
endigma: Why is this typed as `interface{}`? Does it need to be? If it does, it should be `any`.

router/pkg/config/config.go
```diff
@@ -1009,6 +1009,18 @@ type MCPServer struct {
 	BaseURL    string `yaml:"base_url,omitempty" env:"MCP_SERVER_BASE_URL"`
 }
 
+type ConnectRPCConfiguration struct {
```
endigma, line 1045: Please use `envprefix`. See [contributing docs](https://github.com/wundergraph/cosmo/blob/main/router/docs/Configuration.md).

router/pkg/config/config.schema.json
```diff
@@ -2134,6 +2134,64 @@
         }
       }
     },
+    "connect_rpc": {
```
endigma, line 2322: You can add checks to enforce required values and relationships between values here in the schema, it will be enforced when the config is parsed.

router/pkg/connectrpc/samples/service.graphqls
endigma, line 1: If this isn't used anywhere in a test or in the router, I don't believe it should be in here. Perhaps move to `demo/` or somewhere else. 

cc @StarpTech

router/pkg/connectrpc/constructor_validation_test.go
```diff
@@ -0,0 +1,182 @@
...
+import (
+	"net/http"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/zap"
+)
+
+// TestConstructorValidation consolidates all constructor validation tests
```
endigma: What does this comment mean?

router/pkg/connectrpc/error_handling_test.go
```diff
@@ -0,0 +1,376 @@
...
+	graphqlResponse       string
+	expectedConnectCode   connect.Code
+	expectedErrorContains string
+	expectedMetadata      map[string]string
+	expectedGraphQLErrors string // JSON string for GraphQL errors
+	expectedPartialData   string // JSON string for partial data
+}
+
+// TestHTTPStatusToConnectCode tests the mapping of HTTP status codes to Connect error codes
+func TestHTTPStatusToConnectCode(t *testing.T) {
```
endigma: This doesn't actually cover all of the cases in `httpStatusToConnectCode`, however I'd honestly say to just remove this, the tested function is just a switch case.

router/pkg/connectrpc/handler.go
```diff
@@ -0,0 +1,563 @@
...
+func headersFromContext(ctx context.Context) (http.Header, error) {
+	headers, ok := ctx.Value(requestHeadersKey{}).(http.Header)
+	if !ok {
+		return nil, fmt.Errorf("missing request headers")
+	}
+	return headers, nil
+}
+
+// skippedHeaders are headers that should not be forwarded to the GraphQL endpoint
+var skippedHeaders = map[string]struct{}{
```
endigma: Possible to share with MCP?

router/pkg/connectrpc/handler.go
```diff
@@ -0,0 +1,563 @@
...
+	if config.GraphQLEndpoint == "" {
+		return nil, fmt.Errorf("graphql endpoint cannot be empty")
+	}
+
+	if config.HTTPClient == nil {
+		return nil, fmt.Errorf("http client cannot be nil")
+	}
+
+	if config.Logger == nil {
+		config.Logger = zap.NewNop()
```
endigma: This should probably error instead of failing silently, is this behaviour ever desirable outside of tests?

router/pkg/connectrpc/handler.go
```diff
@@ -0,0 +1,563 @@
...
+
+// convertProtoJSONToGraphQLVariables converts proto JSON (snake_case) to GraphQL variables (camelCase)
+func (h *RPCHandler) convertProtoJSONToGraphQLVariables(protoJSON []byte) (json.RawMessage, error) {
+	// Handle empty JSON - return empty object
+	if len(protoJSON) == 0 {
+		return json.RawMessage("{}"), nil
+	}
+
+	// Parse the proto JSON
+	var protoData map[string]interface{}
```
endigma:
```suggestion
	var protoData map[string]any
```

router/pkg/connectrpc/handler.go
```diff
@@ -0,0 +1,563 @@
...
+	}
+
+	return graphqlJSON, nil
+}
+
+// convertKeysRecursive recursively converts all map keys from snake_case to camelCase
+// It handles nested maps and arrays of maps
+func convertKeysRecursive(data interface{}) interface{} {
+	switch v := data.(type) {
+	case map[string]interface{}:
```
endigma:
```suggestion
	case map[string]any:
```

endigma: many cases of this, I recommend you set up [`modernize`](https://pkg.go.dev/golang.org/x/tools/gopls/internal/analysis/modernize#waitgroup) from go vet in whatever tooling you're using to write Go.

router/pkg/connectrpc/handler.go
```diff
@@ -0,0 +1,563 @@
...
+
+// snakeToCamel converts snake_case to camelCase
+func snakeToCamel(s string) string {
+	parts := strings.Split(s, "_")
+	if len(parts) == 1 {
+		return s
+	}
+
+	result := parts[0]
+	for i := 1; i < len(parts); i++ {
```
endigma: This has a modernize warning as well

router/pkg/connectrpc/handler.go
```diff
@@ -0,0 +1,563 @@
...
+		if len(parts[i]) > 0 {
+			result += strings.ToUpper(parts[i][:1]) + parts[i][1:]
```
endigma: strings.Builder may be faster

router/pkg/connectrpc/handler.go
```diff
@@ -0,0 +1,563 @@
...
+	connectErr.Meta().Set(MetaKeyGraphQLErrors, string(errorsJSON))
+	connectErr.Meta().Set(MetaKeyGraphQLPartialData, string(data))
+	connectErr.Meta().Set(MetaKeyHTTPStatus, fmt.Sprintf("%d", httpStatus))
+	
+	// Log warning for partial success
+	var errorMessages []string
+	for _, gqlErr := range errors {
+		errorMessages = append(errorMessages, gqlErr.Message)
+	}
+	h.logger.Warn("NON-CRITICAL GraphQL errors - partial data returned",
```
endigma: Nit: Possible synonym for `NON-CRITICAL`? Would make skimming logs by eye easier if we ever need to.

router/pkg/connectrpc/handler.go
```diff
@@ -0,0 +1,563 @@
...
+
+// executeGraphQL executes a GraphQL query against the router endpoint
+func (h *RPCHandler) executeGraphQL(ctx context.Context, query string, variables json.RawMessage) ([]byte, error) {
+	// Create the GraphQL request
+	graphqlRequest := GraphQLRequest{
+		Query:     query,
+		Variables: variables,
+	}
+
+	requestBody, err := json.Marshal(graphqlRequest)
```
endigma: Making a `json.Encoder` for a byte buffer which can be read directly may be faster and/or lower allocations

router/pkg/connectrpc/handler.go
```diff
@@ -0,0 +1,563 @@
...
+
+	// Success case: Return only the data field
+	// The proto response message expects just the data payload: {...}
+	// Not the GraphQL wrapper: {"data": {...}, "errors": [...]}
+	if len(graphqlResponse.Data) > 0 && string(graphqlResponse.Data) != "null" {
+		return graphqlResponse.Data, nil
+	}
+
+	// Edge case: No errors but also no data (empty response)
+	// Return empty object for backward compatibility
```
endigma: Backward compatibility with what? Does the caller of this function tolerate just getting `{}, nil`?

router/pkg/connectrpc/handler.go
```diff
@@ -0,0 +1,563 @@
...
+	if h.operationRegistry == nil {
+		return fmt.Errorf("operation registry is nil")
+	}
+
+	// This method is no longer functional with service-scoped operations
+	// Operations must be loaded per service using LoadOperationsForService
+	h.logger.Warn("Reload() is deprecated - operations must be loaded per service")
+	
+	return nil
+}
```
endigma: When will it be removed? 😆

router/pkg/connectrpc/handler.go
```diff
@@ -0,0 +1,563 @@
...
+// The returned data should be treated as read-only.
+func (h *RPCHandler) GetOperations() interface{} {
+	if h.operationRegistry == nil {
+		return nil
+	}
+	return h.operationRegistry.GetAllOperations()
+}
+
+// ValidateOperation checks if an operation is available for a specific service
+func (h *RPCHandler) ValidateOperation(serviceName, methodName string) error {
```
endigma: More descriptive name perhaps?

router/pkg/connectrpc/handler_test.go
```diff
@@ -0,0 +1,606 @@
...
+		assert.Contains(t, err.Error(), "operation not found")
+	})
+}
+
+// TestExecuteGraphQL_ForwardsHeadersFromContext tests that headers are properly forwarded from context
+func TestExecuteGraphQL_ForwardsHeadersFromContext(t *testing.T) {
+	logger := zap.NewNop()
+	operationRegistry := NewOperationRegistry(logger)
+
+	t.Run("forwards headers from context", func(t *testing.T) {
```
endigma: Nit: some test names in here don't describe well what the expected behaviour should be, perhaps try naming them structurally with subtests more like:

test execute graphql
    forwarding headers from context
        should forward listed headers
        should not forward unlisted headers

and let that "sentence" (e.g. "test execute graphql forwarding headers from context should forward listed headers" guide your testing structure

router/pkg/connectrpc/handler_test.go
```diff
@@ -0,0 +1,606 @@
...
+			ProtoLoader:       protoLoader,
+		})
+		require.NoError(t, err)
+
+		ctx := context.Background()
+		_, err = handler.executeGraphQL(ctx, "query { test }", nil)
+
+		// Should return an error (not a Connect error, just a regular error)
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), "failed to execute HTTP request")
```
endigma: `assert.ErrContains` or something is a function

router/pkg/connectrpc/handler.go
endigma: maybe try just ranging over the string and building character by character and sending uppercase for next char when you hit an `_`

router/pkg/connectrpc/proto_loader.go
```diff
@@ -0,0 +1,355 @@
...
+		err := pl.files.RegisterFile(protoFd)
+		if err != nil {
+			pl.logger.Warn("file descriptor registration failed in local registry",
+				zap.String("file", string(protoFd.Path())),
+				zap.Error(err))
+			return fmt.Errorf("failed to register file descriptor in local registry: %w", err)
+		}
+		pl.logger.Debug("file descriptor registered successfully in local registry",
+			zap.String("file", string(protoFd.Path())))
+	}
```
endigma, line 319: Why not fail-fast?

router/pkg/connectrpc/server.go
```diff
@@ -0,0 +1,410 @@
...
+				return fmt.Errorf("failed to load operations for service %s: %w", service.FullName, err)
+			}
+			s.logger.Info("loaded operations for service",
+				zap.String("service", service.FullName),
+				zap.Int("count", s.operationRegistry.CountForService(service.FullName)))
+		} else {
+			s.logger.Warn("no operations found for service",
+				zap.String("service", service.FullName))
+		}
+	}
```
endigma: Would it be better to do this in NewServer instead of at start?

router-tests/testdata/connectrpc/services/employee.v1/service.proto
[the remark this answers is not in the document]
asoorm: @endigma 

```
**./demo/pkg/subgraphs/mood/subgraph/generated/generated.go:    """ This mutation update the mood of an employee. """
./demo/pkg/subgraphs/mood/subgraph/schema.graphqls:    """ This mutation update the mood of an employee. """
```

router/pkg/connectrpc/proto_loader.go
asoorm: Yeah - good catch - I think that I initially had in global registry - and I was hacking around to getting it working with local registry. Will clean up.

router/pkg/connectrpc/server.go
asoorm: Current design / thought process:

NewServer() for validating config
Start() for loading services from disk and to start the server.
Reload() to reload services from disk / update the server.

HAppy to change though.

router/core/router.go
asoorm: I did think that, and did look into it briefly - but I got stuck with circular dependency issues. I went for the quick win and copied the MCP Server approach. Further down the line - I think it would be much better approach though.

router/pkg/connectrpc/handler.go
asoorm: Yeah - that's a bad comment - updated.

router/pkg/connectrpc/samples/service.graphqls
asoorm: So I put it there because it's associated with the GraphQL operations in the tests. The operations don't really make sense without the schema. even if it's not used anywhere.

Happy to remove it though -

router/pkg/connectrpc/constructor_validation_test.go
asoorm: This tests whether constructors are validating their inputs. But it's multiple constructors - so that's why I used the word consolidating. I'll make the comment a bit clearer.

router/go.mod
```diff
@@ -72,6 +73,7 @@ require (
 	github.com/hashicorp/go-hclog v1.6.3
 	github.com/hashicorp/go-plugin v1.6.3
 	github.com/iancoleman/strcase v0.3.0
+	github.com/jhump/protoreflect v1.17.0
```
Noroth: I do understand you want to use this library for parsing. My question is, we are already using the connectrpc dependencies and in our engine we use the protocompile `github.com/bufbuild/protocompile`.
Wouldn't it make more sense to use the existing dependencies instead of introducing a new one?

router/pkg/connectrpc/handler.go
```diff
@@ -0,0 +1,665 @@
...
+	Path       []any                  `json:"path,omitempty"`
+	Locations  []GraphQLErrorLocation `json:"locations,omitempty"`
+	Extensions map[string]any         `json:"extensions,omitempty"`
+}
+
+// GraphQLResponse represents a GraphQL response structure
+type GraphQLResponse struct {
+	Errors []GraphQLError  `json:"errors,omitempty"`
+	Data   json.RawMessage `json:"data,omitempty"`
+}
```
Noroth, line 84: Those types (at least part of them) already exist in the `mcpserver` package. Wouldn't it make sense to move them to a shared package and re-use them instead of defining them again?

router/pkg/connectrpc/vanguard_service.go
```diff
@@ -0,0 +1,408 @@
...
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+
+	"connectrpc.com/connect"
+	"connectrpc.com/vanguard"
+	"go.uber.org/zap"
+	"google.golang.org/protobuf/reflect/protoreflect"
```
Noroth, line 14: Here you are in fact using protoreflect from the google library. It's a bit inconsistent

router/pkg/connectrpc/proto_loader.go
```diff
@@ -0,0 +1,391 @@
+package connectrpc
```
Noroth, line 1: It feels a bit like a reimplementation of https://github.com/wundergraph/graphql-go-tools/blob/master/v2/pkg/engine/datasource/grpc_datasource/compiler.go

router/pkg/connectrpc/operation_registry.go
```diff
@@ -0,0 +1,294 @@
...
+				opType = "mutation"
+			case ast.OperationTypeSubscription:
+				opType = "subscription"
+			default:
+				return "", "", fmt.Errorf("unknown operation type")
+			}
+
+			opName := ""
+			if opDef.Name.Length() > 0 {
+				opName = string(doc.Input.ByteSlice(opDef.Name))
```
Noroth: You can use `ByteSliceString`

router/pkg/connectrpc/handler.go
```diff
@@ -0,0 +1,665 @@
...
+	"connectrpc.com/connect"
+	"github.com/jhump/protoreflect/desc"
+	"go.uber.org/zap"
+
+	"github.com/wundergraph/cosmo/router/pkg/httputil"
+)
+
+// httpStatusToConnectCode maps HTTP status codes to Connect error codes
+// Based on Connect RPC specification and common HTTP status code semantics
+func httpStatusToConnectCode(statusCode int) connect.Code {
```
Noroth: Similar method in other file. Mabe create a util file where you put both functions?

router/pkg/connectrpc/operation_registry.go
```diff
@@ -0,0 +1,294 @@
...
+func (r *OperationRegistry) CountForService(serviceName string) int {
+	r.mu.RLock()
+	defer r.mu.RUnlock()
+
+	serviceOps, exists := r.operations[serviceName]
+	if !exists {
+		return 0
+	}
+
+	return len(serviceOps)
```
Noroth: I don't think you need to check `exists` here.

```suggestion
	return len(r.operations[serviceName])
```

router/pkg/connectrpc/handler.go
```diff
@@ -0,0 +1,665 @@
...
+		// Create Connect error with metadata
+		connectErr := connect.NewError(code, fmt.Errorf("GraphQL request failed with HTTP %d", resp.StatusCode))
+		connectErr.Meta().Set(MetaKeyErrorClassification, ErrorClassificationCritical)
+		connectErr.Meta().Set(MetaKeyHTTPStatus, fmt.Sprintf("%d", resp.StatusCode))
+		connectErr.Meta().Set(MetaKeyHTTPResponseBody, string(responseBody))
+
+		h.logger.Error("HTTP error from GraphQL endpoint",
+			zap.Int("status_code", resp.StatusCode),
+			zap.String("connect_code", code.String()),
+			zap.String("response_body", string(responseBody)))
```
Noroth, line 550: Could it potentially be a security issue when you log the whole response body?

router/pkg/connectrpc/handler.go
```diff
@@ -0,0 +1,665 @@
...
+		return nil, connectErr
+	}
+
+	// Parse the GraphQL response to check for errors
+	var graphqlResponse GraphQLResponse
+	if err := json.Unmarshal(responseBody, &graphqlResponse); err != nil {
+		// If we can't parse it, return the raw response (backward compatibility)
+		h.logger.Error("failed to parse GraphQL response",
+			zap.Error(err),
+			zap.String("response_body", string(responseBody)))
```
Noroth: Same here

router-tests/connectrpc/connectrpc_client_test.go
```diff
@@ -0,0 +1,295 @@
...
+// using generated client code to ensure proper multi-protocol support
+func TestConnectRPC_ClientProtocols(t *testing.T) {
+	t.Parallel()
+
+	// Use shared helper for employee GraphQL handler
+	ts := NewTestConnectRPCServer(t, ConnectRPCServerOptions{
+		GraphQLHandler: EmployeeGraphQLHandler(),
+	})
+	
+	err := ts.Start()
```
jensneuse, line 30: feels like there should be a `defer ts.End()` statement

router/pkg/config/config.go
```diff
@@ -1038,10 +1050,11 @@ type Config struct {
...
 	Cluster        Cluster            `yaml:"cluster,omitempty"`
 	Compliance     ComplianceConfig   `yaml:"compliance,omitempty"`
-	TLS            TLSConfiguration   `yaml:"tls,omitempty"`
-	CacheControl   CacheControlPolicy `yaml:"cache_control_policy"`
-	MCP            MCPConfiguration   `yaml:"mcp,omitempty"`
-	DemoMode       bool               `yaml:"demo_mode,omitempty" envDefault:"false" env:"DEMO_MODE"`
+	TLS            TLSConfiguration        `yaml:"tls,omitempty"`
+	CacheControl   CacheControlPolicy      `yaml:"cache_control_policy"`
+	MCP            MCPConfiguration        `yaml:"mcp,omitempty"`
+	ConnectRPC     ConnectRPCConfiguration `yaml:"connect_rpc,omitempty"`
```
jensneuse, line 1089: Do I understand correctly that I can only have a single connect rpc config? Would it make sense to support plural here? (allowing the user to have multiple configurations, e.g. for different use cases, different audiences, with different authentication/authorization?

router/pkg/config/config.go
```diff
@@ -1013,6 +1013,18 @@ type MCPServer struct {
 	BaseURL    string `yaml:"base_url,omitempty" env:"MCP_SERVER_BASE_URL"`
 }
 
+type ConnectRPCConfiguration struct {
+	Enabled            bool             `yaml:"enabled" envDefault:"false" env:"CONNECT_RPC_ENABLED"`
+	Server             ConnectRPCServer `yaml:"server,omitempty" envPrefix:"CONNECT_RPC_SERVER_"`
+	ServicesProviderID string           `yaml:"services_provider_id,omitempty" env:"CONNECT_RPC_SERVICES_PROVIDER_ID"`
+	GraphQLEndpoint    string           `yaml:"graphql_endpoint,omitempty" env:"CONNECT_RPC_GRAPHQL_ENDPOINT"`
```
jensneuse: Why do I need to configure this? Are we not using internal networking to directly call into the graphql handler?

router/pkg/connectrpc/connect_util.go
```diff
@@ -0,0 +1,80 @@
...
+	case http.StatusBadRequest: // 400
+		return connect.CodeInvalidArgument
+	case http.StatusUnauthorized: // 401
+		return connect.CodeUnauthenticated
+	case http.StatusForbidden: // 403
+		return connect.CodePermissionDenied
+	case http.StatusNotFound: // 404
+		return connect.CodeNotFound
+	case http.StatusConflict: // 409
+		return connect.CodeAborted
```
jensneuse, line 24: why aborted for conflict?

router/pkg/connectrpc/handler.go
```diff
@@ -0,0 +1,629 @@
...
+type requestHeadersKey struct{}
+
+// withRequestHeaders adds request headers to the context
+func withRequestHeaders(ctx context.Context, headers http.Header) context.Context {
+	return context.WithValue(ctx, requestHeadersKey{}, headers)
+}
+
+// headersFromContext extracts the request headers from the context
+func headersFromContext(ctx context.Context) (http.Header, error) {
+	headers, ok := ctx.Value(requestHeadersKey{}).(http.Header)
```
jensneuse: if the ctx.Value is nil, will it panic if you cast it to http.Hxeader?

router/pkg/config/config.go
jensneuse: How do we configure different clients/audiences in regards to authentication? E.g. let's say I have 2 different clients, each with different JWT configuration, how does it work?

Do I have to configure header forwarding? or are we forwarding all headers always?

router/pkg/connectrpc/handler_test.go
```diff
@@ -0,0 +1,261 @@
...
+	"go.uber.org/zap"
+)
+
+func TestConvertProtoJSONToGraphQLVariables(t *testing.T) {
+	logger := zap.NewNop()
+	handler := &RPCHandler{logger: logger}
+
+	t.Run("converts snake_case to camelCase", func(t *testing.T) {
+		protoJSON := []byte(`{"employee_id": 1, "first_name": "John"}`)
+		result, err := handler.convertProtoJSONToGraphQLVariables("test.Service", "TestMethod", protoJSON)
```
jensneuse, line 35: do you do this by default? what if the GraphQL server is not using camelCase?

router/pkg/connectrpc/handler.go
```diff
@@ -0,0 +1,629 @@
...
+
+	h.logger.Debug("resolved operation",
+		zap.String("service", serviceName),
+		zap.String("rpc_method", methodName),
+		zap.String("operation", operation.Name),
+		zap.String("type", operation.OperationType))
+
+	// Convert proto JSON (snake_case) to GraphQL variables (camelCase)
+	// Pass service and method info for schema-aware enum detection
+	variables, err := h.convertProtoJSONToGraphQLVariables(serviceName, methodName, requestJSON)
```
jensneuse, line 180: I don't understand why this is happening. Are you assuming that proto is always snake and GraphQL is always camelCase? I don't think we can make such assumptions.

router/pkg/connectrpc/handler.go
```diff
@@ -0,0 +1,629 @@
...
+		return nil, fmt.Errorf("failed to execute GraphQL query: %w", err)
+	}
+
+	return responseJSON, nil
+}
+
+// convertProtoJSONToGraphQLVariables converts proto JSON to GraphQL variables:
+// - Converts field names from snake_case to camelCase
+// - Strips proto enum prefixes (MOOD_HAPPY -> HAPPY) using schema-aware detection
+func (h *RPCHandler) convertProtoJSONToGraphQLVariables(serviceName, methodName string, protoJSON []byte) (json.RawMessage, error) {
```
jensneuse, line 216: I'm not sure if this works for any schema

router/pkg/connectrpc/handler.go
```diff
@@ -0,0 +1,629 @@
...
+		return nil, fmt.Errorf("failed to marshal GraphQL variables: %w", err)
+	}
+
+	return graphqlJSON, nil
+}
+
+// convertKeysRecursive converts map keys from snake_case to camelCase
+// and strips proto enum prefixes using schema information when available
+// Omits fields with empty string values (e.g., from _UNSPECIFIED enum values)
+func (h *RPCHandler) convertKeysRecursive(data any, messageDesc protoreflect.MessageDescriptor) any {
```
jensneuse: I don't think this is right. We cannot convert proto and graphql back and forth without introspecting each schema and doing a lookup on how to name a key.

router/pkg/connectrpc/handler.go
```diff
@@ -0,0 +1,629 @@
...
+			capitalizeNext = false
+		} else {
+			builder.WriteByte(ch)
+		}
+	}
+
+	result := builder.String()
+
+	// Special handling for ID suffix: employeeId -> employeeID
+	if strings.HasSuffix(result, "Id") && len(result) > 2 {
```
jensneuse: why?

router/pkg/connectrpc/handler.go
```diff
@@ -0,0 +1,629 @@
...
+	}
+
+	// Forward headers from the original RPC request
+	headers, err := headersFromContext(ctx)
+	if err != nil {
+		h.logger.Debug("no headers in context", zap.Error(err))
+	} else {
+		// Copy headers, skipping those that shouldn't be forwarded
+		for key, values := range headers {
+			if _, skip := httputil.SkippedHeaders[key]; skip {
```
jensneuse: should we use the `headers.Get()` method instead to ensure proper casing?

router/core/router.go
```diff
@@ -972,6 +970,74 @@ func (r *Router) bootstrap(ctx context.Context) error {
...
+					zap.String("path", provider.Path))
+				break
+			}
+		}
+		if servicesDir == "" {
+			return fmt.Errorf("services storage provider with id '%s' for connect_rpc not found", r.connectRPC.ServicesProviderID)
+		}
+
+		// Discover services using convention-based approach
+		discoveredServices, err := connectrpc.DiscoverServices(connectrpc.ServiceDiscoveryConfig{
```
jensneuse, line 949: should there be a flag to enable discovery?

router/core/router.go
```diff
@@ -972,6 +970,74 @@ func (r *Router) bootstrap(ctx context.Context) error {
...
+		// Determine the router GraphQL endpoint
+		var routerGraphQLEndpoint string
+		if r.connectRPC.GraphQLEndpoint != "" {
+			routerGraphQLEndpoint = r.connectRPC.GraphQLEndpoint
+		} else {
+			routerGraphQLEndpoint = path.Join(r.listenAddr, r.graphqlPath)
+		}
+
+		// Initialize the ConnectRPC server with the services directory
+		serverConfig := connectrpc.ServerConfig{
```
jensneuse, line 963: You're not configuring all fields, e.g. Timeout, which should be configured through config.go

router/core/router.go
```diff
@@ -1472,97 +1538,78 @@ func (r *Router) Shutdown(ctx context.Context) error {
...
 	}
 
-	if r.tracerProvider != nil {
-		wg.Add(1)
-
-		go func() {
-			defer wg.Done()
+	if r.connectRPCServer != nil {
+		wg.Go(func() {
+			if subErr := r.connectRPCServer.Stop(ctx); subErr != nil {
```
jensneuse, line 1515: if an error occured in R1026, connectRPCServer will be nil, so this will panic

router/core/router.go
```diff
@@ -972,6 +970,74 @@ func (r *Router) bootstrap(ctx context.Context) error {
...
+		if r.connectRPC.GraphQLEndpoint != "" {
+			routerGraphQLEndpoint = r.connectRPC.GraphQLEndpoint
+		} else {
+			routerGraphQLEndpoint = path.Join(r.listenAddr, r.graphqlPath)
+		}
+
+		// Initialize the ConnectRPC server with the services directory
+		serverConfig := connectrpc.ServerConfig{
+			ServicesDir:     servicesDir,
+			ListenAddr:      r.connectRPC.Server.ListenAddr,
```
jensneuse, line 965: Is it the right thing to have connectRPC listen on a different addr compared to mount the connect RPC listener on the same addr like the GraphQL handler but have it prefixed with a path or another method?

router/pkg/connectrpc/operation_registry.go
```diff
@@ -0,0 +1,289 @@
...
+// Operations are scoped to their service (package.service) and cached in memory
+// for fast access during request handling.
+//
+// Thread-safety: This registry is designed to be safe for concurrent use with
+// multiple readers and occasional writers. Read operations (Get*, Has*, Count*)
+// can be called concurrently, and write operations (Load*, Clear*) use proper
+// locking to ensure consistency.
+type OperationRegistry struct {
+	// Service-scoped operations: serviceName (package.service) -> operationName -> Operation
+	operations map[string]map[string]*schemaloader.Operation
```
jensneuse, line 17: For read heavy ops, like this one, use sync.Map instead. I've done benchmarking for a similar use case and sync.Map is much better optimized for this use case compared to a map/map with mutex.

Another thought to consider:

If we create the map once and then "only" read from it, even concurrently, we don't need a mutex at all.
Depending on the architecture, it's possible to populate the map at creation time of the rpc handler and then never change it. Instead of modifying operations, you can swap the entire rpc handler and create a new registry. With this approach, no concurrency handling on read is required, so the code is simpler and much more efficient because we can remove locks.

router/pkg/connectrpc/operation_registry.go
```diff
@@ -0,0 +1,289 @@
...
+	return &OperationRegistry{
+		operations: make(map[string]map[string]*schemaloader.Operation),
+		logger:     logger,
+	}
+}
+
+// LoadOperationsForService loads GraphQL operations for a specific service from operation files.
+// Operations are scoped to the service's fully qualified name (package.service).
+// This method is thread-safe and can be called multiple times for different services.
+func (r *OperationRegistry) LoadOperationsForService(serviceName string, operationFiles []string) error {
```
jensneuse: We need to be careful on all such exposed funcs if they (somehow) allow the user to pass path params from the client side. I don't know exactly all cases, so please verify that user input is always validated and that the user cannot form malicious strings, e.g. to somehow explore the file system of the router.

router/pkg/connectrpc/server.go
```diff
@@ -0,0 +1,403 @@
...
+	if config.Logger == nil {
+		return nil, fmt.Errorf("logger is required")
+	}
+
+	if config.RequestTimeout == 0 {
+		config.RequestTimeout = 30 * time.Second
+	}
+
+	// Add protocol if missing
+	if !strings.Contains(config.GraphQLEndpoint, "://") {
```
jensneuse: Why is this necessary? should validation not have caught this?

router/pkg/connectrpc/vanguard_service.go
```diff
@@ -0,0 +1,374 @@
...
+
+// VanguardServiceConfig holds configuration for creating a Vanguard service
+type VanguardServiceConfig struct {
+	Handler     *RPCHandler
+	ProtoLoader *ProtoLoader
+	Logger      *zap.Logger
+}
+
+// VanguardService wraps the RPC handler and creates Vanguard services
+type VanguardService struct {
```
jensneuse, line 39: We should have a better description on what this service is and what it does. The description of the wrapped package "connectrpc.com/vanguard" is pretty good.

router/pkg/connectrpc/vanguard_service.go
```diff
@@ -0,0 +1,374 @@
...
+			zap.String("service_path", servicePath))
+
+		// Configure to always transcode to Connect protocol with JSON codec
+		// This ensures our handler always receives JSON, regardless of the incoming protocol
+		// Use NewServiceWithSchema to provide the schema directly with a custom type resolver
+		vanguardService := vanguard.NewServiceWithSchema(
+			serviceDef.ServiceDescriptor,
+			serviceHandler,
+			vanguard.WithTargetProtocols(vanguard.ProtocolConnect),
+			vanguard.WithTargetCodecs("json"),
```
jensneuse, line 135: should this be hardcoded?

router/pkg/connectrpc/vanguard_service.go
```diff
@@ -0,0 +1,374 @@
...
+			if messageParam == "" {
+				// For methods with no input parameters, use empty JSON object
+				requestBody = []byte("{}")
+			} else {
+				// The message parameter is already URL-decoded by r.URL.Query().Get()
+				requestBody = []byte(messageParam)
+			}
+			vs.logger.Debug("extracted message from GET query parameter",
+				zap.String("message", string(requestBody)))
+		} else {
```
jensneuse, line 197: Is it always POST? could it be OPTIONS? no clue, just asking

router/pkg/httputil/headers.go
```diff
@@ -0,0 +1,27 @@
+package httputil
```
jensneuse: I think we're using very similar code somewhere else in the Router.
Can we consolidate this and go away from calling a package "httputil" (not idiomatic go) and instead give files meaning?

router/pkg/mcpserver/server.go
```diff
@@ -716,7 +695,7 @@ func (s *GraphQLSchemaServer) executeGraphQLQuery(ctx context.Context, query str
 		// Copy all headers from the MCP request
 		for key, values := range headers {
 			// Skip headers that should not be forwarded
-			if _, ok := skippedHeaders[key]; ok {
+			if _, ok := httputil.SkippedHeaders[key]; ok {
```
jensneuse: maybe instead of exposing a map we can expose a more meaningful func that makes it clear that it is intended to skip forwarding headers that are just meant for communication between client and intermediary and would cause problems when forwarded to origin servers.

router-tests/connectrpc/connectrpc_client_test.go
asoorm: I've registered a cleanup func in NewTestConnectRPCServer, but will make explicit.

```
	// Register cleanup
	t.Cleanup(func() {
		ts.Close()
	})
```

router/pkg/config/config.go
asoorm: We support multiple servers as follows:

```

### Directory Structure

The router recursively discovers proto files and operations. Recommended structure:

```
services/
└── employee.v1/
    ├── service.proto
    ├── service.proto.lock.json
    ├── GetEmployee.graphql
    └── UpdateEmployee.graphql
```

For multiple services in the same package:

```
services/
└── company.v1/
    ├── EmployeeService/
    │   ├── employee.proto
    │   └── operations/
    └── DepartmentService/
        ├── department.proto
        └── operations/
```

<Note>
  The combination of proto package name and service name must be unique. Nested proto files in subdirectories are not discovered if a parent directory contains a proto file.
</Note>
```

See accompanying docs: https://github.com/wundergraph/cosmo-docs/pull/201/files

router/pkg/config/config.go
asoorm: I initially thought the same - but I aligned with the MCP Server approach for simplicity - and to get an alpha out the door.

The problem is Circular dependencies between `router/core` and `router/pkg/connectrpc` which would prevent direct internal calls.

router/pkg/connectrpc/connect_util.go
asoorm: This is the spec. https://connectrpc.com/docs/protocol#error-codes

Screenshot 2025-12-23 at 11 56 31

router/pkg/connectrpc/vanguard_service.go
asoorm: Yes, this is intentionally hardcoded. By normalising all incoming requests to Connect protocol with the JSON codec, we simplify the RPCHandler implementation - it only needs to handle one format (JSON) regardless of whether clients send gRPC binary protobuf, gRPC-Web, Connect, or REST requests. Vanguard handles all the protocol transcoding for us.

This design trade-off means implementation simplicity (for alpha release) and makes it easier to debug (JSON being easier to log/inspect) over the performance gains of binary protobuf.

router/pkg/connectrpc/vanguard_service.go
asoorm: Good point. We don't yet support CORS - I'll add it.

router/pkg/connectrpc/operation_registry.go
asoorm: operationFiles paths come from configuration, not user input. The ServicesDir is resolved from `storageProviders.FileSystem` in the router config.

File discovery happens at startup - DiscoverServices() walks the configured directory using filepath.Walk, and generates the file paths internally.

That means no HTTP request params control these paths and users cannot inject arbitrary paths.

So our security boundary is who can modify the router's configuration file, and who has filesystem access to the services directory.

router/pkg/connectrpc/server.go
```diff
@@ -0,0 +1,437 @@
...
+// NewServer creates a new ConnectRPC server and loads all services
+func NewServer(config ServerConfig) (*Server, error) {
+	// Validate configuration
+	if config.ServicesDir == "" {
+		return nil, fmt.Errorf("services directory must be provided")
+	}
+
+	if config.GraphQLEndpoint == "" {
+		return nil, fmt.Errorf("graphql endpoint cannot be empty")
+	}
```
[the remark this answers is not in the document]
asoorm: this is checked in the json schema validator.

router/pkg/connectrpc/operation_registry_test.go
```diff
@@ -0,0 +1,452 @@
...
+					_ = registry.GetAllOperationsForService(serviceName)
+					_ = registry.Count()
+					_ = registry.CountForService(serviceName)
+				}
+			})
+		}
+
+		// Wait for all goroutines to complete
+		wg.Wait()
+	})
```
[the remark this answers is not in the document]
asoorm: The code compiles and tests pass successfully
The wg variable is of type errgroup.Group from golang.org/x/sync/errgroup, not sync.WaitGroup
errgroup.Group.Go() is a valid method that handles goroutine lifecycle management automatically
The pattern used is actually more robust than manual WaitGroup management because errgroup also handles error propagation from goroutines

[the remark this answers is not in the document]
asoorm: What version of Go are you checking against? we are using Go 1.25 - perhaps that's why?

router-tests/connectrpc/connectrpc_client_test.go
```diff
@@ -0,0 +1,369 @@
...
+		// Per GraphQL spec, errors at top level should result in an error
+		require.Error(t, err)
+		
+		var connectErr *connect.Error
+		require.ErrorAs(t, err, &connectErr)
+		assert.Equal(t, connect.CodeUnknown, connectErr.Code())
+		assert.Contains(t, connectErr.Message(), "GraphQL partial success with errors")
+
+		// Capture error structure for golden test
+		errorJSON, err := json.MarshalIndent(map[string]interface{}{
```
StarpTech: Could you add a test how a user (client) would react on multiple GraphQL errors including extension codes?

router-tests/connectrpc/connectrpc_client_test.go
```diff
@@ -0,0 +1,369 @@
...
+			"code":    connectErr.Code().String(),
+			"message": connectErr.Message(),
+		}, "", "  ")
+		require.NoError(t, err)
+		g.Assert(t, "error_http_500", errorJSON)
+	})
+}
+
+// TestConnectRPC_ClientConcurrency tests concurrent requests with generated client
+func TestConnectRPC_ClientConcurrency(t *testing.T) {
```
StarpTech: What does it test? That we can make multiple requests?

router-tests/connectrpc/connectrpc_test.go
```diff
@@ -0,0 +1,73 @@
...
+			t.Logf("  - Service: %s", serviceName)
+		}
+		
+		operationCount := ts.GetOperationCount()
+		t.Logf("Discovered %d operation(s)", operationCount)
+	})
+}
+
+// TestConnectRPC_PredefinedMode tests predefined mode functionality
+func TestConnectRPC_PredefinedMode(t *testing.T) {
```
StarpTech: What is predefined mode and how did you activate here?

router/pkg/connectrpc/helpers_test.go
```diff
@@ -0,0 +1,122 @@
...
+	ResponseBody string
+}
+
+// NewMockGraphQLServer creates a new mock GraphQL server that returns the given response
+func NewMockGraphQLServer(responseBody string) *MockGraphQLServer {
+	server := &MockGraphQLServer{
+		ResponseBody: responseBody,
+	}
+
+	server.Server = httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
StarpTech: Do we need to test scenarios where we need to handle statusCode != 200?

router/pkg/connectrpc/helpers_test.go
```diff
@@ -0,0 +1,122 @@
+package connectrpc
+
+import (
```
StarpTech, line 3: Is there a good reason to put this package into the router rather than the `router-tests` pkg?

router/pkg/connectrpc/operation_loader.go
```diff
@@ -0,0 +1,134 @@
...
+			logger.Warn("failed to read operation file",
+				zap.String("file", filePath),
+				zap.Error(err))
+			continue
+		}
+
+		operationString := string(content)
+
+		// Parse to extract operation name and type
+		opDoc, report := astparser.ParseGraphqlDocumentString(operationString)
```
StarpTech, line 50: We should throw an error when someone is trying to provide multiple operations.

```go
	operationCount := len(opDoc.OperationDefinitions)
	if operationCount != 1 {
		return ast.Document{}, fmt.Errorf("expected exactly one operation definition in file %s, got %d", path, operationCount)
	}
```

router/pkg/connectrpc/operation_loader.go
```diff
@@ -0,0 +1,134 @@
...
+			zap.String("type", opType),
+			zap.String("file", filePath))
+	}
+
+	logger.Info("loaded operations for service",
+		zap.String("service", serviceName),
+		zap.Int("operation_count", len(operations)))
+
+	return operations, nil
+}
```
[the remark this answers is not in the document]
StarpTech: @asoorm

router/pkg/connectrpc/operation_loader.go
```diff
@@ -0,0 +1,134 @@
...
+
+// extractOperationInfo extracts the name and type from an operation document
+func extractOperationInfo(doc *ast.Document) (string, string, error) {
```
StarpTech: We have this already in mcp `getOperationNameAndType` let's build helper functions.

router/pkg/connectrpc/helpers_test.go
asoorm: This file is used for unit tests which test internal package behaviour. Examples `RPCHandler` or `VanguardService`.

my thoughts were that `router-tests` is for integration tests - which actually tests the full router e2e.

So for me - I wasn't so keen on creating dependency or mixing concerns between unit tests and the integration test package.

Please advise if you think I should change approach.

router-tests/connectrpc/connectrpc_test.go
asoorm: Good spot - this is an artefact which needs removing.

router-tests/testdata/connectrpc/services/employee.v1/service.proto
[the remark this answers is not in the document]
asoorm: This is a generated file - so actually we need to fix the source - needs to be addressed as part of a separate chore task.

asoorm:
@coderabbitai please review

router/internal/headers/headers.go
```diff
@@ -0,0 +1,37 @@
...
+	"Proxy-Authenticate": {},
+	"Proxy-Authorization": {},
+	"Te":                 {}, // canonicalized version of "TE"
+	"Trailer":            {}, // not Trailers per URL above; https://www.rfc-editor.org/errata_search.php?eid=4522
+	"Transfer-Encoding":  {},
+	"Upgrade":            {},
+
+	// Content Negotiation. We must never propagate the client headers to the upstream
+	// The router has to decide on its own what to send to the upstream
+	"Content-Type":    {},
```
StarpTech: Headers are missing (see main)

router/pkg/connectrpc/testdata/services/employee.v1/MutationUpdateEmployeeMood.graphql
StarpTech, line 1: Do we use them for unit testing?

router/core/router.go
```diff
@@ -967,12 +968,92 @@ func (r *Router) bootstrap(ctx context.Context) error {
...
+		if err != nil {
+			return fmt.Errorf("failed to discover ConnectRPC services: %w", err)
+		}
+
+		// Determine the router GraphQL endpoint
+		var routerGraphQLEndpoint string
+		if r.connectRPC.GraphQLEndpoint != "" {
+			routerGraphQLEndpoint = r.connectRPC.GraphQLEndpoint
+		} else {
+			routerGraphQLEndpoint, err = url.JoinPath(r.baseURL, r.graphqlPath)
```
StarpTech: We do this already in `Start`, we could store it and reuse it here.

asoorm: Good call - I've moved this logic to NewRouter - and now MCP, ConnectRPC and Playground all use it.

router/pkg/connectrpc/testdata/services/employee.v1/MutationUpdateEmployeeMood.graphql
asoorm: Yes, they are test fixtures - appreciate that `samples` directory is misleading - I will rename to make it clearer.

router/pkg/connectrpc/connect_util.go
```diff
@@ -0,0 +1,80 @@
...
+	case http.StatusServiceUnavailable: // 503
+		return connect.CodeUnavailable
+	case http.StatusInternalServerError: // 500
+		return connect.CodeInternal
+	default:
+		// For any other status code (including 2xx success codes),
+		// return CodeUnknown as a safe default
+		return connect.CodeUnknown
+	}
+}
```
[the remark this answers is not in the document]
asoorm: @coderabbitai - PTAL.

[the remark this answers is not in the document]
asoorm: @coderabbit - Feels like this has already been addressed in 5789d3d7e. PTAL.

router-tests/connectrpc/connectrpc_test_helpers.go
```diff
@@ -0,0 +1,244 @@
...
+}
+
+// ErrorGraphQLHandler returns a handler that responds with a GraphQL error
+func ErrorGraphQLHandler(message string) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(http.StatusOK)
+		_, _ = fmt.Fprintf(w, `{"errors": [{"message": "%s"}]}`, message)
+	}
+}
```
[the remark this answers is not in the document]
asoorm: Test Helper - not a concern

router/pkg/connectrpc/handler.go
```diff
@@ -0,0 +1,685 @@
...
+			}
+		}
+		return false
+	case []any:
+		// Check array elements
+		for _, item := range v {
+			if h.needsTransformation(item, messageDesc) {
+				return true
+			}
+		}
```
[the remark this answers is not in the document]
asoorm: @coderabbitai addressed in af0173e. PTAL

PR #2380, opened 2025-11-28
fix: operation checks for input and argument usages

JivusAyrus:
## Summary by CodeRabbit

* **New Features**
  * Metrics now include explicit nullability (IsNull) and per-source IDs (SubgraphIDs). Public APIs expose type/field/argument/input usage with subgraph attribution and variable remapping.

* **Improvements**
  * More accurate, nullability-aware change reporting; some nested argument/input changes now reported at parent paths. Stable sorting of source IDs to reduce metric cardinality.

* **Chores**
  * Database migrations and materialized view updates to persist IsNull. Config defaults added for metrics collection.

* **Tests**
  * Expanded tests and benchmarks for null propagation, remapping, multi-source scenarios, and scalability.

This PR fixes input and argument schema usage tracking:

- Add missing subgraphIds: Input and argument schema usage entries were missing subgraphIds. This PR ensures they’re recorded.

- Track all inputs/arguments referenced by the operation: Previously, input types, input fields, or arguments that were part of an operation but not used weren’t tracked. That gap made it harder (or impossible) to determine which operations would break when an input or argument changed. This PR now tracks all inputs/arguments of the operation, regardless of whether they are used, improving change-impact detection and breakage analysis.

## Checklist

### DIFF
--- a/cli/src/core/client/client.ts
+++ b/cli/src/core/client/client.ts
@@ -19,7 +19,6 @@ export const CreateClient = (opts: ClientOptions): Client => {
   const transport = createConnectTransport({
     // Requests will be made to <baseUrl>/<package>.<service>/method
     baseUrl: opts.baseUrl,
-
     // You have to tell the Node.js http API which HTTP version to use.
     httpVersion: '1.1',
     nodeOptions: {
--- a/router/pkg/graphqlschemausage/schemausage.go
+++ b/router/pkg/graphqlschemausage/schemausage.go
@@ -481,11 +481,6 @@ type subgraphMapper struct {
 	variableToSubgraphs map[string][]string
 }
 
-// getFieldSubgraphs returns subgraph IDs for a field path
-func (s *subgraphMapper) getFieldSubgraphs(pathKey string) []string {
-	return s.fieldToSubgraphs[pathKey]
-}
-
 // getVariableSubgraphs returns subgraph IDs for a variable
 func (s *subgraphMapper) getVariableSubgraphs(varName string) []string {
 	return s.variableToSubgraphs[varName]


graphqlmetrics/migrations/20251120223951_recreate_gql_schema_usage_5m_90d_mv_with_is_null.sql
```diff
@@ -0,0 +1,52 @@
...
+    ClientVersion,
+    Path,
+    FieldName,
+    NamedType,
+    TypeNames,
+    SubgraphIDs,
+    IsArgument,
+    IsInput,
+    IsIndirectFieldUsage,
+    IsNull
```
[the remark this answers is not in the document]
JivusAyrus: i removed them, can you confirm

controlplane/src/core/services/SchemaUsageTrafficInspector.ts
```diff
@@ -3,6 +3,166 @@ import { ClickHouseClient } from '../clickhouse/index.js';
...
+  OPTIONAL_TO_OPTIONAL_DIFFERENT = 'OPTIONAL_TO_OPTIONAL_DIFFERENT',
+}
+
+/**
+ * Extract base types by removing:
+ * - Trailing "!" (required indicator)
+ * - Array brackets like "[Boolean!]" -> "Boolean!"
+ * - Inner "!" from arrays like "[Boolean!]" -> "Boolean"
+ */
+function extractBaseType(type: string): string {
```
StarpTech: Extract base type and normalized type; this should be possible with GraphQL.js library `parseType`, `getNullableType`, `getNamedType` etc.. The current approach isn't structured and relies heavily on text parsing for AST handling.

StarpTech (changes requested):

router/pkg/graphqlschemausage/schemausage_test.go
```diff
@@ -511,6 +565,2548 @@ func TestGetSchemaUsageInfoInterfaces(t *testing.T) {
 	}
 }
 
+// TestInputUsageWithNullVariables verifies that null variable values are tracked with IsNull flag
```
Noroth: What about interfaces and unions? Inline fragments and fragment spreads? Recursive types

router/pkg/graphqlschemausage/schemausage_test.go
```diff
@@ -4,6 +4,7 @@ import (
 	"bytes"
```
Noroth, line 4: We probably also need a test for nested arguments

```graphql
query ($fooArg: Int!, $barArg: Int!) {
  foo(fooArg: $fooArg) {
    id
    bar(barArg: $barArg) {
       id
    }
  }
}
```

controlplane/src/core/services/SchemaUsageTrafficInspector.ts
JivusAyrus: We are not parsing the schema. We parsing the message we get from the diff.

router/pkg/graphqlschemausage/schemausage_test.go
StarpTech: Tests added, ty

controlplane/src/core/services/SchemaUsageTrafficInspector.ts
StarpTech: Are we able to get access to the raw type string? If yes, we could parse it and handle it with library utilities.

JivusAyrus: Done

StarpTech (approved):
LGTM

PR #2385, opened 2025-12-03
feat(router): provide field arguments in operation context

dkorittki:
@coderabbit summary

## Checklist

--> https://github.com/wundergraph/cosmo-docs/pull/210
- [ ] Wait for [engine pull request](https://github.com/wundergraph/graphql-go-tools/pull/1373) to be merged first, then merge on this

## Motivation and Context

This pull request provides users of Cosmo Streams hooks the ability to access field arguments by their original name. Usually these get lost during the normalization of the operation, where for each argument the operation kit creates a variable:

```graphql
subscription {
  employeeUpdatedMyKafka(employeeID:1) { # maps "employeeID" to variable "a"
    id
  }
}
```

In Cosmo Streams hooks, for example the `OnReceiveEvents` hook, you would need to access this argument by it's mapped variable "a":

```go
func (m *CosmoStreamsModule) OnReceiveEvents(ctx core.StreamReceiveEventHandlerContext, events datasource.StreamEvents) (datasource.StreamEvents, error) {
	employeeID := ctx.Operation().Variables().GetInt("a")

	// do something with employeeID

	return events, nil
}
``` 

This is a problem because
- The DX is bad
- The naming scheme implies order of arguments, something which might change and could be easily forgotten to be adjusted here

One solution is to use variables in the operation:

```graphql
subscription($id:Int!) {
  employeeUpdatedMyKafka(employeeID:$id) {
    id
  }
}

# vars:
# {
#   "id": 1
# }
```

This way one could access the values of an argument by accessing the variable `id`:

```go
func (m *CosmoStreamsModule) OnReceiveEvents(ctx core.StreamReceiveEventHandlerContext, events datasource.StreamEvents) (datasource.StreamEvents, error) {
	employeeID := ctx.Operation().Variables().GetInt("id")

	// do something with employeeID

	return events, nil
}
```

But this assumes the user is in charge of its clients. This is not always the case, especially on public api's. Clients could choose all kinds of variable names a hook developer can't be aware of.

This leaves them only with the option to use the remapped variables, with all it's mentioned downsides.

## Solution

Add `Arguments()` method to `ctx.Operation()` on custom modules interfaces of Cosmo Streams hooks. `ctx.Operation()` provides data of the current operation to hook developers and since field arguments are part of that operation, I think it fits well there.

```go
type OperationContext interface {
	// [...]
	Variables() *astjson.Value
	Arguments() Arguments // this is new
}
```

Given an operation

```graphql
subscription {
  employeeUpdatedMyKafka(employeeID:1) {
    id
  }
}
```

The hook developer can now access the field directly by its path and name

```go
func (m *CosmoStreamsModule) OnReceiveEvents(ctx core.StreamReceiveEventHandlerContext, events datasource.StreamEvents) (datasource.StreamEvents, error) {
	employeeID := ctx.Operation().Arguments().Get("subscription.employeeUpdatedMyKafka.employeeID").GetInt()
	// employeeID is 1

	return events, nil
}
```

Arguments of nested fields are accessable by using a dot notated path of the field

```graphql
subscription {
  rootfield(arg1:"val1") {
    subfield1
    subfield2(arg2: "val2", arg3: "val3")
  }
}
```

```go
func (m *CosmoStreamsModule) OnReceiveEvents(ctx core.StreamReceiveEventHandlerContext, events datasource.StreamEvents) (datasource.StreamEvents, error) {
	arg2 := ctx.Operation().Arguments().Get("subscription.rootfield.subfield2.arg3").GetStringBytes()
        // [...]
}
```

`ctx.Operation().Arguments().Get` returns `*astjson.Value` objects. This way we can expose all kinds of different argument types, including GraphQL arrays and objects. It also makes using `ctx.Operation().Arguments().Get()` feel similar to `ctx.Operation().Variables()`, which also returns `*astjson.Value` objects. Have a look at the unit tests to see how it looks.

### Implementation

Now we know how a hook developer can access field arguments but we how do we actually collect and provide them?

To implement this feature I added a new field to `core.operationContext` to hold the data as a map. The map is created during variable normalization, so it's ready to use when hooks call it. This implementation involves a change on the engine, where we extend the variable extraction visitor of the variable normalizer. [Here is the pull request](https://github.com/wundergraph/graphql-go-tools/pull/1373). That must be merged before we merge this one.

So the actual creation of the field argument map is done by the engine. At first the engine wasn't involved in this feature and the router did it itself with a new walker. After discussing this I changed this to the engine for performance reasons (see comments). The router does not have an existing walker which walks field arguments at the right place, so the data is ready when hooks are called. But the engine does have such a visitor: The variable normalizer. Doing it in the engine prevents us from using a dedicated visitor just for that.

After the engine change the variable normalizers method signature changes and returns the field arg mapping. We store this map on `operationContext` and that's what the `ctx.Operation.Arguments()` (used in hooks) acesses.

### Caching

The results of the variable normalizer are cached for every operation. This is functionality that already exists. I added the field arg map to that cache as well. It's now part of the variable normalizer result and is cached with it.

The cache key generation did not change. The only two factors for a variable normalizer cache key are the operations variables and the normalized string representation. This is sufficient because if these factors do not change, the field argument mapping didn't as well.

### Other Optimisations

The field arg mapping on the engine will only run if actual Cosmo Streams hooks are registered on the router. This is done via function `needFieldArgumentMapping()`. They are the only potential places of accessing field arguments this way, so it wouldn't make sense to process them otherwise.

## Summary by CodeRabbit

* **New Features**
  * Subscription and mutation hooks can now read GraphQL field arguments from the operation context when field-argument mapping is enabled (opt-in).

* **Tests**
  * Added end-to-end and unit tests covering argument access for Start/Publish/Receive hooks and the argument-mapping utility, including rooted, nested, aliased, and variable-backed arguments.

### DIFF
--- a/router/core/arguments_test.go
+++ b/router/core/arguments_test.go
@@ -556,6 +556,20 @@ func TestNewArguments_EmptyMapping(t *testing.T) {
 	assert.Nil(t, result.Get("query.user.id"))
 }
 
+func TestArguments_Get_NonExistentPath(t *testing.T) {
+	vars, err := astjson.ParseBytes([]byte(`{"userId": "123"}`))
+	require.NoError(t, err)
+
+	mapping := astnormalization.FieldArgumentMapping{
+		"query.user.id": "userId",
+	}
+	args := NewArguments(mapping, vars)
+
+	assert.Nil(t, args.Get("query.user.nonexistent"))
+	assert.Nil(t, args.Get("mutation.createUser.id"))
+	assert.Nil(t, args.Get(""))
+}
+
 func assertFieldArgMap(t *testing.T, expected map[string]any, result Arguments) {
 	for path, expectedValue := range expected {
 		jsonValue := result.Get(path)


dkorittki:
Important questions in my head:

- Is it a good idea to create the map during operation parsing? I did it there to avoid it being created on-the-fly for hook developers, to keep accessing the map fast. But now we add some processing to the operation parsing which, depending on the complexity of the query, could take while. Haven't benchmarked this yet.
- Is the way I access root field arguments on the AST document robust? Could this break?

router/core/operation_planner.go
```diff
@@ -92,6 +93,105 @@ func (p *OperationPlanner) preparePlan(ctx *operationContext) (*planWithMetaData
...
+			args[fieldName] = m
+
+			argName := doc.ArgumentNameString(fieldArg)
+			val := doc.Arguments[fieldArg].Value
+			args[fieldName][argName] = getArgValue(doc, val, vars, remapVariables)
+		}
+	}
+
+	return args
+}
```
Noroth: The idea is not bad. I would be careful when adding this to the hot path. There are potentially big queries that will add some latency here. 
Maybe this could be a config value or can we cache that?

Also I would rather recommend that you use our `astvisitor.Walker` and register an `EnterField` visitor. With the walker you will always get the path `walker.Path`.

router/core/operation_planner.go
```diff
@@ -92,6 +93,105 @@ func (p *OperationPlanner) preparePlan(ctx *operationContext) (*planWithMetaData
...
+		fieldName := doc.FieldNameString(fieldRef)
+		fieldArgs := doc.FieldArguments(fieldRef)
+
+		for _, fieldArg := range fieldArgs {
+			m := make(map[string]any, len(fieldArgs))
+			args[fieldName] = m
+
+			argName := doc.ArgumentNameString(fieldArg)
+			val := doc.Arguments[fieldArg].Value
+			args[fieldName][argName] = getArgValue(doc, val, vars, remapVariables)
```
Noroth: map[string]map[string]any might be a bit hard to use. Maybe we could have another data structure and some receiver functions to improve the DX?

You could also have your own implementation for `seq.Iter`

router/core/operation_planner.go
dkorittki: Yeah I will see if I can come with some sort of an opt-in solution: Only if you register your custom module **and** need access to args then this will be executed. Also will switch to using a visitor 👍

router/core/operation_planner.go
dkorittki: Yeah, the map is not the best idea. Also it allows a hook developer to write to it, which is intented and could even cause problems. Will change this to something else.

router/core/operation_planner.go
dkorittki: Fixed. Now we're using a walker/visitor and it's only triggered when Cosmo Streams hooks are actually used by the user.

router/core/operation_planner.go
dkorittki: The map is now hidden behind a new type `Arguments`. A user needs to use the `Get` method to access the values.

dkorittki:
I did a bunch of changes. We are now using a visitor to register all field arguments, including those of nested fields. Its only triggered when a user registers Cosmo Streams hooks. The User does not access the map directly but instead interfaces with a new type `core.Argument`. I will update the pull request description, which shows how it will be used.

router/core/field_argument_visitor.go
```diff
@@ -0,0 +1,135 @@
...
+	argVal := v.operation.Arguments[ref].Value
+
+	resolvedArgVal, err := v.resolveArgValue(argVal)
+	if err != nil {
+		// TODO: Log error somehow
+		return
+	}
+
+	v.fieldArguments[fieldPath][argName] = resolvedArgVal
+}
```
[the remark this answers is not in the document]
dkorittki: Fixed

router/core/field_argument_visitor.go
```diff
@@ -0,0 +1,135 @@
...
+	varName := v.operation.VariableValueNameString(val.Ref)
+	originalVarName := varName
+	if v.remapVariables != nil {
+		if original, ok := v.remapVariables[varName]; ok {
+			originalVarName = original
+		}
+	}
+
+	return v.variables.Get(originalVarName), nil
+}
```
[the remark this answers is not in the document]
dkorittki: Fixed

router/core/field_argument_visitor.go
```diff
@@ -0,0 +1,139 @@
...
+	argName := v.operation.ArgumentNameString(ref)
+	argVal := v.operation.Arguments[ref].Value
+
+	resolvedArgVal, err := v.resolveArgValue(argVal)
+	if err != nil {
+		v.logger.
+			With(zap.String("fieldPath", fieldPath), zap.String("argName", argName)).
+			Warn("failed to resolve argument value", zap.Error(err))
+		return
+	}
```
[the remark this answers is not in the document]
dkorittki: Fixed

router/core/field_argument_visitor.go
```diff
@@ -0,0 +1,139 @@
...
+
+	for _, anc := range v.walker.Ancestors {
+		if anc.Kind == ast.NodeKindField {
+			fieldName := v.operation.FieldNameString(anc.Ref)
+			pathParts = append(pathParts, fieldName)
+		}
+	}
+
+	return strings.Join(pathParts, ".")
+}
```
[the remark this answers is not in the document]
dkorittki: Good catch, fixed

router/core/field_argument_visitor.go
```diff
@@ -0,0 +1,153 @@
...
+	logger         *zap.Logger
+}
+
+func mapFieldArguments(opts mapFieldArgumentsOpts) Arguments {
+	walker := astvisitor.NewWalker(48)
+
+	logger := opts.logger
+	if logger != nil {
+		logger = zap.NewNop()
+	}
```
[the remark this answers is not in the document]
dkorittki: Fixed

router/core/field_argument_visitor.go
```diff
@@ -0,0 +1,139 @@
...
+		fieldArguments: make(map[string]map[string]*astjson.Value),
+		logger:         opts.logger,
+	}
+
+	walker.RegisterEnterDocumentVisitor(visitor)
+	walker.RegisterEnterArgumentVisitor(visitor)
+
+	report := &operationreport.Report{}
+	walker.Walk(opts.operation, opts.definition, report)
+
```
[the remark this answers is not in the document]
dkorittki: Fixed

router/core/arguments.go
```diff
@@ -0,0 +1,55 @@
...
+import "github.com/wundergraph/astjson"
+
+// Arguments allow access to GraphQL field arguments used by clients.
+type Arguments struct {
+	// First key is the path to the field in dot notation
+	// i.e. root_field.subfield1.subfield2.
+	// Second argument is the name of the argument of that field.
+	// This map can be nil if no arguments are mapped.
+	data map[string]map[string]*astjson.Value
+}
```
Noroth, line 27: Technically this could be 

```go
type Arguments map[string]map[string]*astjson.Value
```

router/core/arguments.go
```diff
@@ -0,0 +1,55 @@
...
+		return nil
+	}
+
+	args, found := fa.data[f]
+	if !found {
+		return nil
+	}
+
+	return args[a]
+}
```
Noroth: Add some unit tests for this type. Check if you can provoke panics here.

router/core/field_argument_visitor.go
```diff
@@ -0,0 +1,170 @@
...
+	"go.uber.org/zap"
+)
+
+type fieldArgumentsVisitor struct {
+	walker         *astvisitor.Walker
+	operation      *ast.Document
+	definition     *ast.Document
+	variables      *astjson.Value
+	remapVariables map[string]string
+	fieldArguments map[string]map[string]*astjson.Value
```
Noroth: why not use your arguments type here and have a setter function?

router/core/graph_server.go
```diff
@@ -1495,6 +1496,7 @@ func (s *graphServer) buildGraphMux(
 			WebSocketConfiguration:    s.webSocketConfiguration,
 			ClientHeader:              s.clientHeader,
 			DisableVariablesRemapping: s.engineExecutionConfiguration.DisableVariablesRemapping,
+			MapFieldArguments:         s.subscriptionHooks.needFieldArgumentMapping(),
```
Noroth, line 1542: I think it would be good to also have a disable flag to prevent the visitor from running in general.

router/core/context.go
```diff
@@ -560,6 +564,10 @@ func (o *operationContext) Variables() *astjson.Value {
 	return o.variables
 }
 
+func (o *operationContext) Arguments() *Arguments {
+	return &o.fieldArguments
+}
```
Noroth, line 655: I'm not sure if this is a good idea. An operation can be huge and even if you only register EnterArgument the visitor will still visit every node in the document. It might be better to have some kind of lazy walking here to only request the arguments when you really need them and then store them in the field arguments instead of walking on every request.

Would that make sense?

router/core/arguments.go
dkorittki: I need to hide the map from hook developers, thats why `data` is a private field. The `Arguments` type is exposed to hook developers via `ctx.Operation().Arguments()`. When I change this it would make the map directly accessable to them. I want to hide it to not let them think they can add their own items to it.

Noroth: That makes sense. You could make the type more readable

```go

type fieldData map[string]*astjson.Value

func (f fieldData) set(...) {
...
}

type Arguments {
  data map[string]fieldData
}

func (a *Arguments) internalSet() {
  a.data.set(....)
}

```

router/core/graph_server.go
dkorittki: Yeah it probably makes sense.

Would you make this part of the router configuration or part of Custom Modules code, when you define them?

An alternative would be that I hook myself into a visitor, which already visits arguments. I could save a seperate walk here. I believe there was one visitor, which already does that but need to check if it is viable. What do you think of that?

router/core/context.go
dkorittki: The problem I have with lazy walking the document (i.e. first time a hook accesses field arguments) is that I have to remember the AST, make the walk thread safe (prevent two hooks from walking simultaneously) and remember the result globally for all hooks of that subscription query. It introduces more complex mechanisms compared to the current implementation.

The upside of that is that the ast walking is only performed when really needed, and it's not done during query normalization. But that's only an advantage if walking field arguments is actually expensive. I'll get some big queries and see how it affects performance, so we can be sure lazy walking is a good tradeoff.

dkorittki: I crunched some numbers. There is a benchmark in `router/internal/planningbenchmark/benchmark_test.go`, which contains one test `BenchmarkPlanning` which benchmarks the planning. I utilized this test as a basis and created another identical test but with field argument mapping included. This allowed me to measure the impact of this change compared to the planning step. As schema and operation input I used big production like operations (between 17KB and 37KB per operation). I then ran Go benchmarks for planning with and without field argument mapping.

sec/op was between 0.4 and 1.3ms slower with field argument mappings (1.2 to 2% slower).
B/s throughput was between 9.7 and 14.6KB lower (1.7 to 1.9% slower).
B/op allocated was between 0.3 and 0.6Mi higher (0.3 to 0.8% higher).
allocs/op was between 400 and 1400 more (0.4 to 1.4% higher).

The operations I tested did not contain many field arguments but lots of fields. This made the walker walk a lot of fields but it did not often enter my visitor to create the map for field arguments. I suppose on heavy operations with lots of field arguments we could see an increase in allocations and processing time. But in order to feel that it would involve lots of field arguments and the question is how realistic that is. I assume field arguments are only used for certain fields here and there.

So all in all the performance penalty is not too heavy I think. Given the fact that the field arguments walker only runs when Cosmo Streams Custom Modules are registered I think it's okay to let stay as is? To top it off I will also introduce an option to disable field argument mapping explicitely via config, even if these Custom Modules are registered. What do you think @Noroth ?

dkorittki:
I set this pull request back to draft. Currently I'm working on another approach which avoids a seperate walker entirely.
It will involve engine changes and once both pull requests (this and the engine one) will be ready I will reopen this one.

router/core/context.go
dkorittki: I decided to go a different approach, utilizing the engines existing walkers instead of using a new one.

router/core/operation_processor.go
```diff
@@ -1429,7 +1440,7 @@ func createParseKit(i int, options *parseKitOptions) *parseKit {
 			astnormalization.WithRemoveFragmentDefinitions(),
 			astnormalization.WithRemoveUnusedVariables(),
 		),
-		variablesNormalizer: astnormalization.NewVariablesNormalizer(),
+		variablesNormalizer: astnormalization.NewVariablesNormalizer(options.enableFieldArgumentMapping),
```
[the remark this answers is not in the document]
dkorittki: You need to parse the latest version of `github.com/wundergraph/graphql-go-tools/v2` from the routers go.mod file. There the method interface has been updated.

dkorittki:
Refactored the pull request description with better structure and more context.

dkorittki:
Awaiting a new review here after adressing feedback. Simultaneously awaiting an approval on https://github.com/wundergraph/graphql-go-tools/pull/1373 . I need to merge both in close order, first the engine and then the router. The engine PR is a breaking change as it changes a method signature used in router. This PR adresses this. So once this is approved I keep the PR open and updated until the engine PR has been approved, merged and released. Once that happened I will update go.mod here to use latest engine version.

dkorittki:
@CodeRabbit summary

router/core/arguments.go
dkorittki: fyi left the Arguments a struct with a private field for data but it is it's own type now.

```go
// Arguments allow access to GraphQL field arguments used by clients.
type Arguments struct {
	// mapping maps "fieldPath.argumentName" to "variableName".
	// For example: {"user.posts.limit": "a", "user.id": "userId"}
	mapping astnormalization.FieldArgumentMapping

	// variables contains the JSON-parsed variables from the request.
	variables *astjson.Value
}
```

From graphql-go-tools
```go
package astnormalization

// FieldArgumentMapping maps the path of field arguments inside an operation
// to the name of their mapped variables.
// Key: "rootOperationType.fieldPath.argumentName" (e.g., "query.posts.limit")
// Value: name of variable of field argument after variable normalization.
type FieldArgumentMapping map[string]string
```

router/core/field_argument_visitor.go
dkorittki: Fixed, the engine exposes that type as it does the actual mapping

router/core/arguments.go
dkorittki: Done

router/core/graph_server.go
dkorittki: As now we don't use a dedicated walker anymore (runs inside the engines variable normalizer walker) I think we can omit a new config flag. If someone wants to disable it actively we can always add it afterwards.

router/core/arguments_test.go
```diff
@@ -0,0 +1,576 @@
...
+			// Then normalize variables using VariablesNormalizer which returns the field argument mapping
+			varNorm := astnormalization.NewVariablesNormalizer(
+				astnormalization.VariablesNormalizerOptions{EnableFieldArgumentMapping: true},
+			)
+			result := varNorm.NormalizeOperation(&operation, &schema, rep)
+			require.False(t, rep.HasErrors(), "failed to normalize variables")
+
+			// Use normalized variables (includes both provided and extracted variables)
+			vars, err := astjson.ParseBytes(operation.Input.Variables)
+			require.NoError(t, err)
```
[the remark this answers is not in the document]
dkorittki: Fixed

router/core/cache_warmup.go
```diff
@@ -295,7 +323,7 @@ func (c *CacheWarmupPlanningProcessor) ProcessOperation(ctx context.Context, ope
 		return nil, err
 	}
 
-	_, _, err = k.NormalizeVariables()
+	_, _, _, err = k.NormalizeVariables()
```
Noroth, line 326: nice 🐕 🛷

Noroth:
Closing this for now as this needs some more refinement

PR #2388, opened 2025-12-03
chore: improve freeport usage

ysmolski:
The "makeSafeHttpTestServer" function creates server for testing using 
"httptest.NewUnstartedServer". That method creates listener on ephemeral port.
Before this PR, we were opening another listener using port returned by freeport,
and closing old ephemeral listener. Freeport by itself also checked if port is free
by starting listening on it and then closing it immediately after.
What often has failed is the listening on the port that was just used in another listener.

I have removed the usage of freeport in "makeSafeHttpTestServer" and started
using just ephemeral ports. "freeport" itself mentions that ephemeral ports should be
used whenever possible.

I moved freeport.GetOne closer to actual usage for the router listener. 
I vendored freeport into our codebase. 

In the future, we should not use freeport if we can use ephemeral ports.
The range of ephemeral ports contains at least 15k ports, and that should 
be enought for our needs if we propely close all the connections. Freeport
has only 1 block of 1500 port available and uses tons of resources to check
if the port were free finally.

Closes ENG-8627
Closes ENG-6922

## Summary by CodeRabbit

* **Refactor**
  * Tests now use ephemeral/on‑demand ports throughout to avoid port conflicts.
  * CDN setup API updated to return both server instance and allocated port.

* **Tests**
  * Added a retry framework for resilient test runs with configurable strategies, per-attempt logging and cleanup semantics.
  * Added a robust test port allocation utility with platform-aware ephemeral-range handling and extensive tests.

* **Chores**
  * Replaced external Consul dependency with local test utilities and added a license file.


router-tests/testenv/testenv.go
```diff
@@ -1729,9 +1701,11 @@ func SetupCDNServer(t testing.TB, port int) *httptest.Server {
...
 		require.NoError(t, err)
 		cdnFileServer.ServeHTTP(w, r)
-	}), port)
-
-	return cdnServer
+	})
+	cdnServer = httptest.NewUnstartedServer(handler)
+	cdnServer.Start()
+	port = cdnServer.Listener.Addr().(*net.TCPAddr).Port
+	return
```
Noroth: nitpick and up to you

I'm personally not the biggest fan of `naked` returns. I know they exist and it's fine to use them for short functions.
I do believe having named return values is a nice feature, especially in this case it improves the readability.

My recommendation for named returns is always to provide the values in the return.
```go
return cdnServer, port 
```
Because in this case you don't need to look at the function declaration again to see what you are actually returning.

router-tests/testenv/waitinglistener.go
```diff
@@ -41,7 +43,7 @@ func (l *WaitingListener) Port() int {
 func NewWaitingListener(t *testing.T, waitTime time.Duration) (wl *WaitingListener) {
 	ctx, cancel := context.WithCancel(context.Background())
 	var lc net.ListenConfig
-	listener, err := lc.Listen(ctx, "tcp", "127.0.0.1:0")
+	listener, err := lc.Listen(ctx, "tcp", fmt.Sprintf("127.0.0.1:%d", freeport.GetOne(t)))
```
Noroth: As you mentioned you would rather recommend using ephemeral ports, why did you change this here?

router-tests/testenv/testenv.go
ysmolski: Oh yeah, from the point, it makes sense. Verbosity enhances clarity.

router-tests/testenv/waitinglistener.go
ysmolski: Forgot to revert. This PR had many experiments to debug the issue. I was thinking initially that ephemeral ports is the problem and they can somehow clash with freeports.

router-tests/testenv/testenv.go
```diff
@@ -658,6 +652,7 @@ func CreateTestSupervisorEnv(t testing.TB, cfg *Config) (*Environment, error) {
 		cfg.MCP.Server.ListenAddr = fmt.Sprintf("localhost:%d", freeport.GetOne(t))
 	}
 
+	listenerAddr := fmt.Sprintf("localhost:%d", freeport.GetOne(t))
 	baseURL := fmt.Sprintf("http://%s", listenerAddr)
```
Noroth, line 656: This is a tricky part. Technically it would be great if we could use ephemeral ports here as well but we would need to have a trigger to get the new router port for a newly started supervisor instance and update the testenv.

ysmolski: My opinion is that it is okay to use freeport sparingly. Since we spawn only one router per env (unlike 11 demo servers) and the concurrency of tests is limited to GOMAXPROC, we should be fine. If not fine, I could devote separate PR later (after NY, in a bug squashing stint).

PR #2400, opened 2025-12-09
chore: improve watcher tests

Noroth:
Tests which are using timers can be flaky in general. `time.Sleep` is not optimal to advance time properly. The amount of advanced time varies based on the OS differences and system load. You might expect to wait 10 milliseconds while you might have actually waited 11 or 12 milliseconds.

To provide some improvement, this PR changes the behavior of the watcher tests to be scheduled in their own bubble. with the `synctest` package time.Sleep will advance the time on a fake timer inside of the `syntest.Test()` callback function.

Example:
```go
func TestWithSync(t *testing.T) {
	now := time.Now()
	t.Parallel()
	synctest.Test(t, func(t *testing.T) {
		fmt.Println("Timestamp: ", time.Now().UTC())

		time.Sleep(time.Second * 5)
		fmt.Println("Timestamp: ", time.Now().UTC())
	})

	testMicroseconds := time.Since(now).Microseconds()
	fmt.Println("Test took", testMicroseconds, "microseconds to run")
}
```

The actual test run took only 59 microseconds. 
```
Timestamp:  2000-01-01 00:00:00 +0000 UTC
Timestamp:  2000-01-01 00:00:05 +0000 UTC
Test took 59 microseconds to run
```

## Summary by CodeRabbit

* **Tests**
  * Improved test synchronization and determinism for file-watching scenarios, adding time-synced test harness and synthetic file metadata to reliably cover creates, modifications, renames/moves, deletions, symlinks (including chained cases) and multi-file setups.

* **Chores**
  * Added a configurable file-metadata provider to watcher options to support deterministic timestamps and custom integrations for testing and runtime.

## Checklist

### DIFF
--- a/router/pkg/watcher/watcher_test.go
+++ b/router/pkg/watcher/watcher_test.go
@@ -104,11 +104,11 @@ func TestWatch(t *testing.T) {
 			ctx, cancel := context.WithCancel(context.Background())
 			defer cancel()
 
-			fakeFS := newTimeSyncFS()
+			timeSyncFS := newTimeSyncFS()
 
 			dir := t.TempDir()
 			tempFile := filepath.Join(dir, "config.json")
-			require.NoError(t, fakeFS.WriteFile(tempFile, []byte("a"), 0o600))
+			require.NoError(t, timeSyncFS.WriteFile(tempFile, []byte("a"), 0o600))
 
 			spy := test.NewCallSpy()
 			tickerChan := make(chan time.Time)
@@ -119,7 +119,7 @@ func TestWatch(t *testing.T) {
 				Callback:   spy.Call,
 				TickSource: tickerChan,
 				FileInfoProvider: func(path string) (os.FileInfo, error) {
-					return fakeFS.GetFileInfo(path)
+					return timeSyncFS.GetFileInfo(path)
 				},
 			})
 			require.NoError(t, err)
@@ -134,11 +134,11 @@ func TestWatch(t *testing.T) {
 
 			tempFile2 := filepath.Join(dir, "config2.json")
 
-			require.NoError(t, fakeFS.WriteFile(tempFile2, []byte("b"), 0o600))
+			require.NoError(t, timeSyncFS.WriteFile(tempFile2, []byte("b"), 0o600))
 			sendSyncTick(tickerChan)
 
 			// Move new file ontop of the old file
-			require.NoError(t, fakeFS.Rename(tempFile2, tempFile))
+			require.NoError(t, timeSyncFS.Rename(tempFile2, tempFile))
 			sendSyncTick(tickerChan)
 			spy.AssertCalled(t, 0)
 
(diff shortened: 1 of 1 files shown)


router/pkg/watcher/watcher_test.go
```diff
@@ -806,9 +898,149 @@ func TestCancel(t *testing.T) {
...
+// (which respects synctest's fake time). Then we return our tracked modTime instead of whatever
+// the filesystem reports.
+type fakeFileInfo struct {
+	os.FileInfo
+	modTime time.Time
+}
+
+type fakeFS struct {
+	fileInfos map[string]*fakeFileInfo
+}
```
endigma: maybe there's a better name than `fakeFS`? It's not really _fake_, just hacked to be more precise?

router/pkg/watcher/watcher_test.go
```diff
@@ -806,9 +898,149 @@ func TestCancel(t *testing.T) {
...
+
+func (f *fakeFS) Mkdir(name string, perm os.FileMode) error {
+	err := os.Mkdir(name, perm)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
```
endigma: These could just return the inner function call directly

Noroth:
Rule of thumb: If it looks simple it's most likely gonna be a rabbit hole

router/pkg/watcher/watcher_test.go
Noroth: Ah yes, I initially had some properties which I set in the function and now it's basically just wrapping. Function basically just exists to provide all functions that we are using in the test functions

router/pkg/watcher/watcher_test.go
```diff
@@ -910,13 +910,13 @@ func sendSyncTick(channel chan time.Time) {
 // newFakeFileInfo is a helper function for tests that creates a new fake file info object
 // newFakeFileInfo creates a fakeFileInfo with modTime initialized to time.Now().
 // In synctest, time.Now() returns the fake time controlled by the test.
-func newFakeFileInfo() *fakeFileInfo {
-	return &fakeFileInfo{
+func newFakeFileInfo() *timeSyncFileInfo {
```
endigma: old name still

Noroth: oops

router/pkg/watcher/watcher_test.go
```diff
@@ -806,9 +898,140 @@ func TestCancel(t *testing.T) {
...
+// If timestamps don't advance (filesystem resolution issue) or don't match the test's time control
+// (synctest mismatch), this comparison breaks and the watcher won't detect changes properly.
+//
+// This wrapper fixes it by intercepting file operations and manually bumping modTime using time.Now()
+// (which respects synctest's fake time). Then we return our tracked modTime instead of whatever
+// the filesystem reports.
+type timeSyncFileInfo struct {
+	os.FileInfo
+	modTime time.Time
+}
```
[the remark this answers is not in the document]
Noroth: No that's fine. It's only test related and we don't want the actual file info.

ysmolski:
Good job. One less flaky tezt!

PR #2416, opened 2025-12-16
feat: mask hostname in cli telemetry

pepol:
## Summary by CodeRabbit

* **Chores**
  * Improved CLI telemetry to derive a deterministic, hashed machine identifier (replacing raw hostnames) for more consistent reporting and stronger user anonymity. This change does not alter which events are captured, only how the machine identifier is represented.

## Checklist

### DIFF
--- a/.github/workflows/protographic.yaml
+++ b/.github/workflows/protographic.yaml
@@ -41,7 +41,7 @@ jobs:
       - name: Lint
         run: pnpm run --filter ./protographic lint
 
-      - name: Upload integration results to Codecov
+      - name: Upload test results to Codecov
         uses: ./.github/actions/codecov-upload-pr
         with:
           artifact-name: protographic-tests-coverage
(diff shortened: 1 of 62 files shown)


cli/src/core/telemetry.ts
```diff
@@ -149,13 +150,14 @@ export const captureCommandFailure = async (command: string, error: Error | stri
  * Get CLI metadata to include with all events
  */
 const getMetadata = (): Record<string, any> => {
+  const machineId = crypto.createHash('sha256').update(os.hostname()).digest('hex');
```
StarpTech: How much does it add to startup latency?

pepol: Modified to use non-object approach (`crypto.hash`), which should be faster as per https://github.com/nodejs/performance/issues/136 and https://github.com/nodejs/node/pull/51044.

Either way, based on the benchmark posted for the original object-based solution (https://github.com/nodejs/performance/issues/136#issue-2010764378), the latency overhead is negligible (100k sha1 hashes is ~100ms - with the 41% improvement using the functional approach, it should be about 60ms and that's for 100k hashes, so one hash should be significantly lower, even including the set-up overhead).

PR #2417, opened 2025-12-16
fix: limit the number of entries shown in the table for schema checks on the cli

SkArchon:
## This PR does the following
Introduces a limit parameter for schema checks to pass in a limit, we still do use a default limit if a value was not passed in, we also set a max number.

When determining if the limit was exceeded, we treat the following values as one grouping
```
lintWarnings + lintErrors
breakingChanges + nonBreakingChanges
graphPruneErrors + graphPruneWarnings
```
We however DO NOT treat `compositionErrors` and `compositionWarnings` as one entry because they are displayed in two tables on the CLI at the moment.

However when returning the counts, we return values for every individual attribute, which can be found in the `SchemaCheckCounts` definition. This includes for example `lintWarnings` and `lintErrors` being separate attributes, as they are part of the API, which would be harder to change later if we add `lintWarningsAndErrors` and want to break it into two values.

The CLI also has a new option for changing the limit with a default value, if there are any truncated entries we mention to the user in the CLI output and ask the user to refer the studio dashboard.

Note: This PR does not introduce pagination to studio.

## Summary by CodeRabbit

* **New Features**
  * Added a --limit (-l) option to schema/federated-graph check commands (default 50) with input validation and truncation warnings.
  * Check responses now include summarized counts for linting, breaking/non‑breaking changes, composition and prune findings.
  * When results are truncated, users see a “more entries available” notice pointing to the dashboard for full details.

* **Tests**
  * Added tests covering limit behavior, clamping, and counts reporting.

* **Documentation**
  * Added guidance to run make generate for proto/type generation.

## Checklist

### DIFF
--- a/controlplane/src/core/bufservices/federated-graph/checkFederatedGraph.ts
+++ b/controlplane/src/core/bufservices/federated-graph/checkFederatedGraph.ts
@@ -24,8 +24,7 @@ import {
   newCompositionOptions,
 } from '../../util.js';
 import { UnauthorizedError } from '../../errors/errors.js';
-
-const maxRowLimit = 100_000;
+import { maxRowLimitForChecks } from '../../constants.js';
 
 export function checkFederatedGraph(
   opts: RouterOptions,
@@ -114,7 +113,7 @@ export function checkFederatedGraph(
     );
 
     // If req.limit is not provided, we return all rows
-    const returnLimit = req.limit === undefined ? null : clamp(req.limit, 1, maxRowLimit);
+    const returnLimit = req.limit === undefined ? null : clamp(req.limit, 1, maxRowLimitForChecks);
 
     const counts = {
       compositionErrors: 0,
--- a/controlplane/src/core/bufservices/subgraph/checkSubgraphSchema.ts
+++ b/controlplane/src/core/bufservices/subgraph/checkSubgraphSchema.ts
@@ -27,8 +27,7 @@ import {
   limitCombinedArrays,
 } from '../../util.js';
 import { OrganizationWebhookService } from '../../webhooks/OrganizationWebhookService.js';
-
-const maxRowLimit = 100_000;
+import { maxRowLimitForChecks } from '../../constants.js';
 
 export function checkSubgraphSchema(
   opts: RouterOptions,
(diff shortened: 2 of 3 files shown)


controlplane/src/core/bufservices/subgraph/checkSubgraphSchema.ts
```diff
@@ -251,24 +255,9 @@ export function checkSubgraphSchema(
...
-      hasClientTraffic,
-      checkedFederatedGraphs,
-      lintWarnings,
-      lintErrors,
-      graphPruneWarnings,
-      graphPruneErrors,
-      isCheckExtensionSkipped,
-      checkExtensionErrorMessage,
-    } = await subgraphRepo.performSchemaCheck({
+    const returnLimit = clamp(req.limit ?? defaultRowLimit, 1, maxRowLimit);
```
JivusAyrus: i feel if the limit is not passed, return everything

cli/src/handle-check-result.ts
```diff
@@ -249,21 +249,44 @@ export const handleCheckResult = (resp: CheckSubgraphSchemaResponse) => {
...
+          resp.counts.graphPruneErrors + resp.counts.graphPruneWarnings > rowLimit ||
+          resp.counts.compositionErrors > rowLimit ||
+          resp.counts.compositionWarnings > rowLimit;
+
+        if (hasExceeded) {
+          if (studioCheckDestination !== '') {
+            moreEntriesAvailableMessage += `\n\n`;
+          }
+          moreEntriesAvailableMessage += pc.red(
+            `Some results were truncated due to exceeding the limit of ${rowLimit} rows.`,
```
JivusAyrus, line 266: do you think it would be better if we mention what exceeded the limit?

controlplane/src/core/bufservices/federated-graph/checkFederatedGraph.ts
```diff
@@ -24,6 +25,8 @@ import {
 } from '../../util.js';
 import { UnauthorizedError } from '../../errors/errors.js';
 
+const maxRowLimit = 100_000;
```
JivusAyrus: Nitpick: Should we put this somewhere so both files can use the same constant

cli/src/handle-check-result.ts
SkArchon: I was thinking its fine, since we get them to go to the studio url to view all the results anyway, and this would keep the logic simple.

controlplane/src/core/bufservices/federated-graph/checkFederatedGraph.ts
SkArchon: Updated

PR #2420, opened 2025-12-17
fix: make error accessible to custom modules

SkArchon:
## This PR does the following listed

* Makes error accessible to custom modules, while Error() will return nil since they run too early in our hooks before `SetError()` is called in our router, users can now wrap a custom response body which will be invoked much later after `SetError()` is called in the router

* We remove the `defer propagateSubgraphErrors`, which means that the new behaviour is that in case of an early return due to an error, the final error set to to errors is not guaranteed to be the ones returned from the subgraph (previously this was guaranteed).

## Summary by CodeRabbit

* **New Features**
  * Adds automatic X-Has-Error response header and exposes an error accessor on the request context.

* **Bug Fixes**
  * Ensures upstream/subgraph errors are tracked and surfaced in request handling so they influence response headers and final reporting.

* **Tests**
  * Adds end-to-end tests covering auth failures, upstream errors, successful paths, and header behavior.

## Checklist

### DIFF
--- a/router/core/graphql_handler.go
+++ b/router/core/graphql_handler.go
@@ -179,7 +179,6 @@ func (h *GraphQLHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 		resp, err := h.executor.Resolver.ResolveGraphQLResponse(resolveCtx, p.Response, nil, &respBuf)
 		reqCtx.dataSourceNames = getSubgraphNames(p.Response.DataSources)
-
 		if err != nil {
 			trackFinalResponseError(resolveCtx.Context(), err)
 			h.WriteError(resolveCtx, err, p.Response, w)


router/core/context.go
```diff
@@ -139,6 +139,8 @@ type RequestContext interface {
 	// SetForceSha256Compute forces the computation of the Sha256Hash of the operation
 	// This is useful if the Sha256Hash is needed in custom modules but not used anywhere else
 	SetForceSha256Compute()
+
+	Error() error
```
Noroth, line 144: missing docstring

router/core/context.go
```diff
@@ -473,6 +475,10 @@ func (c *requestContext) SetForceSha256Compute() {
 	c.forceSha256Compute = true
 }
 
+func (c *requestContext) Error() error {
```
Noroth, line 480: missing docstring

router/core/graphql_handler.go
```diff
@@ -190,6 +190,9 @@ func (h *GraphQLHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 		if errs := resolveCtx.SubgraphErrors(); errs != nil {
 			w.Header().Set("Cache-Control", "no-store, no-cache, must-revalidate")
+			// This is recorded in the defer above, but this is used to make sure the error is accessible to
+			// any response writer wrappers set in custom modules that need it when WriteTo is called
+			reqCtx.SetError(errs)
```
StarpTech: I see other paths e.g. `WriteError` where the error would not be set. I think we should call `trackFinalResponseError` in the other places and remove the defer

SkArchon: I applied the change. With the new change, the only difference is that previously if there were subgraph errors they were guaranteed to be the final error set to ctx.error, since it is not deferred anymore, this is not guaranteed.

Also for WriteError we already called `trackFinalResponseError` before WriteError in the existing code.

StarpTech (approved):
LGTM

PR #2436, opened 2025-12-31
feat(cli,protographic): improvements to operation validation and proto generation

asoorm:
This PR is split from #2379 to enable independent review and merging of improvements to proto generation and operation validation. Required before implementing the larger features in #2379.

Ultimately, this PR permits us to have non-standard graphQL variable names / graphql variable names which cannot be derived from proto field names making it simpler to onboard non-standard GraphQL servers.

### Protographic
- Support GraphQL variable name field options where the `graphql_variable_name` doesn't match protojson equivalent of proto message field.

```
message FindEmployeesByCriteriaRequest {
  bool has_pets = 1 [(graphql_variable_name) = "HAS_PETS"];
  Nationality nationality = 2 [(graphql_variable_name) = "Nationality"];
}
```

- Track nested messages independently in lock file

Consider:

```
message FindEmployeesByCriteriaResponse {
  message FindEmployees {
    message Details {
      string forename = 1;
      string first_name = 2;
      string last_name = 3;
    }
    int32 id = 1;
    Details details = 2;
  }
  //  This is a GraphQL query that retrieves a list of employees.
  repeated FindEmployees find_employees = 1;
}
```

**Before**: If multiple `Details` messages existed in different contexts, they shared 
the same lock file entry. This caused field numbering conflicts when fields were added 
or removed in one context but not others:

```
    message Details {
      string forename = 1;
      string first_name = 2;
      string last_name = 4; // ⚠️ Skipped field 3 due to conflict with another Details message
    }
```

After: The lock file now stores the full hierarchical path
(FindEmployeesByCriteriaResponse.FindEmployees.Details), ensuring each Details
message maintains independent field numbering.

### CLI

- Removed the `prefixOperationType` CLI option (breaking change, but OK, because nobody using yet) - it's not actually necessary, because we don't need to derive the operation type from the proto, we only need to execute the trusted documents / Named Operations.

## Context
Split from #2379 to simplify review and get these changes merged independently.

## Summary by CodeRabbit

* **Bug Fixes**
  * Reliable protobuf field-numbering for deeply nested messages via hierarchical path tracking and depth guards.

* **New Features**
  * Preserve original GraphQL variable names on protobuf fields using a custom proto field option; proto headers include the extension only when used.
  * Stricter operation-name validation (must start uppercase; clarified allowed formats).

* **Documentation**
  * Guidance updated to emphasize deterministic proto schema generation and naming constraints.

* **Chores**
  * Removed the CLI option for prefixing operation types.

* **Tests**
  * Added/updated tests for nested field numbering and operation-name validation.

### DIFF
--- a/protographic/src/operations/proto-field-options.ts
+++ b/protographic/src/operations/proto-field-options.ts
@@ -1,7 +1,7 @@
 /**
  * Protocol Buffer field option constants for ConnectRPC integration.
  *
- * See proto/com/wundergraph/connectrpc/options/v1/annotations.proto for full documentation.
+ * These options are defined using `extend google.protobuf.FieldOptions` in generated proto files.
  */
 
 export interface ProtoFieldOption {
--- a/cli/src/commands/grpc-service/commands/generate.ts
+++ b/cli/src/commands/grpc-service/commands/generate.ts
@@ -344,8 +344,8 @@ async function generateFromOperations(
   // Load lock data for field number stability
   let currentLockData = await fetchLockData(lockFile);
 
-  // Process each operation file separately to maintain proto schema consistency
-  // Collect the AST roots instead of proto strings
+  // Process each operation file separately, updating lock data sequentially for field number stability.
+  // Collect AST roots for merging rather than proto strings.
   const roots: protobuf.Root[] = [];
 
   for (const { filename, content } of operationFiles) {


protographic/src/operations/proto-field-options.ts
```diff
@@ -0,0 +1,19 @@
+/**
```
Noroth, line 1: Is this actually something we could put into the global scope of protographic? I feel like options in general could be something that we could consider using more in the future.

StarpTech (approved):
LGTM

PR #2439, opened 2026-01-05
feat: add support for requires in protographic

Noroth:
> [!NOTE]
> This PR has a lot of files changed because the common linting rules were included, which has been missed before.
> A lot of files are simply re-formatted.

This PR introduces support for the `@requires` directive in protographic. Entities that define required and external fields will now generate rpc definitions for lookups.

The implementation is described in the `SDL_TO_PROTO` markdown file. 

A couple new visitors were introduced for handling the selection set within the `@requires` directive.
* The `required-fields-visitor` uses the selection set from the directive to define the structure for the rpcs and proto messages.
* The `abstract-selection-rewriter` is used to normalize the selection set. In general we can imagine that each selection set inside of a requires directive reflects a Graphql operation. Therefore we can also include interfaces and Unions to define the exact fields that we want to use. As interfaces and unions can be nested we need to normalize those to provide all allowed types to generate the proper proto messages.
> [!NOTE]
> Using Inline fragments is currently not yet allowed due to a validation rule. This functionality will be provided in a separate implementation for the engine to reduce the scope of the PRs. The protographic is independent therefore I decided to include the functionality in this PR already.
* The `selection-set-validation-visitor` allows us to apply validation rules on an arbitrary selection set.

## Summary by CodeRabbit

* **New Features**
  * Per-field "required fields" for federated entities with generated RPCs, proto messages, and mappings; selection-set normalization for abstract/interface types; public API to produce deterministic proto output and expose required-field visitors.

* **Improvements**
  * Modular proto-generation utilities, expanded naming helpers and constants, safer validation/traversal, deterministic ordering and wrapper handling for lists.

* **Validation**
  * New rule disallowing abstract types inside @requires selections.

* **Tests**
  * Large expansion of tests covering field sets, federation, proto output, naming, validation, and utilities.

* **Chores**
  * Lint/config, import path and CI workflow updates.

## Checklist

### DIFF
--- a/protographic/src/abstract-selection-rewriter.ts
+++ b/protographic/src/abstract-selection-rewriter.ts
@@ -45,6 +45,13 @@ type GraphQLTypeWithFields = GraphQLObjectType | GraphQLInterfaceType;
  * interface level and within inline fragments for concrete types. This class
  * normalizes such selections by moving interface-level fields into each inline
  * fragment, ensuring consistent selection structure for downstream processing.
+ * This is required to generate the proper proto message definitions as in protobuf
+ * we define abstract types as oneof fields in in a message definition, which defines
+ * the possible concrete types that can be selected.
+ *
+ * As we can select on Interfaces and Unions directly, we need to normalize the selection set
+ * to determine the allowed concrete types that can be selected.
+ *
  *
  * @example
  * Input selection:
@@ -180,7 +187,7 @@ export class AbstractSelectionRewriter {
     this.currentCompositeRoot = fieldType;
     this.compositeTypeStack.push(this.currentCompositeRoot);
 
-    // Only process selection sets for interface types
+    // Only process selection sets for interface and union types
     if (isAbstractType(fieldType)) {
       this.appendValidInlineFragments(ctx.node);
       this.distributeFieldsIntoInlineFragments(ctx.node);
--- a/protographic/src/sdl-to-proto-visitor.ts
+++ b/protographic/src/sdl-to-proto-visitor.ts
@@ -1162,10 +1162,6 @@ Example:
         field.astNode?.directives?.some((d) => d.name.value === REQUIRES_DIRECTIVE_NAME),
       );
 
-      if (requiredFields.length === 0) {
-        continue;
-      }
-
       for (const requiredField of requiredFields) {
         const fieldSet = this.getRequiredFieldSet(requiredField);
         const visitor = new RequiredFieldsVisitor(this.schema, entity, requiredField, fieldSet);
--- a/protographic/tests/abstract-selection-rewriter/01-basics.test.ts
+++ b/protographic/tests/abstract-selection-rewriter/01-basics.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, it } from 'vitest';
-import { buildSchema, parse, print, GraphQLObjectType } from 'graphql';
+import { buildSchema, parse, print, GraphQLObjectType, GraphQLSchema } from 'graphql';
 import { AbstractSelectionRewriter } from '../../src/abstract-selection-rewriter.js';
 
 const schema = buildSchema(
@@ -38,22 +38,10 @@ const schema = buildSchema(
 );
 
 
-/**
- * Helper function to normalize a field set selection
- */
-function normalizeFieldSet(fieldSet: string, typeName: string): string {
+function normalizeFieldSet(fieldSet: string, typeName: string, s: GraphQLSchema = schema): string {
   const doc = parse(`{ ${fieldSet} }`);
-  const objectType = schema.getTypeMap()[typeName] as GraphQLObjectType;
-  const rewriter = new AbstractSelectionRewriter(doc, schema, objectType);
-  rewriter.normalize();
-  return print(doc);
-}
-
-
-function normalizeFieldSetWithComplexInterfaces(fieldSet: string, typeName: string): string {
-  const doc = parse(`{ ${fieldSet} }`);
-  const objectType = schemaWithComplexInterfaces.getTypeMap()[typeName] as GraphQLObjectType;
-  const rewriter = new AbstractSelectionRewriter(doc, schemaWithComplexInterfaces, objectType);
+  const objectType = s.getTypeMap()[typeName] as GraphQLObjectType;
+  const rewriter = new AbstractSelectionRewriter(doc, s, objectType);
   rewriter.normalize();
   return print(doc);
 }
@@ -451,7 +439,7 @@ describe('AbstractSelectionRewriter', () => {
     }
     `;
 
-    const result = normalizeFieldSetWithComplexInterfaces(input, 'Query');
+    const result = normalizeFieldSet(input, 'Query', schemaWithComplexInterfaces);
     expect(result).toMatchInlineSnapshot(`
       "{
         iface {
@@ -480,7 +468,7 @@ describe('AbstractSelectionRewriter', () => {
     }
     `;
 
-    const result = normalizeFieldSetWithComplexInterfaces(input, 'Query');
+    const result = normalizeFieldSet(input, 'Query', schemaWithComplexInterfaces);
     expect(result).toMatchInlineSnapshot(`
       "{
         iface {


protographic/src/selection-set-validation-visitor.ts
```diff
@@ -0,0 +1,240 @@
...
+    }
+
+    const namedType = this.getUnderlyingType(fieldDefinition.type);
+
+    if (this.isAbstractType(namedType)) {
+      this.validationResult.errors.push(
+        `Abstract types are not allowed in requires directives. Found ${namedType.name} in ${this.currentType.name}.${ctx.node.name.value}`,
+      );
+      return BREAK;
+    }
```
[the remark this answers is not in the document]
Noroth: I know, this is temporary. The implementation for the abstract types is quite complex and I decided to implement this in multiple iterations.

protographic/src/sdl-to-proto-visitor.ts
```diff
@@ -1492,13 +1552,23 @@ Example:
...
+    const allFields = Object.values(type.getFields());
+    // Filter out fields that have arguments as those are handled in separate resolver rpcs
+    const validFields = allFields
+      .filter((field) => field.args.length === 0)
+      .filter(
+        (field) =>
+          !field.astNode?.directives?.some(
+            (directive) =>
+              directive.name.value === EXTERNAL_DIRECTIVE_NAME || directive.name.value === REQUIRES_DIRECTIVE_NAME,
+          ),
```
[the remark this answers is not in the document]
Noroth: `@requires` fields are resolved by a different RPC. Therefore it doesn't make sense to include the field in the entity message. RPCs that resolve the entity would otherwise allow to provide a required field without having information about the `@external` fields, which is incorrect. Therefore we need to have RPCs that have `Request` messages that include the `@external` fields.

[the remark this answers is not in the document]
Noroth: no I'm talking to you @coderabbitai. I explained the reason why we filter out required fields.

protographic/src/types.ts
```diff
@@ -1,5 +1,42 @@
...
+  Boolean: 'bool', // Direct mapping
+};
+
+/**
+ * Maps GraphQL scalar types to Protocol Buffer wrapper types for nullable fields
+ *
+ * These wrapper types allow distinguishing between unset fields and zero values
+ * in Protocol Buffers, which is important for GraphQL nullable semantics.
+ */
+export const SCALAR_WRAPPER_TYPE_MAP: Record<string, string> = {
```
asoorm, line 32: `SCALAR_TYPE_MAP` and `SCALAR_WRAPPER_TYPE_MAP` already exist in `operations/type-mapper.ts:20-37` - can we dedupe? Happy for this file to be the source of truth.

protographic/src/proto-utils.ts
```diff
@@ -0,0 +1,389 @@
...
+
+  if (message.reservedNumbers && message.reservedNumbers.length > 0) {
+    messageLines.push(indentContent(indent + 1, `reserved ${message.reservedNumbers};`));
+  }
+
+  const compositeTypeFields = message.fields.filter((f) => f.compositeType);
+  for (const compositeTypeField of compositeTypeFields) {
+    messageLines.push(
+      ...buildCompositeTypeMessage(includeComments, compositeTypeField.compositeType!, indent + 1, seenTypes),
+    );
```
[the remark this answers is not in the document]
Noroth: I think that's fine. It's only creating the messages with oneOf. Those shouldn't be created multiple times on one level of recursion. So they shouldn't be passed to the recursive calls.

protographic/src/abstract-selection-rewriter.ts
```diff
@@ -0,0 +1,492 @@
...
+  private mergeInlineFragments(node: SelectionSetNode): void {
+    const selectedInlineFragments = node.selections.filter((s) => s.kind === Kind.INLINE_FRAGMENT);
+    if (selectedInlineFragments.length === 0) {
+      return;
+    }
+
+    const uniqueInlineFragments: InlineFragmentNode[] = [];
+
+    for (const selectedFragment of selectedInlineFragments) {
+      const unqiueFragment = uniqueInlineFragments.find(
```
asoorm: nit: variable name typo

protographic/src/types.ts
Noroth: Removed them

protographic/src/abstract-selection-rewriter.ts
Noroth: fixed

protographic/tests/abstract-selection-rewriter/01-basics.test.ts
```diff
@@ -0,0 +1,427 @@
...
+        level2: [Level2Interface]
+        dataA: String
+      }
+
+      type Level1TypeB implements Level1Interface {
+        id: ID!
+        level2: [Level2Interface]
+        dataB: String
+      }
+
```
devsergiy, line 280: It could be like this

```
      type Level1TypeC implements Level1Interface {
        id: ID!
        level2: [Level2TypeX]
        dataB: String
      }
```

e.g., level2 has concrete type, not the interface type itself
e.g,. If the field type in the interface is of interface or union type, a concrete type that is a member of this type is a valid field type

protographic/tests/abstract-selection-rewriter/02-advanced.test.ts
```diff
@@ -0,0 +1,862 @@
...
+              }
+            }
+          }
+        }"
+      `);
+    });
+
+    it('should handle two nested interface selections (Managed + Permission)', () => {
+      const input = `
+        departments {
```
devsergiy, line 358: Please try combinations like this

It is valid to have an interface fragment even if the other interface does not implement the interface; the only rule is that there should be an intersection in the implementing types

Also, union fragments are possible inside interface fragments, and vice versa, in case there is an intersection between union members and implementing types

Under object type fragment, it is possible to have fragments of any type, like object, interface, union, if the object type is part of implementing or union types, and of course, the only valid type will be that object type

```
        departments {
          members {
            id
            ... on Managed {
              supervisor
              ... on Permission {
                scope
              }
            }
            ... on Permission {
              scope
              ... on Managed {
                supervisor
                ... on Employee {
                  name
                }
              }
            }
          }
        }
```

protographic/src/abstract-selection-rewriter.ts
```diff
@@ -0,0 +1,562 @@
...
+      return false;
+    }
+
+    if (!isObjectType(type)) {
+      // Non-object types cannot implement interfaces
+      return false;
+    }
+
+    return type.getInterfaces().some((i) => i.name === this.currentType.name);
+  }
```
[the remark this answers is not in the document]
Noroth: You are partly right. You can have a nested selection on interfaces on a field. However the source of truth is the base interface.

If you have 2 interfaces `Iface` and `Iface2` and two types, one that implements both and one that only implements `Iface` and not `Iface2`, nothing stops you from having a query where you select `... on Iface2 { }`. If your query expects and Iface here, you will get all the types that implement `Iface & Iface2` but other types have to be removed by the root definition of the field. Therefore valid types need to be an intersection of the root element.

To determine the valid types based on the current inline fragment, we have another function `getPossibleIntersectingTypes` which will be invoked when we append the valid types to the current document.

protographic/src/sdl-validation-visitor.ts
```diff
@@ -127,10 +133,11 @@ export class SDLValidationVisitor {
...
 
+    // Requires directive support will currently be added. This rule will be removed in the future.
     const requiresRule: LintingRule<Kind.FIELD_DEFINITION> = {
       name: 'use-of-requires',
       description: 'Validates usage of @requires directive which is not yet supported',
-      enabled: true,
+      enabled: false,
       nodeKind: Kind.FIELD_DEFINITION,
       validationFunction: (ctx) => this.validateRequiresDirective(ctx),
     };
```
endigma: Should this be removed immediately?

protographic/src/sdl-to-mapping-visitor.ts
```diff
@@ -115,25 +119,93 @@ export class GraphQLToProtoVisitor {
...
+          continue;
+        }
 
         // Process each @key directive separately
         for (const keyDirective of keyDirectives) {
           const key = this.getKeyFromDirective(keyDirective);
           if (key) {
             // Create entity mapping for each key combination
             this.createEntityMapping(typeName, key);
+            // todo: add required fields mapping
```
endigma: TODO?

protographic/tests/sdl-validation/01-basic-validation.test.ts
```diff
@@ -475,4 +483,106 @@ describe('SDL Validation', () => {
       '[Error] Cycle detected in context: field "baz" is referenced in the following path: "baz.bar.foo"',
     );
   });
+
+  test('should return an error if an abstract type is used in a requires directive', () => {
```
devsergiy, line 487: Don't you already handle that? Is this still an error?

protographic/tests/sdl-to-mapping/03-federation.test.ts
```diff
@@ -986,4 +986,205 @@ describe('GraphQL Federation to Proto Mapping', () => {
       }
     `);
   });
+
+  it('maps entity with required and external fields', () => {
+    const sdl = `
+      directive @key(fields: String!) on OBJECT
+      
+      type Product @key(fields: "id") {
```
devsergiy, line 994: Do we support multiple keys? Should we have a test case for 2 keys?

protographic/tests/proto-utils/build-proto-message.test.ts
```diff
@@ -0,0 +1,802 @@
...
+            }
+          }
+          SearchResult primary_result = 1;
+          SearchResult secondary_result = 2;
+        }
+        "
+      `);
+    });
+
+    it('should count exactly one message definition when multiple fields reference same composite type', () => {
```
devsergiy, line 475: will/should it deduplicate if usages will be at a different nesting level?
e.g.

```
  a {
    u: Union
    other: {
      another: {
        u: Union
      }
    }
  }
```

or one of union members has a field of the same union type it is member of

protographic/tests/field-set/01-basics.test.ts
```diff
@@ -0,0 +1,1156 @@
...
+  ProtoMessageField,
+  RPCMethod,
+  UnionMessageDefinition,
+} from '../../src/types.js';
+import { RequiredFieldMapping } from '../../src/required-fields-visitor.js';
+
+/**
+ * Options for creating a RequiredFieldsVisitor test setup.
+ */
+interface CreateVisitorOptions {
```
devsergiy: Should these types be named like `CreateRequiredFieldsVisitorOptions`?

protographic/tests/field-set/01-basics.test.ts
```diff
@@ -0,0 +1,1156 @@
...
+    expect(resultMessage?.fields).toHaveLength(1);
+    assertFieldMessage(resultMessage?.fields[0], {
+      fieldName: 'details',
+      typeName: 'ListOfDetails',
+      fieldNumber: 1,
+      isRepeated: false,
+    });
+  });
+
+  it('should visit a field set for multiple field selections', () => {
```
devsergiy: >field set for
field set with?

protographic/tests/field-set/01-basics.test.ts
```diff
@@ -0,0 +1,1156 @@
...
+          }
+        }
+        Animal pet = 1;
+        string name = 2;
+      }
+      "
+    `);
+  });
+
+  it('should visit a field set for an interface type with extracted interface field', () => {
```
devsergiy: >extracted
it is probably shared selection

protographic/tests/field-set/01-basics.test.ts
```diff
@@ -0,0 +1,1156 @@
...
+    );
+
+    // Should have only ONE nested Address message, not two
+    expect(fieldsMessage?.nestedMessages).toHaveLength(1);
+    expect(fieldsMessage?.nestedMessages?.[0].messageName).toBe('Address');
+
+    // The single Address message should contain all fields from both selections (city, street, zip)
+    const addressMessage = fieldsMessage?.nestedMessages?.[0];
+    expect(addressMessage?.fields).toHaveLength(3);
+    expect(addressMessage?.fields.map((f) => f.fieldName)).toEqual(expect.arrayContaining(['city', 'street', 'zip']));
```
devsergiy, line 1120: Stupid question: In this case, will it be possible to create address message instances populated with different fields?

protographic/src/sdl-to-mapping-visitor.ts
```diff
@@ -115,25 +119,93 @@ export class GraphQLToProtoVisitor {
...
+          (em) => em.typeName === type.name && formatKeyElements(em.key).join(' ') === normalizedKey,
+        );
+        if (!em) {
+          throw new Error(`Entity mapping not found for type ${type.name} and key ${key}`);
+        }
+
+        em.requiredFieldMappings.push(
+          new RequiredFieldMapping({
+            fieldMapping: value.requiredFieldMapping,
+            request: value.rpc?.request ?? '',
```
devsergiy: What does it mean if we assign an empty string?

protographic/src/sdl-to-mapping-visitor.ts
```diff
@@ -266,7 +347,9 @@ export class GraphQLToProtoVisitor {
 
     for (const fieldName in fields) {
       // Skip special federation fields
-      if (fieldName === '_entities') continue;
+      if (fieldName === '_entities') {
```
devsergiy: As we skip the special field `_entities`, should we also skip `_service`?

protographic/src/sdl-to-proto-visitor.ts
```diff
@@ -208,7 +170,23 @@ export class GraphQLToProtoTextVisitor {
    * @param options - Configuration options for the visitor
    */
   constructor(schema: GraphQLSchema, options: GraphQLToProtoTextVisitorOptions = {}) {
-    const { serviceName = 'DefaultService', packageName = 'service.v1', lockData, includeComments = true } = options;
+    const {
+      serviceName = 'DefaultService',
+      packageName = 'service.v1',
+      lockData,
+      includeComments = true,
+      goPackage,
```
devsergiy: doesn't look like we directly use this properties could we do it like this?

```typescript
const {
  serviceName = 'DefaultService',
  ...rest,
}

const protoOptionsFromLanguageProps = buildProtoOptions(
  {
    ...rest
  }
)
```

protographic/src/sdl-to-proto-visitor.ts
```diff
@@ -731,9 +721,13 @@ export class GraphQLToProtoTextVisitor {
 
     for (const fieldName of orderedFieldNames) {
       // Skip special fields like _entities
-      if (fieldName === '_entities') continue;
+      if (fieldName === '_entities') {
```
devsergiy: same question about `_service` field

protographic/src/sdl-to-proto-visitor.ts
```diff
@@ -1159,12 +1163,56 @@ Example:
...
+      }
+    }
 
     return result;
   }
 
+  private getRequiredFieldSet(field: GraphQLField<any, any>): string {
+    const node = field.astNode?.directives
+      ?.find((d) => d.name.value === REQUIRES_DIRECTIVE_NAME)
+      ?.arguments?.find((arg: ArgumentNode) => arg.name.value === 'fields')?.value as StringValueNode;
```
devsergiy, line 1186: we have such lookups in a few places - should the fields value extraction be unified?

protographic/src/required-fields-visitor.ts
```diff
@@ -0,0 +1,622 @@
...
+      };
+      visit(this.fieldSetDoc, this.visitor);
+    }
+  }
+
+  /**
+   * Normalizes the parsed field set operation by rewriting abstract selections.
+   * This ensures consistent handling of interface and union type selections.
+   */
+  private normalizeOperation(): void {
```
devsergiy: should it be called `normalizeSelectionSet`?

protographic/src/abstract-selection-rewriter.ts
```diff
@@ -0,0 +1,715 @@
+/**
+ * @file abstract-selection-rewriter.ts
+ *
+ * This module provides functionality to normalize GraphQL field set selections
+ * when dealing with abstract types (interfaces). It ensures that fields selected
```
devsergiy: >when dealing with abstract types (interfaces)
interfaces and unions?

protographic/src/abstract-selection-rewriter.ts
```diff
@@ -0,0 +1,715 @@
...
+  GraphQLUnionType,
+  print,
+  isSelectionNode,
+  GraphQLCompositeType,
+  SelectionNode,
+  isCompositeType,
+} from 'graphql';
+import { VisitContext } from './types.js';
+
+type GraphQLTypeWithFields = GraphQLObjectType | GraphQLInterfaceType;
```
devsergiy, line 39: technically union is allowed to have a field `__typename`, so it could be present there

protographic/src/abstract-selection-rewriter.ts
```diff
@@ -0,0 +1,715 @@
...
+  /**
+   * Handles entering an InlineFragment node during AST traversal.
+   *
+   * This method processes inline fragments in two ways:
+   *
+   * 1. For non-interface fragments:
+   *    - Validates if the fragment type implements the current interface
+   *    - Removes invalid fragments (those targeting types that don't implement the interface)
+   *
+   * 2. For interface fragments:
```
devsergiy: 3. for union fragments?

protographic/src/abstract-selection-rewriter.ts
```diff
@@ -0,0 +1,715 @@
...
+   */
+  private inlineFragmentIsValidForCurrentAbstractRoot(node: InlineFragmentNode): boolean {
+    const type = this.schema.getType(node.typeCondition?.name.value ?? '');
+    if (!type) {
+      // Type not found in schema - invalid fragment
+      return false;
+    }
+
+    if (!isObjectType(type)) {
+      // Non-object types cannot implement interfaces
```
devsergiy, line 674: but interface could implement interface

protographic/tests/sdl-validation/01-basic-validation.test.ts
Noroth: technically I'm handling it, but I have decided to include it in the generation part when I have made more progress on the engine side.

protographic/tests/sdl-to-mapping/03-federation.test.ts
Noroth: We already have test cases for that already

protographic/tests/proto-utils/build-proto-message.test.ts
Noroth: If the input defines the "same" abstract type on multiple levels it would not deduplicate. But this is on purpose. For most of our cases messages do not include nested messages. We are using this approach essentially for `@requires`. As the selection set can have multiple levels of nesting we would just continue to create more nested messages as we explicitly define which fields we are selecting.

protographic/tests/field-set/01-basics.test.ts
Noroth: It is technically possible but we would have to introduce something like prefixes. Which could be a bit awkward.

e.g. `HomeAddressAddress`, `WorkAddressAddress`

protographic/src/abstract-selection-rewriter.ts
Noroth: I will have to check this at a later stage. Currently I'm only referring to the classes provided by graphql-js. I will make a TODO

protographic/src/abstract-selection-rewriter.ts
Noroth: That's a tomorrow problem. I'll keep it as a TODO

protographic/src/sdl-to-mapping-visitor.ts
Noroth: It should always be set but I can add a guard here.

protographic/src/sdl-to-mapping-visitor.ts
Noroth: Added

protographic/src/sdl-to-proto-visitor.ts
Noroth: okaaaay, weird syntax but I'll take it.

protographic/src/sdl-to-proto-visitor.ts
Noroth: sure

protographic/src/sdl-to-proto-visitor.ts
Noroth: I can do a bit of a code cleanup in a separate PR.

protographic/tests/field-set/01-basics.test.ts
Noroth: Yes

protographic/src/sdl-to-mapping-visitor.ts
Noroth: removed

protographic/.eslintrc
```diff
@@ -0,0 +1,28 @@
...
+    "unicorn/no-await-expression-member": 0,
+    "unicorn/no-array-push-push": 0,
+    "unicorn/filename-case": 0,
+    "@typescript-eslint/no-unused-vars": 0,
+    "@typescript-eslint/no-non-null-assertion": 0,
+    "unicorn/expiring-todo-comments": 0,
+    "no-useless-constructor": 0,
+    "unicorn/no-process-exit": 0
+  }
+}
```
dkorittki, line 28: Curious about the new linting rules. Are they aligned with other TS projects in Cosmo or where did you get these from?

protographic/src/sdl-to-proto-visitor.ts
```diff
@@ -1159,12 +1145,56 @@ Example:
...
+    const entityTypes = Object.values(typeMap).filter((type) => this.isEntityType(type));
+
+    for (const entity of entityTypes) {
+      const requiredFields = Object.values(entity.getFields()).filter((field) =>
+        field.astNode?.directives?.some((d) => d.name.value === REQUIRES_DIRECTIVE_NAME),
+      );
+
+      if (requiredFields.length === 0) {
+        continue;
+      }
```
dkorittki: Not necessary because if `requiredFields` is 0 then the upcoming for loop would skip and we would continue to next entitity anyways.

protographic/src/sdl-to-proto-visitor.ts
```diff
@@ -1159,12 +1145,56 @@ Example:
...
+        result.messageDefinitions.push(...messageLines);
+      }
+    }
 
     return result;
   }
 
+  private getRequiredFieldSet(field: GraphQLField<any, any>): string {
+    const node = field.astNode?.directives
+      ?.find((d) => d.name.value === REQUIRES_DIRECTIVE_NAME)
```
dkorittki, line 1185: Does protographic validate the SDL to not have multiple requires directives per field? This could would not break because of it, just a question from me that comes up when reading this.

protographic/src/abstract-selection-rewriter.ts
```diff
@@ -0,0 +1,719 @@
...
+// TODO: Check if we need to add Union as well because we could select __typename from a Union inline fragment.
+type GraphQLTypeWithFields = GraphQLObjectType | GraphQLInterfaceType;
+
+/**
+ * Rewrites GraphQL selection sets to normalize abstract type selections.
+ *
+ * When a field returns an interface type, selections can be made both at the
+ * interface level and within inline fragments for concrete types. This class
+ * normalizes such selections by moving interface-level fields into each inline
+ * fragment, ensuring consistent selection structure for downstream processing.
```
dkorittki, line 47: Maybe it's worth at this place to describe why this normalization is necessary and why we can't map it to gRPC directly.

protographic/tests/abstract-selection-rewriter/01-basics.test.ts
```diff
@@ -0,0 +1,499 @@
...
+}
+
+
+function normalizeFieldSetWithComplexInterfaces(fieldSet: string, typeName: string): string {
+  const doc = parse(`{ ${fieldSet} }`);
+  const objectType = schemaWithComplexInterfaces.getTypeMap()[typeName] as GraphQLObjectType;
+  const rewriter = new AbstractSelectionRewriter(doc, schemaWithComplexInterfaces, objectType);
+  rewriter.normalize();
+  return print(doc);
+}
```
dkorittki: Both functions essentially do the same. Maybe its worth unifying them into one and let an function argument decide wether you want to get back the `schema` or `schemaWithComplexInterfaces`. Maybe there is some clever TS way of doing that. This is a bit nitty though, ignore if you want.

protographic/src/abstract-selection-rewriter.ts
```diff
@@ -0,0 +1,719 @@
...
+
+    if (this.compositeTypeStack.length > 0) {
+      this.ancestorStacks.push([...this.compositeTypeStack]);
+      this.compositeTypeStack = [];
+    }
+
+    this.currentCompositeRoot = fieldType;
+    this.compositeTypeStack.push(this.currentCompositeRoot);
+
+    // Only process selection sets for interface types
```
dkorittki:
```suggestion
    // Only process selection sets for interface and union types
```

protographic/.eslintrc
Noroth: copied from cli

protographic/src/sdl-to-proto-visitor.ts
Noroth: We have some validation rules in place but not for this case. When reading the GraphQL schema we assume it's valid because you might not have defined certain directives, which will be inferred during composition time.

PR #2443, opened 2026-01-07
feat: improve `authenticateClient` by avoiding unnecessary token creation

wilsonrivera:
The goal of this PR is to avoid creating Keycloak access tokens every time we call `authenticateClient`.

To achieve this we are taking a look at the existing token (if any) and, if that token still valid, we don't do anything, however, if there is a refresh token, we try to refresh the current access token with it and if everything fails or is unavailable, we fallback to the original flow or creating a new access token.

## Summary by CodeRabbit

* **Bug Fixes**
  * Avoids unnecessary re-authentication by checking token expiration before requesting new credentials.

* **New Features**
  * Automatically refreshes sessions when a valid refresh token is available.

* **Behavior**
  * Falls back to the existing sign-in flow if tokens are invalid or refresh fails.
  * Gracefully handles token decode errors and proceeds with the standard sign-in process.

## Checklist

### DIFF
--- a/studio/src/components/user-menu.tsx
+++ b/studio/src/components/user-menu.tsx
@@ -11,6 +11,44 @@ import Link from "next/link";
 import { ThemeToggle } from "./theme-toggle";
 import { Avatar, AvatarFallback } from "./ui/avatar";
 import { Button } from "./ui/button";
+import { PropsWithChildren } from "react";
+
+const localStorageKeysToRemove = [
+  'graphiql:headers',
+  'graphiql:queries',
+  'graphiql:tabState',
+  'graphiql:variables',
+  'graphiql:query',
+  'graphiql:operationName',
+  'playground:pre-flight:selected',
+  'playground:pre-operation:selected',
+  'playground:post-operation:selected',
+  'playground:script:tabState'
+];
+
+function removeLocalStorageItems() {
+  if (typeof window === "undefined") {
+    return;
+  }
+
+  for (const key of localStorageKeysToRemove) {
+    window.localStorage.removeItem(key);
+  }
+}
+
+const LogoutLink = ({ children }: PropsWithChildren) => {
+  return (
+    <Link
+      onClick={() => {
+        removeLocalStorageItems();
+        resetTracking();
+      }}
+      href={process.env.NEXT_PUBLIC_COSMO_CP_URL + "/v1/auth/logout"}
+    >
+      {children || "Logout"}
+    </Link>
+  );
+}
 
 export const UserMenuMobile = () => {
   const user = useUser();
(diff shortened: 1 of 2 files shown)


controlplane/src/core/services/Keycloak.ts
```diff
@@ -37,6 +38,29 @@ export default class Keycloak {
   }
 
   public async authenticateClient() {
+    if (this.client.accessToken) {
+      // We already have an access token, determine whether the token still valid before trying to authenticate again
+      try {
+        const { exp } = decodeJwt(this.client.accessToken);
```
JivusAyrus, line 44: does the expiry of the access token be the same as the keycloak admin auth?

controlplane/src/core/services/Keycloak.ts
```diff
@@ -37,6 +38,29 @@ export default class Keycloak {
...
+        if (!exp || exp * 1000 > Date.now()) {
+          // Either the access token never expires or it hasn't expired
+          return;
+        }
+
+        if (this.client.refreshToken) {
+          await this.client.auth({
```
JivusAyrus, line 51: When we authenticate using the user creds, will they have permissions to do everything an admin can?

wilsonrivera: No, this is only used on the backend, `client` refers to the `Keycloak client` not anything related to the user.

controlplane/src/core/services/Keycloak.ts
wilsonrivera: Yes

controlplane/src/core/services/Keycloak.ts
```diff
@@ -37,6 +38,29 @@ export default class Keycloak {
...
+            grantType: 'refresh_token',
+            refreshToken: this.client.refreshToken,
+            clientId: 'admin-cli',
+          });
+
+          return;
+        }
+      } catch {
+        // ignore
```
StarpTech: Sur, that we don't want to log them? I'd assume that this only happens when Keycloak is down or an incompatibility with the API and client. Sounds relevant to me.

StarpTech (changes requested):
LGTM

controlplane/src/core/services/Keycloak.ts
```diff
@@ -37,6 +38,29 @@ export default class Keycloak {
...
+          await this.client.auth({
+            grantType: 'refresh_token',
+            refreshToken: this.client.refreshToken,
+            clientId: 'admin-cli',
+          });
+
+          return;
+        }
+      } catch (error: unknown) {
+        this.logger.warn(error, 'Failed to refresh the existing access token, a new token will be requested');
```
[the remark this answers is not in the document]
wilsonrivera: The token is a M2M token which is stored in memory at least for 1 hour, at this point only the token refresh could throw.

PR #2445, opened 2026-01-09
feat: rewarm cache from planner cache on schema change

SkArchon:
## Summary by CodeRabbit

* **New Features**
  * Persistent reload state to preserve plan-cache fallback behavior across restarts.
  * CDN added as a configurable cache-warmup source with automatic fallback.

* **Configuration**
  * New flag to enable/disable in-memory cache fallback (default: true).
  * New flag to enable/disable CDN cache source (default: true).

* **Tests**
  * Expanded coverage for cache-warmup, fallback scenarios, and restart behavior.

* **Chores**
  * Bumped ristretto dependency to v2.4.0.

## Description 
This PR introduces a new mechanism to warm the cache on restarts due to schema changes (or whenever the router execution config changes). Config to enable this

```
cache_warmup:
  enabled: true
  in_memory_fallback: true
```
If in_memory_switchover is not present or false, we default to the current cdn cache warmer (with cosmo), this preserves backwards compatibility.

This is achieved by first adding a new `switchoverConfig` which can keep references across restarts, we use this to store a reference to the planner cache. When the router restarts due to an execution config change, we then loop through the elements in the ristretto planner cache.

Right now ristretto does not support iterating, so we have opened a PR here https://github.com/dgraph-io/ristretto/pull/475. For this PR (which we need to discuss before merging) we have pointed ristretto to the fork from where the PR was opened.

We create a new provider for the cache warmer, for `in_memory_switcher` which extracts the queries from the planner cache upon startup. Note that with this method there is no cache warming upon startup, only on subsequent restarts due to execution config changes.

We also support persisting across restarts on hot config reloading. However we ran into a problem of shutdown occuring first before the new graph mux was created (which is the opposite for schema change updates), which means that the current planner cache gets emptied before we can extract the queries from it. As a solution we preemptively extract the queries from the planner cache before its closed whenever a hot config reloading occurs.

This PR also changes how cache warmer is enabled when used with the cosmo CDN, previously it requested the manifest from the CDN if the cache warmer was enabled in the router `config.yaml`, this meant that ex-cache warmer customers could still access old manifest files fron the cdn (as long as it was present), however now its only enabled if the feature is enabled in cosmo (upon startup we get if cache warmer is enabled from cosmo).

## Checklist


router-tests/cache_warmup_test.go
```diff
@@ -914,6 +916,199 @@ func TestCacheWarmup(t *testing.T) {
...
+
+			writeTestConfig(t, "updated", configFile)
+
+			require.EventuallyWithT(t, func(t *assert.CollectT) {
+				res = xEnv.MakeGraphQLRequestOK(testenv.GraphQLRequest{
+					Query: `query { hello }`,
+				})
+				require.Equal(t, "updated", res.Response.Header.Get("X-Router-Config-Version"))
+				require.Equal(t, "HIT", res.Response.Header.Get("x-wg-execution-plan-cache"))
+			}, 2*time.Second, 100*time.Millisecond)
```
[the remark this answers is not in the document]
SkArchon: its called with `require.EventuallyWithT`

router-tests/cache_warmup_test.go
```diff
@@ -914,6 +916,199 @@ func TestCacheWarmup(t *testing.T) {
...
+			res = xEnv.MakeGraphQLRequestOK(testenv.GraphQLRequest{
+				Query: `{ employees { id } }`,
+			})
+			require.Equal(t, 200, res.Response.StatusCode)
+			require.Equal(t, "updated", res.Response.Header.Get("X-Router-Config-Version"))
+			require.JSONEq(t, employeesIDData, res.Body)
+			require.Equal(t, "HIT", res.Response.Header.Get("x-wg-execution-plan-cache"))
+
+		})
+	})
```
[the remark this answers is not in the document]
SkArchon: Its fine if we reuse from a different test file.

router/core/restart_switchover_config.go
```diff
@@ -0,0 +1,157 @@
...
+	// If the configuration change occurred which disabled or enabled the switchover cache, we need to update the internal state
+	if c.enabled {
+		// Only initialize if its nil (because its a first start or it was disabled before)
+		if c.queriesForFeatureFlag == nil {
+			c.queriesForFeatureFlag = make(map[string]any)
+		}
+	} else {
+		c.queriesForFeatureFlag = nil
+	}
+}
```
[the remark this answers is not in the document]
SkArchon: Whenever this attribute is written, there is no concurrent access, if so would there be an issue?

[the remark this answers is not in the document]
SkArchon: Do you recommend doing this check in set()

```// Double-check after acquiring lock
	if c.queriesForFeatureFlag == nil {
		return
	}```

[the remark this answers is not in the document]
Comment (SkArchon) [reply router/core/restart_switchover_config.go]: how would you recommend avoid this double checking logic everywhere

File: router-tests/cache_warmup_test.go [hunk 4]
```diff
@@ -3,9 +3,15 @@ package integration
 import (
 	"context"
 	"net/http"
+	"os"
+	"path/filepath"
+	"syscall"
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
```
Noroth, line 13: Consider installing an import formatter like gci and configure it on save in your ide.

router-tests/cache_warmup_test.go
```diff
@@ -914,6 +920,241 @@ func TestCacheWarmup(t *testing.T) {
...
+			listenString1 := "after1"
+			updateConfig(t, xEnv, ctx, listenString1, getConfigString(listenString1))
+			res := xEnv.MakeGraphQLRequestOK(testenv.GraphQLRequest{Query: `query { hello }`})
+			assert.Equal(t, "MISS", res.Response.Header.Get("x-wg-execution-plan-cache"))
+
+			// Verify cache persisted on restart
+			listenString2 := "after2"
+			updateConfig(t, xEnv, ctx, listenString2, getConfigString(listenString2))
+			res = xEnv.MakeGraphQLRequestOK(testenv.GraphQLRequest{Query: `query { hello }`})
+			assert.Equal(t, "HIT", res.Response.Header.Get("x-wg-execution-plan-cache"))
```
Noroth: Technically you could already abort here if the first call is not an expected `MISS` using `require` instead of `assert` but your call.

router/go.mod
```diff
@@ -63,7 +63,7 @@ require (
 	github.com/alicebob/miniredis/v2 v2.34.0
 	github.com/caarlos0/env/v11 v11.3.1
 	github.com/cep21/circuit/v4 v4.0.0
-	github.com/dgraph-io/ristretto/v2 v2.1.0
+	github.com/dgraph-io/ristretto/v2 v2.4.0
```
Noroth, line 66: good job!

router/core/cache_warmup_plans.go
```diff
@@ -0,0 +1,21 @@
...
+	queries []*nodev1.Operation
+}
+
+func NewPlanSource(switchoverCacheWarmerQueries []*nodev1.Operation) *PlanSource {
+	return &PlanSource{queries: switchoverCacheWarmerQueries}
+}
+
+func (c *PlanSource) LoadItems(_ context.Context, _ *zap.Logger) ([]*nodev1.Operation, error) {
+	return c.queries, nil
+}
```
Noroth, line 32: Spare us some docs here please 🥺

router/core/graph_server.go
```diff
@@ -107,19 +107,21 @@ type (
...
+	CosmoCacheWarmerEnabled bool
 }
 
 func (b BuildGraphMuxOptions) IsBaseGraph() bool {
 	return b.FeatureFlagName == ""
 }
 
 // newGraphServer creates a new server instance.
-func newGraphServer(ctx context.Context, r *Router, routerConfig *nodev1.RouterConfig, proxy ProxyFunc) (*graphServer, error) {
+func newGraphServer(ctx context.Context, r *Router, routerConfig *nodev1.RouterConfig, proxy ProxyFunc, switchoverConfig *SwitchoverConfig) (*graphServer, error) {
```
Noroth: Why did you put the switchover config into the function parameters? Shouldn't it be part of the `*Router.Config`?

router/core/graph_server.go
```diff
@@ -283,7 +292,7 @@ func newGraphServer(ctx context.Context, r *Router, routerConfig *nodev1.RouterC
 		s.logger.Info("Feature flags enabled", zap.Strings("flags", maps.Keys(featureFlagConfigMap)))
 	}
 
-	multiGraphHandler, err := s.buildMultiGraphHandler(ctx, gm.mux, featureFlagConfigMap)
+	multiGraphHandler, err := s.buildMultiGraphHandler(ctx, gm.mux, featureFlagConfigMap, switchoverConfig, cosmoCacheWarmerEnabled)
```
Noroth: Reading this feels a bit wrong. It looks quite random that the `switchover config` and `enabled` is provided as function parameters.

They go into the `BuildGraphMuxOptions` and I think options in Go should rather be a result of composition.
I would probably update the `BuildGraphMuxOptions` a bit to accept functions to configure the options and pass that. 

```
func WithSwitchoverConfig(enabled bool, config *SOConfig)  func(o *BuildGraphMuxOptions) {
  return func ... {
    if enabled {
      o.SOConfig = config
    }
  }
}
``` .. something like that

File: router/core/restart_switchover_config.go [hunk 10]
```diff
@@ -0,0 +1,168 @@
...
+
+func (s *SwitchoverConfig) UpdateSwitchoverConfig(config *Config, isCosmoCacheWarmerEnabled bool) {
+	s.inMemorySwitchOverCache.updateStateFromConfig(config, isCosmoCacheWarmerEnabled)
+}
+
+func (s *SwitchoverConfig) CleanupFeatureFlags(routerCfg *nodev1.RouterConfig) {
+	s.inMemorySwitchOverCache.cleanupUnusedFeatureFlags(routerCfg)
+}
```
Noroth: Please add docs for exported functions and types at least

router/core/graph_server.go
```diff
@@ -1346,11 +1360,18 @@ func (s *graphServer) buildGraphMux(
...
 		}
 
-		if s.Config.cacheWarmup.Source.Filesystem != nil {
+		switch {
+		case s.cacheWarmup.Source.Filesystem != nil:
 			warmupConfig.Source = NewFileSystemSource(&FileSystemSourceConfig{
 				RootPath: s.Config.cacheWarmup.Source.Filesystem.Path,
 			})
-		} else {
+		// We want to enable this only whenever there is no registering with cosmo
```
Noroth: This comment does not reflect what is going on in the assertion. Switchover can be used when cache warmer is not enabled and we are registered with cosmo.

router/core/restart_switchover_config.go
```diff
@@ -0,0 +1,168 @@
...
+
+func (s *SwitchoverConfig) ProcessOnConfigChangeRestart() {
```
Noroth: `ConfigChangeRestart` sounds a bit weird. Do you mean `RouterInstanceRestart` or something like that?
Referring to the supervisor instance and basically reloading the config and recreation of the graph server and muxes.

router/core/restart_switchover_config.go
```diff
@@ -0,0 +1,168 @@
...
+
+	for ffName := range c.queriesForFeatureFlag {
+		// Skip the base which is ""
+		if ffName == "" {
+			continue
+		}
+		if _, exists := ffNames[ffName]; !exists {
+			delete(c.queriesForFeatureFlag, ffName)
+		}
+	}
```
Noroth: Not sure if I'm missing something but isn't this basically

```suggestion
	for ffName := range c.queriesForFeatureFlag {
		// Skip the base which is ""
		if ffName == "" {
			continue
		}
		if _, exists := routerCfg.FeatureFlagConfigs.ConfigByFeatureFlagName[ffName]; !exists {
			delete(c.queriesForFeatureFlag, ffName)
		}
	}
```

router/pkg/config/config.go
```diff
@@ -982,11 +982,12 @@ type CacheWarmupFileSystemSource struct {
...
-	Source         CacheWarmupSource `yaml:"source"  env:"CACHE_WARMUP_SOURCE"`
-	Workers        int               `yaml:"workers" envDefault:"8" env:"CACHE_WARMUP_WORKERS"`
-	ItemsPerSecond int               `yaml:"items_per_second" envDefault:"50" env:"CACHE_WARMUP_ITEMS_PER_SECOND"`
-	Timeout        time.Duration     `yaml:"timeout" envDefault:"30s" env:"CACHE_WARMUP_TIMEOUT"`
+	Enabled                    bool              `yaml:"enabled" envDefault:"false" env:"CACHE_WARMUP_ENABLED"`
+	Source                     CacheWarmupSource `yaml:"source"  env:"CACHE_WARMUP_SOURCE"`
+	Workers                    int               `yaml:"workers" envDefault:"8" env:"CACHE_WARMUP_WORKERS"`
+	ItemsPerSecond             int               `yaml:"items_per_second" envDefault:"50" env:"CACHE_WARMUP_ITEMS_PER_SECOND"`
+	Timeout                    time.Duration     `yaml:"timeout" envDefault:"30s" env:"CACHE_WARMUP_TIMEOUT"`
+	InMemorySwitchoverFallback bool              `yaml:"in_memory_switchover_fallback" envDefault:"true" env:"CACHE_WARMUP_IN_MEMORY_SWITCHOVER_FALLBACK_ENABLED"`
```
Noroth: Not sure if you want to have `_ENABLED` at the end of your env variable if this is not part of the yaml field name

router/core/restart_switchover_config.go
SkArchon: Not sure what happened here 💀

router/core/graph_server.go
SkArchon: Could you elaborate a bit, `BuildGraphMuxOptions` is created in a loop to the `buildMultiGraphHandler`, so we dont have access to `BuildGraphMuxOptions` here.

However I will convert this function to use its own options for now.

router/core/graph_server.go
```diff
@@ -1296,7 +1319,8 @@ func (s *graphServer) buildGraphMux(
 		DisableExposingVariablesContentOnValidationError: s.engineExecutionConfiguration.DisableExposingVariablesContentOnValidationError,
 		ComplexityLimits:                                 s.securityConfiguration.ComplexityLimits,
 	})
-	operationPlanner := NewOperationPlanner(executor, gm.planCache)
+
+	operationPlanner := NewOperationPlanner(executor, gm.planCache, opts.SwitchoverConfig.inMemorySwitchOverCache.IsEnabled())
```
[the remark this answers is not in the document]
SkArchon: SwitchoverConfig wont be nil so its fine

router/core/graph_server.go
```diff
@@ -120,8 +120,16 @@ func (b BuildGraphMuxOptions) IsBaseGraph() bool {
 	return b.FeatureFlagName == ""
 }
 
+// BuildMultiGraphHandlerOptions contains the configuration options for building a multi-graph handler.
+type BuildMultiGraphHandlerOptions struct {
```
Noroth: nit, I think we don't need to have this exported. It's only used in that single context

router/core/restart_switchover_config.go
```diff
@@ -24,31 +23,36 @@ func NewSwitchoverConfig(logger *zap.Logger) *SwitchoverConfig {
...
 }
 
+// CleanupFeatureFlags cleans up anything related to unused feature flags due to being now excluded
+// from the execution config
 func (s *SwitchoverConfig) CleanupFeatureFlags(routerCfg *nodev1.RouterConfig) {
 	s.inMemorySwitchOverCache.cleanupUnusedFeatureFlags(routerCfg)
 }
 
-func (s *SwitchoverConfig) ProcessOnConfigChangeRestart() {
+func (s *SwitchoverConfig) ProcessOnConfigChangeRouterInstanceRestart() {
```
Noroth:
```suggestion
func (s *SwitchoverConfig) OnRouterConfigReload() {
```

proto/wg/cosmo/node/v1/node.proto
```diff
@@ -43,6 +43,7 @@ message ResponseStatus {
 message RegistrationInfo {
   AccountLimits account_limits = 1;
   string graph_public_key = 2;
+  bool is_cache_warmer_enabled = 3;
```
StarpTech: Why did you add this to the registration endpoint?

router/pkg/config/config.schema.json
```diff
@@ -2430,6 +2430,11 @@
           "duration": {
             "minimum": "1s"
           }
+        },
+        "in_memory_switchover_fallback": {
```
StarpTech: The naming is a bit odd. `switchover` and `fallback` feel redundant.

router/pkg/config/config.schema.json
```diff
@@ -2430,6 +2430,11 @@
...
+          "type": "boolean",
+          "description": "Enable in-memory switchover fallback. When enabled, the router will cache query plans in memory and reuse them on restarts. This is useful when deploying new router configurations without downtime. The default value is true.",
```
StarpTech: You mean across schema updates right?

proto/wg/cosmo/node/v1/node.proto
SkArchon: As discussed will refactor to figure out if the in_memory_fallback should be used based on CDN polling.

router/core/graph_server.go
```diff
@@ -1346,16 +1360,36 @@ func (s *graphServer) buildGraphMux(
...
-		} else {
-			cdnSource, err := NewCDNSource(s.Config.cdnConfig.URL, s.graphApiToken, s.logger)
+		// Enable in-memory switchover fallback when:
+		// - Router has cache warmer with inMemoryFallback enabled, AND
+		// - Either:
+		//   - Using static execution config (not Cosmo): s.selfRegister == nil
+		//   - OR CDN cache warmer is explictly disabled
+		case s.cacheWarmup.InMemoryFallback && (s.selfRegister == nil || !s.Config.cacheWarmup.Source.CdnSource.Enabled):
+			// We first utilize the plan cache (if it was already set, so not on first starts) to create a list of queries
+			// and reset the plan cache to the new plan cache for this start afterwords
```
StarpTech: typo

router/core/cache_warmup_cdn.go
```diff
@@ -55,13 +56,27 @@ func NewCDNSource(endpoint, token string, logger *zap.Logger) (*CDNSource, error
...
 	}, nil
 }
 
-func (c *CDNSource) LoadItems(ctx context.Context, log *zap.Logger) ([]*nodev1.Operation, error) {
+func (c *CDNSource) LoadItems(ctx context.Context, log *zap.Logger) (operations []*nodev1.Operation, err error) {
 	span := trace.SpanFromContext(ctx)
 	defer span.End()
 
+	// Defer fallback to PlanSource if items are nil
+	if c.fallbackSource != nil {
```
StarpTech: Not related to your changes, but do we use a retryable HttpClient? If not, please add it,

router/core/cache_warmup_cdn.go
```diff
@@ -55,13 +56,27 @@ func NewCDNSource(endpoint, token string, logger *zap.Logger) (*CDNSource, error
...
+		defer func() {
+			// If there were no operations loaded from CDN, use the fallbackSource
+			if len(operations) == 0 {
```
StarpTech: Wouldn't it be better designed if the concrete implementation didn't need to know about the fallback? What prevents us from doing it in `unc (w *cacheWarmup) run`?

router/core/reload_persistent_state_test.go
```diff
@@ -0,0 +1,433 @@
...
+
+		cache.updateStateFromConfig(cfg)
+
+		require.Nil(t, cache.queriesForFeatureFlag)
+	})
+}
+
+func TestInMemorySwitchOverCache_GetPlanCacheForFF(t *testing.T) {
+	t.Parallel()
+	t.Run("returns operations for existing feature flag when enabled with ristretto cache", func(t *testing.T) {
```
StarpTech, line 123: Do we have tests that cover the case when the CDN is unavailable and we fallback?

router/core/operation_planner.go
```diff
@@ -21,33 +21,42 @@ type planWithMetaData struct {
...
 	argumentUsageInfo                 []*graphqlmetricsv1.ArgumentUsageInfo
+	content                           string
 }
 
 type OperationPlanner struct {
 	sf             singleflight.Group
 	planCache      ExecutionPlanCache[uint64, *planWithMetaData]
 	executor       *Executor
 	trackUsageInfo bool
+	storeContent   bool
```
StarpTech: Can we be more specific e.g `operationContent` ?

router/core/operation_planner.go
```diff
@@ -81,6 +90,10 @@ func (p *OperationPlanner) preparePlan(ctx *operationContext) (*planWithMetaData
 		schemaDocument:    p.executor.RouterSchema,
 	}
 
+	if opts.storeContent {
+		out.content = ctx.Content()
```
StarpTech, line 94: Can we avoid storing the content? There is an astprinter in the engine that can print the operation from the document.

router/core/restart_switchover_config.go
```diff
@@ -0,0 +1,172 @@
...
+	for k, v := range c.queriesForFeatureFlag {
+		if cache, ok := v.(planCache); ok {
+			switchoverMap[k] = convertToNodeOperation(cache)
+		}
+	}
+	c.queriesForFeatureFlag = switchoverMap
+}
+
+// cleanupUnusedFeatureFlags removes any feature flags from the in-memory switchover cache
+// this is useful in case where the updated execution config excludes a feature flag
```
StarpTech: Could you rephrase? It's a bit hard to understand

router/core/reload_persistent_state.go
```diff
@@ -0,0 +1,172 @@
...
+func (c *InMemorySwitchoverCache) IsEnabled() bool {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+
+	return c.queriesForFeatureFlag != nil
+}
+
+// getPlanCacheForFF gets the plan cache in the []*nodev1.Operation format for a specific feature flag key
+func (c *InMemorySwitchoverCache) getPlanCacheForFF(featureFlagKey string) []*nodev1.Operation {
+	c.mu.RLock()
```
StarpTech, line 87: Why do we need locks for everything?

StarpTech (changes requested):

router/core/cache_warmup_cdn.go
SkArchon: I am guessing you mean for the CDN source, yes we use a retryable client.

router/core/graph_server.go
SkArchon: fixed.

router/core/reload_persistent_state_test.go
SkArchon: Hi yes, it's in `cache_warmup_test.go`.

router/core/restart_switchover_config.go
SkArchon: Done

router/core/operation_planner.go
SkArchon: I didn't look at optimizing it as according to my benchmarking the memory usage did not change by using out.content (as its reference is held in the ast).

Though I did I miss astprinter, will use it instead of assinging out.content, in case an internal change causes `out.content` to use its memory for the string.

router/core/reload_persistent_state.go
SkArchon: Since we can have a schema change and router config change occur at the same time most of these functions need locks.

For router config changes (hot config reload) we extract the queries BEFORE shut down, is initiated, which means a schema change could cause changes to the map while the map contents are overridden.

We can remove the locks for `updateStateFromConfig` since its called on the router / hot config reload start, but its only one function and I kept it for consistency / in case anything changes in the future, since this is only used on hot config reload startup.

router/core/reload_persistent_state.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	switchoverMap := make(map[string]any)
+	for k, v := range c.queriesForFeatureFlag {
+		if cache, ok := v.(planCache); ok {
+			switchoverMap[k] = convertToNodeOperation(cache)
+		}
+	}
+	c.queriesForFeatureFlag = switchoverMap
+}
```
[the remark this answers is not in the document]
SkArchon: This wont get called twice in succession.

router/core/operation_planner.go
SkArchon: We cannot use the astprinter since the astprinter prints additional information which is used in the planner but is not part of the query the user sent, for example `__typename`. The operation hash is built from the normalised query and not the planned query.

Also setting out.content does not increase memory *as long as* the normalization cache is enabled, since the string is the same reference stored in the normalization cache.

router/core/operation_planner.go
SkArchon: Done

router/core/operation_planner.go
SkArchon: We cannot use the astprinter since the astprinter prints additional information which is used in the planner but is not part of the query the user sent, for example __typename. The operation hash is built from the normalised query and not the planned query.

StarpTech: ok

router/core/cache_warmup_cdn.go
SkArchon: Done

router/core/router.go
```diff
@@ -2095,6 +2056,12 @@ func WithConfigPollerConfig(cfg *RouterConfigPollerConfig) Option {
 	}
 }
 
+func WithSwitchoverConfig(cfg *SwitchoverConfig) Option {
```
StarpTech: The name is not clear. A switchover can mean anything. How about `WithMemoryPlanCacheFallback`?

router/pkg/config/config.schema.json
```diff
@@ -2430,6 +2445,11 @@
           "duration": {
             "minimum": "1s"
           }
+        },
+        "in_memory_fallback": {
+          "type": "boolean",
+          "description": "Enable in-memory fallback. When enabled, the router will cache query plans in memory and reuse them on schema changes and hot config reloads. This is useful when deploying new router configurations without downtime. The default value is true.",
```
StarpTech: > This is useful when deploying new router configurations without downtime

This makes no sense. The cache will help with cold starts for operation planning.

StarpTech (changes requested):

router/core/router.go
SkArchon: I think switchover should stay the same, since its goal is to just a struct which contains references to things that should persist across restarts. 

I instead changed 
```
type SwitchoverConfig struct {
	inMemorySwitchOverCache *InMemorySwitchoverCache
}
```
to
```
type SwitchoverConfig struct {
	inMemoryPlanCacheFallback *InMemoryPlanCacheFallback
}
```

router/pkg/config/config.schema.json
SkArchon: Updated

router/core/graph_server.go
```diff
@@ -1346,16 +1360,35 @@ func (s *graphServer) buildGraphMux(
...
-		} else {
+		// Enable in-memory switchover fallback when:
+		// - Router has cache warmer with inMemoryFallback enabled, AND
+		// - Either:
+		//   - Using static execution config (not Cosmo): s.selfRegister == nil
+		//   - OR CDN cache warmer is explictly disabled
+		case s.cacheWarmup.InMemoryFallback && (s.selfRegister == nil || !s.Config.cacheWarmup.Source.CdnSource.Enabled):
+			// We first utilize the existing plan cache (if it was already set, i.e., not on the first start) to create a list of queries
+			// and then reset the plan cache to the new plan cache for this start afterwards.
+			warmupConfig.Source = NewPlanSource(opts.SwitchoverConfig.inMemoryPlanCacheFallback.getPlanCacheForFF(opts.FeatureFlagName))
```
StarpTech: Let's find a different name for `SwitchoverConfig` I find it very unspecific in the core package.

router/core/restart_switchover_config.go
```diff
@@ -0,0 +1,172 @@
...
+			delete(c.queriesForFeatureFlag, ffName)
+		}
+	}
+}
+
+func convertToNodeOperation(data planCache) []*nodev1.Operation {
+	items := make([]*nodev1.Operation, 0)
+
+	// Ensure any buffered writes have been applied
+	data.Wait()
```
StarpTech: Can this block forever or has a ctx variant?

router/core/reload_persistent_state.go
```diff
@@ -0,0 +1,172 @@
...
+			continue
+		}
+		if _, exists := routerCfg.FeatureFlagConfigs.ConfigByFeatureFlagName[ffName]; !exists {
+			delete(c.queriesForFeatureFlag, ffName)
+		}
+	}
+}
+
+func convertToNodeOperation(data planCache) []*nodev1.Operation {
+	items := make([]*nodev1.Operation, 0)
```
StarpTech, line 162: Can we preallocate here based on the cache size?

router/core/router.go
StarpTech: Sounds good.

router/core/restart_switchover_config.go
SkArchon: What I understood is that its sent to a same channel that is also used for Set, so should be fine. but will double check.

router/core/reload_persistent_state.go
SkArchon: By default we cant find a length from the cache, but we should be able be able to use MaxCost() for the capacity, because every item has a cost of 1. Will update.

router/core/graph_server.go
SkArchon: Will change.

router/core/reload_persistent_state.go
StarpTech: I think it's fine then. I'd avoid coupling it with a length that can change independently.

router/core/restart_switchover_config.go
SkArchon: Went through the code again and it seems that no block forever would happen.

router/core/graph_server.go
SkArchon: These are a few suggestions I have, lmk what you think
```
RouterPersistentState
ReloadPersistentState
RouterReloadBridge
```

router/core/reload_persistent_state.go
```diff
@@ -0,0 +1,169 @@
...
+
+	data.IterValues(func(v *planWithMetaData) (stop bool) {
+		items = append(items, &nodev1.Operation{
+			Request: &nodev1.OperationRequest{Query: v.content},
+		})
+		return false
+	})
+	return items
```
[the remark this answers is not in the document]
SkArchon: Theres a nil guard inside ristretto, and we are ok with this

router/core/graph_server.go
SkArchon: Done.

router/core/reload_persistent_state.go
```diff
@@ -0,0 +1,172 @@
...
+	for ffName := range c.queriesForFeatureFlag {
+		// Skip the base which is ""
+		if ffName == "" {
+			continue
+		}
+		if _, exists := routerCfg.FeatureFlagConfigs.ConfigByFeatureFlagName[ffName]; !exists {
+			delete(c.queriesForFeatureFlag, ffName)
+		}
+	}
+}
```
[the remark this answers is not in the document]
SkArchon: This has been fixed pls check

router/go.mod
```diff
@@ -63,7 +63,7 @@ require (
 	github.com/alicebob/miniredis/v2 v2.34.0
 	github.com/caarlos0/env/v11 v11.3.1
 	github.com/cep21/circuit/v4 v4.0.0
-	github.com/dgraph-io/ristretto/v2 v2.1.0
+	github.com/dgraph-io/ristretto/v2 v2.0.0
```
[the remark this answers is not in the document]
SkArchon: Please confirm that this has been fixed

router/go.mod
```diff
@@ -195,3 +195,5 @@ replace (
 // Use what is best for your personal workflow. See CONTRIBUTING.md for more information
 
 //replace github.com/wundergraph/graphql-go-tools/v2 => ../../graphql-go-tools/v2
+
+replace github.com/dgraph-io/ristretto/v2 => github.com/wundergraph/ristretto/v2 v2.0.0-20260118191717-5112a101c70c
```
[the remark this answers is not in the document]
SkArchon: Please confirm that this has been fixed

router/go.mod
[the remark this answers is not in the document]
SkArchon: I think you are looking at an outdated dependency, can you look at the current diff from scratch?

router/go.mod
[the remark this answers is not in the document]
SkArchon: I think you are looking at an outdated dependency, can you look at the current diff from scratch?

PR #2458, opened 2026-01-16
feat: hide posthog errors from cli output if debug not enabled

alepane21:
We want to avoid to print posthog errors in the console if not in the debug session, to avoid confusion about the success of the launched command.
We are using shutdown flush, and for how the library is done, that means that if there is any error, it is going to print it directly to console.log!
So the only sensible way that I found is to fake the response to be always positive.
This will also disable the retry of the requests, but I think is an acceptable compromise.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CLI stability by ensuring telemetry network issues won't disrupt your workflow.

## Checklist

### DIFF
--- a/cli/src/core/telemetry.ts
+++ b/cli/src/core/telemetry.ts
@@ -13,6 +13,49 @@ let client: PostHog | null = null;
 
 let apiClient: ReturnType<typeof CreateClient> | null = null;
 
+type PostHogFetchOptions = {
+  method: 'GET' | 'POST' | 'PUT' | 'PATCH';
+  mode?: 'no-cors';
+  credentials?: 'omit';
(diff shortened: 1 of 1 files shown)


cli/src/core/telemetry.ts
```diff
@@ -51,6 +94,7 @@ export const initTelemetry = () => {
     flushAt: 1, // For CLI, we want to send events immediately
     flushInterval: 0, // Don't wait to flush events
     disableGeoip: false,
+    fetch: safePostHogFetch,
```
SkArchon, line 97: I am assuming this overrides?

endigma (approved):

PR #2470, opened 2026-01-27
feat: expose costs

ysmolski:
Integrate static and dynamic cost calculation into the router. 
Expose the static (estimated) cost value in the modules.
Expose both costs via telemetry and response headers.

Metrics for costs should be enabled individually in the corresponding telemetry section.

Composition was reviewed in another PR and merged here.

I have added end-to-end benchmark for a big query to measure the impact of cost control:
```
     │ SequentialBig │     SequentialBigCostAnalysis      │
     │    sec/op     │   sec/op     vs base               │
*-14     584.1µ ± 0%   586.7µ ± 0%  +0.45% (p=0.000 n=40)

     │ SequentialBig │      SequentialBigCostAnalysis      │
     │      B/s      │     B/s       vs base               │
*-14    5.527Mi ± 0%   5.503Mi ± 0%  -0.43% (p=0.000 n=40)

     │ SequentialBig │   SequentialBigCostAnalysis    │
     │     B/op      │     B/op      vs base          │
*-14    456.1Ki ± 0%   456.1Ki ± 0%  ~ (p=0.822 n=40)

     │ SequentialBig │     SequentialBigCostAnalysis      │
     │   allocs/op   │  allocs/op   vs base               │
*-14     6.444k ± 0%   6.452k ± 0%  +0.12% (p=0.000 n=40)
```

Fixes ENG-8844
Fixes ENG-8986

## Summary by CodeRabbit

* **New Features**
  * Cost control: measure or enforce GraphQL operation costs, with optional response headers and static validation to block over-limit queries.
  * Support for @cost and @listSize directives across composition, federation, and router; propagated through subgraph composition.

* **Configuration**
  * cost_control options (enabled, mode, max_estimated_limit, estimated_list_size, expose_headers) and CostStats toggles.

* **Metrics**
  * New estimated and actual operation cost histograms and recording APIs.

* **Tests / Docs**
  * Extensive directive and cost integration tests; composition architecture docs added.

### DIFF
--- a/router-tests/go.mod
+++ b/router-tests/go.mod
@@ -28,7 +28,7 @@ require (
 	github.com/wundergraph/cosmo/demo/pkg/subgraphs/projects v0.0.0-20250715110703-10f2e5f9c79e
 	github.com/wundergraph/cosmo/router v0.0.0-20260213130455-6e3277e7b850
 	github.com/wundergraph/cosmo/router-plugin v0.0.0-20250808194725-de123ba1c65e
-	github.com/wundergraph/graphql-go-tools/v2 v2.0.0-rc.264
+	github.com/wundergraph/graphql-go-tools/v2 v2.0.0-rc.265
 	go.opentelemetry.io/otel v1.36.0
 	go.opentelemetry.io/otel/sdk v1.36.0
 	go.opentelemetry.io/otel/sdk/metric v1.36.0
--- a/router-tests/go.sum
+++ b/router-tests/go.sum
@@ -134,8 +134,9 @@ github.com/golang/protobuf v1.5.4 h1:i7eJL8qZTpSEXOPTxNKhASYpMn+8e5Q6AdndVa1dWek
 github.com/golang/protobuf v1.5.4/go.mod h1:lnTiLA8Wa4RWRcIUkrtSVa5nRhsEGBg48fD6rSs7xps=
 github.com/google/go-cmp v0.5.6/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
 github.com/google/go-cmp v0.5.9/go.mod h1:17dUlkBOakJ0+DkrSSNjCkIjxS6bF9zb3elmeNGIjoY=
-github.com/google/go-cmp v0.6.0 h1:ofyhxvXcZhMsU5ulbFiLKl/XBFqE1GSq7atu8tAmTRI=
 github.com/google/go-cmp v0.6.0/go.mod h1:17dUlkBOakJ0+DkrSSNjCkIjxS6bF9zb3elmeNGIjoY=
+github.com/google/go-cmp v0.7.0 h1:wk8382ETsv4JYUZwIsn6YpYiWiBsYLSJiTsyBybVuN8=
+github.com/google/go-cmp v0.7.0/go.mod h1:pXiqmnSA92OHEEa9HXL2W4E7lf9JzCmGVUdgjX3N/iU=
 github.com/google/go-containerregistry v0.20.3 h1:oNx7IdTI936V8CQRveCjaxOiegWwvM7kqkbXTpyiovI=
 github.com/google/go-containerregistry v0.20.3/go.mod h1:w00pIgBRDVUDFM6bq+Qx8lwNWK+cxgCuX1vd3PIBDNI=
 github.com/google/renameio v0.1.0/go.mod h1:KWCgfxg9yswjAJkECMjeO8J8rahYeXnNhOm40UhjYkI=
@@ -356,8 +357,8 @@ github.com/wundergraph/astjson v1.1.0 h1:xORDosrZ87zQFJwNGe/HIHXqzpdHOFmqWgykCLV
 github.com/wundergraph/astjson v1.1.0/go.mod h1:h12D/dxxnedtLzsKyBLK7/Oe4TAoGpRVC9nDpDrZSWw=
 github.com/wundergraph/go-arena v1.1.0 h1:9+wSRkJAkA2vbYHp6s8tEGhPViRGQNGXqPHT0QzhdIc=
 github.com/wundergraph/go-arena v1.1.0/go.mod h1:ROOysEHWJjLQ8FSfNxZCziagb7Qw2nXY3/vgKRh7eWw=
-github.com/wundergraph/graphql-go-tools/v2 v2.0.0-rc.264 h1:osns9puiqH4yJfugvIFWKh1CarJIAbBTzuSYM5HyJRI=
-github.com/wundergraph/graphql-go-tools/v2 v2.0.0-rc.264/go.mod h1:gfmmrPd2khZONmwYE8RIfnGjwIG+RqL52jYiBzcUST8=
+github.com/wundergraph/graphql-go-tools/v2 v2.0.0-rc.265 h1:KVmojt3oH13VX8Yr8NZ+fuOiruLyznderHITJs1MyWE=
+github.com/wundergraph/graphql-go-tools/v2 v2.0.0-rc.265/go.mod h1:HjTAO/cuICpu31IfHY9qmSPygx6Gza7Wt9hTSReTI+A=
 github.com/xrash/smetrics v0.0.0-20250705151800-55b8f293f342 h1:FnBeRrxr7OU4VvAzt5X7s6266i6cSVkkFPS0TuXWbIg=
 github.com/xrash/smetrics v0.0.0-20250705151800-55b8f293f342/go.mod h1:Ohn+xnUBiLI6FVj/9LpzZWtj1/D6lUovWYBkxHVV3aM=
 github.com/yosida95/uritemplate/v3 v3.0.2 h1:Ed3Oyj9yrmi9087+NczuL5BwkIc4wvTb5zIM+UJPGz4=
--- a/router/go.mod
+++ b/router/go.mod
@@ -31,7 +31,7 @@ require (
 	github.com/tidwall/gjson v1.18.0
 	github.com/tidwall/sjson v1.2.5
 	github.com/twmb/franz-go v1.16.1
-	github.com/wundergraph/graphql-go-tools/v2 v2.0.0-rc.264
+	github.com/wundergraph/graphql-go-tools/v2 v2.0.0-rc.265
 	// Do not upgrade, it renames attributes we rely on
 	go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.58.0
 	go.opentelemetry.io/contrib/propagators/b3 v1.23.0
--- a/router/go.sum
+++ b/router/go.sum
@@ -119,8 +119,9 @@ github.com/golang/protobuf v1.5.4 h1:i7eJL8qZTpSEXOPTxNKhASYpMn+8e5Q6AdndVa1dWek
 github.com/golang/protobuf v1.5.4/go.mod h1:lnTiLA8Wa4RWRcIUkrtSVa5nRhsEGBg48fD6rSs7xps=
 github.com/google/go-cmp v0.5.6/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
 github.com/google/go-cmp v0.5.9/go.mod h1:17dUlkBOakJ0+DkrSSNjCkIjxS6bF9zb3elmeNGIjoY=
-github.com/google/go-cmp v0.6.0 h1:ofyhxvXcZhMsU5ulbFiLKl/XBFqE1GSq7atu8tAmTRI=
 github.com/google/go-cmp v0.6.0/go.mod h1:17dUlkBOakJ0+DkrSSNjCkIjxS6bF9zb3elmeNGIjoY=
+github.com/google/go-cmp v0.7.0 h1:wk8382ETsv4JYUZwIsn6YpYiWiBsYLSJiTsyBybVuN8=
+github.com/google/go-cmp v0.7.0/go.mod h1:pXiqmnSA92OHEEa9HXL2W4E7lf9JzCmGVUdgjX3N/iU=
 github.com/google/go-containerregistry v0.20.3 h1:oNx7IdTI936V8CQRveCjaxOiegWwvM7kqkbXTpyiovI=
 github.com/google/go-containerregistry v0.20.3/go.mod h1:w00pIgBRDVUDFM6bq+Qx8lwNWK+cxgCuX1vd3PIBDNI=
 github.com/google/renameio v0.1.0/go.mod h1:KWCgfxg9yswjAJkECMjeO8J8rahYeXnNhOm40UhjYkI=
@@ -328,8 +329,8 @@ github.com/wundergraph/astjson v1.1.0 h1:xORDosrZ87zQFJwNGe/HIHXqzpdHOFmqWgykCLV
 github.com/wundergraph/astjson v1.1.0/go.mod h1:h12D/dxxnedtLzsKyBLK7/Oe4TAoGpRVC9nDpDrZSWw=
 github.com/wundergraph/go-arena v1.1.0 h1:9+wSRkJAkA2vbYHp6s8tEGhPViRGQNGXqPHT0QzhdIc=
 github.com/wundergraph/go-arena v1.1.0/go.mod h1:ROOysEHWJjLQ8FSfNxZCziagb7Qw2nXY3/vgKRh7eWw=
-github.com/wundergraph/graphql-go-tools/v2 v2.0.0-rc.264 h1:osns9puiqH4yJfugvIFWKh1CarJIAbBTzuSYM5HyJRI=
-github.com/wundergraph/graphql-go-tools/v2 v2.0.0-rc.264/go.mod h1:gfmmrPd2khZONmwYE8RIfnGjwIG+RqL52jYiBzcUST8=
+github.com/wundergraph/graphql-go-tools/v2 v2.0.0-rc.265 h1:KVmojt3oH13VX8Yr8NZ+fuOiruLyznderHITJs1MyWE=
+github.com/wundergraph/graphql-go-tools/v2 v2.0.0-rc.265/go.mod h1:HjTAO/cuICpu31IfHY9qmSPygx6Gza7Wt9hTSReTI+A=
 github.com/yosida95/uritemplate/v3 v3.0.2 h1:Ed3Oyj9yrmi9087+NczuL5BwkIc4wvTb5zIM+UJPGz4=
 github.com/yosida95/uritemplate/v3 v3.0.2/go.mod h1:ILOh0sOhIJR3+L/8afwt/kE++YT040gmv5BQTMR2HP4=
 github.com/yuin/gopher-lua v1.1.1 h1:kYKnWBjvbNP4XLT3+bPEwAXJx262OhaHDWDVOPjL46M=


router/pkg/config/config.schema.json
```diff
@@ -2683,6 +2683,30 @@
...
+          "additionalProperties": false,
+          "properties": {
+            "enabled": {
+              "type": "boolean",
+              "default": false,
+              "description": "When enabled, static cost is calculated and exposed to modules. Set to true to enable StaticLimit option."
+            },
+            "static_limit": {
+              "type": "integer",
+              "description": "The maximum allowed static (estimated) cost for a query. Queries exceeding this limit will be rejected before execution. If the limit is 0, this limit isn't applied.",
```
StarpTech: We should work with modes here, `enforce,measure` to make it more explicit.

router/core/context.go
```diff
@@ -721,6 +725,19 @@ func (o *operationContext) QueryPlanStats() (QueryPlanStats, error) {
 	return qps, nil
 }
 
+func (o *operationContext) StaticCost() (int, error) {
```
StarpTech: I recommend rethinking the interface we expose in the module for cost Operation().Cost(). To support more features in the future (dynamic costs), we should make it a struct.

StarpTech (changes requested):

router/pkg/config/config.go
```diff
@@ -459,20 +458,32 @@ type ComplexityLimits struct {
...
-	// If the limit is 0, this limit isn't applied.
-	StaticLimit int `yaml:"static_limit,omitempty" envDefault:"0" env:"SECURITY_COST_ANALYSIS_STATIC_LIMIT"`
+	// Mode controls cost analysis behavior:
+	// - "measure": calculates costs without rejecting operations (for monitoring)
+	// - "enforce": calculates costs and rejects operations exceeding the estimated limit
+	Mode CostAnalysisMode `yaml:"mode,omitempty" envDefault:"measure" env:"SECURITY_COST_ANALYSIS_MODE"`
+
+	// EstimatedLimit is the maximum allowed estimated cost for a query.
+	// Requires Mode set to "enforce". Operations exceeding this limit are rejected.
+	EstimatedLimit int `yaml:"estimated_limit,omitempty" envDefault:"0" env:"SECURITY_COST_ANALYSIS_ESTIMATED_LIMIT"`
```
StarpTech: The naming could be clearer. I'd adopt `MaxEstimatedLimit`

router-tests/security_test.go
StarpTech: Please add a test that we log the measurement with the right attributes

router/pkg/config/config.schema.json
```diff
@@ -2691,17 +2691,23 @@
...
+              "description": "Maximum allowed estimated cost for a query. Only enforced when mode is 'enforce'.",
               "default": 0,
               "minimum": 0
             },
-            "list_size": {
+            "estimated_list_size": {
               "type": "integer",
-              "description": "The default assumed size for list fields when no @listSize directive nor slicing argument is provided. Used as a multiplier for list field static costs.",
+              "description": "Default assumed size for list fields when no @listSize directive is specified. Used for estimated cost calculation.",
               "default": 10,
```
StarpTech: We should not enforce such a low limit by default. We keep it open `0`. Same default as [Apollo](https://www.apollographql.com/docs/graphos/routing/security/demand-control#configuring-demand-control).

ysmolski: They don't have defaults for this parameter meaning that they require user to set some value. We could do the same.

router/core/context.go
ysmolski: I made it a struct, but it won't be used for dynamic costs. maybe for something else (like cost layout) later, if users request it..

router/pkg/config/config.schema.json
ysmolski: I have made it a required parameter when Costs Analysis is enabled

router/pkg/config/config.go
ysmolski: Agree. Renamed

router-tests/security_test.go
ysmolski: I have moved all tests into a single file: `costs_test.go`. And it contains tests for metrics.

router/core/operation_processor.go
```diff
@@ -1376,6 +1379,40 @@ func (o *OperationKit) runComplexityComparisons(complexityLimitConfig *config.Co
...
+
+	if estimatedCost > costAnalysis.MaxEstimatedLimit {
+		return &httpGraphqlError{
+			message:    fmt.Sprintf("The estimated query cost %d exceeds the maximum allowed limit %d", estimatedCost, costAnalysis.MaxEstimatedLimit),
+			statusCode: http.StatusBadRequest,
+		}
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
ysmolski: Find a case when CostAnalysis is enabled and yet calculator is nil.

router/pkg/metric/metric_store.go
```diff
@@ -423,9 +447,62 @@ func (h *Metrics) MeasureSchemaFieldUsage(ctx context.Context, schemaUsage int64
...
+		explodeRecordInstrument(ctx, sliceAttr, func(ctx context.Context, newOpts ...otelmetric.RecordOption) {
+			newOpts = append(newOpts, opts...)
+			h.promRequestMetrics.MeasureOperationCostDelta(ctx, delta, newOpts...)
+		})
+	}
+
+	// OTEL metrics
+	opts = append(opts, otelmetric.WithAttributes(sliceAttr...))
+	h.otlpRequestMetrics.MeasureOperationCostDelta(ctx, delta, opts...)
+}
```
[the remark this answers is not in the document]
ysmolski: We calculated it as estimated - actual, that is expected to be higher than 0. If it is less than zero then zero value is okay

router/pkg/metric/metric_store.go
```diff
@@ -78,14 +79,24 @@ var (
 		otelmetric.WithDescription(OperationPlanningTimeHistogramDescription),
 	}
 
+	costBucketBounds = []float64{
+		0, 1, 5, 10,
+		25, 50, 100,
+		250, 500, 1000,
+		2500, 5000, 10000,
+		25000, 50000, 100000,
+	}
```
endigma, line 84: Is there a use for a 0-1, 1-5, 5-10 bucket? Can cost even be this low? What range of values are we suggesting customers aim to cover?

I don't see the use of recording 0, 1, 5, 10, 25 as buckets and also having 100000 on the same scale

ysmolski: If client goes with the default weights (1 for non-scalar fields) then costs for queries can be easy in the `0..25` range. I agree on the redundant granularity of `0, 1, 5, 10` boundaries though. We can go with `0, 5, 10`. I want them being able to distinguish almost zero values and something tangible.

After talking to customers we have wide range of applications in the way how they will use it. This is why I though maybe we could add this bucket boundaries as configurable via config? Is it even possible?

ysmolski: I have reduced the range

endigma: I am not sure about how dynamic buckets would work with cloud, cc @StarpTech @pepol ?

pepol: As far as I can tell, we're not using any bucket-specific operations in cloud. We're using clickhouse histogram functionality and passing in quantile for, e.g. latency, not approaching specific bucket configurations: https://github.com/wundergraph/cosmo/blob/main/controlplane/src/core/repositories/analytics/MetricsRepository.ts#L230

So I'd say similar approach would be used to determine e.g. p99 for costs within controlplane/cloud.

StarpTech: 16 buckets are too high, which can produce a lot of cardinality. We can't use dynamic buckets, but we can define a custom one for the cost metric if it makes sense. In clickhouse we have implemented our own histogram function that only works on the same bucket configuration per metric.

composition/src/errors/errors.ts
```diff
@@ -1694,3 +1694,84 @@ export function oneOfRequiredFieldsError({ requiredFieldNames, typeName }: OneOf
       `".`,
   );
 }
+
+export function listSizeInvalidSlicingArgumentErrorMessage(directiveCoords: string, argumentName: string): string {
+  return ` The "slicingArguments" value "${argumentName}" on "${directiveCoords}" does not reference a defined argument on this field.`;
+}
+
+export function listSizeSlicingArgumentNotIntErrorMessage(
+  directiveCoords: string,
```
Aenimus: Please use the proper type aliases, e.g., `ArgumentName`, `FieldName`, `TypeName`, etc.

composition/src/router-configuration/types.ts
```diff
@@ -93,3 +93,32 @@ export type ConfigurationData = {
...
+  // Key: "TypeName.fieldName" (lookup handle for O(1) merge during normalization)
+  fieldWeights: Map<FieldCoords, FieldWeightConfiguration>;
+
+  // Key: "TypeName.fieldName" (lookup handle for O(1) merge during normalization)
+  listSizes: Map<FieldCoords, FieldListSizeConfiguration>;
+
+  typeWeights: { [key: string]: number };
+
+  // Key: "directiveName.argumentName"
+  directiveArgumentWeights?: { [key: string]: number };
```
Aenimus: Make a type alias for `{DirectiveName}.{ArgumentName}`. You could call it DirectiveArgumentCoords.

composition/src/subgraph/types.ts
```diff
@@ -28,6 +28,7 @@ export type SubgraphConfig = {
   parentDefinitionDataByTypeName: Map<TypeName, ParentDefinitionData>;
   schema: GraphQLSchema;
   schemaNode?: SchemaDefinitionNode | SchemaExtensionNode;
+  costs?: Costs;
```
Aenimus, line 32: Please place properties alphabetically, followed by alphabetical optional properties.

composition/src/subgraph/types.ts
```diff
@@ -45,5 +46,6 @@ export type InternalSubgraph = {
   persistedDirectiveDefinitionDataByDirectiveName: Map<DirectiveName, PersistedDirectiveDefinitionData>;
   schema: GraphQLSchema;
   schemaNode?: SchemaDefinitionNode | SchemaExtensionNode;
+  costs?: Costs;
   url: string;
```
Aenimus, line 50: Please place properties alphabetically, followed by alphabetical optional properties.

Please also fix `url` in the process.

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -394,6 +418,7 @@ export class NormalizationFactory {
   concreteTypeNamesByAbstractTypeName = new Map<string, Set<string>>();
   conditionalFieldDataByCoords = new Map<string, ConditionalFieldData>();
   configurationDataByTypeName = new Map<TypeName, ConfigurationData>();
+  costs: Costs = { fieldWeights: new Map(), listSizes: new Map(), typeWeights: {} };
```
Aenimus: Please can you type this object externally? Then the property maps will also be types. And `typeWeights` can be an explicitly typed `Record<>`

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -1091,13 +1129,40 @@ export class NormalizationFactory {
   }
 
   // returns true if the directive is custom; otherwise, false
+  extractDirectiveArgumentCosts(node: DirectiveDefinitionNode) {
```
Aenimus, line 1154: The veryfirst thing you should check is whether directive is the cost directive. Then you can return early if it's not; and if it is, you never need to check it again. Currently, you check it in every loop iteration.

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -1091,13 +1129,40 @@ export class NormalizationFactory {
...
+      if (!argNode.directives) {
+        continue;
+      }
+      for (const directive of argNode.directives) {
+        if (directive.name.value !== COST) {
+          continue;
+        }
+        const weightArg = directive.arguments?.find((a) => a.name.value === WEIGHT);
+        if (!weightArg || weightArg.value.kind !== Kind.INT) {
+          continue;
```
Aenimus, line 1178: I believe argument validation has already been handled earlier, right?

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -2314,6 +2379,212 @@ export class NormalizationFactory {
...
+      case Kind.OBJECT_TYPE_DEFINITION:
+      case Kind.SCALAR_TYPE_DEFINITION:
+      case Kind.ENUM_TYPE_DEFINITION:
+        this.costs.typeWeights[data.name] = weightValue;
+        break;
+      case Kind.FIELD_DEFINITION: {
+        const typeName = data.renamedParentTypeName || data.originalParentTypeName;
+        const parentTypeData = this.parentDefinitionDataByTypeName.get(typeName);
+        if (parentTypeData?.kind === Kind.INTERFACE_TYPE_DEFINITION) {
+          errorMessages.push(costOnInterfaceFieldErrorMessage(directiveCoords));
```
Aenimus, line 2431: The parent node not existing and the parent node being an interface are two separate and entirely unrelated errors.

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -2314,6 +2379,212 @@ export class NormalizationFactory {
...
+        break;
+      }
+      case Kind.INPUT_VALUE_DEFINITION:
+      case Kind.ARGUMENT: {
+        const ivData = data as InputValueData;
+        if (ivData.isArgument && ivData.fieldName) {
+          const typeName = ivData.renamedParentTypeName || ivData.originalParentTypeName;
+          const parentTypeData = this.parentDefinitionDataByTypeName.get(typeName);
+          if (parentTypeData?.kind === Kind.INTERFACE_TYPE_DEFINITION) {
+            errorMessages.push(costOnInterfaceFieldErrorMessage(directiveCoords));
```
Aenimus, line 2458: See comment about interfaces above

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -2314,6 +2379,212 @@ export class NormalizationFactory {
...
+    };
+
+    for (const argumentNode of args) {
+      const argumentName = argumentNode.name.value;
+
+      // type validation handled upstream
+      if (argumentName === ASSUMED_SIZE && argumentNode.value.kind === Kind.INT) {
+        listSizeConfig.assumedSize = parseInt((argumentNode.value as IntValueNode).value, 10);
+      }
+      if (argumentName === REQUIRE_ONE_SLICING_ARGUMENT && argumentNode.value.kind === Kind.BOOLEAN) {
```
Aenimus: You're falling through here, even though you've already handled the argument. It can't have more than one name at once.

Please consider a switch.

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -2314,6 +2379,212 @@ export class NormalizationFactory {
...
+        listSizeConfig.requireOneSlicingArgument = argumentNode.value.value;
+      }
+
+      if (argumentName === SLICING_ARGUMENTS) {
+        let stringValues: readonly ValueNode[];
```
Aenimus: Please use `Array<ValueNode>`

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -2314,6 +2379,212 @@ export class NormalizationFactory {
...
+      } else {
+        for (const slicingArgName of listSizeConfig.slicingArguments) {
+          const argData = data.argumentDataByName.get(slicingArgName);
+          if (argData?.defaultValue) {
+            errorMessages.push(listSizeAssumedSizeSlicingArgDefaultErrorMessage(directiveCoords, slicingArgName));
+          }
+        }
+      }
+    }
+    const fieldCoord = `${typeName}.${data.name}`;
```
Aenimus: fieldCoords (plural)

composition/src/router-configuration/types.ts
```diff
@@ -93,3 +93,32 @@ export type ConfigurationData = {
...
+  fieldName: FieldName;
+  weight?: number;
+  argumentWeights?: { [key: string]: number };
+};
+
+export type FieldListSizeConfiguration = {
+  typeName: TypeName;
+  fieldName: FieldName;
+  assumedSize?: number;
+  slicingArguments?: string[];
```
Aenimus: `Array<string>`

composition/src/router-configuration/types.ts
```diff
@@ -93,3 +93,32 @@ export type ConfigurationData = {
...
+  sizedFields?: FieldName[];
```
Aenimus: `Array<FieldName>`

composition/src/errors/errors.ts
ysmolski: Done

composition/src/router-configuration/types.ts
ysmolski: done

composition/src/router-configuration/types.ts
ysmolski: done

composition/src/router-configuration/types.ts
ysmolski: done

composition/src/subgraph/types.ts
ysmolski: done

composition/src/subgraph/types.ts
ysmolski: done

composition/src/v1/normalization/normalization-factory.ts
ysmolski: This method had wrong comment on the top. I have fixed it. And we have to iterate like that because cost can be attached to any argument of the directive. I think the current double loop is mandatory

composition/src/v1/normalization/normalization-factory.ts
ysmolski: Yeah, it happens on the layer above, that validates directives types. Should I remove the check for Kind.INT ? I saw the same pattern of safety checks in other places (existing code)...

composition/src/v1/normalization/normalization-factory.ts
ysmolski: fixed

composition/src/v1/normalization/normalization-factory.ts
ysmolski: Here we only care if the parent node is an interface. Should we do a special check if the parent node is not existing here?

composition/src/v1/normalization/normalization-factory.ts
ysmolski: fixed to switch

composition/src/v1/normalization/normalization-factory.ts
ysmolski: used in both places

composition/src/v1/normalization/normalization-factory.ts
ysmolski: To be clarified in other channel.

composition/src/v1/normalization/normalization-factory.ts
ysmolski: I got it. In other place we have a protecting check that breaks early if the parent is not defined. I am adding the same two checks here.

composition/src/v1/normalization/normalization-factory.ts
ysmolski: fixed

composition/src/router-configuration/types.ts
```diff
@@ -93,3 +99,32 @@ export type ConfigurationData = {
...
 };
+
+export type Costs = {
+  // Key: "TypeName.fieldName" (lookup handle for O(1) merge during normalization)
+  fieldWeights: Map<FieldCoords, FieldWeightConfiguration>;
+
+  // Key: "TypeName.fieldName" (lookup handle for O(1) merge during normalization)
+  listSizes: Map<FieldCoords, FieldListSizeConfiguration>;
+
+  typeWeights: { [key: TypeName]: number };
```
Aenimus: Please type this as `Record<TypeName, number>`

composition/src/router-configuration/types.ts
```diff
@@ -93,3 +99,32 @@ export type ConfigurationData = {
...
+
+  // Key: "directiveName.argumentName"
+  directiveArgumentWeights?: { [key: DirectiveArgumentCoords]: number };
```
Aenimus: Please type this as `Record<DirectiveArgumentCoords, number>`

composition/src/router-configuration/types.ts
```diff
@@ -93,3 +99,32 @@ export type ConfigurationData = {
...
+};
+
+export type FieldWeightConfiguration = {
+  typeName: TypeName;
+  fieldName: FieldName;
+  weight?: number;
+  argumentWeights?: { [key: ArgumentName]: number };
```
Aenimus: Please type this as `Record<ArgumentName, number>`

composition/src/subgraph/types.ts
Aenimus: Thanks!

composition/src/subgraph/types.ts
Aenimus: Thanks!

composition/src/types/types.ts
```diff
@@ -1,5 +1,10 @@
+export type ArgumentName = string;
+
 export type ContractName = string;
 
+// {DirectiveName}.{ArgumentName}
+export type DirectiveArgumentCoords = string;
```
Aenimus, line 6: Please type this as `${DirectiveName}.${ArgumentName}` rather than string

shared/src/router-config/builder.ts
```diff
@@ -276,6 +302,7 @@ export const buildRouterConfig = function (input: Input): RouterConfig {
       // https://github.com/wundergraph/cosmo/blob/main/router/core/router.go#L342
       id: subgraph.id,
       childNodes,
+      costConfiguration: subgraph.costs ? costsToCostConfiguration(subgraph.costs) : undefined,
```
Aenimus: Under what circumstances would we not have costs? It's always returned from normalization. Where does it become possibly undefined?

composition/src/v1/normalization/normalization-factory.ts
Aenimus: My mistake, sorry.

composition/src/router-configuration/types.ts
ysmolski: done

composition/src/router-configuration/types.ts
ysmolski: done

composition/src/router-configuration/types.ts
ysmolski: done

composition/src/v1/normalization/normalization-factory.ts
Aenimus: Please add a comment that validation is conducted upstream.

The problem is that validation there can fail, but I believe it would still enter this code(?). So, I think we still have to check the kind to avoid a errors when attempting to parse int.

I could be wrong, and if we never enter this code with an invalid directive, you can remove the check.

composition/src/types/types.ts
ysmolski: Do you mean this:
```
export type DirectiveArgumentCoords = `${DirectiveName}.${ArgumentName}`;
```
https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html

Not sure how to use them because it breaks everything around:
```
src/v1/normalization/normalization-factory.ts:709:67 - error TS2345: Argument of type 'string' is not assignable to parameter of type '`${string}.${string}`'.

709           listSizeFieldMustReturnListOrUseSizedFieldsErrorMessage(directiveCoords, printTypeNode(data.type)),
                                                                      ~~~~~~~~~~~~~~~

src/v1/normalization/normalization-factory.ts:2420:63 - error TS2345: Argument of type 'string' is not assignable to parameter of type '`${string}.${string}`'.

2420           errorMessages.push(costOnInterfaceFieldErrorMessage(directiveCoords));
                                                                   ~~~~~~~~~~~~~~~

src/v1/normalization/normalization-factory.ts:2446:65 - error TS2345: Argument of type 'string' is not assignable to parameter of type '`${string}.${string}`'.

2446             errorMessages.push(costOnInterfaceFieldErrorMessage(directiveCoords));
                                                                     ~~~~~~~~~~~~~~~

src/v1/normalization/normalization-factory.ts:2527:77 - error TS2345: Argument of type 'string' is not assignable to parameter of type '`${string}.${string}`'.

2527               errorMessages.push(listSizeInvalidSlicingArgumentErrorMessage(directiveCoords, slicingArgName));
                                                                                 ~~~~~~~~~~~~~~~

src/v1/normalization/normalization-factory.ts:2534:59 - error TS2345: Argument of type 'string' is not assignable to parameter of type '`${string}.${string}`'.

2534                 listSizeSlicingArgumentNotIntErrorMessage(directiveCoords, slicingArgName, printTypeNode(argData.type)),
                                                               ~~~~~~~~~~~~~~~

src/v1/normalization/normalization-factory.ts:2562:81 - error TS2345: Argument of type 'string' is not assignable to parameter of type '`${string}.${string}`'.

2562             errorMessages.push(listSizeSizedFieldsInvalidReturnTypeErrorMessage(directiveCoords, returnTypeName));
                                                                                     ~~~~~~~~~~~~~~~
```

composition/src/v1/normalization/normalization-factory.ts
Aenimus: I didn't realise that you're checking `@cost` directives on the arguments of other directives; I apologise.

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -737,6 +769,12 @@ export class NormalizationFactory {
         requiredScopes,
       });
     }
+    if (isCost) {
+      this.handleCostDirective({ data, directiveCoords, directiveNode, errorMessages });
+    }
+    if (isListSize && isField) {
```
Aenimus: This should be an else if, because the `directiveName` cannot be two things at once, right?

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -1106,6 +1144,34 @@ export class NormalizationFactory {
...
+        if (directive.name.value !== COST) {
+          continue;
+        }
+        const weightArg = directive.arguments?.find((a) => a.name.value === WEIGHT);
+        if (!weightArg || weightArg.value.kind !== Kind.INT) {
+          continue;
+        }
+        const weightValue = parseInt((weightArg.value as IntValueNode).value, 10);
+        if (!this.costs.directiveArgumentWeights) {
+          this.costs.directiveArgumentWeights = {};
```
Aenimus: Maybe you could just use a map so that it always exists (created at initialization of the parent), and then you can just set a new value any time. You'd then change the logic in the builder to check the map has a size > 0, rather than checking `Object.keys`.

Aenimus: And if you want to stick with a `Record`, does it need to be possibly undefined?

composition/src/v1/normalization/normalization-factory.ts
Aenimus: Thanks!

composition/src/v1/normalization/normalization-factory.ts
Aenimus: Thanks!

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -2330,6 +2397,227 @@ export class NormalizationFactory {
...
+          const fieldWeight = getValueOrDefault(
+            this.costs.fieldWeights,
+            parentFieldCoord,
+            (): FieldWeightConfiguration => ({
+              typeName,
+              fieldName: ivData.fieldName!,
+            }),
+          );
+          if (!fieldWeight.argumentWeights) {
+            fieldWeight.argumentWeights = {};
```
Aenimus: Same comment here about Maps/possibly undefined Records

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -2330,6 +2397,227 @@ export class NormalizationFactory {
...
+          }
+          fieldWeight.argumentWeights[ivData.name] = weightValue;
+        } else {
+          const typeName = ivData.renamedParentTypeName || ivData.originalParentTypeName;
+          const fieldCoord = `${typeName}.${ivData.name}`;
```
Aenimus: ntpick: should be `fieldCoords` (plural)

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -2330,6 +2397,227 @@ export class NormalizationFactory {
...
+        case SLICING_ARGUMENTS: {
+          let argumentValues: ReadonlyArray<ValueNode>;
+          if (argumentNode.value.kind === Kind.LIST) {
+            argumentValues = (argumentNode.value as ListValueNode).values;
+          } else if (argumentNode.value.kind === Kind.STRING) {
+            argumentValues = [argumentNode.value];
+          } else {
+            continue;
+          }
+          if (!listSizeConfig.slicingArguments) {
```
Aenimus: Does this need to be possibly undefined?

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -2330,6 +2397,227 @@ export class NormalizationFactory {
...
+          } else if (argumentNode.value.kind === Kind.STRING) {
+            fieldValues = [argumentNode.value];
+          } else {
+            continue;
+          }
+          if (fieldValues.length < 1) {
+            continue;
+          }
+          hasSizedFields = true;
+          if (!listSizeConfig.sizedFields) {
```
Aenimus: Does this need to be possibly undefined?

shared/src/router-config/builder.ts
ysmolski: I think it originates from here:
```
function constructRouterSubgraph(result: FederationSuccess, s: SubgraphMetadata, index: number): RouterSubgraph {
  const subgraphConfig = result.subgraphConfigBySubgraphName.get(s.name);
  const schema = subgraphConfig?.schema;
  const configurationDataByTypeName = subgraphConfig?.configurationDataByTypeName;
  const costs = subgraphConfig?.costs;
```

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -2330,6 +2397,227 @@ export class NormalizationFactory {
...
+
+    if (!hasSizedFields && !isTypeNodeListType(data.type)) {
+      errorMessages.push(
+        listSizeFieldMustReturnListOrUseSizedFieldsErrorMessage(directiveCoords, printTypeNode(data.type)),
+      );
+    }
+
+    if (
+      listSizeConfig.assumedSize !== undefined &&
+      listSizeConfig.slicingArguments &&
```
Aenimus: For `assumedSize` you're asserting it's not undefined, but for `slicingArguments` you're coercing the value. I believe you could shorten it to:
`if (listSizeConfig.assumedSize && listSizeConfig.slicingArguments?.length > 0)`

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -2330,6 +2397,227 @@ export class NormalizationFactory {
...
+      listSizeConfig.slicingArguments.length > 0
+    ) {
+      if (listSizeConfig.requireOneSlicingArgument !== false) {
```
Aenimus: Don't need the `!== false`

shared/src/router-config/builder.ts
ysmolski: Also I have realized that we need to do the same in the controlplane. I have added it.

composition/src/errors/errors.ts
```diff
@@ -1698,3 +1704,89 @@ export function oneOfRequiredFieldsError({ requiredFieldNames, typeName }: OneOf
...
+  return (
+    ` The "@listSize" directive on "${directiveCoords}" defines both "assumedSize" and "slicingArguments",` +
+    ` but slicing argument "${slicingArgName}" has a default value. When "assumedSize" is used as a fallback` +
+    ` for missing slicing arguments, none of the slicing arguments may have default values.`
+  );
+}
+
+export function costOnInterfaceFieldErrorMessage(directiveCoords: DirectiveArgumentCoords): string {
+  return (
+    ` The "@cost" directive at "${directiveCoords}" is not permitted on fields or arguments of an interface type.` +
```
Aenimus: nitpick: capitalization of interface -> Interface

Aenimus: (3 instances)

composition/src/v1/normalization/normalization-factory.ts
ysmolski: yes, fixed

composition/tests/v1/directives/listSize.test.ts
```diff
@@ -0,0 +1,904 @@
...
+        ),
+      );
+    });
+  });
+
+  describe('validation tests', () => {
+    test('that @listSize with invalid slicingArguments produces an error', () => {
+      const { errors } = normalizeSubgraphFailure(subgraphWithInvalidSlicingArg, ROUTER_COMPATIBILITY_VERSION_ONE);
+      expect(errors).toHaveLength(1);
+      expect(errors[0].message).toContain('does not reference a defined argument');
```
Aenimus: Is there a reason we are not doing strict error matching (see all other tests)? if you change the text of the error message, these tests will fail.

composition/tests/v1/directives/cost.test.ts
```diff
@@ -0,0 +1,936 @@
...
+        expect(costsA?.fieldWeights.get('User.age')).toBeUndefined();
+      });
+    });
+  });
+
+  describe('validation tests', () => {
+    test('that @cost with string weight produces an error', () => {
+      const { errors } = normalizeSubgraphFailure(subgraphWithStringCostWeight, ROUTER_COMPATIBILITY_VERSION_ONE);
+      expect(errors).toHaveLength(1);
+      expect(errors[0].message).toContain('Int!');
```
Aenimus: Same question regarding strict error matching as in the other file.

shared/src/router-config/builder.ts
ysmolski: So I guess it boils down to this call `const subgraphConfig = result.subgraphConfigBySubgraphName.get(s.name);` 
When can it be undefined?

Aenimus: Yeah, that tracks.

composition/src/v1/normalization/types.ts
```diff
@@ -68,6 +68,20 @@ export type HandleSemanticNonNullDirectiveParams = {
   errorMessages: Array<string>;
 };
 
+export type HandleCostDirectiveParams = {
+  data: NodeData | SchemaData;
+  directiveCoords: string;
```
Aenimus: The error you got from changing `DirectiveCoords` comes from here, because this needs to be type `DirectiveCoords`

composition/src/v1/normalization/types.ts
```diff
@@ -68,6 +68,20 @@ export type HandleSemanticNonNullDirectiveParams = {
...
+  directiveNode: ConstDirectiveNode;
+  errorMessages: Array<string>;
+};
+
+export type HandleListSizeDirectiveParams = {
+  data: FieldData;
+  directiveCoords: string;
```
Aenimus: This also needs to be `DirectiveCoords`

composition/src/types/types.ts
Aenimus: Yes, this is what I mean. Please see the other conversation about `normalization/types` on how to rectify this.

router/pkg/config/config.go
```diff
@@ -110,6 +115,7 @@ type Prometheus struct {
 	ConnectionStats     bool        `yaml:"connection_stats" envDefault:"false" env:"PROMETHEUS_CONNECTION_STATS"`
 	Streams             bool        `yaml:"streams" envDefault:"false" env:"PROMETHEUS_STREAM"`
 	EngineStats         EngineStats `yaml:"engine_stats" envPrefix:"PROMETHEUS_"`
+	CostStats           CostStats   `yaml:"cost_stats" envPrefix:"PROMETHEUS_"`
```
endigma: These prefixes can include `COST_STATS_` as well

router/pkg/config/config.go
```diff
@@ -460,6 +471,7 @@ type SecurityConfiguration struct {
 	BlockPersistedOperations    BlockOperationConfiguration `yaml:"block_persisted_operations" envPrefix:"SECURITY_BLOCK_PERSISTED_OPERATIONS_"`
 	ComplexityCalculationCache  *ComplexityCalculationCache `yaml:"complexity_calculation_cache"`
 	ComplexityLimits            *ComplexityLimits           `yaml:"complexity_limits"`
+	CostControl                 *CostControl                `yaml:"cost_control" envPrefix:"SECURITY_COST_CONTROL_"`
```
endigma, line 474: Can you add all the new options to `full.yaml`?

router/core/router.go
```diff
@@ -525,6 +526,18 @@ func NewRouter(opts ...Option) (*Router, error) {
...
+	if ca := r.securityConfiguration.CostControl; ca != nil && ca.Enabled {
+		if ca.Mode != config.CostControlModeMeasure && ca.Mode != config.CostControlModeEnforce {
+			return nil, fmt.Errorf("cost control mode %q is invalid. Valid values are %q and %q", ca.Mode, config.CostControlModeMeasure, config.CostControlModeEnforce)
+		}
+		if ca.Mode == config.CostControlModeEnforce && ca.MaxEstimatedLimit <= 0 {
+			return nil, errors.New("cost control mode is 'enforce' but 'max_estimated_limit' is not set. Please provide a positive value for 'security.cost_control.max_estimated_limit'")
+		}
+		if ca.EstimatedListSize <= 0 {
+			return nil, errors.New("cost control is enabled but 'estimated_list_size' is not set. Please provide a positive value for 'security.cost_control.estimated_list_size'")
+		}
```
endigma: This should be enforced in schema as well

router-tests/go.mod
```diff
@@ -24,7 +24,7 @@ require (
 	github.com/twmb/franz-go v1.16.1
 	github.com/twmb/franz-go/pkg/kadm v1.11.0
 	github.com/wundergraph/astjson v1.1.0
-	github.com/wundergraph/cosmo/demo v0.0.0-20260213130455-6e3277e7b850
+	github.com/wundergraph/cosmo/demo v0.0.0-00010101000000-000000000000
```
endigma: This looks wrong

router-tests/integration_test.go
```diff
@@ -1541,6 +1541,72 @@ func BenchmarkPb(b *testing.B) {
...
+		b.ReportAllocs()
+		b.ResetTimer()
+		for ii := 0; ii < b.N; ii++ {
+			res := xEnv.MakeGraphQLRequestOK(testenv.GraphQLRequest{
+				Query: bigEmployeesQuery,
+			})
+			if len(res.Body) < 3000 {
+				b.Errorf("unexpected result %q, expecting \n\n%q", res.Body, bigEmployeesResponse)
+			}
+		}
```
endigma: Should use `b.Loop`

router/core/factoryresolver.go
```diff
@@ -665,6 +678,50 @@ func (l *Loader) dataSourceMetaData(in *nodev1.DataSourceConfiguration) *plan.Da
...
+	for _, ls := range costConfig.GetListSizes() {
+		listSizes := &plan.FieldListSize{
+			SlicingArguments: ls.GetSlicingArguments(),
+			SizedFields:      ls.GetSizedFields(),
+		}
+		if ls.AssumedSize != nil {
+			listSizes.AssumedSize = int(*ls.AssumedSize)
+		} else {
+			listSizes.AssumedSize = 0
+		}
```
endigma: The default value of this is probably 0 anyways, meaning you can remove the else here

composition/src/v1/normalization/normalization-factory.ts
ysmolski: fixed

composition/src/errors/errors.ts
ysmolski: fixed

composition/src/v1/normalization/types.ts
ysmolski: fixed

composition/src/v1/normalization/normalization-factory.ts
ysmolski: I cannot do that because requireOneSlicingArgument is nullable.

composition/src/types/types.ts
ysmolski: Well, I could go there, but that would be too much of refactoring to take at this point. I would have to change this function signature structure:
```ts
  validateDirective({
    data,
    definitionData,
    directiveCoords,
    directiveNode,
    errorMessages,
    requiredArgumentNames,
  }: ValidateDirectiveParams): Array<string> {
```
and this:
```ts
  validateDirectives(data: NodeData | SchemaData, directiveCoords: string) {
    const undefinedDirectiveNames = new Set<string>();
```
And this is being used in many other parts besides the cost algorithm. I don't think it is a proper place to do this refactoring here.

composition/src/v1/normalization/normalization-factory.ts
ysmolski: I have replaced them with maps in the Costs type and create those maps with the object.

composition/src/v1/normalization/normalization-factory.ts
ysmolski: Ok, converted to map too

composition/src/v1/normalization/normalization-factory.ts
ysmolski: no need. empty array works. fixed

composition/src/v1/normalization/normalization-factory.ts
ysmolski: fixed

composition/src/v1/normalization/normalization-factory.ts
ysmolski: fixed to ` if (listSizeConfig.assumedSize !== undefined && listSizeConfig.slicingArguments.length > 0) {`

composition/tests/v1/directives/listSize.test.ts
ysmolski: improved messages in that file

composition/tests/v1/directives/cost.test.ts
ysmolski: improved messages in that file

composition/src/v1/normalization/types.ts
ysmolski: Not only this place. Some functions are too. See https://github.com/wundergraph/cosmo/pull/2470#discussion_r2922991412

shared/src/router-config/builder.ts
```diff
@@ -276,6 +303,7 @@ export const buildRouterConfig = function (input: Input): RouterConfig {
...
```
Aenimus: What if `costsToCostConfiguration` accepted an undefined input and immediately returned undefined, then everything is handled in the function?

Aenimus: The signature can already return undefined, so it seems fine to me.

composition/src/v1/normalization/normalization-factory.ts
Aenimus: @ysmolski making sure you've seen this.

composition/src/v1/normalization/normalization-factory.ts
Aenimus: Do you mean it has to explicitly be false (not undefined [in GraphQl's case null), else it's an error?

composition/src/types/types.ts
Aenimus: Understood; if it starts creeping out of your scope like this, let's leave it alone.

composition/src/v1/normalization/normalization-factory.ts
Aenimus: Cool.

composition/tests/v1/directives/cost.test.ts
Aenimus: I still see a lot of `.toContain` in the tests. If you look at the other tests, we always use `.toStrictEqual(<Error>)`

composition/src/v1/normalization/types.ts
Aenimus: Understood. Let's not address it here.

composition/src/v1/normalization/normalization-factory.ts
ysmolski: yeah, requireOneSlicingArgument not defined means TRUE in the spec. I could simplify this by removing undefined possibility and assigning true to it in the constructor.

ysmolski: it worked. done

composition/tests/v1/directives/listSize.test.ts
ysmolski: used toStrictEqual in all tests

composition/tests/v1/directives/cost.test.ts
ysmolski: used toStrictEqual in all tests

composition/src/v1/normalization/normalization-factory.ts
ysmolski: Oh. I see now. GH hides that comment, even unresolved under the button.
This check was not needed for cost weight on the field itself. But here, it is on the argument of the directive. And there is not check upstream for complience with INT.  I have added corresponding check along with a test.
Also I have added a test for the weight on the argument of the field.

shared/src/router-config/builder.ts
ysmolski: I guess nothing to be done here

composition/src/router-configuration/types.ts
```diff
@@ -93,3 +99,26 @@ export type ConfigurationData = {
   requireFetchReasonsFieldNames?: Array<FieldName>;
   requires?: RequiredFieldConfiguration[];
 };
+
```
Aenimus, line 102: nitpick: can you alphabetize the properties in all these types, ensuring the optional properties are at the bottom (also alphabetized)

ysmolski: done

shared/src/router-config/builder.ts
ysmolski: ternary op was removed

composition/src/router-configuration/types.ts
Aenimus: Thanks!

composition/src/v1/normalization/normalization-factory.ts
Aenimus: Please let me know the test that slipped by, and I'll look into it. Thanks!

router-tests/go.mod
ysmolski: fixed

router-tests/integration_test.go
ysmolski: fixed and in other benchms too

router/pkg/config/config.go
ysmolski: fixed

router/pkg/config/config.go
ysmolski: fixed for metrics

router/core/factoryresolver.go
ysmolski: removed, thanks!

ysmolski:
> Can you also add cost to traces and add a test for it? Do we have cost for individual subgraph calls or only the overall one as well? If we can break it down it could be nice to have it per-fetch span if possible?

@endigma Traces - we could do that, but I would leave it for later. No customer requested it yet. So we need to release ASAP and add features later.

Costs are not calculated per subgraph yet. At least we do not track that kind of stuff, but it is possible to do in future in the future. That will allow to see costs per fetch.

router/core/router.go
ysmolski: Addressing it atm

ysmolski: Done!

router/pkg/metric/metric_store.go
```diff
@@ -75,6 +79,25 @@ var (
 		otelmetric.WithDescription(OperationPlanningTimeHistogramDescription),
 	}
 
+	costBucketBounds = []float64{
+		0, 5, 10,
```
StarpTech: How confident are you with these numbers? Changing them would be a breaking change, and we will work with them in Cosmo. I also recommend reducing them to 7-8 for fewer cardinality.

```
Bucket 1: cost 0 to 10
Bucket 2: cost 11 to 50
Bucket 3: cost 51 to 200
Bucket 4: cost 201 to 1,000
Bucket 5: cost 1,001 to 5,000
Bucket 6: cost greater than 5,001 to 10,000
Bucket 7: cost greater than 10,000
```

router/core/websocket.go
```diff
@@ -958,6 +958,10 @@ func (h *WebSocketConnectionHandler) parseAndPlan(registration *SubscriptionRegi
 
 	opContext.planningTime = time.Since(startPlanning)
 
+	if err := operationKit.ValidateStaticCost(opContext); err != nil {
```
StarpTech, line 961: Do we have tests for ws?

router/pkg/metric/metric_store.go
ysmolski: Sure, it makes sense to have reduced granularity. If a user needs to know exact numbers he could use headers or custom modules for super accurate tracing. Fixed.

router/core/websocket.go
ysmolski: I have added tests for WS. There is some issue with the complete frames, so tests do not verify those in case of errors. But otherwise both queries and subs are tested.

PR #2472, opened 2026-01-27
feat: client response expressions

SkArchon:
This PR allows for the use of response client expressions, currently we only support header propagation from the subgraph level. Even for cases like setting a static value like "test" for a header on responses, this would run for every subgraph. The problem with this is that the requestContext's expression context has not been updated with information such as errors, response headers, etc. As this is processed after the subgraph level round trippers. (The use case was that the user wants to access the `error` attribute in expressions, which is not populated at this level).

In order to mitigate this we introduce client expressions, which runs once before returning the response.

## Summary by CodeRabbit

* **New Features**
  * Router-level response header rules: set or derive response headers via static values or expressions. Empty results are ignored; evaluation errors are logged without failing requests. Configuration, schema, defaults, and example fixtures updated to expose these rules.

* **Tests**
  * Added comprehensive tests covering static/dynamic expressions, multiple headers, precedence with other header rules, empty/invalid cases, error handling, and logging.

## Checklist


router/pkg/config/config.go
```diff
@@ -349,6 +356,8 @@ type ResponseHeaderRule struct {
 	Name string `yaml:"name"`
 	// Value is the value of the header to set
 	Value string `yaml:"value"`
+
+	Expression string `yaml:"expression"`
```
endigma: This field needs to be added to config schema? And a test? Should it be exclusive with `value` or dependent on `op: set`?

router/core/header_rule_engine.go
```diff
@@ -599,7 +615,22 @@ func (h *HeaderPropagation) getRequestRuleExpressionValue(rule *config.RequestHe
...
 		return "", fmt.Errorf("context cannot be nil")
 	}
-	program, ok := h.compiledRules[rule.Expression]
+	program, ok := h.compiledRequestRules[rule.Expression]
+	if !ok {
+		return "", fmt.Errorf("expression %s not found in compiled rules for header rule %s", rule.Expression, rule.Name)
+	}
+	value, err = expr.ResolveStringExpression(program, reqCtx.expressionContext)
+	if err != nil {
+		return "", fmt.Errorf("unable to resolve expression %q for header rule %s: %s", rule.Expression, rule.Name, err.Error())
```
endigma: Use `%w` for formatting errors as part of another errorf

router/core/header_rule_engine_test.go
```diff
@@ -1056,3 +1056,182 @@ func TestExpression(t *testing.T) {
...
+		reqCtx := &requestContext{}
+		clientCtx := withRequestContext(context.Background(), reqCtx)
+		clientReq, err := http.NewRequestWithContext(clientCtx, "POST", "http://localhost", nil)
+		require.NoError(t, err)
+		reqCtx.expressionContext = expr.Context{Request: expr.LoadRequest(clientReq)}
+		reqCtx.logger = zap.NewNop()
+		reqCtx.responseWriter = rr
+		reqCtx.request = clientReq
+		reqCtx.operation = &operationContext{}
+		reqCtx.subgraphResolver = NewSubgraphResolver([]Subgraph{})
```
endigma: Nit: Can we initialize some of this as part of the struct?

router/pkg/config/config.go
SkArchon: This was old code, previously expression was here, should have been cleaned up.

router/core/header_rule_engine.go
SkArchon: Updated

router/core/header_rule_engine_test.go
SkArchon: Updated

router/core/header_rule_engine.go
```diff
@@ -669,6 +700,24 @@ func createMostRestrictivePolicy(policies []*cachedirective.Object) (*cachedirec
...
+		if err != nil {
+			return fmt.Errorf("failed to evaluate client response header expression for %s: %w", rule.Name, err)
+		}
+		if value != "" {
+			w.Header().Set(rule.Name, value)
+		}
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
SkArchon: There is a nil check on the call site.

endigma (approved):

PR #2485, opened 2026-02-02
feat: sanitize utf8 strings with binary in spans

SkArchon:
Sometimes spans can contain bad strings which has non utf8 characters, when they are serialized into protobuf for transport to collectors, protobuf throws an error because it cannot serialize non utf8 characters.

This means that users will loose these spans, in order to handle this we need to sanitize strings. We do this two ways.

For statuses and names, we simply by default opt to sanitize the string. However as there are a lot of attributes, we allow the user to specify if they want to sanitize the span. In order to do this we refactor the current `OnEnd` function which has redact and add refactor `redact` to its own transformer and add `sanitizeutf8` as another transformer.

## Summary by CodeRabbit

* **New Features**
  * Configurable attribute-processing pipeline for traces.
  * Optional UTF‑8 sanitization for span attributes with toggleable logging.
  * IP anonymization for traces with selectable redact or hash modes.
  * Span error/status descriptions are sanitized before recording.

* **Configuration**
  * New tracing options to enable UTF‑8 sanitization and IP anonymization.

* **Tests**
  * Extensive unit and integration tests for attribute processing, sanitization, anonymization, and status sanitization.

## Checklist

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -28,3 +28,9 @@
 /studio/ @JivusAyrus @wilsonrivera @StarpTech @thisisnithin
 # Cosmo Streams / EDFS
 /router/pkg/pubsub @alepane21 @StarpTech @jensneuse @dkorittki
+
+# generated protobuf files, sourced from /proto/
+/router/gen/
+/connect-go/gen/
+/graphqlmetrics/gen/
+/connect/src/wg/


StarpTech:
Please create a proper PR description that explains the situation and solution.

router/pkg/config/config.go
```diff
@@ -82,6 +87,9 @@ type Tracing struct {
 	OperationContentAttributes bool `yaml:"operation_content_attributes" envDefault:"false" env:"TRACING_OPERATION_CONTENT_ATTRIBUTES"`
 
 	TracingGlobalFeatures `yaml:",inline"`
+
+	// SanitizeUTF8 configures sanitization of invalid UTF-8 sequences in span attribute values
+	SanitizeUTF8 SanitizeUTF8Config `yaml:"sanitize_utf8"`
```
endigma: should use `envPrefix:"SANITIZE_UTF8_`

router/pkg/trace/attributeprocessor/attributes_test.go
```diff
@@ -0,0 +1,256 @@
...
+
+// testAttributes creates a span with the given attributes at creation time and returns the recorded attributes
+func testAttributes(opt trace.TracerProviderOption, attrs ...attribute.KeyValue) []attribute.KeyValue {
+	r := &attrRecorder{}
+	tp := trace.NewTracerProvider(opt, trace.WithSpanProcessor(r))
+	defer func() { _ = tp.Shutdown(context.Background()) }()
+
+	ctx := context.Background()
+	tracer := tp.Tracer("testAttributes")
+	_, s := tracer.Start(ctx, "span name", api.WithAttributes(attrs...))
```
endigma: Best to use `t.Context()` here, and anywhere else context.Background is used in tests

router/pkg/trace/attributeprocessor/attributes_test.go
```diff
@@ -0,0 +1,256 @@
...
+	_, s := tracer.Start(ctx, "span name")
+	s.SetAttributes(attrs...)
+	s.End()
+	return r.attrs
+}
+
+func TestAttributeProcessor(t *testing.T) {
+	t.Parallel()
+
+	t.Run("NoTransformers", func(t *testing.T) {
```
endigma: can you update the test names to reflect what they assert? e.g. "with no transformers, attributes should remain unchanged" as below in your comment.

router/pkg/trace/attributeprocessor/redact.go
```diff
@@ -0,0 +1,57 @@
...
+// Returns an error if the ipAnonymizationMethod is not supported.
+func RedactKeys(keys []attribute.Key, ipAnonymizationMethod IPAnonymizationMethod) (AttributeTransformer, error) {
+	var rFunc func(attribute.KeyValue) string
+
+	switch ipAnonymizationMethod {
+	case Hash:
+		rFunc = func(key attribute.KeyValue) string {
+			h := sha256.New()
+			h.Write([]byte(key.Value.AsString()))
+			return hex.EncodeToString(h.Sum(nil))
```
endigma, line 36: I believe we have access to a pooled xxhash in `graphql-go-tools/v2/pkg/pool` you could possibly use here for speed?

router/pkg/trace/attributeprocessor/redact_test.go
```diff
@@ -0,0 +1,295 @@
...
+	const key = "password"
+	var (
+		name     = attribute.String("name", "bob")
+		eID      = attribute.Int("employee-id", 9287)
+		passStr  = attribute.String(key, "super-secret-pswd")
+		passBool = attribute.Bool(key, true)
+		replaced = attribute.String(key, "[REDACTED]")
+	)
+
+	t.Run("Empty", func(t *testing.T) {
```
endigma: same thing, more descriptive test names please

router/pkg/trace/attributeprocessor/redact.go
SkArchon: My only concern here is the hashes would be different, just in case folks are referring back to a hash for something. Thoughts?

endigma: Can you see if the performance is much better? If not I'd leave it

SkArchon: Seems like theres a difference of about 90 nanoseconds. Ran it a bunch of times, I would keep it as it is for now imo. Lmk if all good.

```
BenchmarkHashComparison/SHA256/127.0.0.1-10         	10520935	       112.2 ns/op	     160 B/op	       3 allocs/op
BenchmarkHashComparison/XXHash64/127.0.0.1-10       	44858205	        26.96 ns/op	      16 B/op	       1 allocs/op
```

endigma: 90ns is irrelevant, keep it as-is sha256 wise, maybe use a pool for sha256 to reduce the allocations though? pretty sure this is basically free performance, if not, leave it

router/pkg/config/config.go
SkArchon: Done, put `TRACING_SANITIZE_UTF8_`

router/pkg/trace/attributeprocessor/attributes_test.go
SkArchon: Done

router/pkg/trace/attributeprocessor/attributes_test.go
SkArchon: Done

router/pkg/trace/attributeprocessor/redact.go
SkArchon: Will keep it as it is

router/pkg/trace/attributeprocessor/redact_test.go
SkArchon: Done

SkArchon:
@endigma let me know if all good

router/pkg/trace/attributeprocessor/attributes_test.go
```diff
@@ -0,0 +1,246 @@
...
+	ctx := context.Background()
+
+	return func(b *testing.B) {
+		b.ReportAllocs()
+		b.ResetTimer()
+		for i := 0; i < b.N; i++ {
+			ac.OnStart(ctx, s)
+			ac.OnEnd(s)
+		}
+	}
```
endigma, line 244: Should use `b.Loop()`

PR #2489, opened 2026-02-04
fix: show posthog errors only on debug logs and don't treat them as important

alepane21:
PostHog is considered important, when really it is not. With this PR I'm changing this so that:
- if at start the PostHog service is not available, we just skip starting PostHog and log it with an Info level
- messages from PostHog component are logged with debug level

## Summary by CodeRabbit

* **Bug Fixes**
  * Usage-tracking setup no longer blocks application startup; initialization proceeds even if tracking cannot be configured.
  * Uptime and tracking are started asynchronously and safely no-op when disabled or unavailable, avoiding startup blocking.

* **Refactor**
  * Reduced log severity for non-fatal tracking issues to cut log noise.
  * Transient tracking errors are logged instead of surfacing as startup failures.

## Checklist

### DIFF
--- a/router-tests/modules/custom-operation-timings/module.go
+++ b/router-tests/modules/custom-operation-timings/module.go
@@ -1,0 +1,48 @@
+package custom_operation_timings
+
+import (
+	"net/http"
+
+	"github.com/wundergraph/cosmo/router/core"
+)
+
+const myModuleID = "operationTimingsModule"
+
+// OperationTimingsModule is a simple module that reads and logs the operation timings
+type OperationTimingsModule struct {
+	ResultsChan chan core.OperationTimings
+}
+
+func (m *OperationTimingsModule) Middleware(ctx core.RequestContext, next http.Handler) {
+	timings := ctx.Operation().Timings()
+
+	if m.ResultsChan != nil {
+		select {
+		case m.ResultsChan <- timings:
+		default:
+			// drop if nobody is listening to avoid blocking the request path
+		}
+	}
+
+	// Call the next handler in the chain or return early by calling w.Write()
+	next.ServeHTTP(ctx.ResponseWriter(), ctx.Request())
+}
+
+func (m *OperationTimingsModule) Module() core.ModuleInfo {
+	return core.ModuleInfo{
+		// This is the ID of your module, it must be unique
+		ID: myModuleID,
+		// The priority of your module, lower numbers are executed first
+		Priority: 1,
+		New: func() core.Module {
+			return &OperationTimingsModule{
+				ResultsChan: make(chan core.OperationTimings, 1),
+			}
+		},
+	}
+}
+
+// Interface guard
+var (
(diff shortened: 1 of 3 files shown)


router/internal/track/track.go
```diff
@@ -49,20 +48,18 @@ func NewUsageTracker(log *zap.Logger, config UsageTrackerConfig) (*UsageTracker,
...
 		Logger:   tracker.posthogLogger(),
-		Endpoint: "https://eu.i.posthog.com",
+		Endpoint: "https://eu.i.posthoga.com",
 	}
 	tracker.client, err = posthog.NewWithConfig("phc_h2Efq192t8Jz2eW14BDRt3I8Vrs2WMd3oQ4KOpMu3xT", cfg)
 	if err != nil {
-		log.Error("failed to create posthog client", zap.Error(err))
-		return nil, err
+		return nil, fmt.Errorf("failed to create posthog client: %w", err)
 	}
```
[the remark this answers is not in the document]
alepane21: Thanks, should be fixed now!

alepane21:
> **Actionable comments posted: 1**
> 
> Caution
> 
> Some comments are outside the diff and can’t be posted inline due to platform limitations.
> 
> ⚠️ Outside diff range comments (1)
> > router/core/router.go (1)> `1400-1424`: _⚠️ Potential issue_ | _🟠 Major_
> > > **Potential nil-pointer panic in Shutdown if usage tracking fails.**
> > > If `track.NewUsageTracker` returns an error at Line 1418, `configureUsageTracking` returns the error at Line 1420. The caller (`Start` at Line 1206-1208) logs the error but continues. However, `r.usage` remains `nil` because it was never assigned after the failure. Later, `Shutdown` calls `r.usage.Close()` at Line 1577, which will panic on a nil receiver.
> > > Consider setting `r.usage` to a no-op tracker when initialization fails:
> > > 
> > > 🐛 Proposed fix
> > > ```diff
> > >  	r.usage, err = track.NewUsageTracker(r.logger, cfg)
> > >  	if err != nil {
> > > +		r.usage = &UsageTrackerNoOp{}
> > >  		return err
> > >  	}
> > > ```
> 
> 🤖 Fix all issues with AI agents

@coderabbitai But Close isn't already checking about nil receiver?

router/core/router.go
```diff
@@ -1417,13 +1417,16 @@ func (r *Router) configureUsageTracking(ctx context.Context) (err error) {
 	}
 	r.usage, err = track.NewUsageTracker(r.logger, cfg)
 	if err != nil {
-		return fmt.Errorf("failed to create usage tracker: %w", err)
+		return err
```
StarpTech: It is better to keep the function self-contained and ignore the error when calling `configureUsageTracking`. I'd also ensure that we always work with a NoOp so we don't need to do nil checks.

alepane21: You are right, way better with the no-op, thank you

PR #2492, opened 2026-02-05
fix: ensure router header expressions are set even on early return errors

SkArchon:
This PR fixes a bug where we do not process router response header expressions on early returns on errors, some of these errors are as follow
* Invalid queries
* Invalid PO operation
* http errors
etc

In addition, we skip processing the header for early returns of **subscription operation types**, this is as **we currently do not process the router response header expressions when the request is successful for subscriptions**, so we dont process them when they are not successful either.

Due to refactoring code, codecov has highlighted quite a few places which did not have code coverage before, I have added tests for those as well. However a few still remain missing as they seem to be "unreachable", and might need more investigation / be covered by unit tests.

## Summary by CodeRabbit

* **New Features**
  * Exposed a configurable header-propagation option in graph/router setup.

* **Bug Fixes**
  * More consistent error-response header behavior; subscriptions no longer propagate custom headers; improved logging for response write/flush failures.

* **Tests**
  * Added extensive tests for header propagation on error responses, persisted-query/JWT error flows, SSE/multipart/body write failures, and subscription flush errors.

* **Tests (helpers)**
  * Added test scaffolding and modules to simulate failing writers, non-flusher writers, and related token/registration test scenarios.

## Checklist


router/core/batch.go
```diff
@@ -197,7 +197,13 @@ func processBatchError(w http.ResponseWriter, r *http.Request, err error, reques
...
 	}
 
-	writeRequestErrors(r, w, statusCode, []graphqlerrors.RequestError{requestError}, requestLogger)
+	writeRequestErrors(writeRequestErrorsParams{
+		request:       r,
+		writer:        w,
+		statusCode:    statusCode,
+		requestErrors: []graphqlerrors.RequestError{requestError},
+		logger:        requestLogger,
+	})
```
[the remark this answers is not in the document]
SkArchon: This is fine because the reqCtx is not present at this moment, thus no exprContext.

SkArchon:
Moved to draft, as working on codecov changes

router/core/errors.go
```diff
@@ -188,66 +188,85 @@ func propagateSubgraphErrors(ctx *resolve.Context) {
...
 
 	// According to the tests requestContext can be nil (when called from module WriteResponseError)
 	// As such we have coded this condition defensively to be safe
-	requestContext := getRequestContext(r.Context())
+	requestContext := getRequestContext(params.request.Context())
 	isSubscription := requestContext != nil && requestContext.operation != nil && requestContext.operation.opType == "subscription"
 
-	wgRequestParams := NegotiateSubscriptionParams(r, !isSubscription)
+	// We only want to apply header propagation for non-subscription operations
+	// In certain cases the requestContext can be nil, e.g.:- when called from the batch handler
```
StarpTech, line 216: I'm curious why the `requestContext` can be nil. Could you check it?

SkArchon: This is called from the batch handlers also which runs before the reqCtx is created.

StarpTech: ok

PR #2528, opened 2026-02-18
fix(cli): prevent hang when displaying large composition errors

cmtm:
The `router compose` command would hang indefinitely when composition produced errors with long messages (e.g., deeply nested unresolvable field paths that can reach 29,000+ characters).

Root cause: cli-table3's wordWrap option has O(n²) or worse performance on large strings. Error messages of ~2,000 chars took ~7 seconds to render; messages of ~29,000 chars would take minutes or longer.

Fix: Remove table formatting for composition errors and print them directly with simple line separators. Tables are still used for warnings, which are typically much shorter.

## Summary by CodeRabbit

* **Style**
  * Improved composition error display format—errors now appear individually with separator lines instead of in a table layout, providing clearer readability while maintaining the same error handling behavior.

### DIFF
--- a/cli/src/commands/router/commands/compose.ts
+++ b/cli/src/commands/router/commands/compose.ts
@@ -211,19 +211,14 @@ export default (opts: BaseCommandOptions) => {
     );
 
     if (!result.success) {
-      const compositionErrorsTable = new Table({
-        head: [pc.bold(pc.white('ERROR_MESSAGE'))],
-        colWidths: [120],
-        wordWrap: true,
-      });
-
       console.log(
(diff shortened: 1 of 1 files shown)


cmtm:
I know this change is opinionated, so feel free to move discussion to an issue if that's preferred.
An alternative solution would be to truncate long error messages, but I feel that would be less useful.

comatory:
@JivusAyrus Does #2634 address this as well?

endigma:
Hi @cmtm, thanks for your contribution! We have an internal PR solving the same issue https://github.com/wundergraph/cosmo/pull/2715 so we'll be closing this in favour of that one.

PR #2532, opened 2026-02-19
fix(router): fix append algorithm producing multiple headers instead of comma-separated

### LINKED ISSUE #2531: response header algortihm append does not append values comma seperated as described in docs
### Component(s)

router

### Component version

latest

### wgc version

latest

### controlplane version

latest

### router version

0.282.0

## Description
router configuration for response headers with algorithm "append" dows not result in a single header with comma seperated values but in multiple headers.

See: https://cosmo-docs.wundergraph.com/router/proxy-capabilities/subgraph-response-header-operations#header-algorithms

## Steps to Reproduce
1. Have multiple subgraphs that set a Edge-Cache-Tag
2. Have a router config like:
...
  headers:
    all:
      response:
        - op: "propagate"
          named: Edge-Cache-Tag
          algorithm: "append"
...
3. Start a combined query
## Expected Result

the response Edge-Cache-Tag Header has a comma seperated, combined list of the tags from the subgraphs

## Actual Result

Multiple Edge-Cache-Tag Headers in the response

### Router configuration

```yaml
...
  headers:
    all:
      response:
        - op: "propagate"
          named: Edge-Cache-Tag
          algorithm: "append"
...
```

### Router execution config

```json

```

### Log output

```shell

```

### Additional context

_No response_

jensneuse:
## Summary by CodeRabbit

* **Bug Fixes**
  * Header propagation now preserves Set-Cookie as separate header lines and concatenates other header values when appending; router-level response header rules now behave consistently with propagation and precedence.

* **Tests**
  * Extensive test coverage added for header propagation behaviors, append/first/last-write semantics, regex matching, defaults/deduplication, error scenarios, canonicalization, and router rule interactions.

Fixes #2531. The response header `algorithm: append` was producing multiple separate HTTP header lines instead of a single comma-separated value as documented. The fix joins all values into one comma-separated string in `applyResponseRuleKeyValue`.

Also adds 30 unit tests for previously uncovered functions (`createMostRestrictivePolicy`, `AddCacheControlPolicyToRules`, `applyResponseRuleKeyValue`, `PropagatedHeaders`, `NewHeaderPropagation`), plus 7 new integration tests that explicitly assert the single-header contract and default value fallback behavior.

## Checklist

### DIFF
--- a/router-tests/header_propagation_test.go
+++ b/router-tests/header_propagation_test.go
@@ -700,10 +700,7 @@ func TestHeaderPropagation(t *testing.T) {
 				})
 				values := res.Response.Header.Values("Set-Cookie")
 				// Set-Cookie must NOT be comma-joined (RFC 6265) — each cookie stays as a separate header
-				require.Len(t, values, 2,
-					"Set-Cookie should produce multiple separate headers, got %d entries: %v", len(values), values)
-				require.Contains(t, values, "session=abc; Path=/")
-				require.Contains(t, values, "lang=en; Path=/")
+				require.ElementsMatch(t, []string{"session=abc; Path=/", "lang=en; Path=/"}, values)
 			})
 		})
 
--- a/router/core/header_rule_engine_test.go
+++ b/router/core/header_rule_engine_test.go
@@ -377,15 +377,10 @@ func TestNewHeaderPropagation(t *testing.T) {
 		require.Error(t, err)
 	})
 
-	t.Run("HasRequestRules on nil receiver returns false", func(t *testing.T) {
+	t.Run("nil receiver returns false for Has*Rules", func(t *testing.T) {
 		t.Parallel()
 		var hp *HeaderPropagation
 		assert.False(t, hp.HasRequestRules())
-	})
-
-	t.Run("HasResponseRules on nil receiver returns false", func(t *testing.T) {
-		t.Parallel()
-		var hp *HeaderPropagation
 		assert.False(t, hp.HasResponseRules())
 	})
 }


router/core/header_rule_engine.go
```diff
@@ -494,7 +494,13 @@ func (h *HeaderPropagation) applyResponseRuleKeyValue(res *http.Response, propag
 		propagation.m.Unlock()
 	case config.ResponseHeaderRuleAlgorithmAppend:
 		propagation.m.Lock()
-		propagation.header[key] = append(propagation.header[key], values...)
+		existing := propagation.header.Get(key)
```
StarpTech: There are headers like `Set-Cookie` that can't be combined as comma separate values and need to be added multiple times instead.

StarpTech (changes requested):

router/core/header_rule_engine_test.go
```diff
@@ -0,0 +1,397 @@
...
+			Enabled: true,
+			Value:   "max-age=300",
+		})
+		require.NotNil(t, result)
+		require.Len(t, result.All.Response, 2)
+		assert.Equal(t, "X-Existing", result.All.Response[0].Named)
+		assert.Equal(t, config.ResponseHeaderRuleAlgorithmMostRestrictiveCacheControl, result.All.Response[1].Algorithm)
+	})
+
+	t.Run("subgraph-specific cache rules", func(t *testing.T) {
```
endigma: Test name should be less vague about it's assertion

router/core/header_rule_engine_test.go
```diff
@@ -0,0 +1,397 @@
...
+		var hp *HeaderPropagation
+		assert.False(t, hp.HasRequestRules())
+	})
+
+	t.Run("HasResponseRules on nil receiver returns false", func(t *testing.T) {
+		t.Parallel()
+		var hp *HeaderPropagation
+		assert.False(t, hp.HasResponseRules())
+	})
+}
```
endigma: What is the purpose of these? Also there are two, also there is identical setup above we could just inline these assertions.

router/core/header_rule_engine_test.go
```diff
@@ -0,0 +1,397 @@
...
+	t.Run("append to empty header", func(t *testing.T) {
+		t.Parallel()
+		prop := newPropagation()
+		rule := &config.ResponseHeaderRule{Algorithm: config.ResponseHeaderRuleAlgorithmAppend}
+		hp.applyResponseRuleKeyValue(nil, prop, rule, "X-Test", []string{"only"})
+		assert.Equal(t, []string{"only"}, prop.header.Values("X-Test"))
+	})
+}
+
+// --- Part 5: PropagatedHeaders ---
```
endigma: Odd comments

router/core/header_rule_engine.go
jensneuse: Good catch. Here's what we found:

**`Set-Cookie` is NOT in `SkippedHeaders`** (`router/internal/headers/headers.go`), so a user could theoretically configure `algorithm: append` on `Set-Cookie`, and this fix would break it by comma-joining values that must remain as separate header lines per RFC 6265.

That said, the `append` algorithm is documented as producing "a comma separated list of values" — the old behavior (multiple header lines) contradicted that contract and broke real use cases like `Edge-Cache-Tag` for CDNs (#2531).

**Three options we see:**

1. **Add `Set-Cookie` to `SkippedHeaders`** — it shouldn't be propagated from subgraphs through the router anyway (breaks cookie scoping/security). Cleanest solution.
2. **Header-specific check in append** — if the header is `Set-Cookie`, fall back to multi-line. More surgical but adds special-case logic in a generic algorithm.
3. **Document the constraint** — `append` always comma-joins; users should not use it for headers that can't be folded (like `Set-Cookie`). Simplest, no code change.

Which approach do you prefer? Or is there a specific use case for propagating `Set-Cookie` from subgraphs that we should consider?

router/core/header_rule_engine_test.go
```diff
@@ -0,0 +1,397 @@
...
+			result, header := createMostRestrictivePolicy(tt.policies)
+			assert.Equal(t, tt.expectedHeader, header)
+			assert.NotNil(t, result)
+
+			// For expires test, verify the earlier time was selected
+			if tt.name == "expires header - earlier wins" {
+				assert.False(t, result.RespExpiresHeader.IsZero())
+				// The earlier one (5 min) should have won
+				assert.True(t, result.RespExpiresHeader.Before(time.Now().Add(6*time.Minute)))
+			}
```
endigma: This is a very weird way to write these tests, perhaps convert to subtests? Or a separate one specifically for this case?

router/core/header_rule_engine.go
```diff
@@ -494,7 +494,13 @@ func (h *HeaderPropagation) applyResponseRuleKeyValue(res *http.Response, propag
...
+		newVal := strings.Join(values, ",")
```
endigma: Can we just join `existing` in here as well?

router-tests/header_propagation_test.go
```diff
@@ -583,6 +583,157 @@ func TestHeaderPropagation(t *testing.T) {
...
+				Subgraphs:     subgraphsOnlyEmployeeSetsHeader,
+			}, func(t *testing.T, xEnv *testenv.Environment) {
+				res := xEnv.MakeGraphQLRequestOK(testenv.GraphQLRequest{
+					Query: queryEmployeeWithHobby,
+				})
+				ch := res.Response.Header.Get(customHeader)
+				// Hobbies responds last and uses the default value
+				require.Equal(t, "default-val", ch)
+			})
+		})
```
endigma, line 768: Is this the correct behaviour? Is not responding with a header at all the "last write" ? Doesn't this mean _every_ subgraph would have to respond with it unless the one that does happens to be the last one called?

router-tests/header_propagation_test.go
```diff
@@ -583,6 +583,157 @@ func TestHeaderPropagation(t *testing.T) {
...
+					}),
+				},
+				Subgraphs: subgraphsPropagateCustomHeader,
+			}, func(t *testing.T, xEnv *testenv.Environment) {
+				res := xEnv.MakeGraphQLRequestOK(testenv.GraphQLRequest{
+					Query: queryEmployeeWithHobby,
+				})
+				values := res.Response.Header.Values(customHeader)
+				require.Equal(t, 1, len(values),
+					"append algorithm should produce a single header with comma-separated values, got %d entries: %v", len(values), values)
```
endigma: `require.Len`

router/core/header_rule_engine.go
StarpTech: I recommend approach 2) that's how Apollo handles it and allows subgraphs to propagate cookie values correctly

router-tests/header_propagation_test.go
```diff
@@ -583,6 +583,157 @@ func TestHeaderPropagation(t *testing.T) {
...
+				require.Equal(t, "employee-value,default-val", values[0])
+			})
+		})
```
endigma, line 661: I can't see in the diff if we have a test where two subgraphs both have the default with append algorithm if it doesn't do `default-val,default-val`

endigma (changes requested):

router-tests/header_propagation_test.go
jensneuse: Yes, this is intentional. When a `default` is configured, a silent subgraph is treated as if it returned the default value — so it counts as a "write". Added a comment to the test clarifying this.

router-tests/header_propagation_test.go
```diff
@@ -583,6 +583,224 @@ func TestHeaderPropagation(t *testing.T) {
...
+			}, func(t *testing.T, xEnv *testenv.Environment) {
+				res := xEnv.MakeGraphQLRequestOK(testenv.GraphQLRequest{
+					Query: queryEmployeeWithHobby,
+				})
+				values := res.Response.Header.Values("Set-Cookie")
+				// Set-Cookie must NOT be comma-joined (RFC 6265) — each cookie stays as a separate header
+				require.Len(t, values, 2,
+					"Set-Cookie should produce multiple separate headers, got %d entries: %v", len(values), values)
+				require.Contains(t, values, "session=abc; Path=/")
+				require.Contains(t, values, "lang=en; Path=/")
```
endigma: You can use `require.Elements` to assert the length and contents in one go

PR #2534, opened 2026-02-19
feat(router): add SetWildcardScope to bypass @requiresScopes checks

### LINKED ISSUE #2490: feat(router): Add SetBypassScopeValidation to skip scope authorization checks
## Summary

We'd like to request a feature to bypass scope validation for requests where a "wildcard" or "admin" scope grants access to all scopes.

## Use Case

In our system, we have a wildcard scope that should grant access to all GraphQL fields regardless of their individual scope requirements. Currently, there's no way to skip scope validation from a custom module - we need to either:
1. Know all required scopes upfront and set them via `SetAuthenticationScopes()`
2. Let authorization fail and parse the `MissingScopes` from the response

Neither approach is ideal for a wildcard scope use case.

## Proposed Solution

Add `SetBypassScopeValidation(bool)` method to:
1. `Authentication` interface in `router/pkg/authentication/authentication.go`
2. `RequestContext` interface in `router/core/context.go`

Then modify `CosmoAuthorizer.getAuth()` in `router/core/authorizer.go` to return this flag, and check it in `AuthorizePreFetch()` and `AuthorizeObjectField()` before performing scope validation.

## Example Usage

```go
func (m *MyModule) Middleware(ctx core.RequestContext, next http.Handler) {
    auth := ctx.Authentication()
    if hasWildcardScope(auth.Scopes()) {
        // Skip all scope checks for this request
        ctx.SetBypassScopeValidation(true)
    }
    next.ServeHTTP(ctx.ResponseWriter(), ctx.Request())
}
```

## Implementation

We have a working implementation in our fork that we'd be happy to contribute as a PR:
- https://github.com/mondaycom/cosmo/tree/feature/bypass-scope-validation

The changes are minimal (~40 lines) and don't affect existing behavior unless explicitly enabled.

## Alternatives Considered

1. **Set all scopes manually** - Requires knowing all scopes from the schema upfront, which isn't exposed via public API
2. **Disable authorization entirely** - Too broad, we still want authentication
3. **Parse MissingScopes from failed responses** - Complex and requires failing first

Would love to hear your thoughts on this approach!

jensneuse:
## Summary by CodeRabbit

* **New Features**
  * Wildcard scope support in router authorization, allowing requests to opt into wildcard access when enabled.
  * Public request context API to enable/disable wildcard scope per request.

* **Tests**
  * Added comprehensive tests validating wildcard-scope behavior across authenticated, unauthenticated, and mixed authorization scenarios.

Adds `SetWildcardScope(bool)` to `RequestContext`, allowing custom modules to mark a request as satisfying all `@requiresScopes` checks without needing to enumerate every scope from the schema. Authentication is still enforced — unauthenticated requests are rejected before scope checks are evaluated. The flag is stored as a Go context value and checked in `CosmoAuthorizer.validateScopes` before the OR-of-AND scope matching logic. Closes #2490.

## Checklist


router/core/authorizer.go
```diff
@@ -77,6 +77,11 @@ func (a *CosmoAuthorizer) AuthorizeObjectField(ctx *resolve.Context, dataSourceI
 	return a.handleRejectUnauthorized(a.validateScopes(ctx, coordinate, required, isAuthenticated, actual))
 }
 
+func hasWildcardScope(ctx context.Context) bool {
+	v, ok := ctx.Value(wildcardScopeKey{}).(bool)
+	return ok && v
+}
+
```
endigma: The setter and getter for a context key should not be in different files

router-tests/modules/set_wildcard_scope_test.go
```diff
@@ -0,0 +1,190 @@
...
..._OR_TYPE"}},{"message":"Unauthorized to load field 'Query.employees.startDate', Reason: not authenticated.","path":["employees",4,"startDate"],"extensions":{"code":"UNAUTHORIZED_FIELD_OR_TYPE"}},{"message":"Unauthorized to load field 'Query.employees.startDate', Reason: not authenticated.","path":["employees",5,"startDate"],"extensions":{"code":"UNAUTHORIZED_FIELD_OR_TYPE"}},{"message":"Unauthorized to load field 'Query.employees.startDate', Reason: not authenticated.","path":["employees",6,"startDate"],"extensions":{"code":"UNAUTHORIZED_FIELD_OR_TYPE"}},{"message":"Unauthorized to load field 'Query.employees.startDate', Reason: not authenticated.","path":["employees",7,"startDate"],"extensions":{"code":"UNAUTHORIZED_FIELD_OR_TYPE"}},{"message":"Unauthorized to load field 'Query.employees.startDate', Reason: not authenticated.","path":["employees",8,"startDate"],"extensions":{"code":"UNAUTHORIZED_FIELD_OR_TYPE"}},{"message":"Unauthorized to load field 'Query.employees.startDate', Reason: not authenticated.","path":["employees",9,"startDate"],"extensions":{"code":"UNAUTHORIZED_FIELD_OR_TYPE"}}],"data":{"employees":[null,null,null,null,null,null,null,null,null,null]}}`, string(data))
```
endigma: Since all the tests use the same query and these responses are bit unwieldy to verify this way, can we make a struct to unmarshal this into so the assertions are more readable?

endigma: Alternatively if these specific errors are unimportant to read can this be a snapshot test?

endigma (changes requested):

jensneuse:
Addressed review feedback:

**@endigma re: context key in different files** — Moved `wildcardScopeKey` type to `authorizer.go` next to `hasWildcardScope`. The setter in `context.go` references it from the same package.

**@endigma re: unwieldy JSON assertions** — Replaced raw JSON string comparisons with structured `graphQLResponse` type. Tests now unmarshal and assert on individual fields (error count, error codes, messages, missing scopes coordinates). Much more readable.

router/core/authorizer.go
jensneuse: Done — moved `wildcardScopeKey` type to `authorizer.go` next to `hasWildcardScope`. The setter in `context.go` references it from the same package.

router-tests/modules/set_wildcard_scope_test.go
jensneuse: Done — replaced raw JSON string comparisons with a `graphQLResponse` struct. Tests now unmarshal and assert on individual fields (error count, error codes, messages, missing scopes coordinates). Much more readable.

Noroth (approved):

router/core/authorizer.go
```diff
@@ -86,6 +93,9 @@ func (a *CosmoAuthorizer) validateScopes(ctx *resolve.Context, coordinate resolv
 	if len(requiredOrScopes) == 0 {
 		return nil
 	}
+	if hasWildcardScope(ctx.Context()) {
+		return nil
+	}
```
endigma, line 91: This calls this function on the unwrapped context from the resolve context, we could really just move this whole context machinery into `core/context.go` next to the setter

router/core/authorizer.go
```diff
@@ -77,6 +77,13 @@ func (a *CosmoAuthorizer) AuthorizeObjectField(ctx *resolve.Context, dataSourceI
 	return a.handleRejectUnauthorized(a.validateScopes(ctx, coordinate, required, isAuthenticated, actual))
 }
 
+type wildcardScopeKey struct{}
+
+func hasWildcardScope(ctx context.Context) bool {
+	v, ok := ctx.Value(wildcardScopeKey{}).(bool)
+	return ok && v
+}
```
endigma: if we want to have this, we should have a getter/setter here on `context.Context` and the outer setter in core/context.go on `resolve.Context`

router-tests/modules/custom-set-wildcard-scope/module.go
```diff
@@ -0,0 +1,32 @@
...
+	"net/http"
+
+	"github.com/wundergraph/cosmo/router/core"
+)
+
+const myModuleID = "setWildcardScopeModule"
+
+type SetWildcardScopeModule struct {
+	Enabled bool `mapstructure:"enabled"`
+}
```
endigma: `Enabled` and the check in `Middleware` serves no purpose

router-tests/modules/set_wildcard_scope_test.go
```diff
@@ -0,0 +1,235 @@
...
+	t.Run("wildcard scope disabled still requires correct scopes", func(t *testing.T) {
+		t.Parallel()
+
+		cfg := config.Config{
+			Graph: config.Graph{},
+			Modules: map[string]any{
+				"setWildcardScopeModule": wildcardModule.SetWildcardScopeModule{
+					Enabled: false,
+				},
+			},
```
endigma: This is equivalent to not enabling the module at all

router-tests/modules/set_wildcard_scope_test.go
```diff
@@ -0,0 +1,235 @@
...
+			require.NoError(t, err)
+			defer res.Body.Close()
+			require.Equal(t, http.StatusOK, res.StatusCode)
+			data, err := io.ReadAll(res.Body)
+			require.NoError(t, err)
+			var resp graphQLResponse
+			require.NoError(t, json.Unmarshal(data, &resp))
+			require.Empty(t, resp.Errors)
+			require.NotEmpty(t, resp.Data)
+		})
```
endigma, line 60: This and other tests: can we break this up a bit with newlines so the series of assertions and reads is clearer

router-tests/modules/set_wildcard_scope_test.go
```diff
@@ -0,0 +1,235 @@
...
+			require.NoError(t, json.Unmarshal(data, &resp))
+			require.Len(t, resp.Errors, 10)
+			for _, e := range resp.Errors {
+				require.Equal(t, "UNAUTHORIZED_FIELD_OR_TYPE", e.Extensions.Code)
+				require.Contains(t, e.Message, "missing required scopes")
+			}
+			require.Len(t, resp.Extensions.Authorization.MissingScopes, 1)
+			require.Equal(t, "Employee", resp.Extensions.Authorization.MissingScopes[0].Coordinate.TypeName)
+			require.Equal(t, "startDate", resp.Extensions.Authorization.MissingScopes[0].Coordinate.FieldName)
+			require.Equal(t, []string{"read:employee"}, resp.Extensions.Authorization.ActualScopes)
```
endigma: We can use `tidwall/gjson` for checks like this without keeping a struct in sync to match this return type. Yes I am aware the struct was a previous suggestion by myself, but subs work has informed my new opinions on this

endigma (changes requested):

PR #2537, opened 2026-02-20
fix(router): do not reuse planners in the plan generator

ysmolski:
Engine 253 has introduce a breaking change that requires that planners should not be reused. Engine did it to simplify certain actions in the planner. Reusing the state make planner more complicated.

PG was the last place where we reused planners. Things change remove the reuse. It does not affect performance of the planners. Added benchmark shows that internal overhead is minimal.

    BenchmarkPlanGenerator-14    1364    2,582,604 ns/op    6,734,461 B/op    11,191 allocs/op

That's ~2.6ms per full plan generation cycle (parsing + normalizing + validating + planning the full.graphql query against 7 subgraphs).

benchstat:

                     │   old.txt.   │            new.txt         │
                     │   sec/op     │   sec/op     vs base       │
    PlanGenerator-14   2.107m ± 1%   2.090m ± 1%  ~ (p=0.187 n=15)

## Summary by CodeRabbit

* **Refactor**
  * More robust plan generation with improved initialization and error propagation for more reliable, fail-fast behavior and clearer error signals.

* **Tests**
  * Added a performance benchmark using comprehensive query test data to exercise plan generation.

* **Chores**
  * Bumped a tooling dependency to the next patch release.

### DIFF
--- a/router/pkg/plan_generator/plan_generator.go
+++ b/router/pkg/plan_generator/plan_generator.go
@@ -49,6 +49,9 @@ type QueryPlanResult struct {
 	Timings  core.OperationTimes `json:"timings,omitempty"`
 }
 
+// PlanGenerator reads GraphQL operation files from cfg.SourceDir,
+// generates query plans for each using the execution config,
+// and writes results to cfg.OutDir as individual files and/or a consolidated report.
 func PlanGenerator(ctx context.Context, cfg QueryPlanConfig) error {
 	if cfg.Concurrency == 0 {
 		cfg.Concurrency = runtime.GOMAXPROCS(0)


router/pkg/plan_generator/plan_generator_test.go
```diff
@@ -407,3 +407,22 @@ func TestPlanGenerator(t *testing.T) {
...
+	tempDir := b.TempDir()
+	cfg := QueryPlanConfig{
+		SourceDir:       path.Join(getTestDataDir(), "queries", "bench"),
+		OutDir:          tempDir,
+		ExecutionConfig: path.Join(getTestDataDir(), "execution_config", "base.json"),
+		Timeout:         "30s",
+		Concurrency:     1,
+	}
+	b.ReportAllocs()
+	b.ResetTimer()
```
endigma: I don't believe this is required with b.Loop

router/pkg/plan_generator/plan_generator.go
```diff
@@ -146,6 +139,15 @@ func PlanGenerator(ctx context.Context, cfg QueryPlanConfig) error {
 
 					queryFilePath := filepath.Join(queriesPath, queryFile.Name())
 
+					// Planners should not be reused.
+					planner, err := pg.GetPlanner()
+					if err != nil {
+						// If we fail to get the planner, we have to cancel the context
+						// to stop this and the other goroutines via ctxError.
+						cancelError(fmt.Errorf("failed to get a planner: %v", err))
```
endigma: Can you use %w to wrap the error?

router/pkg/plan_generator/plan_generator_test.go
ysmolski: i removed reset

router/pkg/plan_generator/plan_generator.go
ysmolski: done and for all other cases in the file

PR #2538, opened 2026-02-20
feat: metrics log exporter for otel to print span information

SkArchon:
This PR allows users to printout metrics that are exported to any OTLP collector, this is a separate exporter that does not wrap existing exporters but simply logs the metrics. Wrapping would mean that there would be logigng for each exporter.

Users have the option of also excluding metrics being logged, this is useful to reduce noise and focus on on or two metrics for investigations.

Example Configuration
```
telemetry:
  metrics:
    otlp:
      debug_exporter:
        enabled: true
        exclude_metrics:
          - "process\\."
          - "server\\.uptime"
          - "router\\.info"
```

This feature while could be used in production, is only meant to be used for debugging purposes.

## Summary by CodeRabbit

* **New Features**
  * Opt-in debug metrics exporter that logs collected metrics for troubleshooting, supports regex-based exclusions, and emits a notice when enabled.
  * Configuration and defaults updated to expose enable/disable, exclude patterns, and a configurable export interval.

* **Tests**
  * End-to-end tests validating debug logging, exclusion behavior (excluded metrics still collected but not logged), and no logging when disabled.

## Checklist


router/pkg/metric/debug_exporter.go
```diff
@@ -0,0 +1,180 @@
...
+		for _, m := range sm.Metrics {
+			if isMetricExcluded(m.Name, s.excludeMetrics) {
+				continue
+			}
+			logMetricData(s.logger, m)
+		}
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
SkArchon: We do not have per exporter debugging anymore. Its a separate exporter on its own.

router/pkg/config/config.go
```diff
@@ -406,33 +410,33 @@ type EngineExecutionConfiguration struct {
...
+	ResolverMaxRecyclableParserSize                        int           `envDefault:"32768" env:"ENGINE_RESOLVER_MAX_RECYCLABLE_PARSER_SIZE" yaml:"resolver_max_recyclable_parser_size,omitempty"`
+	EnableSubgraphFetchOperationName                       bool          `envDefault:"false" env:"ENGINE_ENABLE_SUBGRAPH_FETCH_OPERATION_NAME" yaml:"enable_subgraph_fetch_operation_name"`
+	DisableVariablesRemapping                              bool          `envDefault:"false" env:"ENGINE_DISABLE_VARIABLES_REMAPPING" yaml:"disable_variables_remapping"`
+	EnableRequireFetchReasons                              bool          `envDefault:"false" env:"ENGINE_ENABLE_REQUIRE_FETCH_REASONS" yaml:"enable_require_fetch_reasons"`
+	SubscriptionFetchTimeout                               time.Duration `envDefault:"30s" env:"ENGINE_SUBSCRIPTION_FETCH_TIMEOUT" yaml:"subscription_fetch_timeout,omitempty"`
+	ValidateRequiredExternalFields                         bool          `envDefault:"false" env:"ENGINE_VALIDATE_REQUIRED_EXTERNAL_FIELDS" yaml:"validate_required_external_fields"`
```
SkArchon, line 447: These were not formatted before for whatever reason

router/pkg/config/config.schema.json
```diff
@@ -1238,6 +1238,25 @@
                   "items": {
                     "type": "string"
                   }
+                },
+                "debug_exporter": {
+                  "type": "object",
+                  "description": "Standalone debug metric exporter. When enabled, logs all collected metrics via the router's zap logger at each export interval.",
```
dkorittki:
```suggestion
                  "description": "Standalone debug metric exporter. When enabled, logs all collected metrics via the router's logger at each export interval.",
```

Zap is an implementation details, we don't need to highlight it

router/pkg/metric/debug_exporter.go
```diff
@@ -0,0 +1,181 @@
...
+	"time"
+
+	"go.opentelemetry.io/otel/attribute"
+	sdkmetric "go.opentelemetry.io/otel/sdk/metric"
+	"go.opentelemetry.io/otel/sdk/metric/metricdata"
+	"go.uber.org/zap"
+)
+
+// standaloneDebugExporter is a metric exporter that logs all collected metrics via zap.
+type standaloneDebugExporter struct {
```
dkorittki: Can we drop the "standalone" from the name. Simply calling it `debugExporter`? The standalone kind of feels self explanatory. Or is there a specific reason?

router/pkg/metric/debug_exporter.go
```diff
@@ -0,0 +1,181 @@
...
+	for i, dp := range data.DataPoints {
+		points[i] = formatAttributes(dp.Attributes, timeRange(dp.StartTime, dp.Time), fmt.Sprintf("value=%v", dp.Value))
+	}
+	logger.Info("Metric",
+		zap.String("name", m.Name),
+		zap.String("unit", m.Unit),
+		zap.String("type", typeName),
+		zap.String("temporality", data.Temporality.String()),
+		zap.Bool("monotonic", data.IsMonotonic),
+		zap.Strings("data_points", points),
```
dkorittki: What about defining a sublogger in function `logMetricData` with all common fields like name, unit, type and pass that sublogger to logSum*, logHistogram*, etc. which can enrich the loggers fields and you do the actual logging again in `logMetricData`?

This way the specialised logging functions do not repeat themselves and its more unified.

router/pkg/metric/debug_exporter.go
```diff
@@ -0,0 +1,181 @@
...
+	)
+}
+
+func timeRange(start, end time.Time) string {
+	return fmt.Sprintf("start=%s end=%s", start.Format(time.RFC3339Nano), end.Format(time.RFC3339Nano))
+}
+
+func formatAttributes(attrs attribute.Set, extra ...string) string {
+	var parts []string
```
dkorittki: We already know the length `parts` will have, so we can preallocate.

```suggestion
	parts := make([]string, 0, attrs.Len()+len(extra))
```

router/pkg/metric/meter.go
```diff
@@ -263,30 +263,41 @@ func NewOtlpMeterProvider(ctx context.Context, log *zap.Logger, c *Config, servi
 	}
 
 	if c.OpenTelemetry.TestReader != nil {
-		mp := sdkmetric.NewMeterProvider(append(opts, sdkmetric.WithReader(c.OpenTelemetry.TestReader))...)
-		// Set the global MeterProvider to the SDK metric provider.
-		otel.SetMeterProvider(mp)
+		opts = append(opts, sdkmetric.WithReader(c.OpenTelemetry.TestReader))
```
dkorittki, line 266: Is it wanted that `opts` could have a testreader and debugreader simultaneously? This could happen when `c.OpenTelemetry.TestReader != nil`  and `c.OpenTelemetry.DebugExporter.Enabled == true`. Could be correct, just wanted to highlight it.

dkorittki (commented):
You said in the description that you log a warning when someone enables this feature but I can't find it in the code diff. Does it really exist?

router/pkg/metric/meter.go
SkArchon: Hey yea, this is used by testenv to allow as pass a reader we can hook into for tests. Since we want to be able to cross check metrics against the logs in the debug exporter both need to exist at the same time in testing. Its fine since it doesnt affect the actual runtime.

But will double check.

router/pkg/config/config.schema.json
```diff
@@ -1238,6 +1238,25 @@
...
+                    },
+                    "exclude_metrics": {
+                      "type": "array",
+                      "description": "List of regex patterns for metric names to exclude from debug output.",
+                      "items": {
+                        "type": "string"
+                      }
+                    }
+                  }
                 }
```
endigma: Add to `full.yaml` please

endigma: Also, I think these names could be more descriptive. "standalone debug" doesn't really indicate what it does, it's more like a metrics log exporter?

router/pkg/metric/config.go
```diff
@@ -75,6 +75,13 @@ func (e *EngineStatsConfig) Enabled() bool {
 	return e.Subscription
 }
 
+type DebugMetricsExporterConfig struct {
+	Enabled        bool
+	ExcludeMetrics []*regexp.Regexp
+	// TestExportInterval overrides the export interval for testing. If zero, the default interval is used.
```
endigma: In what way is the interval for testing? Even if it's currently only used for tests I think it should have a proper name

router/pkg/metric/debug_exporter.go
```diff
@@ -0,0 +1,182 @@
...
+	"go.uber.org/zap"
+)
+
+// debugExporter is a metric exporter that logs all collected metrics via zap.
+type debugExporter struct {
+	logger         *zap.Logger
+	excludeMetrics []*regexp.Regexp
+}
+
+func newStandaloneDebugExporter(logger *zap.Logger, excludeMetrics []*regexp.Regexp) sdkmetric.Exporter {
```
endigma: again, maybe `logExporter`? Also, this should not return an interface unless it needs to.

router/pkg/metric/debug_exporter.go
```diff
@@ -0,0 +1,182 @@
...
+	return &debugExporter{
+		logger:         logger.Named("metrics-debug"),
```
endigma: Could possibly make this configurable?

router/pkg/metric/log_exporter.go
```diff
@@ -0,0 +1,182 @@
...
+		logHistogramMetric(logger, m, "Histogram[float64]", data)
+	case metricdata.ExponentialHistogram[int64]:
+		logExpHistogramMetric(logger, m, "ExponentialHistogram[int64]", data)
+	case metricdata.ExponentialHistogram[float64]:
+		logExpHistogramMetric(logger, m, "ExponentialHistogram[float64]", data)
+	case metricdata.Gauge[int64]:
+		logGaugeMetric(logger, m, "Gauge[int64]", data)
+	case metricdata.Gauge[float64]:
+		logGaugeMetric(logger, m, "Gauge[float64]", data)
+	default:
```
endigma, line 110: If these are meant for filtering, I'd suggest not using regex sensitive characters inside them, perhaps like `sum:int64`?

router/pkg/metric/debug_exporter.go
SkArchon: done

router/pkg/config/config.schema.json
SkArchon: done

router/pkg/metric/debug_exporter.go
SkArchon: done

router/pkg/metric/debug_exporter.go
SkArchon: Actually will remove this, does not really add any value.

router/pkg/metric/config.go
SkArchon: done

router/pkg/metric/log_exporter.go
SkArchon: Done

router/pkg/metric/debug_exporter.go
SkArchon: Done

router/pkg/config/config.schema.json
SkArchon: done

router/pkg/config/config.go
```diff
@@ -149,6 +147,11 @@ type Metrics struct {
 	CardinalityLimit int               `yaml:"experiment_cardinality_limit" envDefault:"2000" env:"METRICS_EXPERIMENT_CARDINALITY_LIMIT"`
 }
 
+type MetricsLogExporter struct {
+	Enabled        bool       `yaml:"enabled" envDefault:"false"`
+	ExcludeMetrics RegExArray `yaml:"exclude_metrics,omitempty"`
```
endigma, line 152: Nit: I would suggest `IncludeMetrics` as a mutually exclusive field with this, where anything being in it functions as a whitelist

router/pkg/config/config.schema.json
```diff
@@ -1327,6 +1327,25 @@
...
+                },
+                "log_exporter": {
+                  "type": "object",
+                  "description": "Log exporter. When enabled, logs all collected metrics via the router's logger at each export interval.",
+                  "additionalProperties": false,
+                  "properties": {
+                    "enabled": {
+                      "type": "boolean",
+                      "default": false,
+                      "description": "Enable the log exporter, which will ensure exported metrics are logged. The default value is false."
```
endigma:
```suggestion
                      "description": "Enables the log exporter. The default value is false."
```

router/pkg/metric/log_exporter.go
```diff
@@ -0,0 +1,182 @@
...
+func (s *logExporter) Temporality(_ sdkmetric.InstrumentKind) metricdata.Temporality {
+	return metricdata.CumulativeTemporality
+}
+
+func (s *logExporter) Aggregation(_ sdkmetric.InstrumentKind) sdkmetric.Aggregation {
+	return nil
+}
+
+func (s *logExporter) ForceFlush(_ context.Context) error {
+	return nil
```
endigma, line 43: There is a force logger flush method, do you think that would be useful here?

router/pkg/metric/meter.go
```diff
@@ -263,30 +263,41 @@ func NewOtlpMeterProvider(ctx context.Context, log *zap.Logger, c *Config, servi
...
-
 		opts = append(opts, sdkmetric.WithReader(
-			sdkmetric.NewPeriodicReader(exporter,
+			sdkmetric.NewPeriodicReader(logExp,
 				sdkmetric.WithTimeout(defaultExportTimeout),
-				sdkmetric.WithInterval(defaultExportInterval),
+				sdkmetric.WithInterval(exportInterval),
 			),
 		))
+		log.Warn("Metrics log exporter enabled")
```
endigma: If this is something to be warned about, can we include the negative consequence in the warning to explain why?

SkArchon: I left it as warn so its very obvious to users who might want to have enabled it for debugging purposes, if there was a genuine use case, they can ignore the warn I think.

router/pkg/config/config.go
SkArchon: All metrics are logged by default, why would we need a whitelist

router/pkg/metric/log_exporter.go
SkArchon: You mean Sync()? yea that makes sense

router/pkg/config/config.go
endigma: Because I think most people won't want to log _all_ metrics and are more likely to be interested in "just A, B, C" than "all but C..Z". I suppose they could invert the regex in this case to match anything that's not what they want?

router/pkg/metric/meter.go
endigma: My point stands to add some kind of detail to a warn level message describing why the user should care, this could flag some alerting

router-tests/telemetry/metrics_log_export_test.go
```diff
@@ -0,0 +1,379 @@
...
+					ExcludeMetrics: []*regexp.Regexp{
+						regexp.MustCompile(`^router\.http\.requests$`),
+					},
+					IncludeMetrics: []*regexp.Regexp{
+						regexp.MustCompile(`^router\.http\.request\.duration_milliseconds$`),
+					},
+				},
+			},
+		}, func(t *testing.T, err error) {
+			require.Error(t, err, "metrics log exporter: exclude_metrics and include_metrics cannot be used together, use only one")
```
endigma: This works, but should this be `require.ErrorContains()`?

router/pkg/metric/log_exporter.go
```diff
@@ -0,0 +1,196 @@
...
+	return metricdata.CumulativeTemporality
+}
+
+func (s *logExporter) Aggregation(_ sdkmetric.InstrumentKind) sdkmetric.Aggregation {
+	return nil
+}
+
+func (s *logExporter) ForceFlush(_ context.Context) error {
+	if err := s.logger.Sync(); err != nil {
+		s.logger.Warn("unable to flush metric log export")
```
endigma: nit: could add the flush error, probably doesn't matter though

PR #2544, opened 2026-02-21
fix(router): guard against nil context in getResponseHeaderPropagation

### LINKED ISSUE #2530: Panic in getResponseHeaderPropagation (nil context) on specific query plan – router:latest
### Component(s)

router

### Component version

latest

### wgc version

latest

### controlplane version

latest

### router version

latest

### What happened?

Cosmo Router panics with:

runtime error: invalid memory address or nil pointer dereference

The panic originates from:

github.com/wundergraph/cosmo/router/core.getResponseHeaderPropagation
router/core/header_rule_engine.go:58

This only occurs for one specific GraphQL query shape (out of thousands), and began recently while running:

ghcr.io/wundergraph/cosmo/router:latest

The same query works perfectly when sent directly to the underlying subgraph.

**Stacktrace**
runtime error: invalid memory address or nil pointer dereference

github.com/wundergraph/cosmo/router/core.getResponseHeaderPropagation
    router/core/header_rule_engine.go:58
github.com/wundergraph/cosmo/router/core.(*HeaderPropagation).ApplyResponseHeaderRules
    router/core/header_rule_engine.go:395
github.com/wundergraph/cosmo/router/core.(*engineLoaderHooks).OnFinished
    router/core/engine_loader_hooks.go:141
...

The panic occurs during the OnFinished hook after query resolution.

**Observations**

This only happens for one query plan variant.

Adding fields changes the planner output and avoids the panic.

The panic happens inside:

func getResponseHeaderPropagation(ctx context.Context) *responseHeaderPropagation {
    v := ctx.Value(responseHeaderPropagationKey{})  // panics if ctx == nil

It appears ctx is nil in this execution path.

There is no nil-guard on ctx before calling ctx.Value(...).

This suggests:

A fetch path is completing with a nil context.Context

OnFinished is invoked even when propagation context was never initialized

**Expected Behavior**

If no response header propagation context exists, the router should:

Treat it as a no-op

Not panic

Return the GraphQL response normally

**Actual Behavior**

Router panics

Request returns 500

Recovery middleware catches it, but the request fails

**Suspected Root Cause**

getResponseHeaderPropagation does not guard against:

ctx == nil

missing propagation value in context

A defensive implementation would prevent the crash:

func getResponseHeaderPropagation(ctx context.Context) *responseHeaderPropagation {
    if ctx == nil {
        return nil
    }
    v := ctx.Value(responseHeaderPropagationKey{})
    if v == nil {
        return nil
    }
    return v.(*responseHeaderPropagation)
}

**Notes**

This started recently while running :latest.

Likely a regression introduced in a recent router build.

Pinning to a previous digest appears to avoid the issue.

## Environment
OS: (e.g., "Ubuntu 20.04")
Package Manager: pnpm, npm, yarn, etc
Compiler(if manually compiled): (e.g., "go 14.2")

### Router configuration

```yaml

```

### Router execution config

```json

```

### Log output

```shell

```

### Additional context

_No response_

stakeswky:
## Summary

Fixes #2530

The router panics with `runtime error: invalid memory address or nil pointer dereference` in `getResponseHeaderPropagation` when a fetch path completes with a nil `context.Context`. This happens because `ctx.Value()` is called without checking if `ctx` is nil first.

## Changes

**`router/core/header_rule_engine.go`**
- Added nil guard for `ctx` parameter in `getResponseHeaderPropagation()`
- When `ctx` is nil, returns `nil` (no-op), which causes `ApplyResponseHeaderRules` to return early — matching the expected behavior

**`router/core/header_rule_engine_test.go`**
- Added regression test `TestGetResponseHeaderPropagation_NilContext` to verify nil context does not panic

## Root Cause

As described in the issue, certain query plan variants can result in a fetch path completing with a nil context. The `OnFinished` hook in `engine_loader_hooks.go` then calls `ApplyResponseHeaderRules` which calls `getResponseHeaderPropagation(ctx)`, and `ctx.Value()` panics on nil receiver.

## Before / After

**Before:** Router panics → 500 error → recovery middleware catches it
**After:** Nil context treated as no-op → response returned normally (no header propagation applied)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved header processing stability by adding a defensive check to safely handle missing context.

* **Tests**
  * Added a regression test to ensure header handling does not fail when context is absent.

### DIFF
--- a/router/core/header_rule_engine.go
+++ b/router/core/header_rule_engine.go
@@ -76,6 +76,9 @@ func WithResponseHeaderPropagation(ctx *resolve.Context) *resolve.Context {
 }
 
 func getResponseHeaderPropagation(ctx context.Context) *responseHeaderPropagation {
+	if ctx == nil {
+		return nil
+	}
 	v := ctx.Value(responseHeaderPropagationKey{})
 	if v == nil {
 		return nil
--- a/router/core/header_rule_engine_test.go
+++ b/router/core/header_rule_engine_test.go
@@ -336,6 +336,16 @@ func TestPropagatedHeaders(t *testing.T) {
 	})
 }
 
+func TestGetResponseHeaderPropagation_NilContext(t *testing.T) {
+	t.Parallel()
+
+	// Ensure nil context does not panic (regression test for #2530)
+	assert.NotPanics(t, func() {
+		result := getResponseHeaderPropagation(nil)
+		assert.Nil(t, result)
+	})
+}
+
 func TestNewHeaderPropagation(t *testing.T) {
 	t.Parallel()
 


endigma:
Hi @stakeswky, thanks for your contribution, we'll take a look.

cc @jensneuse

stakeswky:
Thanks! Let me know if you have any questions or feedback.

endigma:
Hi @stakeswky, would you mind updating this PR to the latest main and seeing if the test still fails? I believe we addressed the root cause of this last week.

endigma:
Or, seeing that the test will absolutely still fail, can you retry your original issue?

stakeswky:
Rebased onto the latest `main` just now. The test still fails on the unrebased branch — curious to see if the upstream fix resolves it. Let me know what you find!

endigma:
Hi @stakeswky, are you able to recheck the actual query/plan that prompted you to create this PR? 

We didn’t guard against nil in the function directly because nil context should never have been passed to this function in the first place, so the panic was justified.

I believe this test will still fail because it forces the nil context to be passed directly.

We’ve resolved the root issue now, please let us know and create a new issue if you run into any other problems.

endigma:
Closing for now, please feel free to reopen as an issue if your problem is unresolved

PR #2551, opened 2026-02-24
chore: give hint of configuration on otel invalid utf8 errors

SkArchon:
Right now we have options to sanitize span attributes, however this is off by default. Users right now will get the error message of "string field contains invalid UTF-8" when non utf8 data is set into a span anywhere. However users might not know that a workaround exists (right now for span attributes), without logging the error as is, if the error is related to non utf8 data we will log the error with a hint.

## Summary by CodeRabbit

* **New Features**
  * Added environment variable support and a grouped env-prefix option for configuring the metrics OTLP log exporter (enablement and include/exclude metric lists).

* **Bug Fixes**
  * Improved telemetry error handling and logging for export failures involving invalid UTF-8, emitting a clear hint to enable UTF-8 sanitization.

* **Tests**
  * Added test coverage that simulates invalid UTF-8 export errors to verify sanitization hints are logged.

## Checklist

### DIFF
--- a/aws-lambda-router/CHANGELOG.md
+++ b/aws-lambda-router/CHANGELOG.md
@@ -4,6 +4,12 @@ Binaries are attached to the github release otherwise all images can be found [h
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+# [0.40.0](https://github.com/wundergraph/cosmo/compare/aws-lambda-router@0.39.2...aws-lambda-router@0.40.0) (2026-03-05)
+
+### Features
+
+* add format script + optional git hooks ([#2579](https://github.com/wundergraph/cosmo/issues/2579)) ([edabd2b](https://github.com/wundergraph/cosmo/commit/edabd2bd8f775c5717ca785f16616cff6f7ae039)) (@comatory)
+
 ## [0.39.2](https://github.com/wundergraph/cosmo/compare/aws-lambda-router@0.39.1...aws-lambda-router@0.39.2) (2026-02-24)
 
 ### Bug Fixes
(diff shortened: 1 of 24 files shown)


router/pkg/trace/errors.go
```diff
@@ -0,0 +1,34 @@
...
+	var target invalidUTF8Error
+	if errors.As(err, &target) {
+		return target.InvalidUTF8()
+	}
+	return false
+}
+
+func errHandler(config *ProviderConfig) func(err error) {
+	return func(err error) {
+		if hasInvalidUTF8Error(err) {
```
StarpTech, line 26: Will this be logged once or for every metric emission?

SkArchon: Every metric emission as it stands. It just modifies the error log before printing it.

I could make it once per router run, though there is a chance folks would miss it (if the error is regularly occurring like in the case of the current customer). Wdyt?

StarpTech: Are you saying after every batch export or individual metric point?

endigma:
Is there a downside to turning it on by default? Seems like if it inevitably just makes errors we should do this implicitly.

router/pkg/trace/errors.go
StarpTech: If it is per export, it is fine; otherwise, I'd log it only once.

StarpTech:
> Is there a downside to turning it on by default? Seems like if it inevitably just makes errors we should do this implicitly.

@SkArchon is validating the overhead. If negligible, I'd also suggest enabling it by default.

router/pkg/trace/errors.go
SkArchon: It's per every batch export.

SkArchon:
It takes about `14.70089313` microseconds per request (with no custom metric configuration, more string attributes would mean this increases), lets assume the average is about 140ns per attribute (this is only applied to string attributes). While fast, since this is on the hot path and as only one customer has reported it, I would prefer having the log message, that way when folks do see the log message they are almost likely to go and enable this.

PR #2553, opened 2026-02-25
feat(studio): delete persisted operations

comatory:
@coderabbitai summary

## Checklist

- [x] Documentation has been updated on [https://github.com/wundergraph/cosmo-docs](https://github.com/wundergraph/cosmo-docs/pull/255
).

Adds functionality to delete persisted operations via Studio. The persisted operations are
uploaded via CLI first, only then they can be visible in the Studio.
Organization admins & developers can perform this action.

_Note_: The video displays "retire..." but now we're using delete instead

https://github.com/user-attachments/assets/2d26a6be-0454-4e1f-b38e-a7b511f4f787

## Testing

* ensure you can run cosmo locally
* follow [this guide](https://cosmo-docs.wundergraph.com/tutorial/using-persisted-operations#register-a-persisted-operation) to register and execute a persisted operation
* visit your graph's _clients_ section, open the relevant client which you used when executing the operation above, click _View Operations_
* you should see list of operations, click ▽  toggle button, you should see the clear/trash button
* click it (you might get warning if the operation was executed)

_Try these as well if you have time_:
* use multiple `query ...` definitions in one persisted operation file, UI should be able to handle large volume of them in a single operation
* try retiring an operation that was not executed before (you should see alert dialog in case of missing analytics)
* you should NOT see the button if you're not owner or developer assigned in your org (I was not able to make invitation flow work on my localhost yet, so can't try different roles)

## Considerations

* UI allows to delete operations individually, there's no bulk operation
* operation must first be deleted succesfully, in order to be able to delete
  next operation
* modal dialogs are always shown

> [!NOTE]  
> We believe the limitations above will be a non-issue, once we have support for retiring
> the operations via CLI. Bulk retirement should then be possible via scripting, or any
> other kind of automations.
> We intend to provide ability to override warnings (if operation has traffic) via a flag.

### DIFF
--- a/.github/actions/image-scan/action.yaml
+++ b/.github/actions/image-scan/action.yaml
@@ -9,10 +9,6 @@ inputs:
     description: 'GitHub Token'
     required: true
 
-  skip_dirs:
-    description: 'A comma separated list of folders to ignore'
-    required: false
-
   image_ref:
     description: "The name of the container"
     required: true
@@ -33,7 +29,7 @@ runs:
         exit-code: 1
         severity: 'CRITICAL,HIGH'
         format: 'sarif'
-        trivyignores: '.trivyignore.yaml'
+        trivy-config: 'trivy.yaml'
         limit-severities-for-sarif: true
         output: 'trivy-results.sarif'
 
--- a/trivy.yaml
+++ b/trivy.yaml
@@ -1,0 +1,1 @@
+ignorefile: .trivyignore.yaml


controlplane/test/persisted-operations.test.ts
```diff
@@ -69,233 +78,350 @@ describe('Persisted operations', (ctx) => {
...
-    const deleteFederatedGraphResp = await client.deleteFederatedGraph({
-      name: fedGraphName,
-      namespace: 'default',
+
+      authenticator.changeUserWithSuppliedContext({
+        ...users[TestUser.viewerTimCompanyA]!,
+        rbac: createTestRBACEvaluator(createTestGroup({ role: 'namespace-viewer' })),
+      });
+
+      expect(client.retirePersistedOperation({
```
comatory: This fails with

```
ConnectError {
  "message": "[internal] internal error",
  "name": "ConnectError",
  "rawMessage": "internal error",
  "code": 13,
  "metadata": Headers {},
  "details": [],
}
```

Am I using `authenticator.changeUserWithSuppliedContext` correctly?

controlplane/src/core/bufservices/persisted-operation/check-persisted-operation-traffic.ts
```diff
@@ -0,0 +1,82 @@
...
+      return {
+        response: {
+          code: EnumStatusCode.ERR_NOT_FOUND,
+          details: `Persisted operation ${req.operationId} does not exist`,
+        },
+      };
+    }
+
+    const metricsRepository = new MetricsRepository(opts.chClient);
+    const operationMetrics = await metricsRepository.getPersistedOperationMetrics({
```
StarpTech, line 79: Here, we need to respect the `breaking_change_retention_limit` retention settings.

controlplane/src/core/bufservices/persisted-operation/retirePersistedOperation.ts
```diff
@@ -0,0 +1,145 @@
...
+import { createBlobStoragePath } from './utils.js';
+
+export function retirePersistedOperation(
+  opts: RouterOptions,
+  req: RetirePersistedOperationRequest,
+  ctx: HandlerContext,
+): Promise<PlainMessage<RetirePersistedOperationResponse>> {
+  let logger = getLogger(ctx, opts.logger);
+
+  return handleError<PlainMessage<RetirePersistedOperationResponse>>(ctx, logger, async () => {
```
StarpTech: `retire` is a very uncommon term to say that we delete the operation. Why not just `delete` or `remove`?

controlplane/src/core/bufservices/persisted-operation/deletePersistedOperation.ts
```diff
@@ -0,0 +1,145 @@
...
+            id: retiredOperation.id,
+            operationId: retiredOperation.operationId,
+            operationNames: retiredOperation.operationNames,
+          }
+        : undefined,
+    };
+  });
+}
+
+const removePersistedOperationFromBlobStorage = async ({
```
StarpTech: Please put this into a service. The handler should be free of business logic (goal)

controlplane/src/core/bufservices/persisted-operation/deletePersistedOperation.ts
```diff
@@ -0,0 +1,145 @@
...
+    organizationId,
+    fedGraphId,
+    clientName: operation.clientName,
+    operationId: operation.operationId,
+  });
+
+  try {
+    await blobStorage.deleteObject({
+      key: path,
+      abortSignal: AbortSignal.timeout(10_000),
```
StarpTech: This might be too low. I'd get rid of it entirely. Do you know how this will impact internal retry?

controlplane/src/core/repositories/OperationsRepository.ts
```diff
@@ -162,6 +152,33 @@ export class OperationsRepository {
     };
   }
 
+  public async retirePersistedOperation({
+    operationId,
+  }: {
+    operationId: string;
+  }): Promise<PersistedOperationDTO | undefined> {
+    const operationResult = await this.db.query.federatedGraphPersistedOperations.findFirst({
```
StarpTech: Please run this within a TX.

controlplane/src/core/bufservices/persisted-operation/retirePersistedOperation.ts
comatory: I got influenced by The Guild stack, they [refer to it](https://the-guild.dev/graphql/hive/docs/schema-registry/app-deployments#retire-an-app-deployment) using this name.

I'm ok with changing, is `delete` okay? I'll rename all the mentions of `retire`.

controlplane/src/core/bufservices/persisted-operation/deletePersistedOperation.ts
comatory: I was following conventions from the [publish handler](https://github.com/wundergraph/cosmo/blob/ondrej/eng-8830-controlplane-cli-delete-persistent-operation/controlplane/src/core/bufservices/persisted-operation/publishPersistedOperations.ts#L206) where the `blobStorage` is used within the handler.

I agree with you that those are separate concerns. I'll create a class wrapping this functionality and instantiate it in the handlers. Will replace it in publish handler as well to make it consistent. Thanks for pointing this out.

controlplane/src/core/bufservices/persisted-operation/deletePersistedOperation.ts
comatory: > Do you know how this will impact internal retry?

Not sure how internal retry works, but since I don't know a lot about it I'll remove it like you suggested.

controlplane/src/core/bufservices/persisted-operation/deletePersistedOperation.ts
comatory: Ok my bad, I read the code a bit more and `blobStorage` is already a wrapper around S3, probably not very good idea to create another abstraction on top of it, but I will try moving it to the `OperationsRepository` so it's encapsulated as one action.

controlplane/src/core/repositories/OperationsRepository.ts
comatory: I got a bit confused with this comment but I will resolve it without making changes on my side. Reasoning: the code you were pointing is indeed calling database twice. First call is querying data and second is writing data. I think the transaction only makes sense if multiple writes/updates are involved, which is not the case.

Perhaps you meant to do transaction for writing to database and blob storage, however those are two different storage systems so they can't be part of single database transaction.
We'd have to probably design this differently, and fix all the other places where blob storage is involved and implement some kind of reverting mechanism. I think it's out of scope for this ticket.

I discussed this with @JivusAyrus and he noticed a bad pattern in my code: I'm writing to blob storage and _then_ to database, which could introduce a bug if blob storage succeeded but call to DB failed. I'll call the database first to avoid this problem and blob storage then.

controlplane/src/core/repositories/OperationsRepository.ts
```diff
@@ -162,6 +151,33 @@ export class OperationsRepository {
...
+      },
+    });
+
+    if (!operationResult) {
+      return undefined;
+    }
+
+    await this.db
+      .delete(federatedGraphPersistedOperations)
+      .where(eq(federatedGraphPersistedOperations.operationId, operationId));
```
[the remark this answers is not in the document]
comatory: Addressed in 045a7a8243d2c63510a99f6f73fb4dede15f759e

proto/wg/cosmo/platform/v1/platform.proto
```diff
@@ -1282,6 +1282,21 @@ message PublishPersistedOperationsResponse {
   repeated PublishedOperation operations = 2;
 }
 
+message RetirePersistedOperationRequest {
+  string fedGraphName = 1;
+  string namespace = 2;
+  string operationId = 3;
+}
```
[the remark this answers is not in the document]
comatory: Addressed in 045a7a8243d2c63510a99f6f73fb4dede15f759e

controlplane/src/core/bufservices/persisted-operation/deletePersistedOperation.ts
comatory: Addressed in 332d20cdad95b9a2e2ad85b0f716b793c678af45

controlplane/src/core/bufservices/persisted-operation/retirePersistedOperation.ts
comatory: Addressed in 6c5349440c8cc1075f8842186d7576447b2c9326

Used "delete" word

controlplane/src/core/bufservices/persisted-operation/check-persisted-operation-traffic.ts
comatory: Addressed in a3e412c7d3eddbbcee863f9606a2bc006d666e56

controlplane/src/core/bufservices/persisted-operation/deletePersistedOperation.ts
comatory: I attempted to move this into `OperationStorage` repository, but this approach would require much larger refactoring, as I'd need to pass blob storage to its constructor and update all the call sites and usages of blob storage. Probably out of scope for this feature.

Discussed this quickly with @JivusAyrus as well to get his thoughts on this, I was told repositories are only intended to call the PG database, not the S3 storage, so even putting them together is probably not the best approach, unless we decide to agree that repositories can expand their responsibilites.

I originally put this function there to avoid nesting `try/catch`, as it makes code usually less readable (everything shifts to "right" in the source code) for humans. However, after moving the code around ([see this comment](https://github.com/wundergraph/cosmo/pull/2553#discussion_r2863898479)), it's not so bad actually so I just inlined the functionality into the handler function.

Addressed in ff15c06e744080a7859c55c03b088be1b6001acc

Let me know if you think we should treat repositories differently and happy to plan work around that.

controlplane/src/core/repositories/OperationsRepository.ts
StarpTech: I mean to run a query and delete within a transaction to ensure consistency and avoid roundtrips. Easy change. But also good findings.

comatory:
@StarpTech ready for re-review

comatory:
Docs are here btw https://github.com/wundergraph/cosmo-docs/pull/255

controlplane/src/core/repositories/OperationsRepository.ts
comatory: Ah I got you now. Just realized that the method I was calling is hitting DB separately, that makes sense now. See 78af3afd16b875f5f2d00f793910dade54853093

I tried do some refactoring to avoid duplication so we can still have the benefit of re-using same queries.

comatory:
@StarpTech  please re-review, I believe you wanted [this one](https://github.com/wundergraph/cosmo/pull/2553#discussion_r2864856014) addressed?

studio/src/components/clients/delete-persisted-operation-dialog.tsx
```diff
@@ -0,0 +1,107 @@
...
+            The {pluralizedOperation} {isPlural ? "are" : "is"}{" "}
+            <span className="font-semibold">receiving traffic</span>. Visit{" "}
+            <Link href={metricsLink} className="underline">
+              metrics
+            </Link>{" "}
+            to learn more.
+          </Alert>
+        ) : (
+          <Alert variant="warn">
+            If you are not sending us analytics, we{" "}
```
StarpTech, line 59: I'd make the value prop of sending us data more clear: "If you are not sending us analytics, we cannot guarantee that existing clients won't break" or similiar

controlplane/src/core/repositories/OperationsRepository.ts
StarpTech: Thanks

studio/src/components/clients/delete-persisted-operation-dialog.tsx
comatory: Screenshot 2026-02-27 at 16 26 59

comatory:
@StarpTech copy fixed

PR #2564, opened 2026-02-26
fix: improve router.http.request.duration_milliseconds

alepane21:
When a query plan makes parallel fetches, the metric router.http.request.duration_milliseconds is recorded incorrectly: it reported as latency the time of the slower fetch. With this change I set as duration the timing that I get from the engine for the fetch, that doesn't have this issue.

## Summary by CodeRabbit

* **Tests**
  * Added an integration test ensuring parallel subgraph request durations are reported to Prometheus.

* **Bug Fixes**
  * Improved latency measurement: logging, telemetry and metrics now use fetch timing for more accurate duration reporting.

## Checklist


router/core/engine_loader_hooks.go
```diff
@@ -175,9 +160,11 @@ func (f *engineLoaderHooks) OnFinished(ctx context.Context, ds resolve.DataSourc
 	exprCtx.Subgraph.Name = ds.Name
 	exprCtx.Subgraph.Request.Error = WrapExprError(responseInfo.Err)
 
+	var fetchLatency time.Duration
 	if value := ctx.Value(rcontext.FetchTimingKey); value != nil {
 		if fetchTiming, ok := value.(*atomic.Int64); ok {
-			exprCtx.Subgraph.Request.ClientTrace.FetchDuration = time.Duration(fetchTiming.Load())
+			fetchLatency = time.Duration(fetchTiming.Load())
+			exprCtx.Subgraph.Request.ClientTrace.FetchDuration = fetchLatency
```
SkArchon, line 167: This measures the low level round trip, whereas the current measurement should measure this plus the engine hook logic that would add to the request duration.

However more importantly, would it work with GRPC data sources?

alepane21: Good catch! It would not work with gRPC data sources.

PR #2577, opened 2026-03-03
fix(ci): improve reliability of the router-tests

jensneuse:
## Summary

- All recent CI failures on integration tests were caused by OOM (`signal: killed`), not actual test bugs
- Reduces memory pressure by splitting the largest matrix entry, reducing parallelism (10→4), adding GOMAXPROCS=4, and fixing the Kafka health check
- Increases test timeout (5m→8m) to accommodate lower parallelism

## Changes

1. **Split test matrix**: `'./. ./fuzzquery ./lifecycle ./modules'` → `'./.'` + `'./fuzzquery ./lifecycle ./modules'` (5 jobs instead of 4)
2. **Reduce parallelism**: `--parallel 10` → `--parallel 4` (each parallel test spins up a full router + gRPC plugin subprocess; 10 concurrent with `-race` exhausts 16 GB RAM)
3. **Add GOMAXPROCS=4**: Limits Go scheduler threads, reducing race detector memory overhead
4. **Fix Kafka health check**: `kafka-broker-api-versions.sh --version` only checks binary exists; now uses `--bootstrap-server localhost:9092` to verify broker readiness
5. **Increase test timeout**: 5m → 8m to accommodate fewer parallel tests

## Evidence

Analyzed 7 most recent CI failures — all showed `signal: killed` on plugin subprocesses (OOM), zero actual `--- FAIL` test assertions:

| Run | killed count | test failures |
|-----|-------------|--------------|
| 22612622051 | 3 | 0 |
| 22612367177 | 6 | 0 |
| 22602367418 | 3 | 0 |
| 22593260747 | 4 | 0 |
| 22592631652 | 6 | 0 |
| 22592307099 | 6 | 0 |
| 22591585050 | 6 | 0 |

## Test plan

- [ ] All 5 integration_test matrix jobs pass without `signal: killed`
- [ ] Wall-clock time stays under 25 minutes per job
- [ ] Re-run CI 2-3 times to verify stability

## Summary by CodeRabbit

* **Tests**
  * Added separate "flaky" event test suites for NATS, Kafka and Redis; moved and expanded several scenarios into them, added explicit synchronization after message production, and adjusted read deadline handling.
* **Refactor**
  * Test helpers switched to condition-based waiting tied to engine statistics for more deterministic waits.
* **Chores**
  * CI adjustments: increased test timeouts, reduced parallelism, set GOMAXPROCS, and strengthened Kafka health checks and retries.

### DIFF
--- a/protographic/SDL_PROTO_RULES.md
+++ b/protographic/SDL_PROTO_RULES.md
@@ -293,9 +293,11 @@ message RequireProductNameByIdFields {
 
 Field resolvers allow you to define custom resolution logic for specific fields within a GraphQL type. Using the `@connect__fieldResolver` directive, you can specify which fields should be resolved through dedicated RPC methods, enabling lazy loading, computed fields, or integration with external data sources.
 
-### Basic Field Resolver
+Fields with the `@connect__fieldResolver` directive are excluded from the parent type's Proto message. This is the same behavior as fields with arguments, which are always treated as resolver fields. This ensures that expensive or lazily-loaded fields are not part of the regular response returned by operations — they are only resolved through their dedicated RPC methods.
 
-Fields marked with `@connect__fieldResolver` generate dedicated RPC methods with request and response messages:
+### Basic Field Resolver with Arguments
+
+Fields with arguments are automatically treated as resolver fields. Additionally, fields marked with `@connect__fieldResolver` generate dedicated RPC methods with request and response messages:
 
 ```graphql
 type User {
@@ -344,14 +346,61 @@ message ResolveUserPostsResponse {
 }
 ```
 
+### Field Resolver without Arguments
+
+The `@connect__fieldResolver` directive can also be used on fields **without arguments** to exclude expensive fields from the parent type's message and resolve them separately. In this case, no `Args` message is generated and the request message only contains the `context` field:
+
+```graphql
+type User {
+  id: ID!
+  name: String!
+  avatar: String! @connect__fieldResolver(context: "id")
+}
+
+type Query {
+  user(id: ID!): User
+}
+```
+
+Maps to:
+
+```protobuf
+rpc ResolveUserAvatar(ResolveUserAvatarRequest) returns (ResolveUserAvatarResponse) {}
+
+message ResolveUserAvatarContext {
+  string id = 1;
+}
+
+// Request message only includes context (no field_args)
+message ResolveUserAvatarRequest {
+  repeated ResolveUserAvatarContext context = 1;
+}
+
+message ResolveUserAvatarResult {
+  string avatar = 1;
+}
+
+message ResolveUserAvatarResponse {
+  repeated ResolveUserAvatarResult result = 1;
+}
+
+// Note: avatar is excluded from the User message
+message User {
+  string id = 1;
+  string name = 2;
+}
+```
+
 ### Field Resolver Components
 
-Each field resolver generates four message types:
+For field resolvers **with arguments**, four message types are generated:
 
 1. **Context Message** (`Resolve{Type}{Field}Context`): Contains fields from the parent type needed to resolve the field
 2. **Args Message** (`Resolve{Type}{Field}Args`): Contains the arguments passed to the field
 3. **Result Message** (`Resolve{Type}{Field}Result`): Contains the resolved field value
 4. **Request/Response Messages**: Standard request/response pattern for the RPC method
+
+For field resolvers **without arguments**, the `Args` message is omitted and the request message only contains the `context` field.
 
 ### Context Specification
 
--- a/protographic/src/sdl-to-mapping-visitor.ts
+++ b/protographic/src/sdl-to-mapping-visitor.ts
@@ -1,5 +1,4 @@
 import {
-  ArgumentNode,
   DirectiveNode,
   GraphQLEnumType,
   GraphQLField,
@@ -40,7 +39,7 @@ import {
   formatKeyElements,
   OperationTypeName,
 } from './naming-conventions.js';
-import { REQUIRES_DIRECTIVE_NAME } from './string-constants.js';
+import { CONNECT_FIELD_RESOLVER, REQUIRES_DIRECTIVE_NAME } from './string-constants.js';
 import { RequiredFieldsVisitor } from './required-fields-visitor.js';
 /**
  * Visitor that converts a GraphQL schema to gRPC mapping definitions
@@ -313,7 +312,8 @@ export class GraphQLToMappingVisitor {
       const fields = type.getFields();
 
       for (const field of Object.values(fields)) {
-        if (field.args.length === 0) {
+        const hasDirective = field.astNode?.directives?.some((d) => d.name.value === CONNECT_FIELD_RESOLVER);
+        if (field.args.length === 0 && !hasDirective) {
           continue;
         }
 
--- a/protographic/src/sdl-to-proto-visitor.ts
+++ b/protographic/src/sdl-to-proto-visitor.ts
@@ -1131,7 +1131,7 @@ Example:
       const fields = type.getFields();
 
       for (const field of Object.values(fields)) {
-        if (field.args.length === 0) {
+        if (field.args.length === 0 && !this.findResolverDirective(field)) {
           continue;
         }
 
@@ -1293,21 +1293,28 @@ Example:
       throw new Error(`Invalid field context for resolver. ${error}`);
     }
 
+    const hasArgs = field.args.length > 0;
     const argsMessageName = typeFieldArgsName(methodName);
     const contextMessageName = typeFieldContextName(methodName);
 
-    // Build the arguments message
+    // Build the arguments message (only when field has arguments)
     let fieldNumber = 0;
-    messageLines.push(
-      ...this.buildMessage({
-        messageName: argsMessageName,
-        fields: field.args.map((arg) => ({
-          fieldName: graphqlFieldToProtoField(arg.name),
-          typeName: this.getProtoTypeFromGraphQL(arg.type).typeName,
-          fieldNumber: ++fieldNumber,
-        })),
-      }),
-    );
+    if (hasArgs) {
+      messageLines.push(
+        ...this.buildMessage({
+          messageName: argsMessageName,
+          fields: field.args.map((arg) => {
+            const protoType = this.getProtoTypeFromGraphQL(arg.type);
+            return {
+              fieldName: graphqlFieldToProtoField(arg.name),
+              typeName: protoType.typeName,
+              isRepeated: protoType.isRepeated,
+              fieldNumber: ++fieldNumber,
+            };
+          }),
+        }),
+      );
+    }
 
     // filter the fields in the parent type that are in the context
     const searchFields = context.split(/[\s,]+/).filter((field) => field.length > 0);
@@ -1341,13 +1348,15 @@ Example:
       description: `${CONTEXT} provides the resolver context for the field ${field.name} of type ${parent.name}.`,
     });
 
-    // add the args message to the key message
-    keyMessageFields.push({
-      fieldName: FIELD_ARGS,
-      typeName: argsMessageName,
-      fieldNumber: ++fieldNumber,
-      description: `${FIELD_ARGS} provides the arguments for the resolver field ${field.name} of type ${parent.name}.`,
-    });
+    // add the args message to the key message (only when field has arguments)
+    if (hasArgs) {
+      keyMessageFields.push({
+        fieldName: FIELD_ARGS,
+        typeName: argsMessageName,
+        fieldNumber: ++fieldNumber,
+        description: `${FIELD_ARGS} provides the arguments for the resolver field ${field.name} of type ${parent.name}.`,
+      });
+    }
 
     // build the actual request message
     messageLines.push(
@@ -1357,7 +1366,9 @@ Example:
       }),
     );
 
-    this.processedTypes.add(argsMessageName);
+    if (hasArgs) {
+      this.processedTypes.add(argsMessageName);
+    }
     this.processedTypes.add(requestName);
 
     return messageLines;
@@ -1531,9 +1542,9 @@ Example:
     }
 
     const allFields = Object.values(type.getFields());
-    // Filter out fields that have arguments as those are handled in separate resolver rpcs
+    // Filter out fields that have arguments or @connect__fieldResolver as those are handled in separate resolver rpcs
     const validFields = allFields
-      .filter((field) => field.args.length === 0)
+      .filter((field) => field.args.length === 0 && !this.findResolverDirective(field))
       .filter(
         (field) =>
           !field.astNode?.directives?.some(
--- a/protographic/src/sdl-validation-visitor.ts
+++ b/protographic/src/sdl-validation-visitor.ts
@@ -346,7 +346,15 @@ export class SDLValidationVisitor {
    * @private
    */
   private validateInvalidResolverContext(ctx: VisitContext<FieldDefinitionNode>): void {
-    if (ctx.node.name.value.startsWith('_') || (ctx.node.arguments?.length ?? 0) === 0) {
+    if (ctx.node.name.value.startsWith('_')) {
+      return;
+    }
+
+    const hasArgs = (ctx.node.arguments?.length ?? 0) > 0;
+    const hasResolverDirective = ctx.node.directives?.some((d) => d.name.value === CONNECT_FIELD_RESOLVER) ?? false;
+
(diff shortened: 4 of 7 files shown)


router-tests/observability/main_test.go
```diff
@@ -0,0 +1,12 @@
...
+import (
+	"os"
+	"testing"
+)
+
+func TestMain(m *testing.M) {
+	// Change to router-tests root so relative testdata paths resolve correctly
+	os.Chdir("..")
+	os.Exit(m.Run())
+}
```
endigma: Can testdata for each new sub-suite be colocated into it's folder instead of all the TestMain functions?

router-tests/events/nats_events_test.go
```diff
@@ -1263,87 +1170,6 @@ func TestNatsEvents(t *testing.T) {
 		})
 	})
 
-	t.Run("message with invalid JSON should give a specific error", func(t *testing.T) {
```
endigma, line 1380: Where did this test go?

router/pkg/statistics/engine_stats.go
```diff
@@ -18,10 +18,12 @@ type EngineStatistics interface {
 	SubscriptionCountDec(count int)
 	TriggerCountInc(count int)
 	TriggerCountDec(count int)
+	Wait(ctx context.Context, predicate func(*UsageReport) bool)
 }
 
 type EngineStats struct {
 	mu            sync.Mutex
+	cond          *sync.Cond
```
endigma: If `sync.Cond` is needed here, usually the internal locker is accessed as `cond.L` to make behaviour clear. Is sharing the mu from EngineStats useful here? It's also not clear to me what `mu` is protecting, all attributes are atomic or not mutated as far as I can tell

router-tests/utils.go
```diff
@@ -17,6 +17,9 @@ import (
 
 const (
 	JwksName = "my-jwks-server"
+
+	EmployeesIDData  = `{"data":{"employees":[{"id":1},{"id":2},{"id":3},{"id":4},{"id":5},{"id":7},{"id":8},{"id":10},{"id":11},{"id":12}]}}`
```
endigma: Should this be stored alongside the query/server that produces it?

router/pkg/statistics/engine_stats.go
```diff
@@ -126,6 +163,10 @@ func (s *NoopEngineStats) GetReport() *UsageReport {
 	return nil
 }
 
+func (s *NoopEngineStats) Wait(ctx context.Context, _ func(*UsageReport) bool) {
+	<-ctx.Done()
+}
```
endigma: Should this be waiting on anything? Seems like a noop implementation should either return instantly or never return if `Wait()`-ed unless we implement the synchronization here too

router/pkg/statistics/engine_stats.go
```diff
@@ -62,6 +65,33 @@ func (s *EngineStats) GetReport() *UsageReport {
 	return report
 }
 
+// Wait blocks until predicate returns true for the current stats or ctx is cancelled.
+func (s *EngineStats) Wait(ctx context.Context, predicate func(*UsageReport) bool) {
```
endigma: Could this return the report that satisfied the predicate to avoid weird races where say, a rapidly increasing connection count causes a mismatch between `Wait` and `GetReport`?

router-tests/events/nats_events_test.go
endigma: If the events tests are being fully refactored including removing tests, etc, I believe the PR description should include an overview of test coverage, what was removed or added and why.

router/pkg/statistics/engine_stats.go
```diff
@@ -62,6 +65,30 @@ func (s *EngineStats) GetReport() *UsageReport {
...
+		close(done)
+	}()
+
+	select {
+	case <-done:
+	case <-ctx.Done():
+		// Unblock the goroutine waiting on cond.Wait()
+		s.cond.Broadcast()
+	}
+}
```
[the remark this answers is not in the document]
jensneuse: Fixed in 9725bcc72. The goroutine now checks `ctx.Err()` in the loop condition, and the caller waits for the goroutine to exit after broadcasting. Also changed `Wait` to return `*UsageReport` to avoid the TOCTOU race (see endigma's comment below).

.github/workflows/router-ci.yaml
```diff
@@ -369,7 +370,9 @@ jobs:
 
       - name: Run Integration tests ${{ matrix.test_target }}
         working-directory: ./router-tests
-        run: make test-coverage test_retry_count=0 test_params="-run '^Test[^(Flaky)]' --timeout=5m -p 1 --parallel 10" test_target="${{ matrix.test_target }}"
+        env:
+          GOMAXPROCS: 4
+        run: make test-coverage test_retry_count=0 test_params="-run '^Test[^(Flaky)]' --timeout=8m -p 1 --parallel 2" test_target="${{ matrix.test_target }}"
```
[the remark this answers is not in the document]
jensneuse: Already fixed — the current code uses `-run '^Test' -skip '^TestFlaky'` which correctly includes all non-flaky tests.

router/pkg/statistics/engine_stats.go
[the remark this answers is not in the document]
jensneuse: This was reviewed against an older commit. The `TestFlakyKafkaEvents` and `TestFlakyRedisEvents` functions no longer exist — those tests were folded back into their parent suites.

router-tests/operations/singleflight_test.go
```diff
@@ -563,7 +563,7 @@ func TestSingleFlight(t *testing.T) {
 
 			// Wait for all subscriptions to be established before triggering
 			go func() {
-				xEnv.WaitForSubscriptionCount(uint64(numOfOperations), time.Second*5)
+				xEnv.WaitForSubscriptionCount(uint64(numOfOperations), time.Second*15)
```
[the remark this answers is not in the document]
jensneuse: Already fixed — all four `WaitForSubscriptionCount` calls in singleflight_test.go now use `time.Second*15`.

router-tests/observability/main_test.go
jensneuse: The `TestMain` + `os.Chdir("..")` pattern exists because many tests reference shared `testdata/` files at the router-tests root (e.g. config templates, schemas). Colocating testdata into each subdirectory would require duplicating those shared fixtures or restructuring the test data layout — a larger refactoring that's out of scope for this PR. Happy to discuss as a follow-up.

router-tests/events/nats_events_test.go
jensneuse: Moved to `TestFlakyNatsEvents` at line 1691 (same file). It's a test for invalid JSON messages over NATS subscriptions — it was moved into the flaky suite because it depends on subscription pipeline timing and can intermittently fail under CI pressure. The test content is unchanged.

router/pkg/statistics/engine_stats.go
jensneuse: Good point. Fixed in 9725bcc72 — removed the shared `mu` field and now use `sync.NewCond(&sync.Mutex{})` so the cond owns its own locker. The `cond.L` is accessed directly in `Wait()`. The mu was only there for the cond's locker; the actual stats fields are all atomic.

router-tests/utils.go
jensneuse: Fair point, but `EmployeesIDData` is a shared test constant used across many subdirectories (operations, protocol, security, etc.) — it's the canonical expected response for the employees subgraph fixture. Keeping it in the root `utils.go` avoids import cycles and duplication. The test server that produces it lives in `testenv/` which is a separate package.

router/pkg/statistics/engine_stats.go
jensneuse: Fixed in 9725bcc72. `NoopEngineStats.Wait` now returns immediately with `nil` — since no stats will ever change, blocking until context cancellation was wrong (it would cause tests using the noop implementation to hang).

router/pkg/statistics/engine_stats.go
jensneuse: Great suggestion — fixed in 9725bcc72. `Wait` now returns `*UsageReport` (the report that satisfied the predicate, or the last report on context cancellation). All callers in testenv updated to use the returned report instead of calling `GetReport()` separately.

router-tests/events/nats_events_test.go
jensneuse: Fair feedback. No tests were removed — they were reorganized:

- **Moved to `TestFlakyNatsEvents`**: "message with invalid JSON should give a specific error" and "subscribe_to_multiple_subjects" — both depend on subscription pipeline timing and were intermittently failing under CI pressure. Content is unchanged.
- **Moved to `TestFlakyKafkaEvents`** (later folded back): "every subscriber gets the message" and filter-based tests — same timing sensitivity.  
- **Moved to `TestFlakyRedisEvents`** (later folded back): SSE-based tests with multi-client lifecycle.

The `TestFlaky` prefix gives these tests 3 retries in CI (via `test_retry_count=3`) while the root cause timing issues are addressed. All test logic is preserved.

endigma:
Have you considered using something like a "SyncReporter" that fulfills the Reporter interface in these tests instead of the hacked on synchronisation to EngineStats? 

It would remove a lot of the complexity around awaiting for the engine to report something if it was directly channel based and blocking. 

I don't believe any of the methods on reporter are called on a timer or anything but if they are then it could either be configurable which report functions to care about or separate reporters that block on each signal or a mask of signals to mitigate the issue.

router-tests/observability/graphql_metrics_test.go
```diff
@@ -57,12 +59,12 @@ func TestGraphQLMetrics(t *testing.T) {
 		res := xEnv.MakeGraphQLRequestOK(testenv.GraphQLRequest{
 			Query: `query { employees { id } }`,
 		})
-		require.JSONEq(t, employeesIDData, res.Body)
+		require.JSONEq(t, integration.EmployeesIDData, res.Body)
```
jensneuse: just make this inline

router-tests/observability/main_test.go
endigma: I doubt most things are shared between entirely different suites of tests, if there is anything shared it can be kept but I'd like to see proof of which testdata is referenced so much

router-tests/CLAUDE.md
StarpTech, line 1: Link it to `AGENTS.md` so other agent tools are supported.

router-tests/events/nats_events_test.go
```diff
@@ -60,9 +60,8 @@ const NatsWaitTimeout = time.Second * 30
 
 func assertNatsLineEquals(t *testing.T, reader *bufio.Reader, expected string) {
 	t.Helper()
-	line, _, err := reader.ReadLine()
-	assert.NoError(t, err)
-	assert.Equal(t, expected, string(line))
+	line := testenv.ReadSSELine(t, reader)
+	assert.Equal(t, expected, line)
 }
```
Noroth: This function is identical with the functions in redis and kafka. Should we maybe have one function instead?

router-tests/observability/graphql_metrics_test.go
```diff
@@ -1,6 +1,8 @@
 package integration
 
 import (
+	integration "github.com/wundergraph/cosmo/router-tests"
+
```
Noroth, line 5: package name and import identical. Looks a bit ugly

router-tests/protocol/integration_test.go
```diff
@@ -1,6 +1,8 @@
-package integration
+package integration // trigger CI 2
```
endigma: ci trigger comment should be removed 😄

router-tests/protocol/router_config_test.go
```diff
@@ -14,7 +14,7 @@ import (
 func TestRouterConfigParsing(t *testing.T) {
 	t.Parallel()
 
-	routerConfig, err := execution_config.FromFile("./testdata/routerConfig.json")
+	routerConfig, err := execution_config.FromFile("testdata/routerConfig.json")
```
endigma: nit: this is making a diff for no benefit

router-tests/persisted_operations_test.go
```diff
@@ -17,17 +17,13 @@ import (
 	"github.com/wundergraph/cosmo/router/pkg/config"
 )
 
-// cacheHashNotStored is a constant representing a non-existent entry in Persisted Queries Cache
-// of the testenv. All the pre-stored persistent entries used in testing are located here:
-// ./testenv/testdata/cdn/organization/graph/operations/my-client
-const cacheHashNotStored = "0000000000000000000000000000000000000000000000000000000000000000"
```
Noroth, line 23: Any reason why we deleted the variable and are hardcoding the "sha" now? Also the comment made sense to understand what is happening in the tests.

router/core/router.go
```diff
@@ -1062,8 +1062,10 @@ func (r *Router) bootstrap(ctx context.Context) error {
 		r.connectRPCServer = crpcServer
 	}
 
-	if r.metricConfig.OpenTelemetry.EngineStats.Enabled() || r.metricConfig.Prometheus.EngineStats.Enabled() || r.engineExecutionConfiguration.Debug.ReportWebSocketConnections {
-		r.EngineStats = statistics.NewEngineStats(ctx, r.logger, r.engineExecutionConfiguration.Debug.ReportWebSocketConnections)
+	if _, isNoop := r.EngineStats.(*statistics.NoopEngineStats); isNoop {
+		if r.metricConfig.OpenTelemetry.EngineStats.Enabled() || r.metricConfig.Prometheus.EngineStats.Enabled() || r.engineExecutionConfiguration.Debug.ReportWebSocketConnections {
+			r.EngineStats = statistics.NewEngineStats(ctx, r.logger, r.engineExecutionConfiguration.Debug.ReportWebSocketConnections)
+		}
```
Noroth, line 1068: Can we wrap this whole if in a separate receiver function? This is a bit of mental overload to read

router-tests/testutils/utils.go
```diff
@@ -17,6 +17,8 @@ import (
 
 const (
 	JwksName = "my-jwks-server"
+
+	EmployeesIDData = `{"data":{"employees":[{"id":1},{"id":2},{"id":3},{"id":4},{"id":5},{"id":7},{"id":8},{"id":10},{"id":11},{"id":12}]}}`
```
Noroth, line 21: With the new structure it might make sense to have a package `testutils` or so.

router-tests/protocol/integration_test.go
Noroth: AI comment?

router-tests/events/nats_events_test.go
jensneuse: yes, fixed

router-tests/observability/graphql_metrics_test.go
jensneuse: fixed

router/core/router.go
jensneuse: yes, fixed

router-tests/protocol/integration_test.go
```diff
@@ -1251,7 +1252,7 @@ func TestOperationSelection(t *testing.T) {
 func TestTestdataQueries(t *testing.T) {
 	t.Parallel()
 
-	testDir := filepath.Join("testdata", "queries")
```
endigma, line 1254: 1. filepath.Join is better
2. testDir is not used below in goldie argument

router-tests/testenv/sync_reporter.go
```diff
@@ -0,0 +1,132 @@
...
+	Kind  EventKind
+	Count int
+}
+
+// SyncReporter wraps EngineStats and emits channel-based events for test
+// synchronization. It satisfies statistics.EngineStatistics via delegation
+// to the inner EngineStats, while additionally emitting events on a buffered
+// channel that tests can select on directly.
+type SyncReporter struct {
+	inner  *statistics.EngineStats
```
endigma: Does this need the inner EngineStats with the sync.Cond and etc or could we use purely the channel based reporter in tests?

jensneuse: the inner EngineStats is needed because tests currently depend on reliable snapshot-based waits

endigma: Could wait be a function of syncreporter based on consuming events instead of sync.Cond?

Im not sure the current EngineStats needs to be modified to achieve the desired Wait functionality

endigma: Or even identical sync.Cond mechanism but in a wrapper instead of directly in the one used in production

PR #2579, opened 2026-03-03
feat: add format script + optional git hooks

comatory:
## Summary by CodeRabbit

* **New Features**
  * Added pre-commit and commit-msg hooks for formatting and commit message linting.
  * Introduced unified "format" scripts across packages and a workspace-level format command.

* **Documentation**
  * Updated contributing guide with optional hook setup and installation instructions.

* **Chores**
  * Added Makefile formatting targets and expanded workspace package list.

* **Bug Fixes**
  * Fixed several GraphQL schema/mutation syntax issues in tests and fixtures.

## Checklist

- [ ] Tests or benchmarks have been added or updated. (N/A)
- [ ] Documentation has been updated on [https://github.com/wundergraph/cosmo-docs](https://github.com/wundergraph/cosmo-docs). (N/A)

Adds consistent way of code formatting for both JS/TS packages and Go packages.
JS/TS is relying on prettier, it loads the common configuration from the root
config.
For Go, we're relying on standard `gofmt` features. Both `pnpm` and `make` can be
used. Examples:

`pnpm --filter=./cli format`

or

`make format router`

Additionally, one can run `make format-all` to format the whole monorepo. Some
adjustments might be needed, if we want to ignore certain folders/files, but
it's a good start.

There are also git hook templates which can be installed via `pnpm husky`, in
conjuction with `lint-staged`, any files can be automatically reformatted
before commiting. __This is optional__, you can use the hooks or not, it's
up to you.

One minor adjustment to `controlplane`: running tests will no longer attempt
to format code. This was a bit inconvenient as it would stop you from
running tests while you were working on them, if your code contained formatting
issues.
I removed `prettier` dependency from `controlplane`, as it was causing two
different versions of formatter to be installed. I think we should use just 
a single version.

I had to run the formatter on `controlplane` to make it pass. For `studio` and
other packages, I'll probably run them as separate fixes in other PRs.


package.json
```diff
@@ -23,6 +23,7 @@
     "generate": "pnpm buf generate --path proto/wg/cosmo/platform --path proto/wg/cosmo/notifications --path proto/wg/cosmo/common --path proto/wg/cosmo/node --template buf.ts.gen.yaml && pnpm -r run --filter './connect' build",
     "husky": "husky install",
     "lint:fix": "pnpm run -r --parallel lint:fix",
+    "format": "pnpm run -r --parallel format",
```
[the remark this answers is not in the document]
comatory: It's ok I want to know if it breaks.

.prettierignore
```diff
@@ -22,8 +22,3 @@ logs
 
 CHANGELOG.md
 .husky
-
-
-controlplane/migrations
```
thisisnithin, line 27: These are mostly generated with drizzle, I would keep it ignored

comatory: I just co-located it with package: https://github.com/wundergraph/cosmo/pull/2579/changes#diff-cad0a0212bbb42aac9846fd911010137be6954f1173d928f4526dacfd3491289R1

dkorittki (commented):
Thanks for the work :) Would it be possible to also add formatting as a Makefile target for Go projects` ? It's what Go devs use for most tasks. Something like

```Makefile
format:
	go fmt ./...
```

Affected Makefiles should be in router/, router-tests/, composition-go/ and graphqlmetrics/

aws-lambda-router/package.json
```diff
@@ -11,6 +11,9 @@
     "aws-lambda",
     "serverless"
   ],
+  "scripts": {
+    "format": "find . -type f -name '*.go' -not -path './gen/*' -not -path './vendor/*' -exec gofmt -w {} +"
```
dkorittki: Why not

```suggestion
    "format": "go fmt ./..."
```

Works recursively out of box. I don't think we need to exclude `gen` and `vendor`. Saw this pattern multiple times in Go related directories and just mention this one as an example.

comatory: ah ok, I'll fix it and make it same across the board

composition/CHANGELOG.md
endigma, line 1: We should exclude the changelog from formatting it is automatically generated AFAIK

client-tests/apollo-js/package.json
```diff
@@ -4,6 +4,7 @@
   "description": "Apollo Client tests for GraphQL",
   "main": "index.js",
   "scripts": {
+    "format": "prettier -w .",
```
endigma, line 7: Why in some packages is it `-w -c` and here only `-w`?

Makefile
```diff
@@ -52,6 +52,24 @@ infra-debug-down-v:
...
+		exit 1; \
+	fi; \
+	pnpm --filter "./$$package" --fail-if-no-match run format
+
+ifneq ($(filter format,$(MAKECMDGOALS)),)
+FORMAT_ARGS := $(wordlist 2,$(words $(MAKECMDGOALS)),$(MAKECMDGOALS))
+.PHONY: $(FORMAT_ARGS)
+$(FORMAT_ARGS):
+	@:
+endif
```
endigma, line 71: is this necessary

comatory: I thought it was needed so I can pass package name: `make format <package-name>` and have ability to reject unknown packages. Which exact parts are unneeded.

client-tests/apollo-js/package.json
comatory: `-w -c` is not supposed to be used. You either use write or check. I tried fixing it where I could but maybe some leftovers are still there.

composition/CHANGELOG.md
comatory: Fixed in 11c8af3b5ef4e98188b6db508f9ebc11e2ff94dc

client-tests/apollo-js/package.json
comatory: I'll do a follow up to review these usages.

Makefile
comatory: That extra code is there to allow for positional arguments and only consider the first one. Let me know if you have better solution. I think `make format router` is nicer than `make format package=router` but I'm ok with whatever.

Not a big user of `Make` though, so maybe I'm fighting conventions here. Let me know.

StarpTech (changes requested):
Hi, this is going to touch the history of all files. Can we just introduce without touching all files?

Makefile
endigma: I don't really use the Makefile for anything anymore, I have custom task configs locally. I'll defer to your judgement here w.r.t. usefulness

comatory:
@StarpTech 

> Hi, this is going to touch the history of all files. Can we just introduce without touching all files?

In this phase, it touches just the controlplane, see reasones below. I looked at this closely to see if I can avoid touching the files, but I'm not sure if it will be possible. There are several problems with current formatter:

- mismatch between prettier versions in root `package.json` vs `controlplane/package.json`: this causes different formatting. I mean I could leave it in, but we want consistent formatting across the codebase right?
- `controlplane` enforces formatting in CI in very indirect way, when it runs `pnpm run --filter ./controlplane/emails build`, this script in turns calls `lint:fix,` which runs prettier but in `controlplane` context. This means it forces formatting in CI to be correct. This is good, but we're not consistent in other packages.

So it's really just inconsistency. One of the reasons I'm introducing the `format` scripts is that we could enforce it in CI via unified command. For example, `studio`'s CI pipeline which runs linter does not use prettier at all. So currently, even with the set-up we have, there are 271 badly formatted files for Studio alone, because of the drift.

We can either merge it with the changes we have, or I can remove the `lint:fix` from `controlplane`'s CI. But if we want consistent formatting, we'll have to correct the formatting eventually, so it'll blur history anyway.

Makefile
comatory: I based it on this feedback: https://wundergraph.slack.com/archives/C03PE1KU073/p1772529878252799?thread_ts=1772464905.358179&cid=C03PE1KU073

PR #2583, opened 2026-03-04
chore(router): gofmt check in CI (#2576)

dkorittki:
## Summary by CodeRabbit

* **Chores**
  * Added automated Go code formatting validation to the CI/CD pipeline
  * Added code formatting build targets across project modules

## Checklist

Adds `gofmt` validation in go-linter Github action, which is enforced in router/, router-tests/, aws-lambda-router/, graphqlmetrics/ and composition-go/ directories

### DIFF
--- a/controlplane/package.json
+++ b/controlplane/package.json
@@ -62,7 +62,7 @@
     "@wundergraph/cosmo-connect": "workspace:*",
     "@wundergraph/cosmo-shared": "workspace:*",
     "@wundergraph/protographic": "workspace:*",
-    "axios": "^1.13.5",
+    "axios": "1.13.5",
     "axios-retry": "^4.5.0",
     "bullmq": "^5.10.0",
     "cookie": "^0.7.2",
--- a/playground/package.json
+++ b/playground/package.json
@@ -49,7 +49,7 @@
     "@radix-ui/react-slot": "1.0.2",
     "@radix-ui/react-tabs": "^1.0.4",
     "@radix-ui/react-tooltip": "^1.0.7",
-    "axios": "^1.12.2",
+    "axios": "1.13.5",
     "change-case": "^5.2.0",
     "class-variance-authority": "^0.7.0",
     "clsx": "^2.0.0",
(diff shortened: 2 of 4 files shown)


.github/actions/go-linter/action.yaml
```diff
@@ -21,3 +21,14 @@ runs:
...
+    - name: Check gofmt formatting
+      run: |
+        files=$(gofmt -l .)
+        if [ -n "$files" ]; then
+          echo "The following files are not gofmt-formatted:"
+          echo "$files"
+          exit 1
+        fi
+      shell: bash
+      working-directory: ${{ inputs.working-directory }}
```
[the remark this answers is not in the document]
dkorittki: `gofmt` != `go fmt`. `gofmt` does not understand `./...` as `go fmt` does. `gofmt` works recursively out of the box.

dkorittki:
fyi I removed the Makefile, husky, etc. stuff because https://github.com/wundergraph/cosmo/pull/2579 already adds that

comatory:
If we merge my PR first, you could then just call `make format <package>`, this way we have the formatter centralized. Up to you though.

.github/actions/go-linter/action.yaml
```diff
@@ -21,3 +21,14 @@ runs:
         version: 2025.1.1
         install-go: false
         working-directory: ${{ inputs.working-directory }}
+
+    - name: Check gofmt formatting
```
comatory, line 25: I believe `make format <package-name>` is now implemented. This action could accept package name as an input and just run this make task.

comatory (changes requested):

.github/actions/go-linter/action.yaml
dkorittki: The action intends to fail CI if there are unformatted files and list affected files for convinience. `make format router` would actually format them, then exits with exit code `0`. I'd have to check with git diff inside this action if files actually change, making it more complicated than it is. `gofmt -l` is exactly intended for what this action intends but has no `make` target because a dev would rather use `make format router` locally. Imo it's better to let this stay as is.

PR #2590, opened 2026-03-05
feat: json output support for monograph/subgraph check

comatory:
## Summary by CodeRabbit

* **New Features**
  * Added a JSON output option (-j / --json) for check commands to produce machine-readable results.

* **Tests**
  * Expanded in-process CLI test suites covering many schema-check scenarios, error conditions, and both JSON and human-readable output validation.

* **Chores**
  * Updated formatting tooling and scripts, README formatting, gitignore to ignore coverage, adjusted TypeScript project config, and minor stylistic cleanups.

## Checklist

- [X] Documentation has been updated on https://github.com/wundergraph/cosmo-docs/pull/258

Adds `--json` flag to check commands for monograph/subgraph.
The way I implemented this was to first create comprehensive test suite for existing
`check` subcommand and assert on stdout.
Once that was done, I added JSON support and then I did a somewhat larger refactoring
because the handler function was hard to follow and reason about.

<details>
  <summary>JSON schema of the check result</summary>
  <code lang="graphql">
  export type JsonOutputDescriptor = {
  status: 'error' | 'success';
  code: EnumStatusCode;
  details?: string;
  message?: string;
  url?: string;
  proposals?: {
    message: string;
  };
  traffic?: {
    message: string;
  };
  changes?: {
    breaking: SchemaChange[];
    nonBreaking: SchemaChange[];
  };
  composition?: {
    errors: CompositionError[];
    warnings: CompositionError[];
  };
  lint?: {
    errors: LintIssue[];
    warnings: LintIssue[];
  };
  graphPrune?: {
    errors: GraphPruningIssue[];
    warnings: GraphPruningIssue[];
  };
  extensions?: {
    message: string;
  };
  exceededRowLimit?: boolean;
  rowLimit: number;
  operationUsageStats?: CheckOperationUsageStats;
};
</code>
</details>

> [!CAUTION]
> There are some changes in other CLI files but they are related to formatting. The issue was that some
> files were being ignored by the formatter (tests).

## Question(s) for reviewers?

1. ~~The JSON output is separated into sections, like `lint: { warning: ..., error: ..., success: ...}`. This way I figured it would be nice to just check the `.success` property from DX point of view. I consider `success=false` when either the response has warnings or errors, but Coderabbit raised a point that perhaps that's not correct. What would you propose? My thinking was that it's only successful if no warnings or errors are present, but it depends on what we agree on.~~ @JivusAyrus and me talked, we decided the nested `success` properties are unnecessary.
2. ~~I noticed when I pass `--skip-traffic-check`, the response still contains `operationUsageStats`, which means that `.traffic` property gets passed to the output. I should probably detect the flag and omit it completely, correct? Just want to make sure that `operationUsageStats` are related to this flag.~~ I ended up checking whether `--skip-traffic-check` flag is set and exclude the traffic output. This makes it consistent with stdout.

## How to test

1. Have Cosmo stack running locally with subgraphs (demo subgraphs will work fine)
3. Open studio, enable linter in Policies
4. Pull the branch
5. Run in `cli/` package directory _(running the the source via tsx gets rid of pnpm-related
    error messages which would disallow you piping the JSON to a file)_:

```
./node_modules/.bin/tsx  --env-file=.env src/index.ts \
    subgraph check employeesupdated \
    --schema=../demo/pkg/subgraphs/employeeupdated/subgraph/schema.graphqls \
    --json > /tmp/out.json
```

5. The `/tmp/out.json` should have data in it, like lint warnings (depending on other
   settings you might have)


cli/test/grpc-service.test.ts
```diff
@@ -151,24 +120,15 @@ describe('gRPC Generate Command', () => {
...
-          outputFile,
-        ],
-        {
-          from: 'user',
-        }
-      )).rejects.toThrow('process.exit unexpectedly called with "1"');
+      program.parseAsync(['generate', 'testservice', '-i', 'test/fixtures/full-schema.graphql', '-o', outputFile], {
+        from: 'user',
+      }),
+    ).rejects.toThrow('process.exit unexpectedly called with "1"');
```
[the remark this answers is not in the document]
comatory: I wasn't touching this file, only changing formatting.

cli/src/handle-check-result.ts
```diff
@@ -1,365 +1,738 @@
...
+  shouldOutputJson?: boolean;
+  outFile?: string;
+}): Promise<boolean> => {
+  const jsonBuilder = new JsonOutputBuilder(EnumStatusCode.ERR, rowLimit, outFile);
 
-      if (resp.checkExtensionErrorMessage) {
-        success = false;
-        finalStatement += `\n${logSymbols.error} Subgraph extension check failed with message: ${resp.checkExtensionErrorMessage}`;
-      }
+  let studioCheckDestination = '';
```
StarpTech, line 484: Was it necessary to touch the logic here?

cli/src/handle-check-result.ts
```diff
@@ -1,365 +1,738 @@
...
+  extensions?: {
+    success: boolean;
+    message: string;
+  };
+  exceededRowLimit?: boolean;
+  rowLimit: number;
+  operationUsageStats?: CheckOperationUsageStats;
+};
+
+class JsonOutputBuilder {
```
StarpTech: Could you put this in a separate file with a test?

cli/src/handle-check-result.ts
```diff
@@ -1,365 +1,738 @@
...
 
-export const handleCheckResult = (resp: CheckSubgraphSchemaResponse, rowLimit: number) => {
+export type JsonOutputDescriptor = {
+  status: 'error' | 'success';
+  code: EnumStatusCode;
+  details?: string;
+  message?: string;
+  url?: string;
+  proposals?: {
+    success: boolean;
```
StarpTech: What does it mean?

cli/src/handle-check-result.ts
```diff
@@ -1,365 +1,738 @@
...
+    message: string;
+  };
+  traffic?: {
+    success: boolean;
+    isLinkedToTargetSubgraph: boolean;
```
StarpTech: Do we need to expose it? I'd try to come up with the simplest data model and then iterate on feedback.

cli/src/handle-check-result.ts
comatory: I did not change the logic, this function had hundreds of lines inside of `case` statement and it was hard to follow, so it's more of a refactor. Like I mentioned in the description, I made sure to write unit tests before refactoring.

With JSON output, there is additional path the code can take, so I wanted to avoid adding another levels of `if-else` statements in already large code block.

I think it's now more maintainable, each checking step has its own function (breaking changes etc), I think long term it will be better.

cli/src/handle-check-result.ts
comatory: Yeah no problem.

cli/src/handle-check-result.ts
comatory: Ok I will omit it.

cli/src/handle-check-result.ts
comatory: That's the opened question (see PR description). My thinking was this: the schema checks have different parts, proposals being one of them, others being breaking and non-breaking changes. Some of these have also warnings vs errors. So instead of these being top-level properties, I separated them into buckets and each of them have their success state.

Do you think it's unnecessarily complex? I'm OK with removing these `success` properties.

cli/src/commands/graph/monograph/commands/check.ts
```diff
@@ -21,8 +21,11 @@ export default (opts: BaseCommandOptions) => {
     'This will skip checking for client traffic and any breaking change will fail the run.',
   );
   command.option('-l, --limit [number]', 'The amount of entries shown in the schema checks output.', '50');
+  command.option('-j, --json', 'Prints to the console in json format instead of table');
+  command.option('-o, --out [string]', 'Destination file for the json output.');
```
Noroth, line 25: You could define the format in -o, if we ever decide to use other formats like yaml.

* `-o json`
* `-o yaml`
* `-o table` <- default

and you could have `--output-file` for writing to a file. What do you think?

cli/src/handle-check-result.ts
comatory: Perhaps, next time it's better if I do these changes via stacked PR (with first one just refactor). WDYT?

cli/src/commands/graph/monograph/commands/check.ts
comatory: For sure, we could be more flexible here. I just wonder whether it's in scope of this PR to introduce these additional changes when nobody asked for them yet (I'm not against it however).

@StarpTech told me to make it consistent with other usages of `--out/-o` flag, which already exists in namespaces list subcommand. So changing it here would make it inconsistent IMO.

I'd rather reserve `--format` flag instead for this use case. We could then introduce it to other subcommands if needed. WDYT?

cli/src/handle-check-result.ts
comatory: This logic is also indirectly tested with new unit tests I added, so we test this there thoroughly as well. But it does not hurt to unit test the class itself as well.

cli/src/commands/graph/monograph/commands/check.ts
Noroth: format is fine too. It's just reducing the number of potential arguments.

comatory: We could even select the format automatically based on extension. Lot of possibilities here, thanks for the input.

comatory: I created a ticket for it and we can re-visit this any time in the future: https://linear.app/wundergraph/issue/ENG-9129/cli-provide-more-output-formats

cli/src/handle-check-result.ts
```diff
@@ -1,365 +1,714 @@
...
+
+  addCompositionWarnings(warnings: CompositionError[]): this {
+    this.data.composition = {
+      ...this.data.composition,
+      success: false,
+      errors: [...(this.data.composition?.errors ?? [])],
+      warnings: [...(this.data.composition?.warnings ?? []), ...warnings],
+    };
+    return this;
+  }
```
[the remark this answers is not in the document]
comatory: Waiting for feedback regarding the `success` properties.

cli/src/handle-check-result.ts
```diff
@@ -1,365 +1,714 @@
...
+
+  addLintWarnings(warnings: LintIssue[]): this {
+    this.data.lint = {
+      ...this.data.lint,
+      success: false,
+      errors: [...(this.data.lint?.errors ?? [])],
+      warnings: [...(this.data.lint?.warnings ?? []), ...warnings],
+    };
+    return this;
+  }
```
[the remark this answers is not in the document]
comatory: Waiting for feedback regarding the `success` properties.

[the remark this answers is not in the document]
comatory: we will no longer use `success` prop here

comatory:
> composedSchemaBreakingChanges are not handled

This was not handled by the subgraph check subcommand so far. Could we do it as a follow up PR? Or maybe add it as a new feature, I think the changeset here is already quite big.

JivusAyrus:
> > composedSchemaBreakingChanges are not handled
> 
> This was not handled by the subgraph check subcommand so far. Could we do it as a follow up PR? Or maybe add it as a new feature, I think the changeset here is already quite big.

It is handled by the subgraph check command, it was pushed a few days ago. We will need to do it in this pr, as that would cause a break in functionality.

comatory:
> > > composedSchemaBreakingChanges are not handled
> > 
> > This was not handled by the subgraph check subcommand so far. Could we do it as a follow up PR? Or maybe add it as a new feature, I think the changeset here is already quite big.
> 
> It is handled by the subgraph check command, it was pushed a few days ago. We will need to do it in this pr, as that would cause a break in functionality.

Oh ok I was not aware of that. I will have to check the existing implementation, I did some major refactoring so hopefully I'm able to transfer it.

comatory:
@JivusAyrus Ok I had to backport the changes from main. I guess due to my refactoring, it was not caught as breaking change as we were working on the same file at the same time, anyways it's good you caught it.
You can review the changes in 0655b96e6ae896c7417880956603a74121184477 (`feat: backport composedSchemaBreakingChanges`)

cli/src/handle-check-result.ts
```diff
@@ -1,16 +1,145 @@
...
+      jsonBuilder.setOperationUsageStats(operationUsageStats);
+    }
+    if (!shouldOutputJson) {
+      console.log(warningMessage.join(''));
+    }
+
+    finalStatement = `This check has encountered ${pc.bold(`${totalBreakingChanges}`)} breaking changes${
+      clientTrafficCheckSkipped ? `.` : ` that would break operations from existing client traffic.`
+    }`;
+  } else if (!clientTrafficCheckSkipped) {
```
JivusAyrus: this is not possible, the operations will exist only when breaking changes exist

cli/src/json-check-schema-output-builder.ts
```diff
@@ -0,0 +1,211 @@
...
+  };
+  graphPrune?: {
+    errors: GraphPruningIssue[];
+    warnings: GraphPruningIssue[];
+  };
+  composedSchemaBreakingChanges?: FederatedGraphSchemaChange[];
+  extensions?: {
+    message: string;
+  };
+  exceededRowLimit?: boolean;
```
StarpTech, line 44: What does this mean? Should this really be part of the model?

cli/src/json-check-schema-output-builder.ts
```diff
@@ -0,0 +1,211 @@
...
+  CompositionError,
+  FederatedGraphSchemaChange,
+  GraphPruningIssue,
+  LintIssue,
+  SchemaChange,
+} from '@wundergraph/cosmo-connect/dist/platform/v1/platform_pb';
+
+export type JsonOutputDescriptor = {
+  status: 'error' | 'success';
+  code: EnumStatusCode;
```
StarpTech, line 14: Why should we include the response code?

cli/src/json-check-schema-output-builder.ts
comatory: I figured it's one way to detect whether you got all data. When using console, you get a warning, so this is something similar but for JSON. We can exclude it I don't have strong opinion on this specifically.

cli/src/json-check-schema-output-builder.ts
comatory: I think it's useful to know if you want to differentiate the reasons when `status=error`. There are few variants which are defined by individual `case` statements in `cli/src/handle-check-result.ts`:

- `ERR_SCHEMA_MISMATCH_WITH_APPROVED_PROPOSAL`
- `ERR_INVALID_SUBGRAPH_SCHEMA`
- and eventually `ERR` if none of the above matches

It can be useful to have it if you want to quickly detect these cases, instead of digging through the JSON output. I include the message, but that can be dynamic, where as these codes are stable.

Does this make sense to you?

cli/src/handle-check-result.ts
comatory: Fixed in 66446890b0c33e06dbe717d7d824407f40bd8938

cli/src/json-check-schema-output-builder.ts
comatory: But I can see your point where the code might not tell you much. I think the issue is that we don't expose CLI json schemas, I'd eventually correct it, see: https://linear.app/wundergraph/issue/ENG-9146/clicosmo-docs-document-structured-output-as-json-schema-for-commands

Or we could have map these to a bit more human-like names: 'reason: invalid_subgraph_schema' etc.

StarpTech: Makes sense. I also like your idea regarding JSON schemas, but we should address it at a later stage.

cli/src/json-check-schema-output-builder.ts
StarpTech: Got it.

cli/src/json-check-schema-output-builder.ts
```diff
@@ -0,0 +1,211 @@
...
+    message: string;
+  };
+  traffic?: {
+    message: string;
+  };
+  changes?: {
+    breaking: SchemaChange[];
+    nonBreaking: SchemaChange[];
+  };
+  composition?: {
```
StarpTech, line 28: The file name and class name suggests a generic helper, but it is coupled with checks. Can we rename the file?

comatory: Yeah for sure, it was originally co-located with the function and forgot to rename it. Addressed: 864249975026833d9db8a41fdda08e403fb7e108

cli/src/json-check-schema-output-builder.ts
comatory: Screenshot 2026-03-10 at 13 29 43

To expand a bit more on this, there's `--limit` which can override the default 50 items. So it can tell you whether you're out of bounds for whatever you defined as that limit. The JSON response also includes `rowLimit` telling you what number you've passed to `--limit` flag.

The example above sets limit to `10_000` and the response has 52 breaking and 2 non-breaking changes.

PR #2600, opened 2026-03-06
feat: option to remove nats consumers on shutdown

dkorittki:
## Summary by CodeRabbit

* **New Features**
  * Per-provider option to delete durable NATS (JetStream) consumers on router shutdown; defaults to false.
  * Router will remove tracked durable consumers at shutdown when enabled, preventing accumulation.

* **Configuration**
  * New "consumers.delete_on_shutdown" setting added to NATS provider configuration and schema.

* **Tests**
  * Added end-to-end test verifying durable consumer deletion on shutdown.

## Checklist

Adds a new configuration option to nats providers

```
events:
  providers:
    nats:
      - id: nats
        consumers:
          durable:
            delete_on_shutdown: true
```

to the router. This config tells the router to clean up all durable consumers on the nats server, which the router has created, when the router shuts down. The default is not to clean up to let the router stay identical to previous behaviour.

Usually there is an option for specifying the max age for an inactive consumers on the NATS server but there are users right now having issues with this. We need to investigate into that. Meanwhile this option provides an alternative to that.

### DIFF
--- a/cli/CHANGELOG.md
+++ b/cli/CHANGELOG.md
@@ -4,6 +4,12 @@ Binaries are attached to the github release otherwise all images can be found [h
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+# [0.109.0](https://github.com/wundergraph/cosmo/compare/wgc@0.108.0...wgc@0.109.0) (2026-03-10)
+
+### Features
+
+* json output support for monograph/subgraph check ([#2590](https://github.com/wundergraph/cosmo/issues/2590)) ([5131bba](https://github.com/wundergraph/cosmo/commit/5131bbac8d24115a61468e8ebf9731b4250830eb)) (@comatory)
+
 # [0.108.0](https://github.com/wundergraph/cosmo/compare/wgc@0.107.1...wgc@0.108.0) (2026-03-10)
 
 ### Features
--- a/cli/package.json
+++ b/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "wgc",
-  "version": "0.108.0",
+  "version": "0.109.0",
   "description": "The official CLI tool to manage the GraphQL Federation Platform Cosmo",
   "type": "module",
   "main": "dist/src/index.js",
--- a/controlplane/CHANGELOG.md
+++ b/controlplane/CHANGELOG.md
@@ -4,6 +4,12 @@ Binaries are attached to the github release otherwise all images can be found [h
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+# [0.200.0](https://github.com/wundergraph/cosmo/compare/controlplane@0.199.0...controlplane@0.200.0) (2026-03-10)
+
+### Features
+
+* **keycloak:** update to 26.5.4 ([#2612](https://github.com/wundergraph/cosmo/issues/2612)) ([47a0fb5](https://github.com/wundergraph/cosmo/commit/47a0fb5ce166d2e489b8214fcd7da2e6c0b7f11d)) (@pepol)
+
 # [0.199.0](https://github.com/wundergraph/cosmo/compare/controlplane@0.198.0...controlplane@0.199.0) (2026-03-10)
 
 ### Features
--- a/controlplane/package.json
+++ b/controlplane/package.json
@@ -1,6 +1,6 @@
 {
   "name": "controlplane",
-  "version": "0.199.0",
+  "version": "0.200.0",
   "private": true,
   "description": "WunderGraph Cosmo Controlplane",
   "type": "module",
--- a/keycloak/CHANGELOG.md
+++ b/keycloak/CHANGELOG.md
@@ -4,6 +4,12 @@ Binaries are attached to the github release otherwise all images can be found [h
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+# [0.13.0](https://github.com/wundergraph/cosmo/compare/keycloak@0.12.0...keycloak@0.13.0) (2026-03-10)
+
+### Features
+
+* **keycloak:** update to 26.5.4 ([#2612](https://github.com/wundergraph/cosmo/issues/2612)) ([47a0fb5](https://github.com/wundergraph/cosmo/commit/47a0fb5ce166d2e489b8214fcd7da2e6c0b7f11d)) (@pepol)
+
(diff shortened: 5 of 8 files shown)


router/pkg/pubsub/nats/adapter.go
```diff
@@ -383,6 +401,23 @@ func (p *ProviderAdapter) Shutdown(ctx context.Context) error {
...
+		for _, tc := range consumers {
+			if err := p.js.DeleteConsumer(ctx, tc.streamName, tc.consumerName); err != nil {
+				p.logger.Warn("failed to delete durable consumer on shutdown",
+					zap.String("stream", tc.streamName),
+					zap.String("consumer", tc.consumerName),
+					zap.Error(err),
+				)
+			}
+		}
+	}
```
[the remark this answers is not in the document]
dkorittki: It's true that the leak remains in that case. However I think it is sufficient to log this as a warning, because it is not a fatal error worth bubbling up.

router/pkg/config/config.go
```diff
@@ -616,10 +616,15 @@ type NatsAuthentication struct {
 	NatsTokenBasedAuthentication `yaml:"token,inline"`
 }
 
+type NatsConsumersConfiguration struct {
```
alepane21: I would call this `NatsDurableConsumersConfiguration`

router/pkg/config/config.go
```diff
@@ -616,10 +616,15 @@ type NatsAuthentication struct {
...
+}
+
 type NatsEventSource struct {
-	ID             string              `yaml:"id,omitempty"`
-	URL            string              `yaml:"url,omitempty"`
-	Authentication *NatsAuthentication `yaml:"authentication,omitempty"`
+	ID             string                     `yaml:"id,omitempty"`
+	URL            string                     `yaml:"url,omitempty"`
+	Authentication *NatsAuthentication        `yaml:"authentication,omitempty"`
+	Consumers      NatsConsumersConfiguration `yaml:"consumers,omitempty"`
```
alepane21, line 631: Consumers -> DurableConsumers

router/pkg/config/config.schema.json
```diff
@@ -2580,6 +2580,18 @@
                         }
                       }
                     ]
+                  },
+                  "consumers": {
```
alepane21, line 2584: durable_consumers

router/pkg/config/config.schema.json
```diff
@@ -2580,6 +2580,18 @@
...
+                    "type": "object",
+                    "description": "Configuration for durable JetStream consumers managed by this NATS provider.",
+                    "additionalProperties": false,
+                    "properties": {
+                      "delete_on_shutdown": {
+                        "type": "boolean",
+                        "description": "When enabled, all durable JetStream consumers created by this provider are deleted when the router shuts down. Defaults to false.",
```
alepane21: "when the router shutdowns process is correctly done" or something like that

router/pkg/config/config.schema.json
dkorittki: I was thinking of this section for consumers in general, both non-durables and durables. But I could insert a field inbetween: `consumers.durable. delete_on_shutdown`, where
`consumers` : section for all consumer based settings
`durable` : section for durable consumer specific settings

Opinions?

alepane21: The only thing important to me, is that it should be clear that those options only apply to durable consumers. Maybe we can also keep it like you did at the beginning and just document that "applies only to durable consumers". I think what you proposed is also fine!

dkorittki: Okay :)  I will update the description of the option to make it more clear and use `consumers.durable.delete_on_shutdown`. I think it makes it more future proof

router/pkg/config/config.schema.json
dkorittki: Fixed

router/pkg/config/config.go
dkorittki: Fixed like discussed

router/pkg/config/config.go
dkorittki: Fixed like discussed

router-tests/events/nats_events_test.go
```diff
@@ -1200,6 +1200,63 @@ func TestNatsEvents(t *testing.T) {
 		})
 	})
 
+	t.Run("durable consumer is deleted from nats server on router shutdown", func(t *testing.T) {
```
alepane21, line 1203: I think we should also add a test to verify that without this option, the durable stream stays in the nats server.

router/pkg/pubsub/nats/adapter.go
```diff
@@ -383,6 +378,10 @@ func (p *ProviderAdapter) Shutdown(ctx context.Context) error {
 
 	var shutdownErr error
 
+	if p.consumerConfig.deleteOnShutdown {
```
alepane21, line 381: Just something to be sure: if the configuration is reloaded the shutdown on the adapter is going to be called?

router-tests/events/nats_events_test.go
dkorittki: Good idea

dkorittki: Fixed

router/pkg/pubsub/nats/adapter.go
dkorittki: Good catch! It also fires when the router reloads its execution config, e.g. graph server being swapped. On a router reload websocket connections are being closed. It means the router deletes durable consumers, swaps graph server, then upon first new subscriber recreates durable consumers. It's unnecessarry but shouldn't cause harm either. Due to the adapters lifecycle management I can't easily change this fact without rethinking provider adapter lifecycle in general.

router/pkg/pubsub/nats/adapter.go
```diff
@@ -407,7 +406,74 @@ func (p *ProviderAdapter) Shutdown(ctx context.Context) error {
...
+			return true
+		}
+
+		streamName, ok := value.(string)
+		if !ok {
+			// skip this odd element, should not happen in reality
+			return true
+		}
+
+		err := p.js.DeleteConsumer(ctx, streamName, consumerName)
```
StarpTech, line 463: Durable consumers are global unique right? How does it behave when multiple routers are shutdown? What happens when only one router shuts down but the other remain intact for a while?

dkorittki: No its one consumer per instance. Each router deletes its own consumers on shutdown. If only a portion of routers are shut down the remaining keep their consumers on the nats server. The router uses a name generation scheme including its own hostname, thats why every router instance uses a different consumer name.

StarpTech: OK, great

PR #2611, opened 2026-03-09
feat: slow plan cache

SkArchon:
This PR adds an expensive query cache. We do not use ristretto for this cache and use a simple custom cache, because we do not want to use the same frequency based eviction rules that ristretto uses. Instead this cache (if full) allows adding of queries that are more expensive than the "fastest min" query in the cache.

image

This feature is enabled by default when the in memory fallback cache warmer is enabled.

## Summary by CodeRabbit

* **New Features**
  * Added an expensive-query cache with in-memory fallback, warmup integration, and graceful planner shutdown to protect slow-to-plan queries and repopulate the main cache.

* **Observability**
  * Expose expensive-cache status in debug headers and record hits in spans, OTEL attributes/histograms, and Prometheus metrics.

* **Chores**
  * New configuration options: ExpensiveQueryCacheSize and ExpensiveQueryThreshold.

* **Tests**
  * Comprehensive unit and integration tests covering caching, eviction, reloads, telemetry, shutdown, and concurrency.

## Checklist


router/core/operation_planner.go
```diff
@@ -139,19 +176,34 @@ func (p *OperationPlanner) plan(opContext *operationContext, options PlanOptions
...
+			start := time.Now()
+			prepared, err := p.preparePlan(opContext, operationPlannerOpts{operationContent: p.useFallback})
 			if err != nil {
 				return nil, err
 			}
+			prepared.planningDuration = time.Since(start)
+
 			p.planCache.Set(operationID, prepared, 1)
+			if p.useFallback && p.threshold > 0 && prepared.planningDuration >= p.threshold && prepared.content != "" {
+				p.expensiveCache.Set(operationID, prepared, prepared.planningDuration)
```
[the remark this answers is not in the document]
SkArchon: We need it for cache warming also, where the queries are not necessarily expensive.

router/core/graph_server.go
```diff
@@ -1332,7 +1344,18 @@ func (s *graphServer) buildGraphMux(
...
+		s.logger,
+		executor,
+		gm.planCache,
+		opts.ReloadPersistentState.inMemoryPlanCacheFallback.IsEnabled(),
+		int(s.engineExecutionConfiguration.ExpensiveQueryCacheSize),
+		s.engineExecutionConfiguration.ExpensiveQueryThreshold,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create operation planner: %w", err)
+	}
```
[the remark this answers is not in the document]
SkArchon: This is fine, this is how it is in other error cases before this

router/pkg/slowplancache/slow_plan_cache.go
StarpTech, line 1: Can we move that into a separate package?

router/pkg/config/config.schema.json
```diff
@@ -3325,6 +3325,21 @@
           "default": 1024,
           "description": "The size of the execution plan cache."
         },
+        "expensive_query_cache_size": {
```
StarpTech: `expensive` sounds very weird 😄

router/pkg/otel/attributes.go
```diff
@@ -27,6 +27,7 @@ const (
 	WgRequestError                     = attribute.Key("wg.request.error")
 	WgOperationPersistedID             = attribute.Key("wg.operation.persisted_id")
 	WgEnginePlanCacheHit               = attribute.Key("wg.engine.plan_cache_hit")
+	WgEngineExpensivePlanCacheHit      = attribute.Key("wg.engine.expensive_plan_cache_hit")
```
StarpTech: That adds no value, we already have planner cache hit

router/pkg/slowplancache/slow_plan_cache.go
SkArchon: Done

router/pkg/config/config.schema.json
SkArchon: Renamed it to `PlanFallbackCache` which I thought goes with `PlanCache`

router/pkg/otel/attributes.go
SkArchon: Done, I had it because in case something was served from the planfallback cache and not the base plan cache it would be measurable, but its not too important, removed.

router/pkg/config/config.schema.json
SkArchon: Updated to `SlowPlanCache`

router/pkg/slowplancache/slow_plan_cache_test.go
```diff
@@ -0,0 +1,377 @@
...
+	require.NoError(t, err)
+	defer c.Close()
+
+	c.Set(1, &testPlan{content: "q1"}, 10*time.Millisecond)
+	c.Set(2, &testPlan{content: "q2"}, 20*time.Millisecond)
+	c.Set(3, &testPlan{content: "q3"}, 30*time.Millisecond)
+	c.Wait()
+
+	count := 0
+	c.IterValues(func(_ *testPlan) bool {
```
StarpTech: Go has native support of iterators

SkArchon: Done

router/pkg/config/config.schema.json
```diff
@@ -3325,6 +3325,21 @@
...
+        "slow_plan_cache_size": {
+          "type": "integer",
+          "minimum": 1,
+          "default": 300,
+          "description": "The maximum number of entries in the slow plan cache."
+        },
+        "slow_plan_cache_threshold": {
+          "type": "string",
+          "format": "go-duration",
+          "description": "The minimum planning duration for a query plan to be stored in the slow plan cache, protecting it from TinyLFU cache eviction.",
```
endigma: Needlessly technical description I think

router/pkg/config/config.schema.json
```diff
@@ -3325,6 +3325,21 @@
...
+          "default": "100ms",
+          "duration": {
+            "minimum": "1ns"
```
endigma: Do you think there is a more reasonable minimum we can enforce to prevent misuse? `1ns` is negligibly different from "put everything in the slow plan cache"

router/pkg/slowplancache/slow_plan_cache_test.go
```diff
@@ -0,0 +1,373 @@
...
+	_, ok = c.Get(2)
+	require.True(t, ok)
+}
+
+func TestCache_ConcurrentAccess(t *testing.T) {
+	t.Parallel()
+	c, err := New[*testPlan](100, 0)
+	require.NoError(t, err)
+	defer c.Close()
+	done := make(chan struct{})
```
endigma: Why not a waitgroup? This approach is nearly identical but is based on magic numbers. You can also use range over int

router/pkg/slowplancache/slow_plan_cache_test.go
```diff
@@ -0,0 +1,373 @@
...
+			}
+		}(i)
+	}
+
+	// Concurrent iterators
+	for i := 0; i < 5; i++ {
+		go func() {
+			defer func() { done <- struct{}{} }()
+			for range c.Values() {
+			}
```
endigma: This is probably removed by the compiler

router/pkg/slowplancache/slow_plan_cache.go
```diff
@@ -0,0 +1,244 @@
...
+
+	select {
+	case c.writeCh <- setRequest[V]{waitCh: ch}:
+		<-ch
+	case <-c.done:
+	}
+}
+
+// applySet performs the actual cache mutation. Must only be called from processWrites.
+func (c *Cache[V]) applySet(key uint64, value V, duration time.Duration) {
```
endigma, line 130: Can you add a note for future travellers that this does in fact still hold the Lock while it is running? I think it's probably fine but it's not circumventing Read locking via segmenting the map or anything like that

router/pkg/slowplancache/slow_plan_cache.go
```diff
@@ -0,0 +1,244 @@
...
+	// Reject entries that don't meet the threshold
+	if duration < c.threshold {
+		return
+	}
```
endigma: Should we reject these in Set() directly? No reason to fill up the channel with stuff that's going to get rejected instantly.

router/pkg/slowplancache/slow_plan_cache.go
```diff
@@ -0,0 +1,244 @@
...
+func (c *Cache[V]) processWrites() {
+	defer close(c.done)
+
+	for {
+		select {
+		case req := <-c.writeCh:
+			if req.waitCh != nil {
+				close(req.waitCh)
+				continue
+			}
```
endigma, line 73: Seems like it's closed _before_ the request is processed? (type comment)

router/pkg/slowplancache/slow_plan_cache.go
```diff
@@ -0,0 +1,244 @@
...
+}
+
+// Cache is a bounded map that holds expensive-to-compute values
+// that should not be subject to TinyLFU eviction in the main cache.
+// Writes are buffered through a channel and applied asynchronously by a
+// background goroutine, making Set non-blocking. Reads are protected by a RWMutex.
+// It tracks the minimum-duration entry so that rejection of cheaper entries is O(1).
+type Cache[V any] struct {
+	mu        sync.RWMutex
+	entries   map[uint64]*Entry[V]
```
endigma: `sync.Map` is usually faster in read heavy scenarios, if there's not a specific reason you can't use it I would consider it here

router/pkg/slowplancache/slow_plan_cache_test.go
```diff
@@ -0,0 +1,373 @@
+package slowplancache
```
endigma, line 1: Perhaps adding an insert/retrieval benchmark would be helpful for future iteration?

router/pkg/config/config.schema.json
SkArchon: I would say this allows the user the flexibility to use it as a "cache that catches everything", but not opposed to having a higher number and lowering it based on customer requests. Do you have a number in mind?

router/pkg/slowplancache/slow_plan_cache.go
SkArchon: The cache is slightly more write heavy than read heavy I would say, as its a fallback from the main cache + we attempt to admit it back to the planCache upon hit.

router/pkg/slowplancache/slow_plan_cache.go
SkArchon: Could you elaborate a bit, do you mean add a comment?

router/pkg/slowplancache/slow_plan_cache_test.go
SkArchon: From my testing it seems its not removed, I anyway also added a noop statement (though that might get cleaned up)

router/pkg/slowplancache/slow_plan_cache_test.go
SkArchon: Makes sense, will add this.

router/pkg/config/config.schema.json
endigma: If you want a catch all remove the minimum, it's just annoying in that case to write "1ns", otherwise I would make it like >20ms or something, I don't know what a normal planning time is off the top of my head

router/pkg/slowplancache/slow_plan_cache.go
endigma: the comment for req.waitCh says it is closed after processing

router/pkg/slowplancache/slow_plan_cache.go
endigma: A cache that is more write than read makes no sense, are you sure?

router/pkg/slowplancache/slow_plan_cache.go
SkArchon: Ah ok, the comment should be updated to "if non-nil, will be closed after previous requests are processed"

router/pkg/slowplancache/slow_plan_cache.go
SkArchon: This is only a backup cache, its other / primary job is also to a store the expensive queries and make sure they persist across reloads.

router/pkg/config/config.schema.json
SkArchon: Ok I think we will go with a catch all for now, but if we remove the minimum, they can set a negative number, so we still need to set 0ns / 1ns.

router/pkg/slowplancache/slow_plan_cache_test.go
SkArchon: Done

router/pkg/slowplancache/slow_plan_cache_test.go
SkArchon: Added, also changed to sync.Map since it was a bit faster. (5x on concurrent benchmark, though just a few ns differences on others).

router/pkg/slowplancache/slow_plan_cache_test.go
```diff
@@ -0,0 +1,514 @@
...
+	defer c.Close()
+	var wg sync.WaitGroup
+
+	const (
+		numGoroutines = 2000
+		opsPerRoutine = 5000
+	)
+
+	for g := range numGoroutines {
+		wg.Add(1)
```
endigma: nit: can use wg.Go

PR #2620, opened 2026-03-10
fix(cli): avoid cli-table3 wordWrap deadlock with 4+ subgraphs

### LINKED ISSUE #2619: `wgc router compose` hangs indefinitely when composition errors produce large error text
## Description

`wgc router compose` hangs indefinitely when federation composition produces errors with enough combined text volume. The process prints the error header but never renders the error table:

```
We found composition errors, while composing. Please check the errors below:
█  (hangs here — no table, no exit)
```

The process must be killed with `SIGKILL`; `SIGTERM` and `SIGINT` are ignored.

## Root Cause

`cli-table3`'s `wordWrap: true` option calls `string-width` for every word in every table cell to calculate wrapping. `string-width` executes `emoji-regex` — a 40 KB Unicode regex pattern — on each word. The time this takes grows super-linearly with text volume:

| Error text volume | `table.toString()` time |
|---|---|
| 0.1 MB | 0.2s |
| 0.5 MB | 3s |
| 1 MB | 13s |
| 2 MB | ~90s |
| **9 MB** | **hours (effectively infinite)** |

The trigger is **total error text volume**, not subgraph count. In our case, composition of 4 subgraphs against a large primary schema (~19K lines, ~300 entity types) produces 29 "unresolvable field" errors where the composition engine dumps the full deeply-nested resolution path it attempted. Each error is 100–430 KB, totaling ~9 MB.

## Reproduction

👉 **https://github.com/kamil-gwozdz/cosmo-compose-deadlock-repro**

```bash
git clone https://github.com/kamil-gwozdz/cosmo-compose-deadlock-repro.git
cd cosmo-compose-deadlock-repro
npm install
timeout 60 npx wgc router compose -i compose.yaml
# Exit code 124 = killed by timeout (deadlock confirmed)
```

The repository contains 4 anonymized federated subgraphs that produce ~9 MB of composition errors:

- **subgraph-a** (19K lines): Large primary API with ~300 entity types, heavily interconnected via Connection/Edge patterns
- **subgraph-b** (1.2K lines): Interface resolver claiming `@key(fields: "id")` on ~295 types
- **subgraph-c/d**: Small domain services with `@override` from subgraph-a

The `@key` overlap between subgraph-a and subgraph-b causes "unresolvable field" errors where the composition engine walks paths hundreds of levels deep through the interconnected type graph, producing individual error messages of 100–430 KB each.

## Versions affected

- `wgc@0.102.4`
- `wgc@0.108.0`
- `wgc@0.109.0`

Likely all versions using `cli-table3` with `wordWrap: true`.

## Expected behavior

The composition error table is rendered and the process exits with code 1.

## Actual behavior

The process prints the header and hangs forever on `table.toString()`.

## Workaround

```bash
timeout 60 npx wgc router compose -i compose.yaml
```

## Fix

PR #2620 replaces `cli-table3`'s `wordWrap: true` with a lightweight `wrapText()` utility that wraps text using simple character counting — no `string-width` or `emoji-regex` dependencies. The fix handles 1 MB of text in <100ms vs hours with the original code path.

## Environment

- **OS:** Linux (Debian/Ubuntu in GitHub Codespaces), macOS
- **Node.js:** v18+
- **Terminal:** various (bash, zsh)

kamil-gwozdz:
## Summary

Fixes #2619

`wgc router compose` deadlocks when federation composition produces errors and the config references 4+ subgraphs. The process prints the error header but hangs indefinitely before rendering the error table.

## Root Cause

cli-table3's built-in `wordWrap: true` option triggers an expensive code path through `string-width` → `emoji-regex` for every word in every table cell. With enough error content (4+ subgraphs), this blocks the Node.js main thread indefinitely, preventing signal handling and event loop progress.

## Fix

- Introduced a lightweight `wrapText()` utility (`cli/src/wrap-text.ts`) that wraps text on word boundaries using simple character counting — no heavy regex dependencies.
- Exported a `TABLE_CONTENT_WIDTH` constant (116 = 120 column width − 2 padding − 2 borders) to avoid magic numbers.
- Pre-wrap error/warning messages before inserting them into the table.
- Removed `wordWrap: true` from all affected table configurations.
- Table output looks identical to before; only the wrapping mechanism changes.

## Files Changed

| File | Change |
|------|--------|
| `cli/src/wrap-text.ts` | New standalone utility with `wrapText()` and `TABLE_CONTENT_WIDTH` — zero dependencies |
| `cli/src/utils.ts` | Re-exports `wrapText` for backward compatibility |
| `cli/src/commands/router/commands/compose.ts` | Fixed 4 table instances (errors + warnings for main and feature flag compositions) |
| `cli/src/handle-composition-result.ts` | Fixed 3 table instances (composition errors, deployment errors, warnings) |
| `cli/test/wrap-text.test.ts` | 10 unit tests for `wrapText`, including a regression test verifying 500-word inputs complete in <1s |
| `cli/test/compose-error-table.test.ts` | Integration test reproducing the exact issue — 4 subgraphs with circular `@override`, verifying the error table renders without hanging (5s timeout) |

## Testing

- **`wrap-text.test.ts`** — 10 tests covering word wrapping edge cases, empty inputs, long words, paragraph preservation, and a performance regression test.
- **`compose-error-table.test.ts`** — Integration test using the exact reproduction schemas from the issue (4 subgraphs with circular `@override` between subgraph-b and subgraph-c). Runs `federateSubgraphs`, asserts composition fails, renders the error table through the fixed code path, and verifies output correctness within a 5-second timeout.
- All 11 new tests pass. Existing CLI tests unaffected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CLI composition output: long error and warning messages are now reliably wrapped to prevent hangs and ensure readable table rendering, including very large messages.

* **Tests**
  * Added tests covering wrapping behavior, edge cases, performance, and a regression to ensure fast, correct error reporting.

### DIFF
--- a/cli/test/compose-error-table.test.ts
+++ b/cli/test/compose-error-table.test.ts
@@ -1,117 +1,78 @@
 import { describe, test, expect } from 'vitest';
-import { parse } from 'graphql';
 import Table from 'cli-table3';
-import pc from 'picocolors';
-import { federateSubgraphs, ROUTER_COMPATIBILITY_VERSION_ONE } from '@wundergraph/composition';
 import { wrapText, TABLE_CONTENT_WIDTH } from '../src/wrap-text.js';
 
-// Exact reproduction schemas from https://github.com/wundergraph/cosmo/issues/2619
-// subgraph-b overrides `description` from subgraph-c, while subgraph-c overrides
-// `description` from subgraph-b — a circular override that produces composition errors.
+// Regression test for https://github.com/wundergraph/cosmo/issues/2619
+//
+// The bug: cli-table3's wordWrap option uses string-width → emoji-regex for every
+// word, which grows super-linearly with text volume. With large composition errors
+// (~9 MB in real-world cases), table.toString() hangs indefinitely.
+//
+// The fix: pre-wrap text with our lightweight wrapText() utility instead of relying
+// on cli-table3's built-in wordWrap.
 
-const subgraphASchema = /* GraphQL */ `
-  extend schema
-    @link(url: "https://specs.apollo.dev/federation/v2.5", import: ["@key"])
-
-  type Query {
-    foos: [Foo!]!
+function generateLargeErrorMessage(sizeBytes: number): string {
+  const lines: string[] = ['The field "id" is unresolvable at the following path:'];
+  let depth = 1;
+  while (lines.join('\n').length < sizeBytes) {
+    const indent = ' '.repeat(depth);
+    lines.push(`${indent}type${depth} {`);
+    lines.push(`${indent} edges {`);
+    lines.push(`${indent}  node {`);
+    depth += 3;
+    if (depth > 300) depth = 1;
   }
+  return lines.join('\n').substring(0, sizeBytes);
+}
 
-  type Foo @key(fields: "id") {
-    id: ID!
-    name: String!
-  }
-`;
+describe('Error table rendering with large text (#2619)', () => {
+  test('wrapText handles 1 MB of error text in under 1 second', () => {
+    // In real-world cases, composition produces ~9 MB of errors (29 errors,
+    // each 100-430 KB). cli-table3's wordWrap takes hours on this volume.
+    // Our wrapText must handle it in milliseconds.
+    const errors = Array.from({ length: 10 }, () => generateLargeErrorMessage(100_000));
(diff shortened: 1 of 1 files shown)


endigma:
Hi @kamil-gwozdz, thanks for your contribution! Closing this in favour of https://github.com/wundergraph/cosmo/pull/2715

PR #2624, opened 2026-03-10
fix(composition): reject @override on @key fields

ekroon:
## Summary

Prevents federation composition when a field annotated with `@override` is also part of an `@key` directive. This combination causes query planner failures during entity resolution.

## Problem

When a subgraph overrides a field that is used as an entity key in another subgraph, the query planner can fail with errors like:
```
Cannot query field "databaseId" on type "Query"
```

This occurs because the overridden key field creates an inconsistency in the federation metadata where:
- The field remains in the target subgraph's `keys` configuration
- But disappears from the target subgraph's `RootNodes` 
- This forces a cross-subgraph self-key fetch that breaks entity resolution

## Solution

This PR adds composition-time validation that:
1. Detects when any `@override` directive targets a field that is part of a `@key` directive
2. Reports all violations with clear error messages
3. Blocks composition to prevent runtime failures

## Changes

- Added `overrideOnKeyFieldsError` and `overrideOnKeyFieldErrorMessage` to composition errors
- Implemented validation logic in `batchNormalize` function
- Added test coverage for the new validation

## Testing

- New test: `that an error is returned if @override is used on a @key field`
- All existing override tests continue to pass (30/30)

## Related

- Discovered while investigating planner failures in internal federation graphs
- Alternative mitigation: generic real-schema guard (internal tooling)
- Upstream fix pending in query planner (separate PR in graph-hopper)

## Release Notes

* **New Features**
  * Enhanced composition validation to detect and report errors when @override is applied to fields declared with @key. New validation identifies conflicts across subgraphs and provides detailed error messages with affected fields and source subgraph information.

### DIFF
--- a/composition/src/errors/errors.ts
+++ b/composition/src/errors/errors.ts
@@ -843,6 +843,29 @@ export function duplicateOverriddenFieldsError(errorMessages: string[]): Error {
   );
 }
 
+export function overrideOnKeyFieldErrorMessage(
+  fieldPath: string,
+  sourceSubgraphName: string,
+  targetSubgraphName: string,
+): string {
+  return (
+    ` The field "${fieldPath}" in subgraph "${sourceSubgraphName}" declares an @override directive targeting subgraph "${targetSubgraphName}",` +
+    ` but "${fieldPath}" is part of an @key directive.`
+  );
+}
+
+export function overrideOnKeyFieldsError(errorMessages: string[]): Error {
+  return new Error(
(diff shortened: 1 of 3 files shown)


Aenimus:
Hi,

The problem is not `@override` on a `@key` field—it's that overridden fields are not being treated like they are fully `@external`.

We're aware of this case, but the root fix needs to go deeper and is more nuanced than what's been provided here.

Thanks,

The WunderGraph Team

PR #2625, opened 2026-03-10
fix(mcp): reset registeredTools between reloads to prevent false tool name collisions

### LINKED ISSUE #2623: MCP: registeredTools slice grows unbounded across config reloads causing false tool name collisions
## Bug

When `omit_tool_name_prefix: true` is configured, every router config reload causes **all** operation-derived tools to be falsely reported as colliding with built-in MCP tools. After enough reloads, every operation gets prefixed with `execute_operation_` despite no actual collision.

## Root Cause

`registeredTools` in `mcpserver/server.go` is never reset between reloads. On each reload:

1. `RemoveTools(s.registeredTools...)` removes old tools from the MCP server
2. `registerTools()` appends new tool names to the **same** slice
3. The collision check `slices.Contains(s.registeredTools, operationToolName)` matches every operation against its own name from the previous cycle

## Reproduction

1. Configure the router with `omit_tool_name_prefix: true` and `execution_config.file.watch: true`
2. Touch/modify the execution config file to trigger a reload
3. After the first reload, every operation logs a false collision warning:

```
INFO  core/router.go:1362 Config file changed. Updating server with new config
WARN  mcpserver/server.go:699 Operation name collides with built-in MCP tool, using prefixed name {"operation": "AllEmployeesAndFirstEmployee", "conflicting_tool": "all_employees_and_first_employee", "using_name": "execute_operation_all_employees_and_first_employee"}
WARN  mcpserver/server.go:699 Operation name collides with built-in MCP tool, using prefixed name {"operation": "GetEmployeeById", "conflicting_tool": "get_employee_by_id", "using_name": "execute_operation_get_employee_by_id"}
WARN  mcpserver/server.go:699 Operation name collides with built-in MCP tool, using prefixed name {"operation": "GetEmployees", "conflicting_tool": "get_employees", "using_name": "execute_operation_get_employees"}
WARN  mcpserver/server.go:699 Operation name collides with built-in MCP tool, using prefixed name {"operation": "GetEmployeeStartDate", "conflicting_tool": "get_employee_start_date", "using_name": "execute_operation_get_employee_start_date"}
WARN  mcpserver/server.go:699 Operation name collides with built-in MCP tool, using prefixed name {"operation": "UpdateEmployeeMood", "conflicting_tool": "update_employee_mood", "using_name": "execute_operation_update_employee_mood"}
```

## Impact

- Every operation gets an unintended `execute_operation_` prefix after the first config reload, breaking MCP clients that expect the unprefixed names
- `registeredTools` grows indefinitely (memory leak)
- Log noise on every reload

## Fix

Reset `s.registeredTools = nil` between `RemoveTools` and `registerTools()`:

```go
s.server.RemoveTools(s.registeredTools...)
s.registeredTools = nil

if err := s.registerTools(); err != nil {
    return fmt.Errorf("failed to register tools: %w", err)
}
```

asoorm:
## Summary
- Reset `s.registeredTools = nil` after `DeleteTools` to prevent unbounded slice growth across config reloads
- When `omit_tool_name_prefix: true`, every operation was falsely colliding with its own name from the previous reload cycle
- Changed collision behavior to skip duplicate tools with a warning instead of silently falling back to a prefixed name

Fixes ENG-9158
Closes #2623

## Test plan
- [ ] Configure router with `omit_tool_name_prefix: true` and `execution_config.file.watch: true`
- [ ] Trigger a config reload by modifying the execution config file
- [ ] Verify no false collision warnings are logged after reload
- [ ] Verify tools retain their unprefixed names after reload
- [ ] Verify actual collisions (e.g. operation named `GetSchema` → `get_schema`) are still detected and skipped with a warning

## Summary by CodeRabbit

* **Bug Fixes**
  * Reload now fully clears previously registered tools before re-registering to avoid stale entries.
  * Tool registration now skips operations whose names collide with built-in tools and logs the skip instead of attempting a fallback rename.

* **Tests**
  * Updated test and test name to reflect that colliding tools are skipped (assertions adjusted accordingly).

### DIFF
--- a/router/pkg/mcpserver/server_test.go
+++ b/router/pkg/mcpserver/server_test.go
@@ -149,30 +149,19 @@ func TestReload_ReservedToolNameCollision(t *testing.T) {
 		assert.Equal(t, "get_operation_info", entry.ContextMap()["conflicting_tool"])
 	}
 
-	// "list_employees" should still be registered, "get_operation_info" should not be
-	// in registeredTools as an operation tool (only the internal one).
-	assert.Contains(t, srv.registeredTools, "list_employees")
-	assert.Contains(t, srv.registeredTools, "get_operation_info",
-		"get_operation_info should still be registered as the internal tool")
-
-	// Count occurrences — should appear exactly once (the internal tool, not the operation)
-	count := 0
-	for _, name := range srv.registeredTools {
-		if name == "get_operation_info" {
-			count++
-		}
-	}
-	assert.Equal(t, 1, count,
-		"get_operation_info should appear exactly once in registeredTools")
+	assert.ElementsMatch(t, []string{"get_schema", "list_employees", "get_operation_info"}, srv.registeredTools)
 }
 
-func TestReload_NoPrefixCollisionBetweenOperations(t *testing.T) {
+func TestReload_PrefixModeAvoidsReservedNameCollision(t *testing.T) {
 	core, logs := observer.New(zapcore.DebugLevel)
 	logger := zap.New(core)
 
+	// "GetOperationInfo" snake_cases to "get_operation_info" which is a reserved name.
+	// With the prefix enabled, it becomes "execute_operation_get_operation_info" and no collision occurs.
 	tempDir := t.TempDir()
 	writeOperationFiles(t, tempDir, map[string]string{
-		"ListEmployees.graphql": listEmployeesOp,
+		"GetOperationInfo.graphql": getOperationInfoOp,
+		"ListEmployees.graphql":   listEmployeesOp,
 	})
 
 	schemaDoc, report := astparser.ParseGraphqlDocumentString(testSchema)
@@ -180,7 +169,6 @@ func TestReload_NoPrefixCollisionBetweenOperations(t *testing.T) {
 	err := asttransform.MergeDefinitionWithBaseSchema(&schemaDoc)
 	require.NoError(t, err)
 
-	// With prefix, no collisions expected
 	srv, err := NewGraphQLSchemaServer(
 		"http://localhost:4000/graphql",
 		WithLogger(logger),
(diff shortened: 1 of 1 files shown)


router/pkg/mcpserver/server.go
```diff
@@ -541,12 +542,11 @@ func (s *GraphQLSchemaServer) registerTools() error {
 		if !s.omitToolNamePrefix {
 			toolName = fmt.Sprintf("execute_operation_%s", operationToolName)
 		} else if slices.Contains(s.registeredTools, operationToolName) {
-			s.logger.Warn("Operation name collides with built-in MCP tool, using prefixed name",
+			s.logger.Error("Skipping operation due to tool name collision",
 				zap.String("operation", op.Name),
 				zap.String("conflicting_tool", operationToolName),
-				zap.String("using_name", fmt.Sprintf("execute_operation_%s", operationToolName)),
 			)
```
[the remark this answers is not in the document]
endigma: Colliding tool names would result in a service degradation, this is an error.

router/pkg/mcpserver/server_test.go
```diff
@@ -0,0 +1,207 @@
...
+		id
+		name
+	}
+}`,
+		"ListEmployees.graphql": `query ListEmployees {
+	employees {
+		id
+		name
+	}
+}`,
```
endigma: nit: I'm pretty sure these files are not indentation sensitive so you could intent them wrong in the file and correct here for readability

router/pkg/mcpserver/server_test.go
```diff
@@ -0,0 +1,207 @@
...
+	for _, name := range srv.registeredTools {
+		if name == "get_operation_info" {
+			count++
+		}
+	}
+	assert.Equal(t, 1, count,
+		"get_operation_info should appear exactly once in registeredTools")
+}
+
+func TestReload_NoPrefixCollisionBetweenOperations(t *testing.T) {
```
endigma: This test's assertion seems to be sort of mixed up, comments say a different thing than the test is doing

router/pkg/mcpserver/server_test.go
```diff
@@ -0,0 +1,207 @@
...
+	assert.Contains(t, srv.registeredTools, "get_operation_info",
+		"get_operation_info should still be registered as the internal tool")
+
+	// Count occurrences — should appear exactly once (the internal tool, not the operation)
+	count := 0
+	for _, name := range srv.registeredTools {
+		if name == "get_operation_info" {
+			count++
+		}
+	}
```
endigma: You can assert the full contents of this list, it should be not very many items

endigma (changes requested):
small things

PR #2628, opened 2026-03-11
fix: flag option as experimental

dkorittki:
## Summary by CodeRabbit

* **Configuration**
  * Replaced nested NATS consumer configuration with a single top-level experimental boolean to control deletion of durable NATS consumers on shutdown; example config and schema updated.

* **Tests**
  * Updated tests to use the new top-level flag path for NATS providers.

* **Documentation**
  * Minor schema text fix and the new experimental flag documented in the public schema.

## Checklist

We shortly added a configuration option meant to be experimental but hasn't been flagged so. This pr changes it be making the whole `consumers` option category experimental. Given that this option has only been added today I think we are safe when renaming it.

### DIFF
--- a/cli/CHANGELOG.md
+++ b/cli/CHANGELOG.md
@@ -4,6 +4,10 @@ Binaries are attached to the github release otherwise all images can be found [h
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## [0.109.1](https://github.com/wundergraph/cosmo/compare/wgc@0.109.0...wgc@0.109.1) (2026-03-11)
+
+**Note:** Version bump only for package wgc
+
 # [0.109.0](https://github.com/wundergraph/cosmo/compare/wgc@0.108.0...wgc@0.109.0) (2026-03-10)
 
 ### Features
--- a/cli/package.json
+++ b/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "wgc",
-  "version": "0.109.0",
+  "version": "0.109.1",
   "description": "The official CLI tool to manage the GraphQL Federation Platform Cosmo",
   "type": "module",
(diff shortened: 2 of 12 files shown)


router/pkg/config/config.schema.json
```diff
@@ -2581,24 +2581,10 @@
...
-                            "type": "boolean",
-                            "description": "When enabled, all durable JetStream consumers created by this provider are deleted when the router shuts down normally. Defaults to false.",
-                            "default": false
-                          }
-                        }
-                      }
-                    }
+                  "experiment_delete_durable_consumers_on_shutdown": {
+                    "type": "boolean",
+                    "description": "When enabled, all durable JetStream consumers created by this provider are deleted when the router shuts down normally. NOTE: This option is experimental and may change at any time. Defaults to false.",
```
SkArchon: Nit `This option is experimental and may change in future versions.`. This is how the cardinality limit has been documented.

router/pkg/config/config.go
```diff
@@ -616,19 +616,11 @@ type NatsAuthentication struct {
...
-
 type NatsEventSource struct {
-	ID             string                     `yaml:"id,omitempty"`
-	URL            string                     `yaml:"url,omitempty"`
-	Authentication *NatsAuthentication        `yaml:"authentication,omitempty"`
-	Consumers      NatsConsumersConfiguration `yaml:"consumers,omitempty"`
+	ID                               string              `yaml:"id,omitempty"`
+	URL                              string              `yaml:"url,omitempty"`
+	Authentication                   *NatsAuthentication `yaml:"authentication,omitempty"`
+	DeleteDurableConsumersOnShutdown bool                `yaml:"experiment_delete_durable_consumers_on_shutdown" envDefault:"false"`
```
endigma: This will not have `experiment` in the environment variable

alepane21: Fixed!

PR #2636, opened 2026-03-11
feat(mcp): add OAuth 2.1 authorization with per-tool scope extraction

asoorm:
## Summary
- Add OAuth 2.1 authorization support to the MCP server with JWT validation via JWKS providers
- Implement per-tool scope extraction from `@requiresScopes` directives in the federated graph schema
- Add configurable scope challenge modes (strict, smart, optimistic) for fine-grained access control
- Support configurable scopes for built-in tools (`execute_graphql`, `get_schema`, `get_operation_info`)
- Enforce scope checks at the HTTP transport level per MCP specification

## Summary by CodeRabbit

* **New Features**
  * OAuth 2.1 authentication support for MCP operations with JWT/JWKS token validation
  * Scope-based access control for individual tools and operations
  * Dynamic token rotation without reconnecting
  * RFC 9728-compliant OAuth resource metadata endpoint for server discovery

* **Configuration**
  * New OAuth settings enabling per-operation scope requirements and authorization server configuration

* **Improvements**
  * Enhanced authentication error responses with scope challenge details


router-tests/protocol/mcp_auth_client_test.go
```diff
@@ -0,0 +1,370 @@
...
+
+	// Add Authorization header if token is set
+	if a.token != "" {
+		req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", a.token))
+	}
+
+	resp, err := a.base.RoundTrip(req)
+	// Capture response for error analysis
+	a.lastResponse = resp
+	return resp, err
```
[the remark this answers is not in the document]
asoorm: This is a test - feels like possibly general advice. the MCP StreamableClientTransport makes one request at a time in a test. I dont see the possibility of concurrent callers. and adding mutexes would be dead code.

router/pkg/mcpserver/auth_middleware.go
```diff
@@ -0,0 +1,465 @@
...
+	"github.com/wundergraph/graphql-go-tools/v2/pkg/astparser"
+)
+
+type contextKey string
+
+const (
+	userClaimsContextKey contextKey = "mcp_user_claims"
+	// maxBodyBytes is the maximum size of the request body we'll read for scope checking.
+	// This prevents memory exhaustion from oversized payloads.
+	maxBodyBytes int64 = 1 << 20 // 1 MB
```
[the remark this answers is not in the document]
asoorm: Code path is inside the auth middleware and it's role is to gate requests. so IMO 401 is defensible. A 10MB body limit for scope checking is generous - a legitimate MCP request exceeding that would be unusual. The auth middleware intentionally fails closed - if it can't read or parse the body, it can't determine which scope check to apply, so it rejects the request. Returning 413/400 from inside the auth middleware would leak that the token itself is valid. Bumped the limit from 1MB to 10MB to accomodate large execute_graphql payloads with substantial variables. So at 10MB, this will act as a safety valve against abuse. Legitimate clients should never hit this limit.

router-tests/mcp_auth_e2e_test.go
```diff
@@ -0,0 +1,370 @@
...
+			result, err := mcpClient.CallTool(ctx, "execute_operation_my_employees", map[string]any{
+				"criteria": map[string]any{},
+			})
+
+			// Without authorization configured, this should work
+			require.NoError(t, err)
+			require.NotNil(t, result)
+			t.Logf("✓ Successfully called tool")
+		})
+	})
```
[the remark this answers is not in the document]
asoorm: @coderabbitai Are you sure? Please check again and resolve if this is now fixed.

router-tests/mcp_auth_harness_example.go
```diff
@@ -0,0 +1,242 @@
...
+		fmt.Printf("      \"error\": {\n")
+		fmt.Printf("        \"code\": %d,\n", jsonRPCResp.Error.Code)
+		fmt.Printf("        \"message\": \"%s\",\n", jsonRPCResp.Error.Message)
+		fmt.Printf("        \"data\": {\n")
+		fmt.Printf("          \"required_scopes\": %v\n", jsonRPCResp.Error.Data.RequiredScopes)
+		fmt.Printf("        }\n")
+		fmt.Printf("      }\n")
+		fmt.Printf("    }\n")
+		return fmt.Errorf("insufficient scopes: %v", jsonRPCResp.Error.Data.RequiredScopes)
+	}
```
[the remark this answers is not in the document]
asoorm: File doesn't exist anymore -

router-tests/testutil/oauth_server.go
```diff
@@ -0,0 +1,544 @@
...
+	scope := r.FormValue("scope")
+	if scope == "" {
+		scope = client.Scope
+	}
+	if scope == "" {
+		scope = s.DefaultScopes
+	}
+
+	s.issueTokenResponse(w, clientID, scope)
+}
```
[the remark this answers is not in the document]
asoorm: this is a test utility, not a production AS. Its job is to issue tokens for tests, not to be a spec-compliant OAuth server.

router-tests/testutil/oauth_server.go
```diff
@@ -0,0 +1,544 @@
...
+	s.mu.Unlock()
+
+	// Preserve the state parameter for PKCE / CSRF
+	state := r.URL.Query().Get("state")
+	location := fmt.Sprintf("%s?code=%s", redirectURI, code)
+	if state != "" {
+		location += "&state=" + state
+	}
+
+	http.Redirect(w, r, location, http.StatusFound)
```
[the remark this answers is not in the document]
asoorm: test utility, the redirect URIs are controlled by the test code and will never have existing query params or special characters.

router/pkg/mcpserver/scope_extractor.go
```diff
@@ -0,0 +1,157 @@
...
+	// Start with the first field's OR-groups
+	result := fieldReqs[0].OrScopes
+
+	// Iteratively cross-product with each subsequent field's OR-groups
+	for i := 1; i < len(fieldReqs); i++ {
+		result = crossProduct(result, fieldReqs[i].OrScopes)
+	}
+
+	return result
+}
```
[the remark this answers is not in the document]
asoorm: The number of OR-groups per field is controlled by the platform team via @requiresScopes directives and capped at 16 by the composition layer. For pre-registered operations, this runs once at Reload. For execute_graphql, it only fires on queries touching scoped fields. No action needed - I think we should wait till we get feedback from the field.

router/pkg/mcpserver/server.go
```diff
@@ -548,62 +703,98 @@ func (s *GraphQLSchemaServer) registerTools() error {
...
-			return nil, fmt.Errorf("failed to marshal arguments: %w", err)
+func (s *GraphQLSchemaServer) handleOperation(handler *operationHandler) func(ctx context.Context, request *mcp.CallToolRequest) (*mcp.CallToolResult, error) {
+	return func(ctx context.Context, request *mcp.CallToolRequest) (*mcp.CallToolResult, error) {
+		// Log authenticated user if OAuth is enabled
+		if claims, ok := GetClaimsFromContext(ctx); ok {
+			s.logger.Debug("operation called by authenticated user",
+				zap.String("sub", getClaimString(claims, "sub")),
+				zap.String("email", getClaimString(claims, "email")),
+				zap.String("operation", handler.operation.Name))
 		}
```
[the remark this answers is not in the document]
asoorm: These are Debug-level logs, only visible when the operator explicitly opts in. 
Logging sub at debug level is standard practice. 
Operators who enable debug logging accept the PII trade-off for troubleshooting.

router-tests/cmd/oauth-server/main.go
```diff
@@ -0,0 +1,462 @@
...
+	clients map[string]*client
+	codes   map[string]*authCode
+}
+
+type client struct {
+	id           string
+	secret       string
+	scope        string
+	redirectURIs []string
+}
```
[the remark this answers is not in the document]
asoorm: This is a test oauth server

router-tests/cmd/oauth-server/main.go
```diff
@@ -0,0 +1,462 @@
...
+
+type authCode struct {
+	clientID  string
+	scope     string
+	createdAt time.Time
+}
```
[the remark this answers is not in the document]
asoorm: This is a test oauth server

router-tests/cmd/oauth-server/main.go
```diff
@@ -0,0 +1,438 @@
...
+	code := randomHex(32)
+	h.mu.Lock()
+	h.codes[code] = &authCode{clientID: clientID, scope: scope, createdAt: time.Now()}
+	h.mu.Unlock()
+
+	location := fmt.Sprintf("%s?code=%s", redirectURI, code)
+	if state != "" {
+		location += "&state=" + state
+	}
+	http.Redirect(w, r, location, http.StatusFound)
```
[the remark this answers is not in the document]
asoorm: This is a test oauth server

router-tests/testutil/oauth_server.go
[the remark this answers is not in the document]
asoorm: This is a test oauth server

router-tests/testutil/jwt_helper.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+// NewJWKSTestServer creates a new JWKS test server with RSA keys
+// The server will automatically allocate a free port and return it when the test ends
+func NewJWKSTestServer(t *testing.T) (*JWKSTestServer, error) {
+	t.Helper()
+
+	// Get a free port using the freeport package
+	port := freeport.GetOne(t)
+	portStr := fmt.Sprintf("%d", port)
```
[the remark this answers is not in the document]
asoorm: I think this is a non-issue - this approach is used for Prometheus, MCP server, router listener, exec listener - it's a well established pattern in the code base. if this were a real problem, the entire test suite would be flaky.

router/pkg/config/config.schema.json
```diff
@@ -2439,7 +2437,234 @@
...
+                  "description": "Scopes required to call the get_schema built-in tool. Additive to tools_call scopes. Only relevant when expose_schema is true.",
+                  "items": { "type": "string" }
+                }
+              }
+            },
+            "scope_challenge_include_token_scopes": {
+              "type": "boolean",
+              "default": false,
+              "description": "When true, includes the token's existing scopes in the scope parameter of 403 insufficient_scope responses (workaround for MCP client SDKs that replace rather than accumulate scopes). When false (default), only the scopes required for the operation are returned (RFC 6750 strict)."
+            },
```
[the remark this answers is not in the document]
asoorm: Old comment - closing.

router-tests/testutil/jwt_helper.go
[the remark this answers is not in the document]
asoorm: This file no longer exists on the branch — it was removed in a subsequent revision.

router/pkg/mcpserver/auth_middleware.go
```diff
@@ -0,0 +1,465 @@
...
+		return nil, fmt.Errorf("failed to create authenticator: %w", err)
+	}
+
+	return &MCPAuthMiddleware{
+		authenticator:                    authenticator,
+		enabled:                          enabled,
+		resourceMetadataURL:              resourceMetadataURL,
+		scopes:                           scopes,
+		scopeChallengeIncludeTokenScopes: scopeChallengeIncludeTokenScopes,
+	}, nil
```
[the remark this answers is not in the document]
asoorm: Not needed - NewMCPAuthMiddleware is only called when OAuth is enabled (server.go:226), so tokenDecoder is always set.

client-tests/mcp-ts/mcp-debug-proxy.mjs
asoorm, line 1: We should just move this logging to the underlying MCP Server in debug mode.

```
httputil.DumpRequest / httputil.DumpResponse
```

client-tests/mcp-ts/package.json
```diff
@@ -0,0 +1,20 @@
...
+  "version": "1.0.0",
+  "description": "MCP OAuth scope enforcement E2E tests using the official MCP TypeScript SDK",
+  "type": "module",
+  "scripts": {
+    "test": "vitest run",
+    "test:watch": "vitest watch"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
```
endigma: What is ISC? Cosmo is Apache2 Licensed

client-tests/mcp-ts/package.json
endigma, line 1: Will this be covered by our dependency checks? cc @pepol

router-tests/cmd/oauth-server/main.go
endigma, line 1: Does this need to be externally running in order for some tests to pass? We should run anything needed internally.

router-tests/protocol/mcp_test.go
```diff
@@ -290,7 +292,7 @@ func TestMCP(t *testing.T) {
...
...e\":false,\"properties\":{\"hasPets\":{\"nullable\":true,\"type\":\"boolean\"},\"nationality\":{\"enum\":[\"AMERICAN\",\"DUTCH\",\"ENGLISH\",\"GERMAN\",\"INDIAN\",\"SPANISH\",\"UKRAINIAN\"],\"nullable\":true,\"type\":\"string\"},\"nested\":{\"additionalProperties\":false,\"nullable\":true,\"properties\":{\"hasChildren\":{\"nullable\":true,\"type\":\"boolean\"},\"maritalStatus\":{\"enum\":[\"ENGAGED\",\"MARRIED\"],\"nullable\":true,\"type\":\"string\"}},\"type\":\"object\"}},\"type\":\"object\"}},\"type\":\"object\"}\n```\n\nGraphQL Query:\n```\nquery MyEmployees($criteria: SearchInput) {\n    findEmployees(criteria: $criteria) {\n        id\n        isAvailable\n        currentMood\n        products\n        details {\n            forename\n            nationality\n        }\n    }\n}\n```\n\nUsage Instructions:\n1. Endpoint: https://api.example.com/graphql\n2. HTTP Method: POST\n3. Headers Required:\n   - Content-Type: application/json; charset=utf-8\n\nRequest Format:\n```json\n{\n  \"query\": \"<operation_query>\",\n  \"variables\": <your_variables_object>\n}\n```\nImportant Notes:\n1. Use the query string exactly as provided above\n2. Do not modify or reformat the query string"
```
endigma: Can we use a snapshot for this? Or backticks? I don't see how I could ever see what's wrong from such a huge oneline comparison.

router-tests/testutil/auth_helpers.go
endigma, line 1: Maybe https://github.com/containers/image/blob/main/docker/wwwauthenticate.go ?

router-tests/testutil/oauth_server.go
```diff
@@ -0,0 +1,567 @@
...
+}
+
+// OAuthTestServer is a minimal OAuth 2.1 Authorization Server for integration tests.
+//
+// Supported endpoints:
+//   - GET  /.well-known/jwks.json                   — JWKS for token verification
+//   - GET  /.well-known/oauth-authorization-server   — AS metadata (RFC 8414)
+//   - POST /token                                    — Token endpoint (client_credentials + authorization_code)
+//   - POST /register                                 — Dynamic client registration (RFC 7591)
+//   - GET  /authorize                                — Authorization endpoint (auto-approves for testing)
```
endigma: You could put these comments next to the route definitions so it's more easily updated. I tend to not like huge doc comments on types describing their methods, these descriptions don't show in LSP and need to be manually maintained.

router-tests/protocol/mcp_auth_client_test.go
endigma, line 1: Seems like this and it's sibling should be in a subdirectory? I don't think `.` is run in CI anymore.

router/pkg/config/config.schema.json
```diff
@@ -2047,15 +2047,15 @@
...
+                        "default": "2m",
                         "duration": {
                           "minimum": "0s"
                         }
                       },
                       "interval": {
                         "type": "string",
                         "description": "Token refill interval for the rate limiter.",
-                        "default": "1m",
+                        "default": "30s",
```
endigma, line 2058: This change seems unrelated, is it just a small fix?

router/pkg/config/config.go
```diff
@@ -1124,7 +1124,50 @@ type MCPConfiguration struct {
...
+	// ScopeChallengeIncludeTokenScopes controls whether the server includes the token's existing scopes
+	// in the scope parameter of 403 insufficient_scope responses.
+	// When false (default), only the scopes required for the operation are returned (RFC 6750 strict).
+	// When true, the token's existing scopes are unioned with the required scopes.
+	// This is a workaround for MCP client SDKs that replace rather than accumulate scopes.
+	ScopeChallengeIncludeTokenScopes bool `yaml:"scope_challenge_include_token_scopes" envDefault:"false" env:"SCOPE_CHALLENGE_INCLUDE_TOKEN_SCOPES"`
+	// MaxScopeCombinations sets the upper limit on the number of OR-group combinations
+	// produced when computing the Cartesian product of @requiresScopes across fields.
+	// Defaults to 2048 if not set. Increase for complex RBAC configurations.
+	MaxScopeCombinations int `yaml:"max_scope_combinations" envDefault:"0" env:"MAX_SCOPE_COMBINATIONS"`
```
endigma: Why doesn't it `envDefault` to 2048? Is the default somewhere nested further in?

client-tests/mcp-ts/mcp-debug-proxy.mjs
endigma, line 1: I don't mind this existing, but I would prefer if it was a very thin Go server instead of TypeScript. I'm sure Claude could handle the port and Go is far less susceptible to just.. problems.. long term. Perhaps it can be put in with the testing OAuth server.

router/mcp.oauth.config.yaml
endigma, line 1: This is a router config? Can you add a makefile target somewhere to use it? Can it go with the other 2 example configs in the router directory?

client-tests/mcp-ts/mcp.test.config.yaml
```diff
@@ -0,0 +1,67 @@
+# MCP OAuth Scope Enforcement — Test Configuration
+# yaml-language-server: $schema=../../router/pkg/config/config.schema.json
+#
+# This config is used by the E2E tests in McpOAuthScopes.test.ts.
+# It adds built-in tool scope requirements on top of the base MCP OAuth setup.
+#
+# Usage:
+#   1. Start the test OAuth server:
+#        go run ./router-tests/cmd/oauth-server
```
endigma: If the router in a debug configuration could need this to work maybe we should consider running it in docker compose, or as a cosmo or router level makefile target

endigma (changes requested):

client-tests/mcp-ts/package.json
asoorm: Great catch - 

For context - Left over from npm init defaults. This started as throwaway so reviewers could familiarise with the MCP Auth flow - instead of watching a Loom. Thought it was useful enough to keep, but I never revisited the scaffolding. Will switch to Apache 2 to match Cosmo.

router-tests/cmd/oauth-server/main.go
asoorm: It's following same pattern as `router-tests/cmd/jwks-server` and `router-tests/cmd/config-validator`. Dev helpers for running the router locally against a real Authorization Server, not test dependencies.

Go tests spin OAuth up in-process via httptest + the jwks package.

I agree it doesn't belong in router-tests, but then we should refactor the other packages if we want to move it as separate task?

router-tests/testutil/auth_helpers.go
asoorm: Nice find!

router-tests/protocol/mcp_auth_client_test.go
asoorm: Thanks for finding this.

router-tests/testutil/oauth_server.go
asoorm: Good call. Dropped the per-route list from the type doc - the route path strings are self-describing.

router/pkg/config/config.schema.json
asoorm: It is just a small fix. config.schema.json has 10s/1m for refresh_unknown_kid.max_wait/interval respectively. But the Go envDefault tags on RefreshUnknownKID are 2m/30s.

I hit the mismatch while merging my MCP Oauth JWKS config against the existing one and aligned the schema to what the ccode actually uses.

asoorm:
```
type RefreshUnknownKID struct {
	Enabled  bool          `yaml:"enabled" envDefault:"false"`
	MaxWait  time.Duration `yaml:"max_wait" envDefault:"2m"`
	Interval time.Duration `yaml:"interval" envDefault:"30s"`
	Burst    int           `yaml:"burst" envDefault:"2"`
}
```

router-tests/cmd/oauth-server/main.go
endigma: if it's prior art, just put it alongside those, we can move them all together later if its an issue, i was more just commenting on whether or not it needs to be run for router-tests to pass

router/mcp.oauth.config.yaml
asoorm: moved to router directory

router-tests/cmd/mcp-debug-proxy/main.go
```diff
@@ -0,0 +1,95 @@
...
+	logHeaders("  ←", resp.Header)
+	if len(body) > 0 {
+		log.Printf("  ← body: %s", formatBody(body, maxBody))
+	}
+}
+
+func logHeaders(prefix string, h http.Header) {
+	for k, vs := range h {
+		for _, v := range vs {
+			log.Printf("%s %s: %s", prefix, k, v)
```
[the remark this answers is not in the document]
asoorm: The whole purpose of this dev-proxy is to capture tokens. That's why it's separated from the main router binary. It's only designed as a convenience for developers.

router-tests/protocol/mcp_auth_client_test.go
```diff
@@ -0,0 +1,178 @@
+package integration
```
Noroth, line 1: Can we reconsider the names of the files? They look identical while one file is actually a test file and the other one only has a mock client implementation. If you use a file picker with Cmd+P for instance you cannot really distinguish between the files as only the `o` is different.

router-tests/testutil/oauth_server.go
```diff
@@ -0,0 +1,559 @@
...
+	keyID    string
+	issuer   string
+	audience string
+	jwksURL  string
+	server   *http.Server
+	storage  jwkset.Storage
+
+	mu      sync.RWMutex
+	clients map[string]*OAuthClient // clientID → client
+	codes   map[string]*authCode    // code → pending auth code
```
Noroth: Nit:

Unless we want to have unicode characters in strings (for displaying something in stdout) we should avoid them in comments.

```suggestion
	clients map[string]*OAuthClient // clientID -> client
	codes   map[string]*authCode    // code -> pending auth code
```

router-tests/testutil/oauth_server.go
```diff
@@ -0,0 +1,559 @@
...
+
+// NewOAuthTestServer creates and starts a minimal OAuth 2.1 AS on a random port.
+func NewOAuthTestServer(t *testing.T, opts *OAuthTestServerOptions) (*OAuthTestServer, error) {
+	t.Helper()
+
+	if opts == nil {
+		opts = &OAuthTestServerOptions{}
+	}
+
+	port := freeport.GetOne(t)
```
Noroth: We should try to avoid freeport because it can be quite flaky. Can we use httptest instead if we need a random port?

router/core/router.go
```diff
@@ -981,6 +981,19 @@ func (r *Router) bootstrap(ctx context.Context) error {
 			mcpOpts = append(mcpOpts, mcpserver.WithCORS(*r.corsOptions))
 		}
 
+		// Add OAuth configuration if enabled
+		if r.mcp.OAuth.Enabled {
```
Noroth: We should start cleaning up this huge function step by step. You could factor out the `if r.mcp.Enabled` in line 945 and put the whole block in a separate receiver function.

router/pkg/mcpserver/auth_middleware.go
```diff
@@ -0,0 +1,332 @@
...
+}
+
+// MCPAuthMiddleware provides HTTP-level authentication and scope enforcement for MCP.
+type MCPAuthMiddleware struct {
+	authenticator                    authentication.Authenticator
+	resourceMetadataURL              string
+	scopes                           config.MCPOAuthScopesConfiguration
+	scopeChallengeIncludeTokenScopes bool
+	toolScopesMu                     sync.RWMutex
+	toolScopes                       map[string][][]string // toolName → OR-of-AND scope groups
```
Noroth:
```suggestion
	toolScopes                       map[string][][]string // toolName -> OR-of-AND scope groups
```

router/pkg/mcpserver/auth_middleware.go
```diff
@@ -0,0 +1,332 @@
...
+				m.sendInsufficientScopeResponse(w, m.scopes.Initialize, tokenScopes, missing)
+				return
+			}
+		}
+
+		// Parse JSON-RPC body for method-level scope checks (SSE/GET requests have no body)
+		var body []byte
+		if r.Method == http.MethodPost && r.Body != nil {
+			limitedReader := io.LimitReader(r.Body, maxBodyBytes+1)
+			body, err = io.ReadAll(limitedReader)
```
Noroth: nit: variable only used one time

```suggestion
			body, err = io.ReadAll(io.LimitReader(r.Body, maxBodyBytes+1))
```

router/pkg/mcpserver/auth_middleware.go
```diff
@@ -0,0 +1,332 @@
...
+	if len(m.scopes.Initialize) > 0 {
+		authHeader += fmt.Sprintf(`, scope="%s"`, strings.Join(m.scopes.Initialize, " "))
+	}
+	if m.resourceMetadataURL != "" {
+		authHeader += fmt.Sprintf(`, resource_metadata="%s"`, m.resourceMetadataURL)
+	}
+	if err != nil {
+		desc := strings.ReplaceAll(err.Error(), `"`, `'`)
+		authHeader += fmt.Sprintf(`, error_description="%s"`, desc)
+	}
```
Noroth: From the context that I saw this is fine, but be careful with passing errors directly to this function because some errors might be wrapped types and calling err.Error() will unwrap the errors and you might leak information that you don't intend to.

Probably better to change it to errorDescription string as done in `writeScopeChallenge`

router/pkg/mcpserver/auth_middleware.go
```diff
@@ -0,0 +1,332 @@
...
+			limitedReader := io.LimitReader(r.Body, maxBodyBytes+1)
+			body, err = io.ReadAll(limitedReader)
+			if err != nil {
+				m.sendUnauthorizedResponse(w, fmt.Errorf("failed to read request body"))
+				return
+			}
+			if int64(len(body)) > maxBodyBytes {
+				m.sendUnauthorizedResponse(w, fmt.Errorf("request body too large"))
+				return
+			}
```
Noroth, line 147: Can this actually happen? Is this an edge case and do you test it?

router/pkg/mcpserver/scope_challenge.go
```diff
@@ -0,0 +1,88 @@
+package mcpserver
+
+// SatisfiesAnyGroup checks whether tokenScopeSet satisfies at least one AND-group
+// in the OR-of-AND scope requirements. Returns true if no requirements exist.
+func SatisfiesAnyGroup(tokenScopeSet map[string]struct{}, orScopes [][]string) bool {
```
Noroth: I don't think this function needs to be exported? Or is this supposed to be used in custom modules?

router/pkg/mcpserver/scope_challenge.go
```diff
@@ -0,0 +1,88 @@
...
+
+// BestScopeChallenge picks the AND-group closest to the client's current scopes.
+// Returns the complete AND-group that the client should request, or nil if any
+// group is already satisfied.
+//
+// Algorithm:
+//  1. For each AND-group, count how many scopes the token is missing.
+//  2. If any group has 0 missing, return nil (already satisfied).
+//  3. Pick the group with the fewest missing scopes (ties: first group wins).
+func BestScopeChallenge(tokenScopes []string, combinedOrScopes [][]string) []string {
```
Noroth: Same thing here does this function have to be exported?

router/pkg/mcpserver/auth_middleware.go
```diff
@@ -0,0 +1,332 @@
...
+
+	"github.com/wundergraph/graphql-go-tools/v2/pkg/astparser"
+)
+
+type contextKey string
+
+const (
+	userClaimsContextKey contextKey = "mcp_user_claims"
+	// maxBodyBytes prevents memory exhaustion from oversized payloads.
+	maxBodyBytes int64 = 10 << 20 // 10 MB
```
Noroth: Erm ackshually.... it's M`i`B 😄 

```suggestion
	maxBodyBytes int64 = 10 << 20 // 10 MiB
```

router/pkg/mcpserver/auth_middleware_test.go
```diff
@@ -0,0 +1,700 @@
...
+				assert.Contains(t, wwwAuth, tt.wantWWWAuthenticatePrefix)
+			}
+		})
+	}
+}
+
+func TestMCPAuthMiddleware_PerToolScopes(t *testing.T) {
+	t.Parallel()
+
+	const testMetadataURL = "http://localhost:5025/.well-known/oauth-protected-resource/mcp"
```
Noroth: Hardcoded port seems a bit odd in parallel tests is this safe?
Multiple tests in this file

router/pkg/mcpserver/execute_graphql_scope_test.go
```diff
@@ -0,0 +1,205 @@
...
+	"github.com/stretchr/testify/assert"
+
+	"github.com/wundergraph/cosmo/router/pkg/authentication"
+	"github.com/wundergraph/cosmo/router/pkg/config"
+)
+
+func TestMCPAuthMiddleware_ExecuteGraphQLScopes(t *testing.T) {
+	t.Parallel()
+
+	const testMetadataURL = "http://localhost:5025/.well-known/oauth-protected-resource/mcp"
```
Noroth: More hard coded ports and parallel tests

router/pkg/mcpserver/execute_graphql_scope_test.go
```diff
@@ -0,0 +1,205 @@
...
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/wundergraph/cosmo/router/pkg/authentication"
+	"github.com/wundergraph/cosmo/router/pkg/config"
+)
+
+func TestMCPAuthMiddleware_ExecuteGraphQLScopes(t *testing.T) {
```
Noroth: It would be better if you had proper names for test functions in general or introduce a subtest. Something like `t.Run("Should expect x by doing y with conditions z")`

router/pkg/mcpserver/scope_challenge.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	return combinedOrScopes[bestIdx]
+}
+
+// BestScopeChallengeWithExisting returns the challenge scopes, optionally including
+// the token's existing scopes. When includeExisting is true, the result is the union
+// of the token's current scopes and the best AND-group, deduplicated. This works
+// around MCP client SDKs that replace rather than accumulate scopes on re-authorization.
+func BestScopeChallengeWithExisting(tokenScopes []string, combinedOrScopes [][]string, includeExisting bool) []string {
```
Noroth: Also exported

router/pkg/mcpserver/scope_extractor.go
```diff
@@ -0,0 +1,173 @@
...
+	TypeName  string
+	FieldName string
+	OrScopes  [][]string
+}
+
+// ScopeExtractor walks operations and extracts per-field scope requirements
+// from FieldConfigurations.
+type ScopeExtractor struct {
+	// scopeIndex maps "TypeName.FieldName" to OR-of-AND scope groups for O(1) lookup.
+	scopeIndex           map[string][][]string
```
Noroth, line 26: nit: You could introduce a type alias for [][]string to make it a bit more readable and potentially have receiver functions to help with indexing and length checks

router/pkg/mcpserver/scope_extractor.go
```diff
@@ -0,0 +1,173 @@
...
+		walker:     &walker,
+		operation:  operation,
+		definition: e.schemaDoc,
+		scopeIndex: e.scopeIndex,
+	}
+
+	walker.RegisterEnterFieldVisitor(v)
+
+	report := &operationreport.Report{}
+	walker.Walk(operation, e.schemaDoc, report)
```
Noroth, line 73: You don't check the report for errors

router/pkg/mcpserver/scope_extractor.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+	result := make([][]string, 0, total)
+	for _, groupA := range a {
+		for _, groupB := range b {
+			merged := mergeAndDedup(groupA, groupB)
+			result = append(result, merged)
+		}
+	}
+	return result, nil
+}
```
Noroth, line 162: Just a question, we have Or groups but in this case when a = 0 and b > 0 then we have a total of 0 and not b. But maybe this is right?

router/pkg/mcpserver/scope_extractor.go
```diff
@@ -0,0 +1,173 @@
...
+			seen[s] = struct{}{}
+			result = append(result, s)
+		}
+	}
+	for _, s := range b {
+		if _, ok := seen[s]; !ok {
+			seen[s] = struct{}{}
+			result = append(result, s)
+		}
+	}
```
Noroth, line 179: You could do this in one loop with a counter variable that is the size of the bigger slice and then just doing a length check for a and b

router/pkg/mcpserver/server.go
```diff
@@ -203,14 +219,81 @@ func NewGraphQLSchemaServer(routerGraphQLEndpoint string, opts ...func(*Options)
...
-	// Create the MCP server
-	mcpServer := server.NewMCPServer(
-		"wundergraph-cosmo-"+strcase.ToKebab(options.GraphName),
-		"0.0.1",
-		// Prompt, Resources aren't supported yet in any of the popular platforms
-		server.WithToolCapabilities(true),
-		server.WithPaginationLimit(100),
-		server.WithRecovery(),
+	// Create a cancellable context for managing the server lifecycle
+	ctx, cancel := context.WithCancel(context.Background())
```
Noroth: Isn't this a router scope? Why context.Background? Also why did you put the ctx in the struct? I don't think it is used somewhere?

router/pkg/mcpserver/server.go
```diff
@@ -432,77 +578,75 @@ func (s *GraphQLSchemaServer) Stop(ctx context.Context) error {
...
-
 	}
 
 	// Get operations filtered by the excludeMutations setting
 	operations := s.operationsManager.GetFilteredOperations()
 
 	graphqlOperationNames := make([]string, 0, len(operations))
 
+	// Build per-tool scope map for the auth middleware
+	toolScopes := make(map[string][][]string)
```
Noroth, line 645: [][]string again that could be type aliased.

router/pkg/mcpserver/server.go
```diff
@@ -556,62 +700,94 @@ func (s *GraphQLSchemaServer) registerTools() error {
...
 		}
 
+		jsonBytes := request.Params.Arguments
+
 		// Validate the JSON input against the pre-compiled schema derived from the operation input type
 		if handler.compiledSchema != nil {
 			if err := s.schemaCompiler.ValidateInput(jsonBytes, handler.compiledSchema); err != nil {
-				return mcp.NewToolResultErrorFromErr("Input validation Error", err), nil
+				return &mcp.CallToolResult{
+					Content: []mcp.Content{&mcp.TextContent{Text: fmt.Sprintf("Input validation error: %v", err)}},
```
Noroth, line 785: Is it fine to leak this error?

router/pkg/mcpserver/server.go
```diff
@@ -621,13 +797,10 @@ func (s *GraphQLSchemaServer) handleOperation(handler *operationHandler) func(ct
...
+func (s *GraphQLSchemaServer) handleGraphQLOperationInfo() func(ctx context.Context, request *mcp.CallToolRequest) (*mcp.CallToolResult, error) {
+	return func(ctx context.Context, request *mcp.CallToolRequest) (*mcp.CallToolResult, error) {
 		var input GraphQLOperationInfoInput
-		inputBytes, err := json.Marshal(request.GetArguments())
-		if err != nil {
-			return nil, fmt.Errorf("failed to marshal input arguments: %w", err)
-		}
+		inputBytes := request.Params.Arguments
 		if err := json.Unmarshal(inputBytes, &input); err != nil {
 			return nil, fmt.Errorf("failed to unmarshal input arguments: %w. Ensure you provide {\"operationName\": \"<n>\"}", err)
```
Noroth: you can use backticks  `` to avoid escaping the quotes

router/pkg/mcpserver/server.go
Noroth, line 833: Could we use the package text/template for this?

Noroth (changes requested):
Mostly nits

router/pkg/mcpserver/auth_middleware.go
asoorm: If a client sends a huge request, we only read up to the limit `LimitReader` - and accept the call as valid. LimitReader doesn't return an error if we go over the limit. so we still need a defensive guard. Do we need a test for this?

endigma (approved):
approved, my additional comments are part of ludwig's review

router/pkg/mcpserver/auth_middleware_test.go
asoorm: Yeah - it's only used to populate the WWW-Authenticate header. `resource_metadata` value. Nothing binds to the port in the tests - that said - the port number is obviously confusing - so will swap for https://example.test/... url to make it clear.

router/pkg/mcpserver/execute_graphql_scope_test.go
asoorm: As above -

router/pkg/mcpserver/scope_extractor.go
asoorm: I like your thoughts - but I think the blast radius for addressing this will ripple too far for marginal readability improvement. names already carry the semantics - `OrScopes` and `combinedOrScopes`. Are you happy for me to leave as-is for now?

router/pkg/mcpserver/scope_extractor.go
asoorm: Good catch - I've added some tests to make these edge cases explicit. But an empty OR list means that there is no valid way to pass.

router/pkg/mcpserver/server.go
asoorm: Same push-back as earlier - would prefer to leave as-is.

router/pkg/mcpserver/server.go
asoorm: Checked the lib - should be safe - it is just input validation against the JSON schema. Nothing server-side leaks

Noroth (approved):

PR #2667, opened 2026-03-19
chore(demo): fix code generation for cost directives

ysmolski:
Include new directives into the schemas to make gqlgen happy.

Fixes ENG-9243

## Summary by CodeRabbit

* **Chores**
  * Bumped Go module and developer-tooling dependencies to newer versions.

* **New Features**
  * Introduced @cost and @listSize GraphQL directives to annotate cost and list-sizing metadata.

* **Configuration**
  * GraphQL codegen configured to skip runtime handling for the new directives.

* **CI**
  * Added a Demo CI workflow to run generation and checks for demo changes.

* **Tests**
  * Updated test/demo GraphQL schemas and test configs, including several field nullability and list-element signature adjustments.


demo/pkg/subgraphs/employees/subgraph/schema.graphqls
```diff
@@ -18,6 +18,21 @@ extend schema
...
+  | OBJECT
+  | SCALAR
+
+directive @listSize(
+  assumedSize: Int,
+  slicingArguments: [String!],
+  sizedFields: [String!],
+  requireOneSlicingArgument: Boolean = true
+) on FIELD_DEFINITION
+
```
Noroth, line 35: Thinking of this in general. I think we had a couple of changes in Cosmo where our demo project was actually affected, but as we don't automatically check it, they ended up to be silent errors.

I think we should have a github action for the demo to run a few checks. We should regenerate every subgraph, check that generation is successful and validate that git is not dirty after generation.

Noroth: I also think, that I'm `thinking` a lot, omg

ysmolski: I wonder if should do the same for config we use in testing. Those configs are generated from the demo too.

ysmolski: I did that as well

PR #2677, opened 2026-03-23
feat(router): entity caching with L1/L2, shadow mode, and analytics

jensneuse:
## Summary

- **Entity-level caching** for federated GraphQL: caches individual entities by `@key` fields at the router level, so a `Product` cached by one query is reused by every query that needs it — regardless of query shape or subgraph
- **Two-layer architecture**: L1 (per-request dedup via `sync.Map`) eliminates redundant entity fetches within a single request; L2 (Redis or in-memory ristretto) shares cached entities across requests with per-entity TTLs via `@entityCache(maxAge)`
- **Shadow mode** for safe rollout: cache reads/writes happen but cached data is never served — compares cached vs fresh responses field-by-field to measure staleness before going live
- **Analytics pipeline**: entity cache hit/miss/stale metrics exported via OpenTelemetry for observability dashboards
- **Circuit breaker** on L2: prevents cascade failures when Redis is degraded
- **Composition support**: new `@entityCache` and `@entityCacheTTL` directives validated and propagated through the composition layer
- **~40K lines** across composition (TypeScript), engine (graphql-go-tools), router (Go), and proto layers

Companion PR: [graphql-go-tools #1259](https://github.com/wundergraph/graphql-go-tools/pull/1259)

## Documentation

All entity caching docs live in [`docs/entity-caching/`](docs/entity-caching/):

| Document | Description |
|----------|-------------|
| [`engineering-brief.md`](docs/entity-caching/engineering-brief.md) | Full engineering brief — architecture, design decisions, review focus areas |
| [`directives.md`](docs/entity-caching/directives.md) | `@entityCache` and `@entityCacheTTL` directive definitions and usage |
| [`configuration.md`](docs/entity-caching/configuration.md) | Router config reference (L1/L2, circuit breaker, env vars) |
| [`analytics.md`](docs/entity-caching/analytics.md) | Entity cache analytics and observability |
| [`analytics-rfc.md`](docs/entity-caching/analytics-rfc.md) | RFC for the entity cache analytics pipeline |

Start with the **engineering brief** for a full overview of the design and review focus areas.

## Test plan

- [ ] Composition directive validation tests (`composition/tests/v1/directives/entity-caching.test.ts`)
- [ ] Router entity cache unit tests (`router/core/executor_entity_cache_test.go`)
- [ ] Entity cache package tests (memory, redis, circuit breaker)
- [ ] Integration tests (`router-tests/entity_caching/`)
- [ ] Standard subgraph entity caching integration tests
- [ ] Shadow mode correctness verification
- [ ] CI green on all platforms

## Summary by CodeRabbit

* **New Features**
  * Entity caching: directives for object/root/mutation caching with L1 (per-request) and L2 (memory/Redis), per-subgraph overrides, shadow mode, partial loads, and circuit-breaker handling
  * Request-scoped field deduplication via request-scoped keys
  * Playground: cache-mode controls, cache badges, cache details/keys viewer, Cache Explorer UX
  * Cache analytics: snapshotting and observable metrics; demo & benchmark tooling

* **Documentation**
  * New RFCs, guides, and engineering briefs covering entity caching, directives, config, analytics, and demos

* **Tests**
  * Extensive unit/integration suites and router tests covering caching scenarios and backends


composition/src/errors/errors.ts
```diff
@@ -1705,6 +1705,89 @@ export function oneOfRequiredFieldsError({ requiredFieldNames, typeName }: OneOf
   );
 }
 
+// Entity caching directive error messages.
+// These are reported during composition when subgraph schemas use entity caching directives
+// incorrectly. Each error corresponds to a validation rule in validateAndExtractEntityCachingConfigs().
+
+// @entityCache requires the type to be a federation entity (must have @key)
+export function entityCacheWithoutKeyErrorMessage(typeName: string): string {
+  return `Type "${typeName}" has @entityCache but no @key directive.`;
```
Aenimus: What happens if the entity has `@key` directive(s) but can't resolve its own key(s)?

composition/src/errors/errors.ts
```diff
@@ -1705,6 +1705,89 @@ export function oneOfRequiredFieldsError({ requiredFieldNames, typeName }: OneOf
...
+}
+
+// @queryCache is restricted to Query fields — Mutation/Subscription fields use @cachePopulate or @cacheInvalidate
+export function queryCacheOnNonQueryFieldErrorMessage(fieldCoords: string): string {
+  return `@queryCache is only valid on Query fields, found on "${fieldCoords}".`;
```
Aenimus: `The directive "@queryCache" must only be defined on query root fields; however, it was defined on field "${fieldCoords}".`

Also: have you checked what happens if the query root type is renamed?

composition/src/utils/string-constants.ts
```diff
@@ -64,6 +67,7 @@ export const FRAGMENT_SPREAD_UPPER = 'FRAGMENT_SPREAD';
 export const FROM = 'from';
 export const HYPHEN_JOIN = `\n -`;
 export const ID_SCALAR = 'ID';
```
Aenimus, line 69: Please make sure these additions are alphabetical. Currently, they are not.

composition/src/v1/constants/directive-definitions.ts
```diff
@@ -101,6 +111,29 @@ export const AUTHENTICATED_DEFINITION: DirectiveDefinitionNode = {
   repeatable: false,
 };
 
+// @cacheInvalidate on FIELD_DEFINITION
```
Aenimus: You must namespace all these directives.

Aenimus (changes requested):
Review ongoing, but mandatory changes already required.

composition/src/v1/constants/directive-definitions.ts
```diff
@@ -501,6 +567,21 @@ export const INTERFACE_OBJECT_DEFINITION: DirectiveDefinitionNode = {
   repeatable: false,
 };
 
+// @is(field: String!) on ARGUMENT_DEFINITION
```
Aenimus: You only allow top level mapping?

composition/src/v1/warnings/warnings.ts
```diff
@@ -190,6 +190,40 @@ export function singleSubgraphInputFieldOneOfWarning({
   });
 }
 
+// Warns when a @queryCache field cannot construct complete cache keys from its arguments.
+// This happens when the entity has @key fields that don't correspond to any argument (by name or @is mapping).
+// The field can still populate the cache (writes), but cannot serve reads from cache.
+export function incompleteQueryCacheKeyMappingWarning(
+  subgraphName: string,
```
Aenimus: Please us the proper aliases, e.g., `FieldName`, `SubgraphName`,  and `TypeName`etc.

composition/src/v1/warnings/warnings.ts
```diff
@@ -190,6 +190,40 @@ export function singleSubgraphInputFieldOneOfWarning({
...
+  fieldCoords: string,
+  entityType: string,
```
Aenimus: `entityTypeName: TypeName`

composition/src/v1/warnings/warnings.ts
```diff
@@ -190,6 +190,40 @@ export function singleSubgraphInputFieldOneOfWarning({
...
+  unmappedKeyField: string,
```
Aenimus: `unmappedFieldName: FieldName`

composition/src/v1/warnings/warnings.ts
```diff
@@ -190,6 +190,40 @@ export function singleSubgraphInputFieldOneOfWarning({
...
+): Warning {
+  return new Warning({
```
Aenimus, line 207: Why is this only a warning? It's an error IMO.

composition/src/v1/warnings/warnings.ts
```diff
@@ -190,6 +190,40 @@ export function singleSubgraphInputFieldOneOfWarning({
...
+      ` or use @is(field: "${unmappedKeyField}") to enable cache reads.`,
+    subgraph: {
+      name: subgraphName,
+    },
+  });
+}
+
+// Warns when @is(field: "x") is used on an argument already named "x" — the auto-mapping
+// would produce the same result, so @is adds no value and can be safely removed.
+export function redundantIsDirectiveWarning(subgraphName: string, argumentName: string, fieldCoords: string): Warning {
```
Aenimus: This is a better example of what constitutes a warning—something mostly harmless but should/could be cleaned up.

composition/src/v1/constants/directive-definitions.ts
jensneuse: explain why? and what namespace do you propose?

composition/src/utils/string-constants.ts
jensneuse: is there a comment or documentation indicating that the order should be alphabetical? if not, should I add one?

composition/src/v1/constants/directive-definitions.ts
Aenimus: We namespace all directives that were not previously defined in federation.

This clarifies intent, and reduces the chance to clash. This is particularly important for common names like `@is` or `@cacheInvalidate`.

I'd propose `@openfed__`, which is what we already use.

Alternatively `@wg__` might be considered.

composition/src/v1/warnings/warnings.ts
jensneuse: the queryCache directive has 2 modes:
a) the field is cached as a root field
b) the field is cached as an entity (under the condition that the mapping is correct)
which means that it can be intentional to not have a mapping or it might not be possible, but it can be possible that it was a mistake, as such, I think it's ok to make it a warning because I don't want to block the user if they don't want to or cannot treat the field as an entity

composition/src/utils/string-constants.ts
Aenimus: No, but I think it's immediately clear looking at the file.

composition/src/v1/warnings/warnings.ts
Aenimus: If you're sure.

But it also looks like you've verified that you only allow mapping of top-level fields, but key fields can be arbitrarily nested.

docs/entity-caching/directives.md
```diff
@@ -0,0 +1,478 @@
+# Entity Caching Directives
```
Aenimus, line 1: When you namespace the directives, remember update all references, such as here.

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -445,6 +476,13 @@ export class NormalizationFactory {
...
+  // Populated in Phase 1 of validateAndExtractEntityCachingConfigs().
+  // Used as a lookup in Phase 2 to verify that @queryCache/@cacheInvalidate/@cachePopulate
+  // fields return entity types that have opted into caching.
+  entityCacheConfigByTypeName = new Map<
+    string,
+    { maxAgeSeconds: number; includeHeaders: boolean; partialCacheLoad: boolean; shadowMode: boolean }
+  >();
   errors = new Array<Error>();
   entityDataByTypeName = new Map<string, EntityData>();
   entityInterfaceDataByTypeName = new Map<string, EntityInterfaceSubgraphData>();
```
Aenimus: string -> `TypeName`

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -445,6 +476,13 @@ export class NormalizationFactory {
   directiveDefinitionDataByName = initializeDirectiveDefinitionDatas();
   doesParentRequireFetchReasons = false;
   edfsDirectiveReferences = new Set<string>();
+  // Populated in Phase 1 of validateAndExtractEntityCachingConfigs().
+  // Used as a lookup in Phase 2 to verify that @queryCache/@cacheInvalidate/@cachePopulate
+  // fields return entity types that have opted into caching.
+  entityCacheConfigByTypeName = new Map<
+    string,
```
Aenimus: string -> `TypeName`

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -445,6 +476,13 @@ export class NormalizationFactory {
...
+    { maxAgeSeconds: number; includeHeaders: boolean; partialCacheLoad: boolean; shadowMode: boolean }
+  >();
   errors = new Array<Error>();
   entityDataByTypeName = new Map<string, EntityData>();
```
Aenimus: string -> `TypeName`

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -3472,6 +3510,504 @@ export class NormalizationFactory {
...
+   *   @cachePopulate(maxAge: Int)      — on Mutation/Subscription fields: writes the returned entity into cache
+   *   @is(field: String!)             — on arguments: maps a query argument to an entity @key field for cache key construction
+   *
+   * Processing happens in three phases:
+   *   Phase 1: Collect @entityCache from entity types → populates entityCacheConfigByTypeName lookup
+   *   Phase 2: Process root type fields (Query/Mutation/Subscription) for @queryCache, @cacheInvalidate, @cachePopulate, @is
+   *   Phase 3: Attach validated configs to their respective ConfigurationData entries for serialization
+   */
+  validateAndExtractEntityCachingConfigs() {
+    const entityCacheConfigs: EntityCacheConfig[] = [];
```
Aenimus: Convention is `Array<T>` not `T[]`

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -3472,6 +3510,504 @@ export class NormalizationFactory {
...
+    const rootFieldCacheConfigs: RootFieldCacheConfig[] = [];
+    const cachePopulateConfigs: CachePopulateConfig[] = [];
+    const cacheInvalidateConfigs: CacheInvalidateConfig[] = [];
+
+    // Phase 1: Extract @entityCache from object types.
+    // This must run before Phase 2 because @queryCache/@cacheInvalidate/@cachePopulate validate
+    // that the return entity type has @entityCache via the entityCacheConfigByTypeName lookup.
+    for (const [typeName, parentData] of this.parentDefinitionDataByTypeName) {
```
Aenimus, line 3757: We loop through parent data already. We don't need a separate loop. Just call a function when the switch is an `Object` type.

Aenimus: i think you should take inspiration from other functions that extract directives. Search for `extract` and it should help.

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -3472,6 +3510,504 @@ export class NormalizationFactory {
...
+        continue;
+      }
+      const entityCacheDirectives = parentData.directivesByName.get(ENTITY_CACHE);
+      if (!entityCacheDirectives || entityCacheDirectives.length < 1) {
+        continue;
+      }
+      // Rule 1: Must have @key
+      if (!this.keyFieldSetDatasByTypeName.has(typeName)) {
+        this.errors.push(
+          invalidDirectiveError(ENTITY_CACHE, typeName, '1st', [entityCacheWithoutKeyErrorMessage(typeName)]),
```
Aenimus: Why are you hard-coding all these ordinals?

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -3472,6 +3510,504 @@ export class NormalizationFactory {
...
+          configData.cacheInvalidateConfigurations.push(ci);
+        }
+      }
+    }
+  }
+
+  // Recursively unwraps NonNullType wrappers to check if the underlying type is a list.
+  // Used to determine whether a @queryCache field returns a single entity (cache-readable)
+  // or a list (cache writes only, no key-based lookups).
+  isListType(typeNode: TypeNode): boolean {
```
Aenimus: This helper already exists `isTypeNodeListType`

StarpTech:
@claude review

router-tests/entity_caching_standard_subgraphs_test.go
```diff
@@ -0,0 +1,291 @@
...
+	return c
+}
+
+// entityCachingConfig returns RouterOptions that enable entity caching with
+// the given MemoryEntityCache as the default L2 cache.
+func entityCachingConfig(cache *entitycache.MemoryEntityCache) []core.Option {
+	return []core.Option{
+		core.WithEntityCaching(config.EntityCachingConfiguration{
+			Enabled: true,
+			L1: config.EntityCachingL1Configuration{
```
SkArchon, line 30: I set this to false (L1 Cache Enabled) and the tests still succeeded

router-tests/entity_caching/entity_caching_test.go
```diff
@@ -0,0 +1,1190 @@
+package entity_caching
```
SkArchon, line 1: I tried disabling L2 by setting

```L2: config.EntityCachingL2Configuration{
				Enabled: false,
			},
```

I noticed that the following tests still succeeded

```
disabled_caching_does_not_cache
partial_cache_load
negative_cache_caches_null
root_field_different_args
query_cache_shadow_mode
mutation_populates_cache
mutation_populate_writes_to_cache
shadow_mode_with_failing_cache
negative_cache_ttl_expiry
cache_populate_maxage_override
```

I believe some test cases are false positives, like "partial_cache_load". Would be good to check.

router-tests/entity_caching/entity_caching_test.go
```diff
@@ -0,0 +1,1190 @@
...
+
+	itemsModel "github.com/wundergraph/cosmo/router-tests/entity_caching/subgraphs/items/subgraph/model"
+	"github.com/wundergraph/cosmo/router-tests/testenv"
+	"github.com/wundergraph/cosmo/router/core"
+	nodev1 "github.com/wundergraph/cosmo/router/gen/proto/wg/cosmo/node/v1"
+	"github.com/wundergraph/cosmo/router/pkg/config"
+	"github.com/wundergraph/graphql-go-tools/v2/pkg/engine/resolve"
+)
+
+func TestEntityCaching(t *testing.T) {
```
SkArchon, line 20: Could we name the tests a bit more descriptively like in the `entity_caching_standard_subgraphs_test.go` test (e.g :- `shadow mode always fetches from subgraph`).

router-tests/entity_caching/entity_caching_test.go
```diff
@@ -0,0 +1,1190 @@
...
+			// createItem has @cachePopulate(maxAge: 60)
+			res := xEnv.MakeGraphQLRequestOK(testenv.GraphQLRequest{
+				Query: `mutation { createItem(name: "Foobar", category: "test") { id name category } }`,
+			})
+			require.Contains(t, res.Body, `"Foobar"`)
+			require.Contains(t, res.Body, `"createItem"`)
+		})
+	})
+
+	t.Run("l1_deduplication", func(t *testing.T) {
```
SkArchon: This is a false posive, when L1.Enabled = false, it also succeeds, I believe this is because the engine combines "a" and "b" into one query when sending the request to the subgraph. (I believe this applies to other L1 tests as well).

router-tests/entity_caching/entity_caching_test.go
```diff
@@ -0,0 +1,1190 @@
...
+	itemsModel "github.com/wundergraph/cosmo/router-tests/entity_caching/subgraphs/items/subgraph/model"
+	"github.com/wundergraph/cosmo/router-tests/testenv"
+	"github.com/wundergraph/cosmo/router/core"
+	nodev1 "github.com/wundergraph/cosmo/router/gen/proto/wg/cosmo/node/v1"
+	"github.com/wundergraph/cosmo/router/pkg/config"
+	"github.com/wundergraph/graphql-go-tools/v2/pkg/engine/resolve"
+)
+
+func TestEntityCaching(t *testing.T) {
+	t.Parallel()
```
SkArchon, line 21: Could we have maybe a different structure for the tests for L1, L2, Shadow, etc. Like

```
	t.Run("L1 Tests (or a better name)", func(t *testing.T) {
        t.Run("deduplication", func(t *testing.T) {
            ...
        })
        ...
    })
    t.Run("L2Tests (or a better name)", func(t *testing.T) {
       ...
```

I think this would help identify easily identify what we are looking at

SkArchon (changes requested):

proto/wg/cosmo/node/v1/node.proto
```diff
@@ -191,6 +196,52 @@ message EntityInterfaceConfiguration {
...
 }
 
+// Entity type caching configuration (from @entityCache directive)
+message EntityCacheConfiguration {
+  string type_name = 1;
+  int64 max_age_seconds = 2;
+  bool include_headers = 3;
+  bool partial_cache_load = 4;
+  bool shadow_mode = 5;
+  int64 negative_cache_ttl_seconds = 6;
```
Noroth: Maybe it's just me but I would have expected something else with the naming.
Negative seconds sounds a bit like you have to configure -10s to do something.

maybe something like `null_entity_cache_ttl_seconds`?

proto/wg/cosmo/node/v1/node.proto
```diff
@@ -191,6 +196,52 @@ message EntityInterfaceConfiguration {
...
+message EntityCacheFieldMapping {
+  string entity_key_field = 1;
+  repeated string argument_path = 2;
+}
+
+// Mutation/subscription cache population configuration (from @cachePopulate directive)
+message CachePopulateConfiguration {
+  string field_name = 1;
+  string operation_type = 2;
+  optional int64 max_age_seconds = 3;
```
Noroth: Just a question, this field is supposed to be optional and the other `max_age_seconds` are required values? in `RootFieldCacheConfiguration` and `EntityCacheConfiguration`

When this one is omitted we have unlimited TTL? What if we provide negative numbers here?

Are the other fields defined as int64 to allow -1 for unlimited? Or should they always have positive >0 values?

From just the review perspective it looks a bit inconsistent, maybe some comments would help here

docs/entity-caching/directives.md
```diff
@@ -0,0 +1,478 @@
...
+) on OBJECT
+```
+
+**Not repeatable.** Each entity type may have at most one `@entityCache` directive.
+
+**Arguments:**
+
+| Argument | Type | Default | Description |
+|----------|------|---------|-------------|
+| `maxAge` | Int! | required | TTL in seconds. How long a cached entity remains valid. |
```
Noroth: What if negative number?

router/core/executor.go
```diff
@@ -252,3 +257,58 @@ func (b *ExecutorConfigurationBuilder) buildPlannerConfiguration(ctx context.Con
...
+			}
+			result[subgraphName][ec.TypeName] = &resolve.EntityCacheInvalidationConfig{
+				CacheName:                   resolveEntityCacheProviderID(cfg, subgraphName, ec.TypeName),
+				IncludeSubgraphHeaderPrefix: ec.IncludeHeaders,
+			}
+		}
+	}
+	if len(result) == 0 {
+		return nil
+	}
```
Noroth: We can return early by checking the length to not create the map.

And we can use the `Get...` receiver functions in proto to make sure we check the receiver for `nil`

```suggestion
	if cfg == nil || !cfg.Enabled || len(engineConfig.GetDatasourceConfigurations() == 0) {
		return nil
	}

	result := make(map[string]map[string]*resolve.EntityCacheInvalidationConfig)
	for _, ds := range engineConfig.GetDatasourceConfigurations() {
		subgraphName := subgraphNameByID(subgraphs, ds.Id)
		for _, ec := range ds.GetEntityCacheConfigurations() {
			if _, ok := result[subgraphName]; !ok {
				result[subgraphName] = make(map[string]*resolve.EntityCacheInvalidationConfig)
			}
			result[subgraphName][ec.TypeName] = &resolve.EntityCacheInvalidationConfig{
				CacheName:                   resolveEntityCacheProviderID(cfg, subgraphName, ec.TypeName),
				IncludeSubgraphHeaderPrefix: ec.IncludeHeaders,
			}
		}
	}

	return result
}
```

router/core/executor_entity_cache_test.go
```diff
@@ -0,0 +1,142 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	nodev1 "github.com/wundergraph/cosmo/router/gen/proto/wg/cosmo/node/v1"
+	"github.com/wundergraph/cosmo/router/pkg/config"
+	"github.com/wundergraph/graphql-go-tools/v2/pkg/engine/resolve"
+)
+
+func TestResolveEntityCacheProviderID(t *testing.T) {
```
Noroth, line 16: Tests could be t.Parallel()

router/core/factoryresolver.go
```diff
@@ -723,6 +818,10 @@ func (l *Loader) dataSourceMetaData(in *nodev1.DataSourceConfiguration) *plan.Da
 	return out
 }
 
+func (l *Loader) resolveEntityCacheProviderID(subgraphName, typeName string) string {
+	return resolveEntityCacheProviderID(l.entityCachingConfig, subgraphName, typeName)
+}
```
Noroth: I personally would rather inline this.

router/core/graph_server.go
```diff
@@ -22,6 +22,7 @@ import (
 	"github.com/golang-jwt/jwt/v5"
 	"github.com/klauspost/compress/gzhttp"
 	"github.com/klauspost/compress/gzip"
+	"github.com/wundergraph/graphql-go-tools/v2/pkg/engine/resolve"
```
Noroth: If possible run gci on this file

router/core/graph_server.go
```diff
@@ -104,7 +105,11 @@ type (
 		traceDialer             *TraceDialer
 		connector               *grpcconnector.Connector
 		circuitBreakerManager   *circuit.Manager
-		headerPropagation       *HeaderPropagation
+		headerPropagation          *HeaderPropagation
+		entityCacheInstances       map[string]resolve.LoaderCache
+		entityCacheKeyInterceptors []EntityCacheKeyInterceptor
+		otlpEntityCacheMetrics     *rmetric.EntityCacheMetrics
+		promEntityCacheMetrics     *rmetric.EntityCacheMetrics
```
Noroth: Not related to this PR but I think we should refactior the metrics and have essentially one field that contains all the instantiated metrics instead of appending more and more metrics. In another PR I mean.

router/core/graph_server.go
```diff
@@ -193,8 +198,24 @@ func newGraphServer(ctx context.Context, r *Router, routerConfig *nodev1.RouterC
...
-		storageProviders:  &r.storageProviders,
-		headerPropagation: r.headerPropagation,
+		storageProviders:           &r.storageProviders,
+		headerPropagation:          r.headerPropagation,
+		entityCacheKeyInterceptors: r.entityCacheKeyInterceptors,
+	}
+
+	entityCacheInstances, err := r.buildEntityCacheInstances()
+	if err != nil {
+		cancel()
```
Noroth, line 217: I'm not sure but I don't think we need to cancel here because we error out and the router fails to start anyways.

router/core/graph_server.go
```diff
@@ -543,6 +568,68 @@ func (s *graphServer) setupEngineStatistics(baseAttributes []attribute.KeyValue)
 	return nil
 }
 
+func (s *graphServer) validateEntityCacheOverrides(
```
Noroth: What exactly is happening here? This function doesn't return an error or changes anything if I'm not mistaken. Is it just for logging purposes?

router/core/router.go
```diff
@@ -1282,6 +1291,105 @@ func (r *Router) buildClients() error {
...
+		cache = entitycache.NewCircuitBreakerCache(cache, entitycache.CircuitBreakerConfig{
+			Enabled:          true,
+			FailureThreshold: l2Cfg.CircuitBreaker.FailureThreshold,
+			CooldownPeriod:   l2Cfg.CircuitBreaker.CooldownPeriod,
+		})
+	}
+	return cache, nil
+}
+
+func (r *Router) findRedisClient(providerID string) (rd.RDCloser, error) {
```
Noroth: This is creating a redis client right? `Find` sounds a bit wrong here.

router/pkg/entitycache/circuit_breaker_test.go
```diff
@@ -0,0 +1,238 @@
+package entitycache
```
Noroth, line 1: Would also probably nice to have some concurrent tests here

router/pkg/entitycache/memory.go
```diff
@@ -0,0 +1,109 @@
...
+	m := &MemoryEntityCache{}
+	cache, err := ristretto.NewCache(&ristretto.Config[string, []byte]{
+		NumCounters:        numCounters,
+		MaxCost:            maxSizeBytes,
+		BufferItems:        64,
+		IgnoreInternalCost: true,
+		OnEvict: func(item *ristretto.Item[[]byte]) {
+			m.len.Add(-1)
+		},
+	})
```
Noroth, line 38: You could also use the ristrettor cache metrics here. Then you could get the number of entries from there

router/pkg/entitycache/redis.go
```diff
@@ -0,0 +1,81 @@
...
+var _ io.Closer = (*RedisEntityCache)(nil)
+
+type RedisEntityCache struct {
+	client    redis.UniversalClient
+	keyPrefix string
+}
+
+func NewRedisEntityCache(client redis.UniversalClient, keyPrefix string) *RedisEntityCache {
+	return &RedisEntityCache{client: client, keyPrefix: keyPrefix}
+}
```
Noroth, line 22: Should we check that `keyPrefix` is not empty string here?

router/pkg/entitycache/redis_test.go
```diff
@@ -0,0 +1,161 @@
...
+	"testing"
+	"time"
+
+	"github.com/alicebob/miniredis/v2"
+	"github.com/redis/go-redis/v9"
+	"github.com/stretchr/testify/require"
+	"github.com/wundergraph/graphql-go-tools/v2/pkg/engine/resolve"
+)
+
+func newTestRedisCache(t *testing.T, prefix string) (*RedisEntityCache, *miniredis.Miniredis) {
```
Noroth, line 14: Should we test the behavior when the client is closed and we still try to read something?

router/pkg/metric/entity_cache_metrics.go
```diff
@@ -0,0 +1,228 @@
...
+}
+
+func (m *EntityCacheMetrics) RecordSnapshot(ctx context.Context, snapshot resolve.CacheAnalyticsSnapshot) {
+	for _, event := range snapshot.L1Reads {
+		cacheType := cacheTypeFromEntityType(event.EntityType)
+		switch event.Kind {
+		case resolve.CacheKeyHit:
+			m.recordRequestStat(ctx, "hits", "l1", cacheType)
+		case resolve.CacheKeyMiss:
+			m.recordRequestStat(ctx, "misses", "l1", cacheType)
```
Noroth: We have constants for hits and misses in `otel/attributes.go`. We could re-use them

router-tests/entity_caching/redis_test.go
```diff
@@ -0,0 +1,123 @@
...
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/wundergraph/graphql-go-tools/v2/pkg/engine/resolve"
+)
+
+func TestRedis(t *testing.T) {
+	t.Parallel()
+
+	t.Run("basic_miss_then_hit", func(t *testing.T) {
```
Noroth, line 15: It would be good in general to have a more descriptive name for tests, like `Should first miss and then hit the cache` or something like that. It would be good to have a short description of the expected test outcome

Noroth (changes requested):

PR #2681, opened 2026-03-23
fix: documenting and updating span / metric error cases

SkArchon:
The goal of this PR is to identify if there any scenarios where we don't mark spans as errors, as well as update documentation.

This PR introduces

* Client Disconnections: error is recorded but span is not marked as error
* GRPC service connectivity issues: the span is set to error
* Updated documentation

@StarpTech should we allow Client Disconnections not being recorded as an error for spans be based on a flag? I could imagine users where they want it to be treated as errors (for spans).

## Release Notes

* **Documentation**
  * Added documentation clarifying how client disconnections vs server-side failures affect span error status and error metrics reporting.

* **Bug Fixes**
  * Client disconnections are now excluded from error spans, error metrics, and error-level logging; only server-side failures increment error counts.

* **Tests**
  * Added test coverage for client disconnection telemetry, error handling behavior, and schema usage metrics export.

## Checklist

## Open Source AI Manifesto

### DIFF
--- a/cli/CHANGELOG.md
+++ b/cli/CHANGELOG.md
@@ -4,6 +4,10 @@ Binaries are attached to the github release otherwise all images can be found [h
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## [0.112.3](https://github.com/wundergraph/cosmo/compare/wgc@0.112.2...wgc@0.112.3) (2026-03-24)
+
+**Note:** Version bump only for package wgc
+
 ## [0.112.2](https://github.com/wundergraph/cosmo/compare/wgc@0.112.1...wgc@0.112.2) (2026-03-20)
 
 ### Bug Fixes
--- a/cli/package.json
+++ b/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "wgc",
-  "version": "0.112.2",
+  "version": "0.112.3",
   "description": "The official CLI tool to manage the GraphQL Federation Platform Cosmo",
   "type": "module",
   "main": "dist/src/index.js",
--- a/controlplane/CHANGELOG.md
+++ b/controlplane/CHANGELOG.md
@@ -4,6 +4,12 @@ Binaries are attached to the github release otherwise all images can be found [h
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## [0.209.1](https://github.com/wundergraph/cosmo/compare/controlplane@0.209.0...controlplane@0.209.1) (2026-03-24)
+
+### Bug Fixes
+
+* handle arrayJoin of empty arrays in clickhouse queries ([#2682](https://github.com/wundergraph/cosmo/issues/2682)) ([733c14b](https://github.com/wundergraph/cosmo/commit/733c14b4710ed093f9ccee19836054f7e92647a4)) (@JivusAyrus)
+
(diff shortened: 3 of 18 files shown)


router/core/context.go
```diff
@@ -278,6 +278,11 @@ type requestContext struct {
 	customFieldValueRenderer resolve.FieldValueRenderer
 	// forceSha256Compute indicates whether the Sha256Hash of the operation should definitely be computed
 	forceSha256Compute bool
+	// clientDisconnected indicates that the request context was canceled (context.Canceled)
+	// during request processing. This most commonly happens when a client disconnects, but can
+	// also occur during graceful server shutdown. In either case, the work was interrupted, not
+	// failed, so it should not inflate error metrics or mark spans as ERROR.
+	clientDisconnected bool
```
Noroth: It looks a bit generic. Would it make more sense to have a field that rather indicates whether to treat the current context span as an error?

router/core/graphql_prehandler.go
```diff
@@ -457,9 +457,17 @@ func (h *PreHandler) Handler(next http.Handler) http.Handler {
 		// and enrich the context to make it available in the request context as well for metrics etc.
 		next.ServeHTTP(ww, r)
 
-		// Mark the root span of the router as failed, so we can easily identify failed requests
+		// Mark the root span of the router as failed, so we can easily identify failed requests.
+		// Client disconnections are not server-side errors and should not mark the root span as ERROR.
 		if requestContext.error != nil {
-			rtrace.AttachErrToSpan(trace.SpanFromContext(r.Context()), requestContext.error)
+			fromContext := trace.SpanFromContext(r.Context())
```
Noroth: please call it `contextSpan` or something like that but not `fromContext` that is a bad naming

router/core/errors.go
```diff
@@ -111,17 +111,29 @@ func logInternalErrorsFromReport(report *operationreport.Report, requestLogger *
...
 	if requestContext == nil {
 		return
 	}
 
+	// Client disconnections are not server-side errors. We still record the error
+	// on the request context for response handling and access logging, but we do not
+	// mark the span as ERROR or increment error metrics.
+	if errors.Is(err, context.Canceled) {
+		requestContext.SetError(err)
+		requestContext.clientDisconnected = true
```
Noroth: Looking at this I'm not even sure if we need the field at all. You could have a helper function for the context to check if the error is context.Canceled.

logic-wise you always set the error in the requestContext so instead you can just assign the error with `SetError` and then if requestContext.IsCanceled() or so.

What do you think?

router/pkg/grpcconnector/grpccommon/grpc_plugin_client.go
```diff
@@ -55,6 +57,8 @@ func WithReconnectConfig(reconnectTimeout time.Duration, pingInterval time.Durat
 
 var _ grpc.ClientConnInterface = &GRPCPluginClient{}
 
+const pluginNotActiveErrorMessage = "plugin is not active"
```
Noroth: maybe?

```suggestion
var pluginNotActiveErr = errors.New("plugin is not active")
```

router/core/context.go
SkArchon: Will remove

router/pkg/grpcconnector/grpccommon/grpc_plugin_client.go
SkArchon: I had this initially but was not sure if it was less nicer to do err.Error(), switched.

router/core/errors.go
SkArchon: Makes sense, though I didnt add a separate function.

router/core/graphql_prehandler.go
SkArchon: Done

Issue #2395, opened 2025-12-04
Rate Limiter per client.


rajendersaini:
### Component(s)

router

### Is your feature request related to a problem? Please describe.

Ability for us to use rate limiter per client

### Describe the solution you'd like

Ability for us to use rate limiter per client

Example let's say jwt have client_id, we can pull that via prefix and override the base config if we want to otherwise it uses default request rate limits.

rajendersaini:
https://github.com/wundergraph/cosmo/pull/2409

endigma:
Hi @rajendersaini, this should be resolved when https://github.com/wundergraph/cosmo/pull/2683 is released.

PR #2683, opened 2026-03-24
feat(router): support per-key rate-limit overrides with regex pattern matching

endigma:
Adds an `overrides` array to `simple_strategy` in the rate limiter config. Each override specifies a `matching` regex pattern tested against the resolved rate-limit key; the first match wins and unmatched keys fall back to global defaults.

Includes documentation updates to the hardening guide and configuration reference.

## Summary by CodeRabbit

* **New Features**
  * Per-key rate limit overrides via regex matching with first-match precedence and fallback to global defaults.
  * Overrides specify per-pattern rate, burst, and period values.

* **Documentation**
  * Expanded rate limiting docs and hardening guide with guidance, examples, and an updated example that derives the key suffix from the X-Api-Key header.

* **Tests**
  * Added tests for key-suffix extraction, override matching, precedence, and expected rate-limit behavior.

## Checklist

- [x] Documentation has been updated on [https://github.com/wundergraph/docs-website](https://github.com/wundergraph/docs-website).


endigma:
Resolves #2395

router/core/ratelimiter.go
```diff
@@ -60,53 +89,84 @@ type CosmoRateLimiter struct {
...
-	str, err := expr.ResolveStringExpression(c.keySuffixProgram, rc.expressionContext)
+
+	suffix, err = expr.ResolveStringExpression(c.keySuffixProgram, rc.expressionContext)
 	if err != nil {
-		return "", fmt.Errorf("failed to resolve key suffix expression: %w", err)
+		return "", "", fmt.Errorf("failed to resolve key suffix expression: %w", err)
 	}
-	buf := bytes.NewBuffer(make([]byte, 0, len(ctx.RateLimitOptions.RateLimitKey)+len(str)+1))
+
+	buf := bytes.NewBuffer(make([]byte, 0, len(ctx.RateLimitOptions.RateLimitKey)+len(suffix)+1))
```
SkArchon: Maybe move `len(ctx.RateLimitOptions.RateLimitKey)+len(suffix)+1)` to a different variable (with comment) so its easier to read

router/core/ratelimiter.go
```diff
@@ -60,53 +89,84 @@ type CosmoRateLimiter struct {
...
 		Period: ctx.RateLimitOptions.Period,
 	}
-	key, err := c.generateKey(ctx)
+
+	key, suffix, err := c.generateKey(ctx)
 	if err != nil {
 		return nil, err
 	}
+
+	limit := c.resolveLimit(suffix, defaultLimit)
```
SkArchon, line 120: Maybe make this getLimit and pass in the resolveContext.RateLimitOptions to it instead of the defaultLimit (so we construct only on when no overrides are found)

SkArchon:
Will take another look when the PR is ready for review

router/core/ratelimiter.go
endigma: kinda micro optimization and would hurt readability I think

router/core/ratelimiter.go
endigma: switched to just doing a string concat

router-tests/security/ratelimit_test.go
```diff
@@ -684,8 +699,90 @@ func TestRateLimit(t *testing.T) {
...
+			require.Contains(t, res.Body, `"remaining":3`)
+			require.NotContains(t, res.Body, `"errors"`)
+
+			res, err = xEnv.MakeGraphQLRequestWithHeaders(testenv.GraphQLRequest{
+				Query:     `query ($n:Int!) { employee(id:$n) { id details { forename surname } } }`,
+				Variables: json.RawMessage(`{"n":1}`),
+			}, map[string]string{
+				"X-Client-ID": "premium-user",
+			})
+			require.NoError(t, err)
```
SkArchon, line 780: nit: maybe also good to validate until premium user also gets errorred out.

router/pkg/config/config.go
```diff
@@ -634,12 +634,20 @@ type RedisConfiguration struct {
...
+	Burst                          int                 `yaml:"burst" envDefault:"10" env:"RATE_LIMIT_SIMPLE_BURST"`
+	Period                         time.Duration       `yaml:"period" envDefault:"1s" env:"RATE_LIMIT_SIMPLE_PERIOD"`
+	RejectExceedingRequests        bool                `yaml:"reject_exceeding_requests" envDefault:"false" env:"RATE_LIMIT_SIMPLE_REJECT_EXCEEDING_REQUESTS"`
+	RejectStatusCode               int                 `yaml:"reject_status_code" envDefault:"200" env:"RATE_LIMIT_SIMPLE_REJECT_STATUS_CODE"`
+	HideStatsFromResponseExtension bool                `yaml:"hide_stats_from_response_extension" envDefault:"false" env:"RATE_LIMIT_SIMPLE_HIDE_STATS_FROM_RESPONSE_EXTENSION"`
+	Overrides                      []RateLimitOverride `yaml:"overrides,omitempty"`
+}
+
+type RateLimitOverride struct {
+	Matching string        `yaml:"matching"`
```
SkArchon, line 647: nit: spacing before `yaml:"matching"`

docs-website/router/security/hardening-guide.mdx
```diff
@@ -87,6 +87,35 @@ rate_limit:
     key_prefix: cosmo_rate_limit
 ```
 </Note>
+
+### Per-Key Rate Limit Overrides
+
+You can apply different rate limits to specific keys by defining overrides under `simple_strategy`. Each override includes a `matching` field, a [Go regex](https://pkg.go.dev/regexp/syntax) tested against the suffix portion of the rate-limit key (the result of `key_suffix_expression`), not the full prefixed Redis key. The first matching override wins; unmatched keys fall back to the global defaults.
```
SkArchon, line 93: Would be good to have an example / describe an example, something similar to the test you have.

endigma: does that work? i tailored the example to the security context a bit as well

router-tests/security/ratelimit_test.go
endigma: done

rajbayer:
@endigma  - thank you so much. This is great.

PR #2686, opened 2026-03-24
fix(router)!: prevent set from being forwarded to client

SkArchon:
This PR refactors the response side of header forwarding

> [!WARNING]
> **THIS IS A BREAKING CHANGE**

Which includes

* "set" is now not propagated to the client, instead the goal of set is to add a header to the response from the subgraph, so it looks like the set header "came from the subgraph directly".

* Since "set" is now acts like a header that came directly from the subgraph, if users want to they "technically" can add a propagate op to send this header to the client. Though this is still not recommended (mentioned in the docs), as the same limitations of how set was before applies. (Limitations such as if no subgraph runs nothing will get propagated).

* Previously if a user uses "set" to set a Cache-Control header, this would hard override any values set from the cache control configuration without looking at what is the most restrictive configration. This has been removed. Since "set" now acts like a header sent directly from the subgraph, it would behave the same way as it would for an actual Cache Control header that was sent from a subgraph.

**Why these changes?**
* Currently "set" returns values to the client by default, but the API for this is highly confusing, this is as "set" is interpreted as "set" a header to the response from the subgraph. It is also not document correctly / clearly. This also leads to weird and potentially unexpected behaviour

For example
```
 headers:
  subgraphs:
    products:
      response:
        - op: "set"
          name: "X-Powered-By2232"
          value: "wundergraph"
```
The header would be sent to the client ONLY IF the products subgraph was called. Likewise when using `headers.all` for the same, in case of errors the client header is not set. This PR's goal is to realign set to mean "set (or add) a header to the subgraph response headers".

If user's still want to do this they can use the following alternative (currently only supports expressions).

```
headers:
  router:
    response:
      - name: X-Demo-Header
        # The expression can also be dynamic and not just a static value
        expression: "'static-value'"
```

* Cache Control override
Previously a user was allowed to use the "set" header to do a hard override on the Cache-Control header. Usually we would only pick the most restrictive cache control header, however upon usage of set, whatever was provided would be overridden. After discussing internally we decided to not allow users a hard override for Cache-Control headers, as it could be a potential security issue (i.e. :- When a subgraph wants to expire its data, it's not expired because the "set" override set a higher ttl, and we don't use the lower ttl because of the "set" override for the header)

## Summary by CodeRabbit

* **Documentation**
  * Clarified response header rule semantics and execution order; added guidance and examples showing how injected headers affect processing.
* **Bug Fixes / Behavior**
  * Response-header "set" operations are internal-only (not forwarded) unless explicitly followed by a "propagate".
  * Injected Cache-Control values now feed into the restrictive cache-selection algorithm so per-subgraph injections influence final client cache headers.

## Checklist

## Open Source AI Manifesto


router/core/header_rule_engine.go
```diff
@@ -456,13 +468,10 @@ func (h *HeaderPropagation) OnOriginResponse(resp *http.Response, ctx RequestCon
...
-		if rule.Name == cacheControlKey {
-			// Handle the case where the cache control header is set explicitly
-			propagation.setCacheControl = true
-		}
-		propagation.m.Unlock()
+		// Inject the value into the subgraph response headers so it looks like it
+		// came from the subgraph. Downstream rules (propagate, cache control
+		// algorithm, etc.) will process it naturally.
+		res.Header.Set(rule.Name, rule.Value)
 		return
```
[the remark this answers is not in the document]
SkArchon: That is fine

docs-website/router/proxy-capabilities/adjusting-cache-control.mdx
```diff
@@ -152,4 +152,23 @@ headers:
...
+    - name: "specific-subgraph"
+      value: "max-age=5400, public"
+```
+
+<Warning>
+As such we recommend not using this method and simply sticking to using the `subgraph` configuration in `cache_control_policy`, as it makes things more explicit and clearer.
+</Warning>
+
+<Note>
+  Until router version VERSION, setting a `Cache-Control` header would do a hard override of whatever value would be computed by the `cache_control_policy` configuration. This has been changed since.
```
endigma: VERSION? how do we chicken-egg this with the release now that docs are in tree?

docs-website/router/proxy-capabilities/adjusting-cache-control.mdx
```diff
@@ -152,4 +152,23 @@ headers:
...
+cache_control_policy:
+  enabled: true
+  value: "max-age=180, public"
+  subgraphs:
+    - name: "specific-subgraph"
+      value: "max-age=5400, public"
+```
+
+<Warning>
+As such we recommend not using this method and simply sticking to using the `subgraph` configuration in `cache_control_policy`, as it makes things more explicit and clearer.
```
endigma:
```suggestion
We do not recommend using this method, simply sticking to using the `subgraph` configuration in `cache_control_policy` makes things more explicit and clearer.
```

docs-website/router/proxy-capabilities/subgraph-response-header-operations.mdx
```diff
@@ -90,12 +86,59 @@ Currently, we support the following header rules:
...
+Within each scope, rules execute in definition order. This means a `set` rule defined before a `propagate` rule in the same scope will inject the value into the subgraph response before the `propagate` rule evaluates it.
 
-  * `value` - The value to set for the header
+## Response Header Set
+
+The `set` operation injects a header value into the subgraph response, making it appear as if the subgraph returned it. This value is then processed by downstream rules (e.g., `propagate` rules or the [restrictive cache control algorithm](/router/proxy-capabilities/adjusting-cache-control)).
 
 <Note>
-  Go canonicalizes headers by default e.g. `x-my-header` to `X-My-Header.` Write your rule accordingly or use `(?i)``^X-Test-.*` flags to make your regex case insensitive.
+The `set` value is **not** forwarded to the client response.
```
endigma:
```suggestion
The `set` value is **not** forwarded to the client response unless it is propagated.
```

docs-website/router/proxy-capabilities/subgraph-response-header-operations.mdx
```diff
@@ -90,12 +86,59 @@ Currently, we support the following header rules:
...
 </Note>
+
+
+<Warning>
+You can use `set` to set a header value and then use a `propagate` rule to forward that same header to the client. However, we would caution against this, as the header will not be propagated if no subgraph was actually called.
+</Warning>
```
endigma, line 111: Can we merge all this into one block? It's kinda whiplash to read "is **not** forwarded**" -> "do xyz to forward it"

docs-website/router/proxy-capabilities/subgraph-response-header-operations.mdx
```diff
@@ -90,12 +86,59 @@ Currently, we support the following header rules:
...
+```
+
+In this example, when a request hits `specific-subgraph`, the `Cache-Control: max-age=5400` value is injected into the subgraph response. The restrictive cache control algorithm then processes this value alongside other subgraph responses to compute the final `Cache-Control` header sent to the client. This is equivalent to doing
+
+```
+cache_control_policy:
+  enabled: true
+  value: "max-age=180, public"
+
+```
```
endigma: If we don't want people to do this, perhaps we should not include so many searchable examples of doing it in the docs

router-tests/operations/singleflight_test.go
```diff
@@ -1057,23 +1057,23 @@ func TestSingleFlight(t *testing.T) {
...
 			responses := runConcurrentSingleflightRequests(t, xEnv, `{ employee(id: 1) { id } }`, 5)
 			for i, res := range responses {
-				require.Equal(t, "value-a", res.Response.Header.Get("X-Header-A"),
-					"response %d missing X-Header-A", i)
-				require.Equal(t, "value-b", res.Response.Header.Get("X-Header-B"),
-					"response %d missing X-Header-B", i)
+				require.Equal(t, "", res.Response.Header.Get("X-Header-A"),
+					"response %d: set response headers should not be forwarded to the client", i)
+				require.Equal(t, "", res.Response.Header.Get("X-Header-B"),
+					"response %d: set response headers should not be forwarded to the client", i)
```
endigma, line 1072: `require.Empty` I think?

docs-website/router/proxy-capabilities/subgraph-response-header-operations.mdx
```diff
@@ -90,12 +86,59 @@ Currently, we support the following header rules:
 
   * `default` - Fallback to this value when the `named`, `matching` or `rename` header could not be found.
 
-* `set` - Sets a header on the request forward to the subgraph. You must set the following values:
```
endigma, line 93: Does the page still flow properly after this change? We still have `set`?

docs-website/router/proxy-capabilities/subgraph-response-header-operations.mdx
```diff
@@ -90,12 +86,59 @@ Currently, we support the following header rules:
...
+<Warning>
+You can use `set` to set a header value and then use a `propagate` rule to forward that same header to the client. However, we would caution against this, as the header will not be propagated if no subgraph was actually called.
+</Warning>
+
+### Configuration
+
+* `name` - The name of the header to set
+* `value` - The value to set for the header
+
+### Example: Setting Cache Control on a Subgraph Response
```
endigma: Perhaps we can replace this with a use case `set` is actually ideal for? We can feature config like this elsewhere as a footgun example but I think showing an example then a warning to not do it is bad practice

endigma (changes requested):

docs-website/router/proxy-capabilities/adjusting-cache-control.mdx
SkArchon: Do a release -> then update and merge since its only my pr ideally -> and then merge
Either way can open a new PR if the number is wrong worst case

endigma: You mean releasing from the PR branch?

SkArchon: More like timing manually computing the release number based on what commits are present after the last release, but probably just easier to open two PRs.

docs-website/router/proxy-capabilities/subgraph-response-header-operations.mdx
SkArchon: Makes sense, I still kept it but without the example to redirect them to the correct page, since users could end up on the page with the mental model of searching for setting cache control headers.

docs-website/router/proxy-capabilities/subgraph-response-header-operations.mdx
SkArchon: Added back with correct text.

docs-website/router/proxy-capabilities/subgraph-response-header-operations.mdx
SkArchon: Example removed and simplified.

endigma (approved):

router-tests/operations/singleflight_test.go
Noroth: You gave a thumbs up but it's still require.Equal

Noroth (approved):

PR #2714, opened 2026-03-31
feat: upgrade otel dependencies with backward compatible semconv attributes

Noroth:
This PR is the first step of our OTEL upgrade. It ensures that the dependencies are upgraded to the latest version while we are still able to drop/rewrite any newly introduced/altered attributes. This is necessary because various telemetry systems currently rely on the attributes emitted by the router.

In the future we will need to slowly migrate to the new attributes

## Summary by CodeRabbit

* **Chores**
  * Updated Go toolchain and bumped many dependencies and build tools (telemetry, metrics, gRPC, protobufs, Prometheus, JWT).

* **Improvements**
  * Reduced telemetry noise by dropping HTTP-instrumentation metrics and filtering specified span attributes.
  * Remapped semantic-convention attribute keys for downstream compatibility.
  * Added HTTP request/response content-length attributes and adjusted OTLP exporter defaults for traces/metrics.

* **Tests**
  * Added telemetry test helpers and expanded/robustified telemetry and metric tests.

## Checklist

## Open Source AI Manifesto


router/pkg/trace/filtering_provider.go
```diff
@@ -0,0 +1,97 @@
...
+
+func (s *filteringSpan) SetAttributes(kv ...attribute.KeyValue) {
+	n := 0
+	for i := range kv {
+		if _, drop := semconvDropKeys[kv[i].Key]; !drop {
+			kv[n] = kv[i]
+			n++
+		}
+	}
+	s.Span.SetAttributes(kv[:n]...)
```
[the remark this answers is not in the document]
Noroth: No it doesn't `Span.SetAttributes` is actually appending attributes in the underlying OTEL logic. We just have to make sure we don't append attributes we don't want to have

router/go.mod
endigma, line 71: Would be good to bump this as well, it had an otel issue

Noroth: This bumps the minimum required patch version of Go. This is actually a good thing but I believe we should do this in a separate PR as there are more deps that could be bumped.

router/pkg/trace/semconv_processor.go
```diff
@@ -0,0 +1,80 @@
...
+// Shutdown implements [trace.SpanProcessor].
+func (s *semconvProcessor) Shutdown(ctx context.Context) error { return nil }
+
+// OnEnd implements [trace.SpanProcessor].
+func (*semconvProcessor) OnEnd(s trace.ReadOnlySpan) {
+	kind := s.SpanKind()
+	attributes := s.Attributes()
+
+	for i := range attributes {
+		if mappedKey, ok := semconvCommonMapping[attributes[i].Key]; ok {
```
SkArchon, line 62: Would be good to add a comment here why we override, instead of adding (so both attributes are available)

router-tests/security/circuit_breaker_test.go
```diff
@@ -939,6 +943,10 @@ func TestCircuitBreaker(t *testing.T) {
 							Name:  testutils.ToPtr("otel_scope_name"),
 							Value: testutils.ToPtr("cosmo.router.prometheus"),
 						},
+						{
```
SkArchon, line 946: This is the only new attribute from the upgrade? (Excluding the mappings)

Noroth: Yeah it's a new scope attribute, which you cannot configure to not be emitted.

router/pkg/trace/semconv_processor.go
Noroth: I added some more comments to explain the current and future state

router/pkg/metric/meter.go
```diff
@@ -390,6 +391,13 @@ func defaultPrometheusMetricOptions(ctx context.Context, serviceInstanceID strin
 		// In a custom View function, we need to explicitly copy the name, description, and unit.
 		s := sdkmetric.Stream{Name: i.Name, Description: i.Description, Unit: i.Unit}
 
+		// We currently don't want to expose the otelhttp scope metrics.
+		// TODO: Once we want to expose the otelhttp scope metrics, we need to .
```
alepane21: We need to ...?

router/pkg/trace/handler.go
```diff
@@ -45,7 +45,7 @@ func NewTracingHandler(s TracingHandlerOpts) func(next http.Handler) http.Handle
 		// Disable built-in metricStore through NoopMeterProvider
 		otelhttp.WithMeterProvider(sdkmetric.NewMeterProvider()),
 		otelhttp.WithSpanNameFormatter(s.SpanNameFormatter),
-		otelhttp.WithTracerProvider(s.TracerProvider),
+		otelhttp.WithTracerProvider(&FilteringTracerProvider{TracerProvider: s.TracerProvider}),
```
alepane21, line 48: Can we have this to be configurable per exporter? When exporting toward external services (like newrelic or datadog) I think it would be useful to allow the standard names, to make it easier to integrate with external services.

alepane21 (changes requested):

Noroth:
The current retry logic has some regressions because the request body is not rewinded automatically. As a workaround on each retry we have to rewind the body manually before calling the underlying transport.

I opened an issue in the otel repository and the maintainers are taking a look currently
https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8759

router/pkg/metric/meter.go
Noroth: yes 😄

router/pkg/trace/handler.go
Noroth: I plan take a look at more configuration options in the future. Sadly configuring individual exporters is not as easy with the otel framework

endigma (approved):

alepane21 (approved):

pepol (approved):

PR #2715, opened 2026-03-31
fix(cli): replace cli-table3 wordWrap with CLITable wrapper

### LINKED ISSUE #2619: `wgc router compose` hangs indefinitely when composition errors produce large error text

iliya-i:
## Summary                                                                                                            

Fixes #2619  `cli-table3`'s built-in `wordWrap` option uses `string-width` which executes `emoji-regex` on every word in every table cell. Rendering time grows super-linearly with text volume, causing the CLI process to hang indefinitely when composition errors produce large output. 

This PR introduces a `CLITable` wrapper class that replaces `cli-table3`'s `wordWrap` with a lightweight `wrapText` utility that breaks text on word boundaries using simple character counting. The wrapper automatically wraps all string cells based on their column width, so developers never need to think about it. 

* `cli/src/wrap-text.ts`: text wrapping utility 
* `cli/src/cli-table.ts`: `CLITable` class that wraps `cli-table3`, automatically applies `wrapText` per column based on `colWidths`, omits `wordWrap` and `wrapOnWordBoundary` from the type to prevent misuse 
* Replaced all `new Table` / `wordWrap: true` usages 
* Column overhead is derived from the `cli-table3` instance style

Based on the root cause analysis and `wrapText` utility from @kamil-gwozdz in #2620.

## Summary by CodeRabbit

* **New Features**
  * Deterministic CLI text wrapping and a unified table renderer for clearer console output across commands

* **Bug Fixes**
  * Prevented hangs and improved performance when rendering very large error messages

* **Tests**
  * Added regression and performance tests to validate wrapping behavior and guard against future hangs

## Checklist

## Open Source AI Manifesto


cli/src/cli-table.ts
```diff
@@ -0,0 +1,44 @@
...
+          })
+        : row;
+      this.table.push(wrapped);
+    }
+  }
+
+  toString(): string {
+    return this.table.toString();
+  }
+}
```
endigma, line 44: If you do `export ... as Table` instead of changing every import to `CLITable` this would shrink the diff by a lot

iliya-i: @endigma I made it different on purpose, since exporting as `Table` could be confused with the original `cli-table3` package, but I can change it feel strongly about reducing the diff?

endigma: more a nit than anything, up to you

cli/test/cli-table.test.ts
```diff
@@ -0,0 +1,52 @@
...
+      head: ['COUNT', 'AUTO', 'MESSAGE'],
+      colWidths: [10, null, 30],
+    });
+
+    table.push([42, 'not wrapped', 'wrapped']);
+
+    expect(spy).toHaveBeenCalledTimes(1);
+    expect(spy).toHaveBeenCalledWith('wrapped', 30 - DEFAULT_OVERHEAD);
+
+    spy.mockRestore();
```
comatory: `vite.config.ts` has `restoreMocks: true` so this should not be needed

cli/src/wrap-text.ts
```diff
@@ -0,0 +1,30 @@
+/**
+ * Wraps text to fit within a given width, breaking on word boundaries.
+ * This avoids cli-table3's built-in wordWrap which can deadlock due to
+ * expensive string-width/emoji-regex evaluations on large inputs (#2619).
+ */
+export function wrapText(text: string, maxWidth: number): string {
```
comatory, line 6: The function is used internally within just the table class right? Why not make it part of the class itself. It does not have access to `this` so it could be a static method. We can keep separate tests for it in extra describe block in `cli-table.test.ts` WDYT?

comatory (approved):
Just a few suggestions

cli/src/wrap-text.ts
iliya-i: this way it's a pure function and it's much easier to test

iliya-i:
https://github.com/wundergraph/cosmo/issues/2619#issuecomment-4178136933

PR #2724, opened 2026-04-01
feat: allow custom scope claim name

alepane21:
Some systems don't use JWT claim "scope" as the claim where they put scopes. In alternative we saw is "scp". This additional config will keep scope as the default value, but will allow users to choose for a different claim if needed.

## Summary by CodeRabbit

* **New Features**
  * Configurable JWT claim name for authorization scopes via a new scope_claim option (default: "scope"); access control respects this setting.

* **Tests**
  * Added integration and unit tests verifying custom scope-claim parsing and scope enforcement with non-default claim names.

* **Documentation**
  * Updated docs and examples to document scope_claim and when to set a custom claim (e.g., "scp").

## Checklist

## Open Source AI Manifesto


router/pkg/authentication/authentication.go
```diff
@@ -63,20 +75,27 @@ func (a *authentication) SetScopes(scopes []string) {
...
 		return nil
 	}
 	return strings.Split(scopes, " ")
 }
 
+func (a *authentication) scopeClaimKey() string {
+	if a == nil || len(a.scopeClaim) == 0 {
+		return DefaultScopeClaim
+	}
+	return a.scopeClaim
```
endigma: Why validate this every time instead of setting it up in a constructor/setter?

endigma (changes requested):

router/pkg/authentication/authentication.go
alepane21: You are right, it was useless complexity

router/pkg/authentication/authentication.go
```diff
@@ -55,6 +60,13 @@ func (a *authentication) Claims() Claims {
 	return a.claims
 }
 
+func (a *authentication) SetScopesClaim(scopeClaim string) {
+	if a == nil {
+		return
+	}
+	a.scopeClaim = scopeClaim
+}
```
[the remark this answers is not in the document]
alepane21: Don't need the white space trimming

router-tests/security/authentication_test.go
```diff
@@ -606,6 +606,57 @@ func TestAuthentication(t *testing.T) {
 			require.Equal(t, `{"data":{"employees":[{"id":1,"startDate":"January 2020"},{"id":2,"startDate":"July 2022"},{"id":3,"startDate":"June 2021"},{"id":4,"startDate":"July 2022"},{"id":5,"startDate":"July 2022"},{"id":7,"startDate":"September 2022"},{"id":8,"startDate":"September 2022"},{"id":10,"startDate":"November 2022"},{"id":11,"startDate":"November 2022"},{"id":12,"startDate":"December 2022"}]}}`, string(data))
 		})
 	})
+	t.Run("scopes required valid token from configured custom scope claim", func(t *testing.T) {
```
endigma: this is a very confusing test name, can you try rephrasing it as action + condition + expectation?

router-tests/security/authentication_test.go
```diff
@@ -606,6 +606,57 @@ func TestAuthentication(t *testing.T) {
...
+			require.NoError(t, err)
+			header := http.Header{
+				"Authorization": []string{"Bearer " + token},
+			}
+			res, err := xEnv.MakeRequest(http.MethodPost, "/graphql", header, strings.NewReader(employeesQueryRequiringClaims))
+			require.NoError(t, err)
+			defer res.Body.Close()
+			require.Equal(t, http.StatusOK, res.StatusCode)
+			require.Equal(t, testutils.JwksName, res.Header.Get(xAuthenticatedByHeader))
+			data, err := io.ReadAll(res.Body)
```
endigma: Is this weird way of making requests required for the test or something outdated? We usually just use MakeGraphQLRequestOK?

router/pkg/authentication/authentication.go
```diff
@@ -42,6 +46,7 @@ type Authentication interface {
 type authentication struct {
 	authenticator string
 	claims        Claims
+	scopeClaim    string
```
endigma, line 47: Should we make it clear that this only supports top level claims, not JSONPath, etc? Or should it support path access?

router/pkg/authentication/authentication.go
```diff
@@ -113,5 +126,7 @@ func Authenticate(ctx context.Context, authenticators []Authenticator, p Provide
 }
 
 func NewEmptyAuthentication() Authentication {
-	return &authentication{}
+	return &authentication{
+		scopeClaim: DefaultScopeClaim,
+	}
 }
```
endigma: Is there a reason we have to set scopeClaim via the setter instead of here in the constructor? Does it change?

endigma (changes requested):

router/pkg/authentication/authentication.go
alepane21: I kept the same behaviour that we are already using on the interface (like SetScopes). I wouldn't change it, honestly.

router/pkg/authentication/authentication.go
alepane21: Specified in the schema and documentation that only top levels claims are supported. I think that supporting JSONPath would be and overkill, even the systems that decided to use scp are a minority.

router-tests/security/authentication_test.go
alepane21: just copied from the one over it, fixed sorry!

router-tests/security/authentication_test.go
alepane21: it was the same logic used around it, fixed

router/pkg/authentication/authentication.go
endigma: I would, these interface methods are meant for mocking the individual authenticator, scopeClaim is an implementation detail of this specific authenticator. I'd include it in the constructor.

SetScopes/Scopes are imperative, and _per request_.

router/pkg/config/config_test.go
```diff
@@ -1533,6 +1533,41 @@ authentication:
 		require.NoError(t, err)
 	})
 
+	t.Run("verify scope claim can be configured", func(t *testing.T) {
```
endigma: Nit: I would prefer removing "verify" here

router/pkg/authentication/authentication.go
```diff
@@ -31,6 +33,8 @@ type Authentication interface {
 	// Claims returns the claims of the authenticated request, as returned by
 	// the Authenticator.
 	Claims() Claims
+	// SetScopesClaim sets the claim key used by Scopes and SetScopes.
+	SetScopesClaim(scopeClaim string)
```
endigma: I would remove this and add it as a parameter of NewEmptyAuthentication. This is bloating the interface as seen in FakeAuthenticator, and is highly specific to _this_ authenticator implementation.

endigma (changes requested):
Just small thing

endigma (approved):

PR #2739, opened 2026-04-02
feat(manifest): support PO and PQL manifest for feature flags

StarpTech:
## Summary

Previously, persisted operations were validated only against the base federated graph schema at upload time, making feature flags and persisted operations incompatible for FF-specific queries. Operations using fields introduced by a feature flag subgraph would be rejected during upload.

This PR makes them work together end-to-end: the controlplane validates against all enabled FF schemas, tracks which schemas each operation is valid on, and dynamically filters the PQL manifest based on feature flag state. The router already handles per-mux schema validation gracefully, and now correctly re-warms all feature flag muxes when the manifest is updated.

### Controlplane

**Validation against feature flag schemas** — when publishing persisted operations, the controlplane now validates each operation against the base graph schema first, then falls back to all enabled FF composition schemas. An operation is accepted if valid on at least one. If all operations pass base validation, FF schemas are never fetched (short-circuit).

**Validity tracking** — each operation is stored with a `validOnBaseGraph` boolean and links in a new `persistedOperationToFeatureFlags` junction table recording which FFs validated it.

**Manifest filtering** — the PQL manifest now only includes servable operations: those valid on the base graph, or linked to at least one enabled FF. This is non-destructive — operations stay in the DB and blob storage regardless of FF state.

**Manifest regeneration on FF state changes** — toggling or deleting a feature flag now regenerates the PQL manifest for all affected federated graphs. FF-only operations appear/disappear automatically based on FF state.

**Error messages** — when an operation is rejected, the error lists which schemas were checked (base graph + FF names) and shows the base graph's validation error.

### Router

**Per-mux manifest re-warm** — previously, `SetOnUpdate` only supported a single callback, so only the last-built mux (random feature flag) got its plan cache re-warmed on manifest update. Now each mux registers its own warmup function, and the graph server composes them into a single `SetOnUpdate` callback that runs all warmups sequentially. This provides natural backpressure — at most one warmup runs at a time.

**Per-mux schema validation** — the manifest is shared across all muxes. Operations valid on a FF schema but invalid on the base schema are resolved from the manifest on both muxes, but per-mux schema validation rejects them on the base mux. Warmup skips invalid operations with a warning.

**Config reload** — on config reload, the new graph server registers a new composed callback via `SetOnUpdate`, atomically replacing the old one. No listener lists or cleanup needed.

### Database

New column on `federated_graph_persisted_operations`:
- `valid_on_base_graph` (`boolean NOT NULL DEFAULT true`) — backward compatible, existing operations default to base-valid

New junction table `persisted_operation_to_feature_flags`:
- `persisted_operation_id` → FK cascade to operations
- `feature_flag_id` → FK cascade to feature flags
- All cascade deletes are handled: operation deleted, FF deleted, graph deleted, org deleted

## Behavior Matrix

| Scenario | Manifest includes operation? |
|----------|-----|
| Operation valid on base graph | Always |
| Operation valid on enabled FF only | Yes |
| Operation valid on disabled FF only | No |
| FF disabled | FF-only ops removed from manifest |
| FF re-enabled | FF-only ops restored in manifest |
| FF deleted | FF-only ops removed, junction rows cascade-deleted |
| Base-valid ops when FF changes | Unaffected |

## Checklist

## Open Source AI Manifesto

## Summary by CodeRabbit

* **New Features**
  * Persisted operations are validated against the base schema and enabled feature-flag schemas; an operation is accepted if valid on any schema.
  * Manifests and operation listings now track whether operations are base-valid or tied to feature flags; manifests regenerate when flags are enabled, disabled, or deleted.
  * Router maintains separate plan caches per feature flag to serve flag-specific operations.

* **Documentation**
  * Updated docs and CLI guidance describing validation, manifest, and flag behavior.


router/internal/persistedoperation/pqlmanifest/store.go
```diff
@@ -18,56 +17,53 @@ type Manifest struct {
...
 	default:
 		s.logger.Debug("Skipping PQL manifest update signal, worker is busy")
 	}
 }
 
-// Close stops the update worker goroutine.
+// Close stops the worker goroutine.
 func (s *Store) Close() {
 	close(s.updateCh)
 }
```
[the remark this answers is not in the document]
StarpTech: We already ensure this in Shutdown. it is then only place where we call it

controlplane/src/core/repositories/FederatedGraphRepository.ts
```diff
@@ -955,6 +956,37 @@ export class FederatedGraphRepository {
...
+      .where(
+        and(
+          eq(federatedGraphsToFeatureFlagSchemaVersions.baseCompositionSchemaVersionId, data.baseSchemaVersionId),
+          eq(schemaVersion.organizationId, this.organizationId),
+          eq(graphCompositions.isComposable, true),
+          eq(graphCompositions.isFeatureFlagComposition, true),
+          eq(featureFlags.isEnabled, true),
+        ),
+      )
+      .execute();
```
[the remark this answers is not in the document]
StarpTech: he scenario of multiple rows per (baseSchemaVersionId, featureFlagId) cannot occur in practice:                                         
                                                                                                                                           
  1. Every composeAndDeployGraphs call (the only entry point) always runs the base composition first, which creates a new schema version   
  and updates federatedGraphs.composedSchemaVersionId — even on failure (line 842-858: "Always update the federated schema after composing,
   even if the schema is not composable").                                                                                                 
  2. FF compositions run after, using the just-updated fedGraph.composedSchemaVersionId as baseCompositionSchemaVersionId. This value is
  unique per composeAndDeployGraphs call.                                                                                                  
  3. If the base composition fails, continue parentLoop at line 1688-1689 skips all remaining FF compositions for that graph.
  4. The entire flow runs in a single transaction (line 1556), so concurrent calls cannot interleave base and FF compositions.             
                                                                                                                                           
  Since each composeAndDeployGraphs invocation produces a unique baseCompositionSchemaVersionId, there is at most one fgffsv row per       
  featureFlagId for any given baseSchemaVersionId. The current query already correctly returns one row per feature flag, and adding a MAX()
   subquery would add complexity without fixing a real bug

controlplane/src/core/bufservices/feature-flag/deleteFeatureFlag.ts
```diff
@@ -157,6 +158,21 @@ export function deleteFeatureFlag(
       compositionWarnings.push(...composition.compositionWarnings);
     });
 
+    // Regenerate PQL manifests so that operations only valid on the deleted FF
+    // are excluded. Junction entries are cascade-deleted by the FK.
+    for (const graph of federatedGraphs) {
```
JivusAyrus, line 163: shouldnt we also validate if the current ops are valid?

JivusAyrus (changes requested):

controlplane/src/core/bufservices/feature-flag/deleteFeatureFlag.ts
StarpTech: Why is it relevant when deleting a flag?

JivusAyrus: if a op was on that feature flag, in that case that operations should be removed right?

JivusAyrus (approved):

Noroth (approved):

PR #2749, opened 2026-04-08
feat(router): add TLS support for NATS event source

### LINKED ISSUE #2582: feat(router): add TLS support for NATS event source (skip verify, custom CA, mTLS)
### Component(s)

router

The NATS event source (`events.providers.nats`) has no TLS support. The `NatsEventSource` config struct only has `id`, `url`, and `authentication` fields — there is no `tls` block. This means all NATS connections are plaintext, making it impossible to use NATS in environments that require encrypted transport (e.g., compliance requirements, production infrastructure with TLS-only brokers).

Add a `tls` block to the NATS event source config supporting all standard TLS modes:

| Mode | Config |
|---|---|
| No TLS | omit `tls` block |
| TLS + skip verify (self-signed/dev) | `insecure_skip_verify: true` |
| TLS + custom CA (1-way TLS) | `ca_file: /path/to/ca.crt` |
| mTLS (mutual TLS) | `ca_file` + `cert_file` + `key_file` |

Example config:
```yaml
events:
  providers:
    nats:
      - id: my-nats
        url: nats://localhost:4222
        tls:
          ca_file: /path/to/ca.crt
          cert_file: /path/to/client.crt
          key_file: /path/to/client.key
```

Implementation requires:
1. New `NatsTLSConfiguration` struct in `config.go`
2. `TLS *NatsTLSConfiguration` field on `NatsEventSource`
3. TLS option building in `buildNatsOptions` using `nats.Secure(tlsCfg)`
4. Update `config.schema.json` and `testdata/config_full.json` golden fixture

Adding the NATS server's CA to the host OS system cert pool. This is operationally impractical in containerized/ephemeral environments.

### Additional context

The Kafka event source already has a (minimal) `tls.enabled` flag. This proposal brings NATS to full parity and beyond, covering all 4 standard TLS modes.

vatsalpatel:
Add a `tls` block to `NatsEventSource` supporting four modes: no TLS (omit block), skip verify, custom CA (1-way TLS), and mTLS.

 - New `NatsTLSConfiguration` struct with `insecure_skip_verify`, `ca_file`, `cert_file`, and `key_file` fields
 - TLS option building in `buildNatsOptions` via `nats.Secure(tlsCfg)`; cert/key pair validated together, missing CA file surfaced at startup
 - `config.schema.json` updated with `tls` object (`additionalProperties: false`)
 - Golden fixture updated
 - Tests for all modes and error paths in both `config_events_nats_test.go` and `provider_builder_test.go`

Closes #2582

## Summary by CodeRabbit

* **New Features**
  * Added optional TLS support for NATS event providers: skip-server-cert verification, custom CA, and mutual TLS (client cert/key). Schema and sample config updated.

* **Tests**
  * Added unit tests for TLS success and failure cases, covering validation for missing/mismatched fields and unknown attributes.

* **Documentation**
  * Updated NATS provider docs with TLS reference, usage examples for modes, and a warning about disabling certificate verification.

## Checklist


vatsalpatel:
Docs PR: https://github.com/wundergraph/cosmo-docs/pull/261

alepane21:
> Docs PR: [wundergraph/cosmo-docs#261](https://github.com/wundergraph/cosmo-docs/pull/261)

Hi @vatsalpatel,
thank you for your contribution!
We recently moved documentation to this repo, inside the `docs-website` folder, so the external PR is not needed anymore, you can just add the docs changes inside this PR.

vatsalpatel:
Updated docs-website @alepane21 
Should I close the cosmo-docs PR?

docs-website/router/configuration.mdx
```diff
@@ -1746,6 +1750,60 @@ events:
...
+
+The NATS provider supports four TLS modes:
+
+| Mode | Config |
+|---|---|
+| No TLS | Omit the `tls` block |
+| TLS with skip verify | `insecure_skip_verify: true` |
+| TLS with custom CA | `ca_file: /path/to/ca.crt` |
+| Mutual TLS (mTLS) | `ca_file` + `cert_file` + `key_file` |
+
```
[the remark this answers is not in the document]
vatsalpatel: Fixed

alepane21:
> Updated docs-website @alepane21 Should I close the cosmo-docs PR?

Don't worry, I did it! Thanks again.

dkorittki:
Hey @vatsalpatel, thanks for your contribution! Just wanted to let you know I am looking into your pull request and will provide feedback very soon.

router/pkg/config/config.schema.json
```diff
@@ -2550,6 +2550,33 @@
...
+                      "cert_file": {
+                        "type": "string",
+                        "description": "Path to the client certificate file (PEM) for mTLS.",
+                        "format": "file-path"
+                      },
+                      "key_file": {
+                        "type": "string",
+                        "description": "Path to the client private key file (PEM) for mTLS.",
+                        "format": "file-path"
+                      }
```
dkorittki, line 2821: Can you add a conditional json schema validation rule that requires if either `cert_file` or `key_file` is set, that the other one is set as well? This matches what you are doing in provider_builder.go but allows the router to fail early during config validation. Also can you provide a unit test for this?

router/pkg/config/config.schema.json
```diff
@@ -2550,6 +2550,33 @@
...
+                        "description": "Skip server certificate verification. Not recommended for production use."
+                      },
+                      "ca_file": {
+                        "type": "string",
+                        "description": "Path to a custom CA certificate file (PEM) used to verify the server certificate.",
+                        "format": "file-path"
+                      },
+                      "cert_file": {
+                        "type": "string",
+                        "description": "Path to the client certificate file (PEM) for mTLS.",
```
dkorittki: Can you mention the dependency between `cert_file` and `key_file` in both fields `description` field? i.e.
```suggestion
                        "description": "Path to the client certificate file (PEM) for mTLS. If set 'key_file' has to be set as well.",
```

router/pkg/pubsub/nats/provider_builder.go
```diff
@@ -114,6 +117,44 @@ func buildNatsOptions(eventSource config.NatsEventSource, logger *zap.Logger) ([
...
+			tlsCfg.RootCAs = pool
+		}
+
+		if eventSource.TLS.CertFile != "" || eventSource.TLS.KeyFile != "" {
+			if eventSource.TLS.CertFile == "" || eventSource.TLS.KeyFile == "" {
+				return nil, fmt.Errorf("both cert_file and key_file must be provided for mTLS on NATS provider %q", eventSource.ID)
+			}
+			if eventSource.TLS.CaFile == "" {
+				return nil, fmt.Errorf("ca_file is required when mTLS credentials are configured for NATS provider %q", eventSource.ID)
+			}
```
dkorittki: Is there a specific reason you require a custom CA certificate in mtls scenarios? Technically it works without one. In that case the router would use the hosts default CA pool. Maybe a user signed the nats servers certificates using a well-known CA or maintains his own PKI and has CA certs installed in the host CA pool.

router/pkg/pubsub/nats/provider_builder_test.go
```diff
@@ -67,6 +76,114 @@ func TestBuildNatsOptions(t *testing.T) {
...
+			URL: "nats://localhost:4222",
+			TLS: &config.NatsTLSConfiguration{
+				InsecureSkipVerify: true,
+			},
+		}
+		logger := zaptest.NewLogger(t)
+
+		opts, err := buildNatsOptions(cfg, logger)
+		require.NoError(t, err)
+		require.Greater(t, len(opts), 7) // base options + TLS option
```
dkorittki: I think this assertion is not strong enough. It would be better to actually verify that opts create the nats configuration we expect. For example like this

```go
opts, err := buildNatsOptions(cfg, logger)
require.NoError(t, err)

natsOpts := nats.GetDefaultOptions()
for _, opt := range opts {
	require.NoError(t, opt(&natsOpts)) // apply opts on top of natsOpt
}

require.True(t, natsOpts.Secure)
require.NotNil(t, natsOpts.TLSConfig)
require.True(t, natsOpts.TLSConfig.InsecureSkipVerify)
```

The same goes for tests `mtls with ca file succeeds` and `ca file only succeeds`.

router/pkg/pubsub/nats/provider_builder_test.go
```diff
@@ -67,6 +76,114 @@ func TestBuildNatsOptions(t *testing.T) {
 	})
 }
 
+func TestBuildNatsOptionsWithTLS(t *testing.T) {
+	t.Run("insecure skip verify", func(t *testing.T) {
```
dkorittki:
```suggestion
	t.Run("insecure skip verify is allowed", func(t *testing.T) {
```

dkorittki (changes requested):
Thanks again for contribution, really appreciate it! The implementation looks good and works as expected. I left some comments on areas which need some touches.

In general what it lacks are some integration tests, which verify tls works against a real nats server instance. The right place to do this would be the `router-tests` module but I think it's asked too much for, since it would become rather complex. I will brainstorm a few ideas on how to solve this but meantime I want to provide some feedback to you already

vatsalpatel:
Hey @dkorittki 
Thank you for the review!
I have made all of the requested changes
- Making CA cert optional for mTLS
- More explicit assertations for nats options that are generated
- Improved documentation listing explicit dependencies of TLS files

For the integration test, I did some research and there are 2 options we can go for
1. We can use https://pkg.go.dev/github.com/nats-io/nats-server/v2/test which provides an easy to setup a test server
```go
  opts := &server.Options{
      Host:      "127.0.0.1",
      Port:      -1,
      NoLog:     true,
  }
  s := natstest.RunServer(opts)
  t.Cleanup(s.Shutdown)
  s.ReadyForConnections(5 * time.Second)
```
2. We can use `testcontainers-go`, but this will require docker to be running.

I think option 1 seems better, however we will have to add `nats-server` pkg to `go.mod`. If that is fine with you, I will update the integration tests. 

Also FYI, looks like the `router CI` does setup a nats container and run integration tests against it.
```yaml
    services:
      nats:
        image: ghcr.io/wundergraph/cosmo/nats:2.11.0-alpine
```

dkorittki:
We have the [router-tests](https://github.com/wundergraph/cosmo/tree/main/router-tests) module, which contains all of our integration tests for the router. It configures the router to connect to a real nats instance but there is no easy way of enabling TLS for specific tests, to test TLS connectivity. The best option I see is going forward without such a test for now and I'll add one very soon in a seperate pull request. I verified that your changes work and not having a test for it temporarily is no big risk from my point of view. I'll review the latest changes from you and once approved I'll merge it as is

dkorittki (approved):

router/pkg/config/config.go
```diff
@@ -683,11 +683,19 @@ type NatsAuthentication struct {
 	NatsTokenBasedAuthentication `yaml:"token,inline"`
 }
 
+type NatsTLSConfiguration struct {
+	InsecureSkipVerify bool   `yaml:"insecure_skip_verify,omitempty" envDefault:"false"`
```
SkArchon: I am not 100% sure but I think envDefault does not work when `NatsTLSConfiguration` is used as a pointer type struct.

router/pkg/config/config.go
```diff
@@ -683,11 +683,19 @@ type NatsAuthentication struct {
 	NatsTokenBasedAuthentication `yaml:"token,inline"`
 }
 
+type NatsTLSConfiguration struct {
```
SkArchon, line 687: This is a bit different from the KafkaTLSConfig which has an explicit `Enabled` flag, I would suggest that we align them together.

router/pkg/pubsub/nats/provider_builder.go
```diff
@@ -114,6 +117,41 @@ func buildNatsOptions(eventSource config.NatsEventSource, logger *zap.Logger) ([
 		}
 	}
 
+	if eventSource.TLS != nil {
```
SkArchon, line 120: In the schema config we don't seem to have a required field, what happens if a user does
```
  tls: {}
```

SkArchon (changes requested):

dkorittki:
Good points from @SkArchon . @vatsalpatel can you address them as well?

router/pkg/config/config.go
vatsalpatel: Added `Enabled` flag

router/pkg/config/config.go
vatsalpatel: Yeah makes sense. Removed

router/pkg/pubsub/nats/provider_builder.go
vatsalpatel: I think the idea was that this should use the OS trust store certificates. Adding explicit `Enabled` means this check is now updated to require `TLS` to be non nil AND `Enabled` to be true

vatsalpatel:
Thank you for the review @SkArchon 
I have added the `Enabled` flag and made the other requested changes as well. I have also updated the test cases and the docs. 
Please let me know if any other changes are required!

CC: @dkorittki

router/pkg/config/config.go
dkorittki: @SkArchon introducing an `enabled` flag for consistency seems like a good idea but there are things to consider. Implementation wise we could omit creating a `tls.Config` object for the nats client used by the router, if `enabled` is `false`. There are some gotchas, though.

A NATS server can be configured to require TLS. During connection negotation it lets the client know that. The client then upgrades the connection to TLS regardless of what TLS config we passed to it. So even if you set `nats.tls.enabled: false` it would still upgrade to TLS in that case, which is a bit confusing. I believe we can't prevent that because it happens inside the nats library. This is different to Kafka, there is no protocol level negotation. So we can't prevent a drift in behavior between both `enabled` implementations.

Another thing to consider is you can tell the nats client to use TLS by setting the url scheme to `tls://` . Consider this contradictory config:

```yaml
events:
  providers:
    nats:
      - id: default
        url: "tls://localhost:4222"
        tls:
          enabled: false
```

That's a scenario not possible with Kafka either. How should the router behave in that case @SkArchon ? My opinion is the router should fail early, latest when creating the `tls.Config` object for the nats client. The router should log an error that the config is contradictory.

dkorittki: @vatsalpatel We will have an internal discussion on what we think is the right way to solve this. I'll keep you updated

SkArchon: Hi @vatsalpatel could you re move the enabled flag, we decided to not go with it

router/pkg/pubsub/nats/provider_builder.go
SkArchon: Since we did not want to go with the enabled flag, you will probably need to validate this, maybe adding in the config schema at least one of them need to be specified.

router/pkg/config/config.go
dkorittki: Just to add to that: We decided to not go with the `enabled` flag because it would make things more confusing than helpful.

Setting `enabled: false` might still result in tls connections, if the server requires TLS. Clients are expected to upgrade the connection regardless of user settings. This is by spec (see `tls_required` on https://docs.nats.io/reference/reference-protocols/nats-protocol#syntax).

Setting `enabled: true` is ambigious to setting `url: tls://your-nats-server`. 

The only benefit would have been an identical config layout to Kafka settings but Kafka and NATS work fundamentally different in this regard (connection upgrade vs fixed ports), so keeping the config design in sync just does not make sense.

Hence we would like the `enabled` flag to be absent. You can revert your last commit basically. Sorry for the inconvinience!

dkorittki:
Hey @vatsalpatel , thanks for the work you put in here! Just two more things. Can you remove the `enabled` flag? See https://github.com/wundergraph/cosmo/pull/2749#discussion_r3122882935. Sorry for the inconvinience. Also can you rename the config parameter `insecure_skip_verify` to `insecure_skip_ca_verification`. This keeps it consistent to other tls sections we have in the router config.

router/pkg/pubsub/nats/provider_builder.go
vatsalpatel: Verified, its enabling tls with system trust store. Added a test case for this

router/pkg/config/config.go
vatsalpatel: Pushed the changes reverting the `enabled` flag

vatsalpatel:
Hey @dkorittki and @SkArchon 
I have removed the `enabled` flag and also renamed the `insecure_skip_ca_verification` field. Please review the changes again

dkorittki:
Looking good from my side but still need a rereview from @SkArchon

SkArchon (approved):
Approved, Dominik will add a schema validation

router/pkg/config/config.go
dkorittki: thanks

dkorittki:
Merged. It will be part of the next router release. Thanks again for the contribution @vatsalpatel
I linked the pull request containing integration tests, which will be ready for review soon

vatsalpatel:
Thank you @dkorittki and @SkArchon for the review and the feedback!
Looking forward to the integration PR as well. Let me know if I can help in any way

PR #2753, opened 2026-04-09
feat(router): add security complexity mode

AdrienPoupa:
This adds a new option, `mode`, to the complexity limits. Similarly to the `mode` option recently introduced in Cost Control, this lets the user choose between measurement or enforcement for complexity limits.

This is useful to fine tune the limits, and will allow us to:
1. Enable the limits in measure mode to record the highest values in our production environments
2. Adjust the values, then enforce the limits

I have tested my changes locally and validated that requests that would have been blocked in enforce mode are allowed in measure mode.

Disclaimer: I used AI (Claude Opus 4.6) to generate the majority of this code. I have followed the HumanOSS guidelines.

## Summary by CodeRabbit

* **New Features**
  * Added SECURITY_COMPLEXITY_MODE (`measure` | `enforce`, default `enforce`) to control whether complexity is measured only or enforced; missing mode defaults to `enforce`.

* **Documentation**
  * Updated configuration docs, examples, schema, and env-var mapping to include the new mode and its behavior.

* **Tests**
  * Added tests for measure vs. enforce behavior, telemetry attributes, cache hit/miss behavior, config validation, and regression coverage.

## Checklist

## Open Source AI Manifesto


router/pkg/config/config.go
```diff
@@ -1390,5 +1403,10 @@ func LoadConfig(configFilePaths []string) (*LoadResult, error) {
 		cfg.Config.SubgraphErrorPropagation.AllowedExtensionFields = unique.SliceElements(append(cfg.Config.SubgraphErrorPropagation.AllowedExtensionFields, "code", "stacktrace"))
 	}
 
+	// Default complexity limits mode to "enforce" for backward compatibility
+	if cfg.Config.SecurityConfiguration.ComplexityLimits != nil && cfg.Config.SecurityConfiguration.ComplexityLimits.Mode == "" {
+		cfg.Config.SecurityConfiguration.ComplexityLimits.Mode = ComplexityLimitsModeEnforce
+	}
```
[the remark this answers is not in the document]
AdrienPoupa: I'm not sure that's a real issue. The JSON schema validation before the YAML is unmarshalled into the config struct. The schema at line 2941 defines `"enum": ["measure", "enforce"]` for the mode field. Any invalid value is rejected with "value must be one of". LoadConfig returns an error and the router won't start. The post-processing code at line 1407 only runs if validation passed, so it only ever sees "", "measure", or "enforce".

The same pattern applies to CostControl.Mode; it also has no runtime validation beyond the JSON schema enum, and nobody has added one there either.

docs-website/router/configuration.mdx
```diff
@@ -2145,10 +2147,15 @@ The configuration for adding a complexity limits for queries. We currently expos
 - **Root Fields in Query**
 - **Root Field Aliases in Query**
 
-For all of the limits, if the limit is 0, or `enabled` isn't true, the limit isn't applied. All of them have the same configuration fields:
+For all of the limits, if the limit is 0, or `enabled` isn't true, the limit isn't applied. 
+The `mode` field controls whether limits are enforced or only measured. 
+When set to `measure`, complexity is calculated and reported via telemetry but requests are not rejected. 
+When set to `enforce` (the default), requests exceeding limits are rejected with an error. 
+All the limits have the same configuration fields:
```
endigma: Nit: Can this be a bit more structured? Header for fields, bullet points for the statements?

router/pkg/config/config.go
```diff
@@ -496,7 +496,20 @@ type ComplexityCalculationCache struct {
 	CacheSize int64 `yaml:"size,omitempty" envDefault:"1024" env:"SECURITY_COMPLEXITY_CACHE_SIZE"`
 }
 
+// ComplexityLimitsMode defines how complexity limits behave.
+type ComplexityLimitsMode string
+
+const (
+	ComplexityLimitsModeMeasure ComplexityLimitsMode = "measure"
+	ComplexityLimitsModeEnforce ComplexityLimitsMode = "enforce"
+)
```
endigma, line 506: It's good to have a dedicated `ComplexityLimitsModeUnset/Invalid = ""` for these pseudo-enums

router/pkg/config/config.go
```diff
@@ -1390,5 +1403,10 @@ func LoadConfig(configFilePaths []string) (*LoadResult, error) {
 		cfg.Config.SubgraphErrorPropagation.AllowedExtensionFields = unique.SliceElements(append(cfg.Config.SubgraphErrorPropagation.AllowedExtensionFields, "code", "stacktrace"))
 	}
 
+	// Default complexity limits mode to "enforce" for backward compatibility
+	if cfg.Config.SecurityConfiguration.ComplexityLimits != nil && cfg.Config.SecurityConfiguration.ComplexityLimits.Mode == "" {
+		cfg.Config.SecurityConfiguration.ComplexityLimits.Mode = ComplexityLimitsModeEnforce
```
endigma: is `envDefault` not sufficient?

endigma (changes requested):
Few things, otherwise looks good. I would expect that config_defaults.json gets updated as well, but I don't see that in the diff?

router/pkg/config/config.go
endigma: This code path is also reported as not covered by CodeCov, likely because the default is set earlier

docs-website/router/configuration.mdx
AdrienPoupa: I refactored this, let me know if you had something different in mind.

router/pkg/config/config.go
AdrienPoupa: Done. I had taken inspiration from cost control, which does not have an unset entry, we might want to address it there as well:
```go
type CostControlMode string

const (
	CostControlModeMeasure CostControlMode = "measure"
	CostControlModeEnforce CostControlMode = "enforce"
)
```

router/pkg/config/config.go
AdrienPoupa: That's what I thought initially. I then discovered that ComplexityLimits is a *ComplexityLimits pointer field on SecurityConfiguration without an envPrefix tag. Because of this, `env.Parse` cannot traverse into the nil pointer to resolve the envDefault:"enforce" tag; it only works when the parent field has an envPrefix that lets the library discover and instantiate the struct. Adding the `envPrefix` was a more invasive, but let me know if you'd rather do this instead.

AdrienPoupa:
> Few things, otherwise looks good. I would expect that config_defaults.json gets updated as well, but I don't see that in the diff?

I applied the same pattern as `CostControl`:
```json
    "ComplexityLimits": null,
    "CostControl": null,
```
That makes it default to `enforce`. I can revisit that as needed, but I thought it made sense to keep the changes minimal.

router/pkg/config/config.go
endigma: I would like the envPrefix tag either way actually, I've been working to add it whenever we touch anything in config. That will probably also fix the config_defaults issue for both cost control and this new mode.

docs-website/router/configuration.mdx
endigma: looks fine to me

router/pkg/config/config_test.go
```diff
@@ -1688,6 +1688,98 @@ access_logs:
...
+  complexity_limits:
+    mode: "${SECURITY_COMPLEXITY_MODE}"
+    depth:
+      enabled: true
+      limit: 5
+`)
+		cfg, err := LoadConfig([]string{f})
+		require.NoError(t, err)
+		require.Equal(t, ComplexityLimitsModeMeasure, cfg.Config.SecurityConfiguration.ComplexityLimits.Mode)
+	})
```
endigma: Should this use the actual envar instead of an interpolation? We already know interpolation works.

endigma (changes requested):

endigma:
Also, would you mind doing merge to update instead of re-squashing down to one commit? We squash at the end anyway so the commit history doesn't matter, and it bugs the GitHub review UI out a bit.

router/pkg/config/config_test.go
AdrienPoupa: Done

AdrienPoupa:
It turned out the extra check in `config.go` was redundant. I investigated adding support for `envDefault` but I don't feel confident making those changes here; I'd rather remove support for the environment variable if it causes too many issues and discrepancies with the existing configuration.

router/pkg/config/config_test.go
```diff
@@ -1688,6 +1688,100 @@ access_logs:
...
+  complexity_limits:
+    depth:
+      enabled: true
+      limit: 5
+`)
+		cfg, err := LoadConfig([]string{f})
+		require.NoError(t, err)
+		require.Equal(t, ComplexityLimitsModeUnset, cfg.Config.SecurityConfiguration.ComplexityLimits.Mode)
+	})
+}
```
endigma: This is not testing anything valuable

endigma (changes requested):
I'm fine with no environment variable here, if it's running into a pre-existing issue.

Please remove it entirely including docs and the now weird test

router/pkg/config/config_test.go
AdrienPoupa: removed

router/pkg/config/config_test.go
```diff
@@ -1688,6 +1688,84 @@ access_logs:
...
+    depth:
+      enabled: true
+      limit: 5
+`)
+		_, err := LoadConfig([]string{f})
+		require.ErrorContains(t, err, "at '/security/complexity_limits/mode'")
+		require.ErrorContains(t, err, "value must be one of")
+	})
+
+	t.Run("default mode is unset and treated as enforce at runtime", func(t *testing.T) {
```
endigma: This name describes behaviour it does not assert

endigma (changes requested):
One thing, looking good

router/pkg/config/config_test.go
AdrienPoupa: renamed the test

endigma (approved):

PR #2761, opened 2026-04-14
feat(protographic): add support for interface objects

Noroth:
This PR adds support in protographic to generate rpc and message types for interface object lookups. 

## Summary by CodeRabbit

* **New Features**
  * Interfaces with `@key` are treated as federated entities: an interface-level lookup is generated alongside lookup operations for each implementing type, including support for multiple and composite `@key` directives.

* **Tests**
  * Added extensive tests covering interface entity lookups, differing/multiple/composite keys, and multiple implementing types with snapshot validations.

* **Documentation**
  * Documented interface-object lookup behavior, interface wrapper response semantics, and that `@requires` is not supported for interfaces.

## Checklist

## Open Source AI Manifesto


protographic/SDL_PROTO_RULES.md
```diff
@@ -723,7 +798,7 @@ message ResolveUserCommentsResult {
 
 ### Interfaces
 
-GraphQL interfaces are mapped to Protocol Buffer messages with a `oneof` field containing all implementing types:
+GraphQL interfaces are mapped to Protocol Buffer messages with a `oneof` field containing all implementing types. Interfaces can also define `@key` directives to become interface objects — see [Interface Object Lookups](#interface-object-lookups) for details.
```
dkorittki, line 801: Maybe better use "Entity Interfaces" wording here instead of "Interface Object"? Matches to what Apollo uses in their docs https://www.apollographql.com/docs/graphos/schema-design/federated-schemas/entities/interfaces

protographic/src/sdl-to-mapping-visitor.ts
```diff
@@ -123,21 +125,25 @@ export class GraphQLToMappingVisitor {
...
+      // Process each @key directive separately
+      for (const keyDirective of keyDirectives) {
+        const key = this.getKeyFromDirective(keyDirective);
+        if (key) {
+          // Create entity mapping for each key combination
+          this.createEntityMapping(typeName, key);
         }
+      }
 
+      if (isObjectType(type)) {
```
dkorittki, line 146: For consistency reasons maybe move all object conditions, which need to match in order to create `require` Lookups, to here. Because right now the "this is an object" condition has to match here while other conditions such as "has required directive" and "has fields" are matched inside `this.createRequiredFieldsMapping`, making the conditional filtering happen at two places.

protographic/src/sdl-to-proto-visitor.ts
```diff
@@ -602,7 +602,7 @@ export class GraphQLToProtoTextVisitor {
       }
 
       // Skip non-object types
```
dkorittki:
```suggestion
      // Skip non-entity types
```

Sounds more precise to me given that directly under this if condition you filter for types not having `@key` directives, making this a filter for everything not an entity (normal and interface entities)

dkorittki (approved):
lgtm with some suggestions

protographic/src/sdl-to-mapping-visitor.ts
Noroth: I'll leave it as it is

PR #2781, opened 2026-04-21
feat(router): support costs on arguments of directives

ysmolski:
This PR supports cost weights on the arguments of the directives. When a customer applies such a directive with the specific non-null argument to a field, then its cost will be added to the cost of the field.

Composition now tracks weights for arguments of directives for each field. This is a static data so it goes into the config of a schema. It is needed for the engine to understand where directives' arguments with costs were used. 
Router/engine uses it to take those weights into costs accounting.

If such directive was applied to the implementing type of some interface, and if the interface type is being used in the operation, then all the implementing types are taken into an account and maximum weight is picked among them.

I have added just a simple e2e test to verify that feature is working. More extended tests are implemented in the corresponding [engine PR](https://github.com/wundergraph/graphql-go-tools/pull/1465).

Example how cost from the directive could be added to the `employees' field of the `Engineer` type:
```graphql
directive @expensiveOp(applied: Boolean = true @cost(weight: 22)) on FIELD_DEFINITION

type Engineer implements RoleType {
  departments: [Department!]!
  title: [String!]!
  employees: [Employee!]! @goField(forceResolver: true) @expensiveOp
  engineerType: EngineerType!
}
```

See [documentation preview](https://wundergraphinc-yury-eng-8700-router-support-cost-on-the-arg.mintlify.app/router/security/cost-control#weights-on-directive-arguments) for more.

## Summary by CodeRabbit

* **New Features**
  * Directive-argument weights now contribute to field cost calculations and are propagated through configuration and serialization so costs reflect directive argument values.

* **Documentation**
  * Added guidance, examples, and activation rules for directive-argument weights; clarified behavior for directives applied to interface-implementing types.

* **Tests**
  * Added/updated tests for directive-argument weighting, tightened assertions to ensure no top-level errors, and adjusted expected cost results.

* **Demo**
  * Added an example directive and applied it to sample fields to demonstrate directive-argument weighting.


router-tests/security/costs_test.go
```diff
@@ -343,6 +343,33 @@ func TestOperationCost(t *testing.T) {
...
+						ExposeHeaders:     true,
+					}
+				},
+			}, func(t *testing.T, xEnv *testenv.Environment) {
+				res := xEnv.MakeGraphQLRequestOK(testenv.GraphQLRequest{
+					Query: `{ employee(id:1) { role { employees { id } } } }`,
+				})
+				require.Contains(t, res.Body, `"data":`)
+
+				// "@expensiveOp(applied: Boolean = true @cost(weight: 22))" is applied on
```
SkArchon: Nit: Looks good to me, however maybe we could link this to (as a comment) to schema.graphqls in employees.

SkArchon (approved):
Looked at router

router-tests/security/costs_test.go
ysmolski: tests are done yet. only composition is ready for review

ysmolski: To address the original comment, there is a comment in the employees' schema.

router-tests/security/costs_test.go
```diff
@@ -76,6 +76,7 @@ func TestOperationCost(t *testing.T) {
 					Query: `{ employee(id:1) { id details { forename surname } } }`,
 				})
 				require.Contains(t, res.Body, `"data":`)
+				require.NotContains(t, res.Body, `"error":`)
```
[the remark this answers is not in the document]
ysmolski: why it won't catch?

composition/src/v1/normalization/normalization-factory.ts
Aenimus, line 736: I think here you can add `if (!isField) { return errorMessages; }` 

Then you don't have to recheck the condition constantly. WDYT?

Aenimus: There might be something else you have to do before you return it's hard to tell which curly braces are where. But you can definitely short-circuit here.

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -2522,36 +2535,61 @@ export class NormalizationFactory {
...
+    const parentTypeData = this.parentDefinitionDataByTypeName.get(typeName);
+    // Directive argument weights should only be on concrete type fields.
+    if (!parentTypeData || parentTypeData.kind === Kind.INTERFACE_TYPE_DEFINITION) {
+      return;
+    }
+
+    // Determine which arguments are active (non-null) on this directive usage.
+    // An argument is active if it has an explicit non-null value or
+    // if it has a default value and was not explicitly set to null.
+    const explicitArgs = new Map<string, ConstValueNode>();
```
Aenimus: By active do we mean required? Or?

Can we rename to `requiredArgNodeByName` or something to that effect?

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -2522,36 +2535,61 @@ export class NormalizationFactory {
...
+    if (directiveNode.arguments) {
+      for (const arg of directiveNode.arguments) {
```
Aenimus: you can remove the wrapping if with `const arg of directiveNode.arguments ?? []`

ysmolski: thanks! done.

composition/src/v1/normalization/normalization-factory.ts
ysmolski: i have reworded that comment and renamed the var to your suggestion along with other var below.

composition/src/v1/normalization/normalization-factory.ts
ysmolski: yep, fixed

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -2522,36 +2538,59 @@ export class NormalizationFactory {
...
+    data,
+    definitionData,
+    directiveName,
+    directiveNode,
+  }: RecordDirectiveWeightOnFieldParams) {
+    const typeName = data.renamedParentTypeName || data.originalParentTypeName;
+    const parentTypeData = this.parentDefinitionDataByTypeName.get(typeName);
+    // Directive argument weights should only be recorded for concrete type fields.
+    if (!parentTypeData || parentTypeData.kind === Kind.INTERFACE_TYPE_DEFINITION) {
+      return;
```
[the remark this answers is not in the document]
ysmolski: It would not work that way. I have added calrifying comment why

composition/src/v1/normalization/normalization-factory.ts
Aenimus: Perfect thanks

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -2472,6 +2484,20 @@ export class NormalizationFactory {
     data.nullLevelsBySubgraphName.set(this.subgraphName, levels);
   }
 
+  getOrCreateFieldWeight(typeName: TypeName, fieldName: FieldName): FieldWeightConfiguration {
```
Aenimus, line 2487: I don't think this method does anything useful.

Aenimus: I suppose the intention is to reduce repetition?

ysmolski: yeah, sugar syntax to reduce the repeatable chunk of code

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -2522,36 +2539,68 @@ export class NormalizationFactory {
...
+    data,
+    definitionData,
+    directiveName,
+    directiveNode,
+  }: RecordDirectiveWeightOnFieldParams) {
+    // This method walks every argument defined on the directive and records a directive weight
+    // on the field only when all these conditions hold:
+    // 1. The argument of the directive has a cost weight assigned
+    // 2. The argument is non-null
+    // 3. The parent type is not the interface type.
```
Aenimus: nitpick: not the interface type -> not an Interface type

composition/tests/v1/directives/cost.test.ts
```diff
@@ -20,6 +22,16 @@ import {
   schemaToSortedNormalizedString,
 } from '../../utils/utils';
 
+function fieldWeight(typeName: TypeName, fieldName: FieldName, overrides: Record<string, unknown> = {}) {
```
Aenimus: Please can we add a return type to this function?

composition/src/v1/normalization/normalization-factory.ts
ysmolski: fixed

composition/tests/v1/directives/cost.test.ts
ysmolski: added

composition/src/v1/normalization/normalization-factory.ts
ysmolski: perhaps this could be a free function? I dont know what is the best practice in TS for this kind of constructor-like behavior.

composition/tests/v1/directives/cost.test.ts
Aenimus: Thanks!

Aenimus (approved):
LGTM from composition side.

demo/pkg/subgraphs/employees/subgraph/schema.graphqls
```diff
@@ -41,6 +41,9 @@ directive @goField(
 
 directive @openfed__requireFetchReasons repeatable on FIELD_DEFINITION | INTERFACE | OBJECT
 
+# To demonstrate that cost can be applied to the argument of a directive
+directive @expensiveOp(applied: Boolean = true @cost(weight: 22)) on FIELD_DEFINITION
```
Noroth: What if applied = false?

router/core/factoryresolver.go
```diff
@@ -684,7 +684,7 @@ func (l *Loader) dataSourceMetaData(in *nodev1.DataSourceConfiguration) *plan.Da
 		return out
 	}
 	for _, fw := range costConfig.GetFieldWeights() {
-		w := &plan.FieldWeight{}
+		w := &plan.FieldCost{}
```
Noroth: If possible give variables better names. 
```suggestion
		fieldCost := &plan.FieldCost{}
```

demo/pkg/subgraphs/employees/subgraph/schema.graphqls
ysmolski: Then the cost will account it. It won't be applied only if the value is `null`. I guess it was not the ideal name for the example.

ysmolski: Renaming that arg with additional comment:

```
# To demonstrate that cost can be applied to the argument of a directive.
# If coefficient is set to null, then weight won't be added.
directive @expensiveOp(coefficient: Int = 2 @cost(weight: 22)) on FIELD_DEFINITION
```

router/core/factoryresolver.go
ysmolski: renamed it along with the `fw`.

Noroth (approved):

endigma (approved):

PR #2788, opened 2026-04-22
chore(router): bump nats client lib + add router-tests to verify nats tls

dkorittki:
This adds integration tests to the PR https://github.com/wundergraph/cosmo/pull/2749 from @vatsalpatel .

I made a noteworthy design decision for these new tests: Currently nats tests use the nats instance from docker-compose or as defined in CI. In order to test TLS I would have to configure the instance to allow for TLS and worse: require mTLS. To avoid this instead I spawn an in-process nats server configured to require TLS, instead of using the one from docker-compose or CI. This allows developers to keep their plain text connection. Also other nats test focusing on other things don't need to bother with TLS. The newly added tests are the only ones who need to bother with TLS/mTLS.

In order to use a nats in-process server the router-tests go.mod needs a new dependency `github.com/nats-io/nats-server/v2` (and nats.go upgraded from v1.35.0 → v1.50.0 as a transitive requirement — this is a backwards-compatible minor version bump). To keep version parity between router and router-tests I updated the router nats dependencies, too.

## Summary by CodeRabbit

* **Tests**
  * Added an end-to-end NATS TLS test validating connectivity and event delivery across TLS, custom-CA verification, mutual TLS, tls:// URL handling, and startup-failure scenarios; includes utilities to generate and use temporary TLS certificates during tests.

* **Chores**
  * Updated NATS libraries and multiple Go ecosystem dependencies for compatibility and security.

## Checklist

## Open Source AI Manifesto


dkorittki:
fyi I force pushed to rebase the actual change onto main after the original pull request got merged to main.

router-tests/events/nats_tls_test.go
```diff
@@ -0,0 +1,376 @@
...
+			require.JSONEq(t,
+				`{"employeeUpdated":{"id":3,"details":{"forename":"Stefan","surname":"Avram"}}}`,
+				string(args.dataValue),
+			)
+		})
+
+		require.NoError(t, client.Close())
+		testenv.AwaitChannelWithT(t, EventWaitTimeout, clientErrCh, func(t *testing.T, err error) {
+			require.NoError(t, err)
+		})
```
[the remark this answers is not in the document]
dkorittki: Fixed

router-tests/events/nats_tls_test.go
```diff
@@ -0,0 +1,376 @@
...
+		natsConn := connectInsecureTLSNATSClient(t, natsPlainURL(srv))
+
+		// Router uses the tls:// URL scheme; InsecureSkipCaVerification is required because the
+		// embedded server uses a self-signed cert not in the system trust store.
+		routerTLS := &config.NatsTLSConfiguration{InsecureSkipCaVerification: true}
+
+		testenv.Run(t, &testenv.Config{
+			RouterConfigJSONTemplate: testenv.ConfigWithEdfsNatsJSONTemplate,
+			ModifyEventsConfiguration: func(cfg *config.EventsConfiguration) {
+				cfg.Providers.Nats = tlsNATSEventSources(tlsSchemeURL, routerTLS)
```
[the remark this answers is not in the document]
dkorittki: I modified the test to only verify the router connects to the server with a `tls://` url configured on the router. I don't test wether this url alone is sufficient for activating tls connection support on the router. This is because without any provided TLS config the router would use the host CA pool and would ultimately fail in these tests because it's a self signed cert. There is no way to modify the host ca pool for this test specifically without it becoming hacky.

router-tests/events/nats_tls_test.go
```diff
@@ -0,0 +1,290 @@
...
+			)
+		})
+
+		require.NoError(t, client.Close())
+		testenv.AwaitChannelWithT(t, EventWaitTimeout, clientErrCh, func(t *testing.T, err error) {
+			require.NoError(t, err)
+		})
+	}
+
+	t.Run("router connects when server requires TLS", func(t *testing.T) {
```
SkArchon, line 167: Do you think we should have a example test where the nats server upgrades the client, when tls is not expected from the client?

router-tests/events/nats_tls_test.go
```diff
@@ -0,0 +1,290 @@
+package events_test
```
SkArchon, line 1: Also are we addressing the `tls: {}` case in this PR? IIRC we wanted to block it?

router-tests/events/nats_tls_test.go
SkArchon, line 167: Maybe some failure cases when cert is invalid / mismatching?

router-tests/events/nats_tls_test.go
```diff
@@ -0,0 +1,290 @@
...
+
+		testenv.AwaitChannelWithT(t, EventWaitTimeout, resultCh, func(t *testing.T, args natsSubscriptionArgs) {
+			require.NoError(t, args.errValue)
+			require.JSONEq(t,
+				`{"employeeUpdated":{"id":3,"details":{"forename":"Stefan","surname":"Avram"}}}`,
+				string(args.dataValue),
+			)
+		})
+
+		require.NoError(t, client.Close())
```
SkArchon: We are closing twice I think with line 141?

router-tests/events/nats_tls_test.go
```diff
@@ -0,0 +1,290 @@
...
+	return strings.Replace(s.ClientURL(), "tls://", "nats://", 1)
+}
+
+// connectToNATS creates a nats connection and connects to serverURL.
+// By default, it skips server certificate verification.
+// It allows to pass extraOpts, which override or extend defaults options.
+func connectToNATS(t *testing.T, serverURL string, extraOpts ...nats.Option) *nats.Conn {
+	t.Helper()
+	opts := []nats.Option{
+		nats.Secure(&tls.Config{InsecureSkipVerify: true}), //nolint:gosec // test helper only
```
SkArchon: I dunno if this is possible, but do u think its a potential "false positive", because this is also called for the "url test"?

SkArchon (changes requested):
Changes look ok in general, few questions

router-tests/events/nats_tls_test.go
dkorittki: Configuring router nats connection with `tls: {}` means the router uses TLS to connect to NATS. It's the same as using `tls://` as the url scheme, so it's ambiguous. Not ideal but it also doesn't hurt anyone. I'll address this in a seperate pull request.

router-tests/events/nats_tls_test.go
dkorittki: `connectToNATS` is only used for the test helper publishing messages into nats. It's not used to configure anything on the nats server or the router client. `InsecureSkipVerify: true` is set here to avoid having the test helper awkwardly pass the CA cert around, which would involves passing around a custom `tls.Config`. Does this answer your question?

router-tests/events/nats_tls_test.go
dkorittki: Yes, line 141 is a safety net to ensure we definetely close the connection on test exit. The one on line 158 actually verifies we can cleanly close without errors. I modified the code so we only have one place of closing the connection (inside t.Cleanup).

router-tests/events/nats_tls_test.go
dkorittki: Done

dkorittki: Done. Added a test which verifies the router fails when configured for tls but the server does not support it.

SkArchon (approved):
lgtm

PR #2789, opened 2026-04-22
feat: improve OIDC connection behavior

wilsonrivera:
https://github.com/user-attachments/assets/72ca2235-9468-4cb9-83dd-c853b5082757

## Summary by CodeRabbit

* **New Features**
  * New OIDC provider management UI: connect, disconnect, info, mapper editing dialogs and a card showing provider details.
* **Bug Fixes**
  * Clarified OIDC provider deletion error message wording.
* **Refactor**
  * Centralized endpoint URL validation to a shared absolute-URL validator.
  * Extracted OIDC UI into a dedicated component, simplifying the settings page.

## Checklist

- [ ] Documentation has been updated on [https://github.com/wundergraph/docs-website](<https://github.com/wundergraph/docs-website>).
- [X] I have read the [Contributors Guide](<https://github.com/wundergraph/cosmo/blob/main/CONTRIBUTING.md>).

## Open Source AI Manifesto

This project follows the principles of the [Open Source AI Manifesto](<https://human-oss.dev>). Please ensure your contribution aligns with its principles.

The goal of this PR is to improve how users connect OIDC providers to their organization with better error handling. The code have also been moved to dedicated componets for a cleaner structure.

These changes will allow us further evolve how users manage their OIDC integration in the future.


studio/src/lib/zod.ts
```diff
@@ -0,0 +1,41 @@
...
+      ctx.addIssue({
+        code: z.ZodIssueCode.custom,
+        message: 'Must be a valid absolute URL starting with https://',
+      });
+      return;
+    }
+
+    try {
+      const url = new URL(val); // Ensure that the value is a valid absolute URL
+      if (url.hostname === 'localhost') {
```
comatory: Why do we detect `localhost` specifically? What if `127.0.0.1` is used? I know you're just transferring existing code, but still, I want to make sure it makes sense.

studio/src/lib/zod.ts
```diff
@@ -0,0 +1,41 @@
...
+        if (url.protocol !== 'http:' && url.protocol !== 'https:') {
```
comatory, line 19: Wouldn't it be possible to just do `if (!url.protocol)` ?

studio/src/lib/zod.ts
```diff
@@ -0,0 +1,41 @@
...
+    if (!val) {
+      ctx.addIssue({
+        code: z.ZodIssueCode.custom,
+        message: 'Must be a valid absolute URL starting with https://',
+      });
+      return;
+    }
+
+    try {
+      const url = new URL(val); // Ensure that the value is a valid absolute URL
```
comatory, line 17: I'm not sure what runtimes (browser version) we support with Studio but have you considered using [URL.parse](https://developer.mozilla.org/en-US/docs/Web/API/URL/parse_static)?

If URL can't be parsed it does not throw but returns `null`. If we need to support older browsers, let's use `new URL`.

Note: I looked at monorepo and we don't have any `browserslist`, we might want to consider adding it.

studio/src/components/oidc/oidc-card.tsx
```diff
@@ -0,0 +1,98 @@
...
+            <Badge variant="outline">Enterprise feature</Badge>
+          </CardTitle>
+          <CardDescription>
+            Connecting an OIDC provider allows users to automatically log in and be a part of this organization.{' '}
+            <Link href={docsBaseURL + '/studio/sso'} className="text-sm text-primary" target="_blank" rel="noreferrer">
+              Learn more
+            </Link>
+          </CardDescription>
+        </div>
+        {!oidc ? (
```
comatory: Can we reverse the condition? 🙏 
I think it might read better if we do `oidc ? ... : ...`

studio/src/components/oidc/update-mappers-dialog.tsx
```diff
@@ -0,0 +1,218 @@
...
+      { mappers },
+      {
+        onSuccess(data) {
+          if (data.response?.code === EnumStatusCode.OK) {
+            toast({
+              description: 'Group mappers updated successfully.',
+              duration: 4000,
+            });
+
+            refetch().finally(() => {
```
comatory, line 105: What happens if `refetch` fails?

studio/src/components/oidc/update-mappers-dialog.tsx
```diff
@@ -0,0 +1,218 @@
...
+    }
+  }, [open, currentMappers]);
+
+  const updateMappers = () => {
+    if (isPending) {
+      return;
+    }
+
+    setPending(true);
+    mutate(
```
comatory, line 95: It seems like you never set `setPending=true`. I think `mutate` takes callback `onSettled` so you could probably use `onSettled:() => setPending(false)` and avoid calling the setter in multiple places.

studio/src/components/oidc/update-mappers-dialog.tsx
```diff
@@ -0,0 +1,218 @@
...
+          <DialogDescription>Map your groups to cosmo groups.</DialogDescription>
+        </DialogHeader>
+
+        <div className="grid grid-cols-2 justify-between gap-x-2 px-1 text-sm font-bold">
+          <span>Group in Cosmo</span>
+          <span>Group in the provider</span>
+        </div>
+
+        {isLoading ? (
+          <div className="flex items-center justify-center p-1">
```
comatory, line 143: Nitpick: I noticed one thing in the video you provided. The error causes the form to push down the content. Could we instead reserve some space for the error? Like minimum height so the error that's most likely to happen won't push the content down in the dialog?

studio/src/components/oidc/oidc-card.tsx
```diff
@@ -0,0 +1,98 @@
...
+          <Button className="md:ml-auto" type="submit" variant="default" asChild>
+            <Link href={calURL} target="_blank" rel="noreferrer">
+              Contact us
+            </Link>
+          </Button>
+        ) : (
+          <div className="ml-auto flex gap-x-3">
+            <OIDCInfoDialog
```
comatory, line 44: This gets hard to read, as I open the individual component files . I assume the same solution was used previously, and you're just extracting the former huge component to more smaller components (which is good 👍).
The way I read it is “I have a card component but wait, four dialogs are rendered” 🤔 I assume that only single dialog out of these four can be rendered in single time? 

If so, could we model it so it's explicit? Because I see we pass `open` prop in some cases, but in other we let the dialog component itself decide whether it is opened or not, e.g. with complicated conditions such as `isProviderConnected && isAdmin && open`.

I think it would be much easier to reason about if we handled the open state in this component. We could model four states (`info-dialog-opened`, `update-mappers-dialog-opened`, `connect-oidc-dialog-opened`, `disconnect-oidc-dialog-opened`) which would reflect states in which dialog should get opened. This can be solved by using `useReducer`.

I see two further parts for refactoring:
1. Passing callback to dialog components which dispatches an event and stores the opened state
2. Move logic from dialogs up the tree into this component (possibly wrapping it in hooks) _or_ moving the logic into high order component above. I'd still use the reducer for clarity.

The rendering logic in this component would then have a `switch` statement and render just one dialog at a time instead of passing `open` prop. It'd make it obvious what needs to happen in order to see a dialog. I know it's a big ask and requires more refactoring, but I think it'd be worth it for maintainability.

studio/src/components/oidc/update-mappers-dialog.tsx
```diff
@@ -0,0 +1,218 @@
...
+            <CgSpinner className="size-6 animate-spin" />
+          </div>
+        ) : isError || !organizationGroups?.groups ? (
```
comatory, line 146: Could we avoid nested ternaries? It's difficult to red which state renders what. Could we instead assign the different states to variables and use `if ... else if ... ` during render?

I think it's ok to split this component further to individual parts that get rendered based on these conditions. No need to create separate file, just additional React components in the same file.

comatory (changes requested):
I'm really happy we're refactoring these huge components and I welcome better error messages. I think we could push it a bit further, WDYT?

studio/src/lib/zod.ts
wilsonrivera: Protocol can be `mailto:`, `file:`, `ws:` etc

studio/src/components/oidc/update-mappers-dialog.tsx
wilsonrivera: The only thing that happens is the user wouldn't see the updated information for the OIDC, not a fatal error

PR #2820, opened 2026-05-05
fix: feature flags not being reloaded

comatory:
We need to call identify and reload the feature flags after the tracking is initialized

@coderabbit summary

## Checklist

## Open Source AI Manifesto

## Summary by CodeRabbit

* **Chores**
  * Improved analytics identification flow so users are consistently identified and feature flags are reloaded immediately across all identification paths, yielding more reliable feature-flag behavior.
  * Analytics setup now runs only when properly configured and a secondary analytics provider is limited to production, improving telemetry consistency and avoiding accidental non-production tracking.

### DIFF
--- a/studio/src/lib/track.ts
+++ b/studio/src/lib/track.ts
@@ -18,7 +18,15 @@ const resetTracking = () => {
   posthog.reset();
 };
 
-const identify = ({
+const setupReo = (email: string) => {
+  // Identify with Reo
+  window.Reo?.identify({
+    username: email,
+    type: 'email',
+  });
+};
+
+const setupPosthog = ({
   email,
   id,
   organizationId,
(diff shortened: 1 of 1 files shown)


studio/src/lib/track.ts
```diff
@@ -48,25 +48,32 @@ const identify = ({
     // with the right data
     posthog.reset();
   } else if (distinctId === email) {
+    posthog.identify(email, {
```
alepane21: Why? Is this to be sure to add the email prop? Why do we need it as a pro if we already have it as an id?
From official doc (https://posthog.com/docs/product-analytics/identify):
> You only need to call identify once per session, and you should **avoid calling it multiple times unnecessarily**.
> If you call identify multiple times with the same data without reloading the page in between, PostHog will ignore the subsequent calls.

So this should not be an issue, but I think you can avoid call it again, unless there is something that I'm missing.

comatory: Ah, sorry that was leftover from debugging. Good catch I don't think we need it, so I removed it.

I don't think `posthog.identify` should get called twice. This `else-if` branch has `return` at the end.

alepane21: What I meant is that if we are here, in this if branch, identify has already be called previously (because get_distinct_id is already the email)

comatory: fixed

alepane21 (approved):

Aenimus (approved):
LGTM!

PR #2892, opened 2026-05-26
feat: support manifest for local development

Noroth:
## Summary by CodeRabbit

* **New Features**
  * Router can load execution config from a local manifest directory and optionally hot-reload it; supports skipping missing or ignoring specific feature flags and configurable watch intervals.

* **Chores**
  * Added tooling/CLI to generate manifest directory layouts and Make targets to run them.
  * Updated ignore rules to exclude manifest directories.

* **Documentation**
  * Added docs and examples for manifest-based config and hot-reload behavior.

* **Tests**
  * Added tests covering manifest hot-reload scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/wundergraph/cosmo/pull/2892?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

## Checklist

## Open Source AI Manifesto


router/core/router.go
```diff
@@ -1652,6 +1615,150 @@ func (r *Router) Start(ctx context.Context) error {
...
+	}
+
+	return w, nil
+}
+
+func (r *Router) buildManifestConfigWatcher(ctx context.Context, ll *zap.Logger) (watcher.WatcherFunc, error) {
+	w, err := watcher.New(watcher.Options{
+		Logger:   ll,
+		Paths:    []string{filepath.Join(r.manifestConfig.Path, "mapper.json")},
+		Interval: r.manifestConfig.WatchInterval,
```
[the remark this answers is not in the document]
Noroth: Yes that's intended. Only watch the mapper.json

router/pkg/routerconfig/routerconfig.go
```diff
@@ -10,3 +19,132 @@ func VersionPath(version int) string {
...
+}
+
+// AssembleConfig assembles the router execution config from the base config and the feature flag configs.
+// The base config is the latest.json file in the manifest directory.
+// The feature flag configs are the feature-flags/<feature-flag-name>.json files in the manifest directory.
+// The rules are the rules for skipping missing feature flags and ignored feature flags.
+func AssembleConfig(basePath string, mapper map[string]string, rules *AssembleConfigRules) (*nodev1.RouterConfig, error) {
+	if rules == nil {
+		rules = &AssembleConfigRules{}
+	}
```
dkorittki: Why not pass `rules` as object instead of pointer and avoid this problem alltogether?

router/pkg/routerconfig/routerconfig.go
```diff
@@ -1,6 +1,15 @@
 package routerconfig
```
dkorittki, line 18: Imo it's good if we godoc new packages, to explain their purpose.
Is `routerconfig` a package which should cover all ways the router can build the engine config? From it's  name I assume yes. If it's only for assembling engine configs from local files the name would be too broad.

router/core/router.go
```diff
@@ -1096,6 +1115,31 @@ func (r *Router) bootstrap(ctx context.Context) error {
 	return nil
 }
 
+func (r *Router) assembleStaticExecutionConfigFromManifest() (*nodev1.RouterConfig, error) {
```
dkorittki: What about moving all the stuff what happens in this function into the `routerconfig.AssembleConfig` method? That package knows best what to do when: path validation, mapper parsing + execution config assembly. Then the router boostrap code in the core package only needs to call something like this

```go
if routerShouldUseLocalManifest {
    executionConfig, err := routerconfig.AssembleConfig(manifestPath)
    // [...]
}
```

router/pkg/config/config.go
```diff
@@ -1041,6 +1041,15 @@ type ExecutionConfig struct {
...
 	FallbackStorage FallbackExecutionConfigStorage `yaml:"fallback_storage,omitempty" envPrefix:"EXECUTION_CONFIG_FALLBACK_STORAGE_"`
+	Manifest        ExecutionConfigManifest        `yaml:"manifest,omitempty" envPrefix:"EXECUTION_CONFIG_MANIFEST_"`
+}
+
+type ExecutionConfigManifest struct {
+	Path                    string        `yaml:"path,omitempty" env:"PATH"`
+	SkipMissingFeatureFlags bool          `yaml:"skip_missing_feature_flags" envDefault:"false" env:"SKIP_MISSING_FEATURE_FLAGS"`
+	IgnoredFeatureFlags     []string      `yaml:"ignored_feature_flags,omitempty" env:"IGNORED_FEATURE_FLAGS"`
+	Watch                   bool          `yaml:"watch,omitempty" envDefault:"false" env:"WATCH"`
+	WatchInterval           time.Duration `yaml:"watch_interval,omitempty" envDefault:"1s" env:"WATCH_INTERVAL"`
```
dkorittki, line 1052: Would be good to have some unit tests asserting the router unmarshalles the config as expected

dkorittki (approved):
LGTM with some minor improvement ideas, feel free to implement them at will.

In general I feel a bit confused about the `manifest` name for the new config type. One is called "static execution config" and the other is called "manifest execution config". To me both are static and manifests, so these do not distinguish them for me. Maybe we can elaborate on the name again or describe it somewhere central in a godoc.

Also it would be great to have a place in our docs where we explain this concept to users. Currently we only have the new parameters listed on the routers config page.

router/pkg/config/config.go
Noroth: I think we are good here with the existing fixtures. There is no special unmarshaling logic here.

router/core/router.go
Noroth: I think this is fine. Otherwise we have to define similar types in routerconfig as well that we are using in the function from the router.Config to avoid cyclic imports.

Noroth: Or we pass them all directly. I'll take a look

PR #2922, opened 2026-06-04
feat: check auth before sending `WhoAmI` request for telemetry

wilsonrivera:
## Summary by CodeRabbit

* **Refactor**
  * Telemetry now more resilient: CLI continues to operate and report anonymized metadata when authentication is unavailable.
  * Authentication flow improved to allow optional silent checks/refreshes so operations can proceed with reduced capabilities without surfacing errors to users.

## Checklist

## Open Source AI Manifesto

We still send two requests to the `WhoAmI` endpoint when running the `auth whoami` command due to how telemetry is setup.

Previously, telemetry would send `WhoAmI` always before the access token was refreshed and ignored the API key completely; now the access token is refreshed by telemetry (if needed), this will help us prevent getting so many `invalid token` requests.


cli/src/commands/auth/utils.ts
```diff
@@ -137,10 +137,11 @@ export const startPollingForAccessToken = async ({
 
 // checks if either of access token or api key are present
 // if not, it will try to refresh the access token
-export async function checkAuth() {
+// if `showErrorMessage` is `false`, any and all error message will be skipped
+export async function checkAuth(showErrorMessage = true): Promise<void> {
```
comatory: Could we use JSDoc annotation here? This function is imported in bunch of places and it would be nice to have it displayed in editors:

```
/**
* Checks if either of access token or api key are present.
* If not, it will try to refresh the access token
*
* @param [showErrorMessage] - when `false`, any and all error message will be skipped
*/
```

cli/src/core/telemetry.ts
```diff
@@ -129,6 +130,22 @@ const getIdentity = async (): Promise<TelemetryIdentity> => {
       };
     }
 
+    try {
+      await checkAuth(false);
+    } catch {
+      // ignore
+    }
+
+    if (!config.apiKey) {
```
comatory: Would it make sense to add helper to `commands/auth/utils.ts` called `isAuthenticated`[^1] which would check for presence of `config.apiKey`? I'd include the comments below on top of the signature of the function.

```suggestion
    if (!isAuthenticated) {
```

[^1]: I think in this case it could be either a constant or function, does not matter.

cli/src/core/telemetry.ts
```diff
@@ -129,6 +130,22 @@ const getIdentity = async (): Promise<TelemetryIdentity> => {
...
+      /**
+       * `checkAuth` perform access token refresh and updates the value for `config.apiKey` accordingly  (if needed);
+       * however, if the value for `config.apiKey` remains empty, that means the user is not authenticated or the
+       * token refresh failed, for this reason, we can assume the call to the `WhoAmI` RPC will fail and we can
+       * just exit early
+       */
+      return { organizationId: 'anonymous' };
```
comatory: This looks a bit magic to me, until I make a search for this pattern in the CLI codebase. Could we please create un-mutable object and re-use it?

```
const UNAUTHENTICATED_CONTEXT_METADATA = Object.freeze({ organizationId: 'anonymous' });
```
[^1]

Then we could replace all the occurences of this pattern and I think it would be much more clearer that we treat `anonymous` as a specific value for telemetry.

[^1]: Not sure if the pattern for constants is all upper-case letters or regular camel-case, but you get the idea.

comatory (changes requested):

cli/src/core/telemetry.ts
```diff
@@ -124,9 +126,17 @@ export const initTelemetry = () => {
 const getIdentity = async (): Promise<TelemetryIdentity> => {
   try {
     if (!apiClient) {
-      return {
-        organizationId: 'anonymous',
-      };
+      return UNAUTHENTICATED_CONTEXT_METADATA;
+    }
+
+    if (!(await isAuthenticated())) {
```
comatory, line 132: There was a `!config.apiKey` previously. I know I asked for a change here but I meant wrapping that code only in a function or variable to make it more readable.

Won't this change behaviour? Previously, the code would be synchronous as it only checked for a missing value but now it calls the authentication function basically. I just want to ensure that the condition is the same.

cli/src/core/telemetry.ts
```diff
@@ -124,9 +126,17 @@ export const initTelemetry = () => {
...
+      /**
+       * `checkAuth` perform access token refresh and updates the value for `config.apiKey` accordingly  (if needed);
+       * however, if the value for `config.apiKey` remains empty, that means the user is not authenticated or the
+       * token refresh failed, for this reason, we can assume the call to the `WhoAmI` RPC will fail and we can
+       * just exit early
+       */
```
comatory: I think we should move this JS doc comment to annotate `isAuthenticated` function instead.

comatory (changes requested):
There's just one thing I'm not sure about.

comatory (approved):

PR #2928, opened 2026-06-08
fix(router/connectrpc): map GraphQL enum values to proto names on response path

### LINKED ISSUE #2924: ConnectRPC transcoder fails to map GraphQL enum strings to proto-prefixed values
### Component(s)

router

### Component version

0.321.0

### wgc version

0.121.2

### controlplane version

n/a — reproduction uses a static execution_config.file. No controlplane, no Cosmo Cloud.

### router version

0.321.0

## Description
Given a GraphQL enum declared in the subgraph SDL:

```graphql
enum DocumentStatus {
  active
  archived
}

type Document @key(fields: "id") {
  id: ID!
  title: String!
  status: DocumentStatus!
}
```

`wgc grpc-service generate` (via `@wundergraph/protographic`) produces the corresponding proto enum, applying the `SCREAMING_SNAKE_CASE` enum-name prefix per [Google's protobuf style guide](https://protobuf.dev/programming-guides/style/) and inserting an `*_UNSPECIFIED` zero value:

```proto
enum DocumentStatus {
  DOCUMENT_STATUS_UNSPECIFIED = 0;
  DOCUMENT_STATUS_active      = 1;
  DOCUMENT_STATUS_archived    = 2;
}
```

For a GraphQL response value of `"active"`, the two wires diverge as follows:

| Layer | What it emits / expects |
|---|---|
| GraphQL (`:3002/graphql`) | `"status": "active"` — bare GraphQL enum name |
| Proto descriptor (from `@wundergraph/protographic` / `wgc grpc-service generate`) | enum value names are prefixed: `DOCUMENT_STATUS_active`, `DOCUMENT_STATUS_archived`, … |
| proto3-JSON canonical | `"status": "DOCUMENT_STATUS_active"` ↔ integer `1` |
| proto3-binary canonical | varint tag `0x18` (field 3, wire-type 0) followed by varint `1` |
| Cosmo router Connect-JSON wire | `"status":"active"` — bare GraphQL name passed through. The proto3-JSON parser matches the JSON string against the enum value names in the descriptor; `"active"` is not in that list (the descriptor only knows `DOCUMENT_STATUS_active`), so it falls back to the index-0 value `DOCUMENT_STATUS_UNSPECIFIED`. |
| Cosmo router Connect-binary wire | field 3 of `Document` absent — no tag, no value bytes. proto3-binary parsers treat a missing field as "use the type's default"; for enums that's the index-0 value, `DOCUMENT_STATUS_UNSPECIFIED` (indistinguishable on the wire from a deliberately-unset enum). |

Why the two wires fail differently:

- The proto3-JSON parser looks up the JSON string against the enum value names declared in the descriptor. `"active"` is not in the lookup table (the descriptor only knows `DOCUMENT_STATUS_active`), so the parser falls back to the proto3-JSON default — the index-0 entry. `@wundergraph/protographic` produces the prefix per [Google's protobuf style guide](https://protobuf.dev/programming-guides/style/); the router does not apply the inverse mapping when reading the GraphQL response.
- The proto3-binary parser sees no tag for field 3 and leaves the typed field at its declared default. Working hypothesis: the binary encoder shares the GraphQL→proto enum mapping path with the JSON encoder and, on a lookup miss, elects to skip the field rather than write the integer.

The defect is localized to enum fields. Scalars (`string`, `bool`), nested messages, and `repeated` lists in the same response message round-trip correctly. The identical query through the router's GraphQL endpoint returns the correct enum string.

## Steps to Reproduce

A self-contained reproduction lives at <https://github.com/wilsonjsq/cosmo-connectrpc-enum-bug>: two containers (a mock GraphQL subgraph and the stock router image), one composition step via `wgc router compose`, and one zero-dependency Node script that probes the three wires. `git clone` it and the commands below run as-is from the repo root.

Layout:

```
.
├─ subgraph.graphql                       # federation v2 SDL with one enum
├─ subgraph-server.mjs                    # mock subgraph, no deps
├─ services/document/service.proto        # proto matching protographic's output
├─ services/document/GetDocument.graphql  # the operation under test
├─ supergraph-config.yaml                 # `wgc router compose` input
├─ router.yaml                            # stock router config (inlined under Router configuration below)
├─ docker-compose.yml                     # subgraph + router
└─ proof.mjs                              # three wire probes + verdict
```

Commands:

```bash
# 1. Compose the supergraph locally — writes execution-config.json.
wgc router compose -i supergraph-config.yaml -o execution-config.json

# 2. Bring up the two-container stack.
docker compose up -d
docker compose logs router --tail=20    # expect "Router started"

# 3. Run the wire-level probe.
node proof.mjs

# 4. Tear down.
docker compose down
```

## Expected Result

A typed SDK reading the response should observe `DocumentStatus.DOCUMENT_STATUS_active` (= 1). Equivalently, on the wire:

- **JSON**: `"status": "DOCUMENT_STATUS_active"` (canonical proto3-JSON), or the integer `1`.
- **Binary**: the `Document` message contains the bytes `18 01` — tag `0x18` (field 3, wire type 0) followed by varint `1`.

## Actual Result

Captured live against `ghcr.io/wundergraph/cosmo/router:latest` (router `0.321.0`):

```
=== wire-level reproduction =====================================
Proto enum: DOCUMENT_STATUS_UNSPECIFIED=0  DOCUMENT_STATUS_active=1  DOCUMENT_STATUS_archived=2

(a) /3002/graphql body:        {"data":{"document":{"id":"1","title":"Quarterly Report","status":"active"}}}
(b) /5026 Connect-JSON body:   {"document":{"id":"1","title":"Quarterly Report","status":"active"}}
(c) /5026 Connect-binary (23 bytes) field 3 of Document:
    field 3 ABSENT (proto3 default 0 = DOCUMENT_STATUS_UNSPECIFIED)

GraphQL:                       "active"   (expected: "active")
Connect-JSON enum:             "active"   (proto3-JSON expects "DOCUMENT_STATUS_active")
Connect-binary enum tag:       field 3 ABSENT (proto3 default 0 = DOCUMENT_STATUS_UNSPECIFIED)

```

The `Document` message on the binary wire is 23 bytes; field 1 (`id`, scalar string) and field 2 (`title`, scalar string) consume all of them along with their tags. Only field 3 (the enum) is missing — surgical evidence that the defect is enum-specific, not a general encoder fault.

---

### Environment information

```
Host OS:        macOS (any host with Docker works; the stack itself runs in containers)
Docker Engine:  28.3.2
Node (host):    22.x (for `wgc router compose` and `proof.mjs`)
wgc:            0.121.2
Router image:   ghcr.io/wundergraph/cosmo/router:latest
  digest:       sha256:2e50309f55975fd13e99f9086d2191e6cfb614a0ceb011003959dc2e9211239f
  built:        2026-05-29
  version:      0.321.0
```

### Router configuration

```yaml
version: "1"

listen_addr: "0.0.0.0:3002"

execution_config:
  file:
    path: "/etc/router/execution-config.json"

log_level: "info"

traffic_shaping:
  all:
    request_timeout: 30s

connect_rpc:
  enabled: true
  server:
    listen_addr: "0.0.0.0:5026"
  storage:
    provider_id: "fs-services"
  graphql_endpoint: "http://localhost:3002/graphql"

storage_providers:
  file_system:
    - id: "fs-services"
      path: "/etc/router/services"
```

### Router execution config

```json
Generated locally by `wgc router compose -i supergraph-config.yaml -o execution-config.json` (~2 KB). Reproducible from the inputs above; not inlined for brevity.
```

### Log output

```shell
Router startup is clean (one offline-mode warning, then `Router started`):


{"level":"info","msg":"Config file provided. ...","service_version":"0.321.0"}
{"level":"warn","msg":"No graph token provided. The following Cosmo Cloud features are disabled. ...","features":["Schema Usage Tracking","Persistent operations","Advanced Request Tracing","Cosmo Cloud Tracing","Cosmo Cloud Metrics"]}
{"level":"info","msg":"discovered service","full_name":"repro.v1.DocumentService","dir":"/etc/router/services/document","proto_files":1,"operation_files":1}
{"level":"info","msg":"ConnectRPC server ready","listen_addr":"0.0.0.0:5026","services":1,"operations":1}
{"level":"info","msg":"Server initialized and ready to serve requests","listen_addr":"0.0.0.0:3002","playground":true,"introspection":true}
{"level":"info","msg":"Router started","component":"supervisor"}


No error or warning is logged when the enum mismatch occurs — the field is silently dropped on encode.
```

### Additional context

- **Repro repo**: <https://github.com/wilsonjsq/cosmo-connectrpc-enum-bug>. ~390 lines across 9 files; runs end-to-end in under a minute on a warm Docker daemon.
- **Likely root-cause surface**: `router/pkg/connectrpc/` — `handler.go` (Connect-JSON) and `vanguard_service.go` (Connect-binary + the gRPC family via Vanguard). Both encoders read the GraphQL enum string from the upstream response and would need to apply the same name mapping that `@wundergraph/protographic` applies at codegen time when producing the proto enum value names — i.e. `SCREAMING_SNAKE_CASE(EnumName) + "_" + GraphQLEnumValue`. For the binary path, writing the integer tag (`0x18 0x01` for field 3 = `active`) on lookup hit, and a hard error rather than silent omission on lookup miss, would also be defensible as a safety net. Happy to convert `proof.mjs` (or a table-driven variant covering all seven transport modes) into a regression test against the router test suite.
- Happy to provide additional captures (different enums, different schema shapes, the same proof script run against `createGrpcTransport({httpVersion:"2"})` and `createGrpcWebTransport` in both JSON and binary modes) on request.

asoorm:
## Summary

Fixes the ConnectRPC transcoder dropping GraphQL enum values on the response path (community issue #2924, ENG-9687).

The router had asymmetric enum handling: the request path strips proto enum prefixes (`MOOD_HAPPY` -> `HAPPY`) but the response path returned GraphQL data raw. The Vanguard transcoder then parses that JSON against the method's output descriptor, which only knows the prefixed value names (`MOOD_HAPPY`). A bare value like `HAPPY` misses the lookup:

| Wire | Result before this fix |
| -- | -- |
| Connect-JSON | lookup miss -> falls back to index-0 `*_UNSPECIFIED` |
| Connect-binary / gRPC | enum can't be encoded -> field silently omitted -> decodes as `*_UNSPECIFIED` |

## The fix

Adds `convertGraphQLResponseToProtoJSON`, the inverse of `convertProtoJSONToGraphQLVariables`, wired into `HandleRPC` between the GraphQL call and the return. It:

- Re-adds the prefix descriptor-driven: iterates `enumDesc.Values()`, strips each value's prefix via the existing `stripEnumPrefixWithType`, and matches the GraphQL value - symmetric with the request path, avoiding any case-conversion mismatch with protographic's value naming.
- Handles repeated enums and nested messages (incl. abstract types, which protographic emits as `oneof`).
- Maps null/absent enums to `*_UNSPECIFIED` (inverse of the request path's `*_UNSPECIFIED` -> `""`).
- On a lookup miss, leaves the value as-is so the downstream proto3-JSON parse surfaces a hard error rather than silently substituting a wrong value.

No behavior change for non-enum fields: every other GraphQL scalar already round-trips as valid proto3-JSON.

## Scope

Enums only - per protographic's scalar map, every non-enum value is already valid proto3-JSON. This gap existed in both the original `eng-8277` (structpb) branch and merged main (Vanguard) - never a regression.

## Tests

- RED -> GREEN: `TestConvertGraphQLResponseToProtoJSON` (asserts `HAPPY` -> `MOOD_HAPPY` via `test.EmployeeService.GetEmployee`) now passes.
- All existing `TestConvertProtoJSONToGraphQLVariables` request-path cases still pass.

Follow-ups (tracked in ENG-9687): extend test coverage for null->`UNSPECIFIED`, nested/repeated, case-preserving values, and a wire-level regression test.

Closes #2924

## Summary by CodeRabbit

* **Bug Fixes**
  * Restored proto-style enum prefixes in RPC responses so enum values returned from GraphQL are rewritten to expected proto names; handles nullable enums (maps null to proto zero value), repeated enums, nested fields, and preserves case segments.

* **Tests**
  * Added regression test covering enum conversion behavior, nested/repeated cases, null handling, and unknown-value passthrough.

* **Chores**
  * Expanded test schema with document-related enums/messages to exercise conversion logic.

### DIFF
--- a/router/pkg/connectrpc/handler.go
+++ b/router/pkg/connectrpc/handler.go
@@ -188,7 +188,17 @@ func (h *RPCHandler) HandleRPC(ctx context.Context, serviceName, methodName stri
 		return nil, fmt.Errorf("failed to execute GraphQL query: %w", err)
 	}
 
-	return responseJSON, nil
+	// Convert the GraphQL response back to proto JSON. This is the inverse of the request-path
+	// transform: the subgraph emits bare GraphQL enum values (e.g. "HAPPY"), but the Vanguard
+	// transcoder parses this JSON against the method's output descriptor, which only knows the
+	// proto-prefixed value names (e.g. "MOOD_HAPPY"). Without re-adding the prefix the enum is
+	// dropped on the binary/gRPC wire and falls back to *_UNSPECIFIED on the JSON wire.
+	protoResponseJSON, err := h.convertGraphQLResponseToProtoJSON(serviceName, methodName, responseJSON)
+	if err != nil {
+		return nil, fmt.Errorf("failed to convert GraphQL response to proto JSON: %w", err)
+	}
+
+	return protoResponseJSON, nil
 }
 
 // convertProtoJSONToGraphQLVariables processes proto JSON for GraphQL compatibility.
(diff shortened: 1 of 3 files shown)


router/pkg/connectrpc/handler.go
```diff
@@ -412,6 +422,157 @@ func toUpperSnakeCase(s string) string {
...
+	default:
+		return data
+	}
+}
+
+// addProtoEnumPrefixToValue handles a single field value: enums get the prefix re-added (handling
+// both scalar and repeated enums), nested messages recurse with their own descriptor, and any other
+// value passes through unchanged.
+func (h *RPCHandler) addProtoEnumPrefixToValue(value any, fieldDesc protoreflect.FieldDescriptor) any {
+	if fieldDesc != nil {
```
Noroth: nit: you can invert the if to have less nesting

router/pkg/connectrpc/handler_test.go
```diff
@@ -176,3 +176,103 @@ func TestConvertProtoJSONToGraphQLVariables(t *testing.T) {
...
 	})
 }
+
+// TestConvertGraphQLResponseToProtoJSON covers the response direction (GraphQL -> proto),
+// the inverse of convertProtoJSONToGraphQLVariables. The GraphQL subgraph emits bare enum
+// values (e.g. "HAPPY"), but the proto descriptor only knows the prefixed value names
+// (e.g. "MOOD_HAPPY"). Without re-adding the prefix, the Vanguard transcoder cannot map the
+// value and silently drops the enum field on the binary wire / falls back to *_UNSPECIFIED on
+// the JSON wire. Regression test for https://github.com/wundergraph/cosmo/issues/2924.
+func TestConvertGraphQLResponseToProtoJSON(t *testing.T) {
```
Noroth, line 186: nit: you can add t.Parallel() to the tests

Noroth (approved):

PR #2934, opened 2026-06-09
chore: add hostname tag to Sentry config

gausie:
Tags each Sentry event with the hostname of the machine generating the telemetry, making it easier to filter and correlate events by server instance.

## Summary by CodeRabbit

* **Chores**
  * Telemetry now adds the machine hostname as a tag for error reporting across core and worker processes.
  * This improves traceability of reports to specific hosts without changing external behavior or public APIs.

### DIFF
--- a/controlplane/src/core/composition/composeGraphs.worker.ts
+++ b/controlplane/src/core/composition/composeGraphs.worker.ts
@@ -11,6 +11,7 @@
  * is safe for local modules. Value imports from npm packages are fine.
  */
 import { randomUUID } from 'node:crypto';
+import os from 'node:os';
 import { printSchemaWithDirectives } from '@graphql-tools/utils';
 import {
   federateSubgraphsContract,
@@ -87,6 +88,8 @@ if (SENTRY_ENABLED && SENTRY_DSN) {
     enableLogs: SENTRY_ENABLE_LOGS,
     spotlight: process.env.NODE_ENV !== 'production',
   });
+
+  Sentry.setTag('hostname', os.hostname());
 }
 
 function parseGRPCMapping(mappings: string): GRPCMapping {
--- a/controlplane/src/core/sentry.config.ts
+++ b/controlplane/src/core/sentry.config.ts
@@ -1,3 +1,4 @@
+import os from 'node:os';
 import process from 'node:process';
 import * as Sentry from '@sentry/node';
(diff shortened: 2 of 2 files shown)


pepol (approved):

controlplane/src/core/composition/composeGraphs.worker.ts
```diff
@@ -11,6 +11,7 @@
  * is safe for local modules. Value imports from npm packages are fine.
  */
 import { randomUUID } from 'node:crypto';
+import os from 'node:os';
```
comatory, line 14: nitpick

```suggestion
import { hostname } from 'node:os';
```

controlplane/src/core/composition/composeGraphs.worker.ts
```diff
@@ -90,6 +91,8 @@ if (SENTRY_ENABLED && SENTRY_DSN) {
     enableLogs: SENTRY_ENABLE_LOGS,
     spotlight: process.env.NODE_ENV !== 'production',
   });
+
+  Sentry.setTag('hostname', os.hostname());
```
comatory, line 95: nitpick

```suggestion
  Sentry.setTag('hostname', hostname());
```

comatory (approved):

controlplane/src/core/composition/composeGraphs.worker.ts
gausie: Is this a repository convention? I typically wouldn't decontextualise an unusual function like this, when someone sees `os.hostname()` they immediately know what's going on.

Aenimus: I think Ondřej just prefers scoped imports but it's not inherently bad or prescribed or anything. @comatory can tell you more

comatory: Nope just preference,I'm just used to doing it because of tree shaking... but this is backend code so not that important, so it's all good (that's why `nitpick`)

I'll think about enforcing this via linter in `studio/` codebase though.

gausie: Yeah agreed. In particular for `node:` scoped imports, i tend to always import the namespace. As you said, they can't be tree shaken.

PR #2935, opened 2026-06-09
refactor(controlplane): convert # private methods to TypeScript private in CompositionService

gausie:
## Summary

- Converts all ECMAScript `#private` methods in `CompositionService` to TypeScript `private` methods
- This makes them visible on the prototype at runtime, allowing the `@traced` class decorator to instrument them with Sentry spans
- More files to follow as part of this initiative

## Why

ECMAScript `#` private methods use slot-based access that bypasses prototype reflection, so `@traced` cannot see or wrap them. TypeScript `private` provides the same compile-time enforcement while remaining on the prototype at runtime.

## Summary by CodeRabbit

* **Refactor**
  * Internal refactoring of composition service implementation with no user-visible changes.


Aenimus (approved):
LGTM!

PR #2777, opened 2026-04-20
feat(router): entity caching with L1/L2, shadow mode, and analytics

jensneuse:
## Summary

- **Entity-level caching** for federated GraphQL: caches individual entities by `@key` fields at the router level, so a `Product` cached by one query is reused by every query that needs it — regardless of query shape or subgraph
- **Two-layer architecture**: L1 (per-request dedup via `sync.Map`) eliminates redundant entity fetches within a single request; L2 (Redis or in-memory ristretto) shares cached entities across requests with per-entity TTLs via `@entityCache(maxAge)`
- **Shadow mode** for safe rollout: cache reads/writes happen but cached data is never served — compares cached vs fresh responses field-by-field to measure staleness before going live
- **Analytics pipeline**: entity cache hit/miss/stale metrics exported via OpenTelemetry for observability dashboards
- **Circuit breaker** on L2: prevents cascade failures when Redis is degraded
- **Composition support**: new `@entityCache` and `@entityCacheTTL` directives validated and propagated through the composition layer
- **~40K lines** across composition (TypeScript), engine (graphql-go-tools), router (Go), and proto layers

Companion PR: [graphql-go-tools #1259](https://github.com/wundergraph/graphql-go-tools/pull/1259)

## About this PR

Supersedes #2677 (closed by GitHub as "no shared history with main"). The original `jensneuse/entity-caching` branch was rooted at an unsigned `Initial commit` while `wundergraph/cosmo:main` is rooted at the GPG-signed equivalent — same content at the root, but every commit SHA differs all the way down, so GitHub refused to compare or merge them.

This branch is a **single squashed commit** containing the full diff from `origin/main` (`d86216fc0`) to `jensneuse/entity-caching` (`e5ac2b97`), applied on top of current main. The original feature branch is preserved at `jensneuse/entity-caching` for full per-commit history if needed during review.

The squash was verified to revert nothing: `entity-caching`'s most recent main-equivalent commit matches the current `origin/main` tip exactly, so no recent main work is lost.

## Documentation

All entity caching docs live in [`docs/entity-caching/`](docs/entity-caching/):

| Document | Description |
|----------|-------------|
| [`engineering-brief.md`](docs/entity-caching/engineering-brief.md) | Full engineering brief — architecture, design decisions, review focus areas |
| [`directives.md`](docs/entity-caching/directives.md) | `@entityCache` and `@entityCacheTTL` directive definitions and usage |
[...]

PR #2944, opened 2026-06-10
feat(composition): entity caching directives, proto config types, and generated bindings

SkArchon:
## What's included
This PR splits out the composition parts from from #2777 (`jensneuse/entity-caching-v2`). It does some additional cleanup and changes on top of it. The most notable difference would be removing the `auto mapping` feature for the `queryCache` directive.

It adds the following directives
`@openfed__entityCache`
`@openfed__queryCache` (with `@openfed__is`)
`@openfed__requestScoped`
`@openfed__cacheInvalidate`
`@openfed__cachePopulate`

I've also gotten claudes help to split the pr into separate commits per directive so they are easier to reason about. (Any changes though will be new commits). I have also extracted the directives from the MD in the older PR to here: https://app.notion.com/p/wundergraph/Directives-37bb19e0a0ec80b9b505d896beed5897

## Summary by CodeRabbit

* **New Features**
  * Added entity caching directives with configurable TTL and negative-cache controls
  * Added query result caching with automatic and explicit entity key field mapping
  * Added cache invalidation and population configuration for mutation and subscription operations
  * Added request-scoped field support for federated schema composition


SkArchon:
@coderabbitai full review

composition/src/errors/errors.ts
```diff
@@ -2050,3 +2050,293 @@ export function nonEqualComposeDirectiveMajorVersionError(directiveName: Directi
...
+}
+
+// Shared validation for maxAge across @openfed__entityCache, @openfed__queryCache, and @openfed__cachePopulate.
+export function maxAgeNotPositiveIntegerErrorMessage(directiveName: string, value: number): string {
+  return `@${directiveName} maxAge must be a positive integer, got "${value}".`;
+}
+
+// Validation for @openfed__entityCache negativeCacheTTL. 0 disables negative caching; negative values rejected.
+export function negativeCacheTTLNotNonNegativeIntegerErrorMessage(directiveName: string, value: number): string {
+  return `@${directiveName} negativeCacheTTL must be a non-negative integer, got "${value}".`;
```
wilsonrivera: Could you make the message above and this one be `positive integer` or `non-negative integer`?

composition/src/utils/string-constants.ts
```diff
@@ -79,6 +83,7 @@ export const INT_SCALAR = 'Int';
 export const INTERFACE = `Interface`;
 export const INTERFACE_UPPER = 'INTERFACE';
 export const INTERFACE_OBJECT = 'interfaceObject';
+export const IS = 'openfed__is';
```
wilsonrivera: Can you prefix the names of the `openfed__` constants with `OPENFED_`

composition/src/v1/constants/directive-definitions.ts
```diff
@@ -818,3 +829,125 @@ export const TAG_DEFINITION: DirectiveDefinitionNode = {
   name: stringToNameNode(TAG),
   repeatable: true,
 };
+
+// @openfed__entityCache(maxAge: Int!, negativeCacheTTL: Int = 0, includeHeaders: Boolean = false, partialCacheLoad: Boolean = false, shadowMode: Boolean = false) on OBJECT
+export const ENTITY_CACHE_DEFINITION: DirectiveDefinitionNode = {
```
wilsonrivera: Same here, could you prefix the constants with `OPENFED_`?

wilsonrivera (changes requested):

composition/src/errors/errors.ts
SkArchon: So "Zero" is also a valid number but is not a positive integer, should I leave it as is, or make it more verbose by saying "must be zero or a positive integer"?

wilsonrivera: I think `zero or a positive integer` is better as `non-negative` doesn't imply zero is allowed

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -3984,6 +4045,1257 @@ export class NormalizationFactory {
...
+        }
+      }
+    };
+
+    walkSelections(operationDef.selectionSet.selections, entityTypeName, '');
+    return result;
+  }
+
+  // Returns the named type name after stripping NonNull and List wrappers.
+  getNamedTypeName(typeNode: TypeNode): string {
```
gausie: This function is an exact duplicate of `getTypeNodeNamedTypeName`

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -3984,6 +4045,1257 @@ export class NormalizationFactory {
...
+        newConfigurationData(false, typeName),
+      );
+      configurationData.entityCaching = { ...configurationData.entityCaching, requestScopedFields: fields };
+    }
+  }
+
+  // Extracts @openfed__entityCache from OBJECT types into ConfigurationData (nested under entityCaching,
+  // matching extractRequestScopedFields). The directive is only valid on entities (@key); the maxAge and
+  // negativeCacheTTL bounds are validated here since they are semantic, not expressible in the directive
+  // definition. Populates entityCacheConfigByTypeName so later caching directives can verify their return
```
gausie: Comments like this should be `/** */` jsdoc style, though I also wonder if this is the correct comment for this function anyway

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -3984,6 +4045,1257 @@ export class NormalizationFactory {
...
+    );
+    configurationData.entityCaching = {
+      ...configurationData.entityCaching,
+      rootFieldCacheConfigurations: [...(configurationData.entityCaching?.rootFieldCacheConfigurations ?? []), config],
+    };
+  }
+
+  // Extracts @openfed__cacheInvalidate from Mutation/Subscription fields. The return type must be a cached
+  // entity (@key + @openfed__entityCache). A non-Mutation/Subscription placement (including non-root fields,
+  // where operationType is undefined) is reported, never silently ignored.
```
gausie: This comment is duplicated below. One must be wrong. The remaining one should use jsdoc

proto/wg/cosmo/node/v1/node.proto
```diff
@@ -91,6 +91,95 @@ message DataSourceConfiguration {
   repeated EntityInterfaceConfiguration entity_interfaces = 14;
   repeated EntityInterfaceConfiguration interface_objects = 15;
   CostConfiguration cost_configuration = 16;
+  // Entity caching configuration (e.g. request-scoped fields from @openfed__requestScoped).
+  EntityCaching entity_caching = 21;
```
gausie: Just out of interest, why do we jump from 16 to 21 here?

gausie (changes requested):
I note that these directives don't actually do anything yet but would now be considered valid by the composition pipeline right? Not a big deal

proto/wg/cosmo/node/v1/node.proto
SkArchon: Good catch, will rename to 17.

wilsonrivera (approved):
LGTM

composition/src/v1/normalization/normalization-factory.ts
SkArchon: Done, but didn't document arg params, thought that overkill there.

PR #2950, opened 2026-06-10
fix(router): fix inefficient trigger id generation

dkorittki:
~~Do not merge before https://github.com/wundergraph/graphql-go-tools/pull/1531 is merged~~ --> released in v2.4.6, go.mod is updated

This pull request fixes a bug where the router creates more connections to message brokers (aka providers in Router terms) than necessary. The bug is only present on websocket subscriptions for Cosmo Streams. On GraphQL subscription subprotocol you can provide `initial_payload`. The router uses this data alongside query extensions from the subscription GraphQL query to feed the engines trigger hash generator. It means different initial payloads or query extensions meant the router producing different triggers. A trigger causes the router to create a new connection to the message broker.

For regular websocket subscriptions this makes sense because these two fields can influence the data a subgraph returns. You need to have different connections to have the router be able to receive different messages and deduplicate to the correct subscribers.

For Cosmo Streams this does not make sense, because the initial payload and query extensions don't influence a message brokers messages it sends to the router. Only the subject and provider-id do. It results in the router creating multiple triggers and thus connections to the broker, which all receive the same data. If clients subscribing to the router use individual headers and place them into `initial_payload` it means a 1 to 1 ratio of subscribers on the router and connections to Redis.

To combat this there's a way now in the engine that allows datasources to provide the hash input by themselves. This pull request implements the interface for that on the pubsub datasource. Now we only use a provider-id and subject/topic/channel to generate the trigger id. This means every subscriber using a subscription which shares the same subject and provider-id shares the same trigger.

## Summary by CodeRabbit

* **Improvements**
  * Reduced redundant triggers for GraphQL subscriptions: subscriptions with identical targets or matching connection properties now share triggers, and NATS subscriptions are established more reliably to avoid publish/subscribe races.
* **Tests**
  * Added end-to-end tests that validate trigger initialization, deduplication behavior, and expected trigger counts across multiple subscription scenarios.

## Checklist

## Open Source AI Manifesto


router/pkg/pubsub/datasource/subscription_datasource.go
```diff
@@ -88,17 +88,22 @@ func (s *PubSubSubscriptionDataSource[C]) SetHooks(hooks Hooks) {
 	s.hooks = hooks
 }
 
+func (s *PubSubSubscriptionDataSource[C]) ProvideTriggerHashInput(ctx *resolve.Context, input []byte, xxh *xxhash.Digest) error {
+	return s.triggerHashInput(ctx, input, xxh)
```
endigma: Could this be a method on `s.Adapter`? Seems to be defined in:

- `router/pkg/pubsub/kafka/engine_datasource_factory.go`
- `router/pkg/pubsub/nats/engine_datasource_factory.go`
- `router/pkg/pubsub/redis/engine_datasource_factory.go`

Inside their respective `ResolveDataSourceSubscription` just to be called here, maybe it could be part of the adapter interface? I would just prefer to see a static function over a value function whenever possible.

endigma (changes requested):

router/pkg/pubsub/datasource/subscription_datasource.go
dkorittki: As discussed to keep this fix simple I'll do it in a seperate pull request

endigma (approved):

alepane21 (approved):

PR #2945, opened 2026-06-10
feat(router): entity caching with L1/L2, shadow mode, and per-request cache controls (3/6)

SkArchon:
Part 2 of 4 — split out from #2777 (`jensneuse/entity-caching-v2`). Stacked on #2944.

## What's included
- **router/pkg/entitycache**: in-memory L1, Redis L2, circuit breaker
- **router/core**: `factoryresolver` mapping of proto cache config to planner metadata, graph server wiring, per-request cache-control headers (`graphql_handler`), plan generator nil guard, config schema + defaults
- **metrics/otel**: entity cache metrics (opt-in telemetry) and attributes
- **graphql-go-tools** bumped to PR wundergraph/graphql-go-tools#1259 HEAD (resolver/planner side of the feature) — the two must land together
- **router-tests/entity_caching**: full integration suite with self-contained gqlgen subgraphs, Redis tests, and a compose tool for `testdata/config.json`
- **demo**: cachegraph/cachegraph_ext/viewer subgraphs, cache-demo runner, latency injector, router cache configs (demo go.mod moves in lockstep with router deps; demo is path-replaced into router-tests builds)

## Excluded
`router/internal/graphiql/graphiql.html` (playground embed) ships in part 3/4 — this PR runs with the previous embedded playground.

## Verification
`go build ./...` passes in `router/` and `demo/`; `go vet ./entity_caching/...` passes in `router-tests/`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Update: test subgraphs coupled to demo
The six entity_caching test subgraphs were moved into `demo/pkg/subgraphs/cachetest/` and are imported by the suite — same coupling pattern as the main suite (`testenv` → `demo/pkg/subgraphs`). router-tests no longer carries subgraph implementations. Pure move: schemas/resolvers/data unchanged, `make compose` reproduces an identical `testdata/config.json`, full suite green.

## Diff cleanup
- The no-op-tracer-provider optimization (`HasNonDefaultExporter` + bootstrap guard) moved to the benchmark/tooling PR (6/6) — it is a general tracing perf change motivated by benchmarking, not entity caching.
- Reverted ~1.5k lines of `demo/pkg/subgraphs/projects/generated` proto churn — leftover `reserved` markers from fields added and removed during development; the GraphQL schema is unchanged.
[...]

PR #2952, opened 2026-06-10
feat(router): entity cache OTEL/Prometheus metrics (4/6)

SkArchon:
Part 4 of 6 — split out from #2777. Stacked on #2945.

## What's included
- **router/pkg/metric/entity_cache_metrics**: cache hit/miss/latency instruments recorded from the per-request cache analytics snapshot
- **otel attributes** for cache level/source/name/entity type (shared constants so recording sites can't drift)
- **`entity_caching_stats` opt-in config keys** for Prometheus and OTLP (yaml + env vars, schema, defaults) — opt-in per #2777 review #21
- **graph server wiring**: metric setup/shutdown and handler snapshot recording; cache analytics collection enables only when metrics are configured

The tree at this PR's tip is byte-identical to the pre-split router branch — the split is purely for reviewability.


SkArchon:
Auto-closed as merged by a force-push mishap. Replaced by #2956.

PR #2957, opened 2026-06-11
feat(controlplane): always sample batch subgraph publish traces

gausie:
## Problem

Batch subgraph publishes go through the `PublishFederatedSubgraphs` RPC, which is rare, slow, and frequently hits the ~90s request timeout. In prod, `SENTRY_TRACES_SAMPLE_RATE` is below 1.0, so the flat `tracesSampleRate` drops most **successful** batch publishes — only the timed-out ones reliably show up in Sentry. That makes it hard to track the end-to-end composition cost of a publish over time.

## Change

Replace the flat `tracesSampleRate` with a `tracesSampler` that:

- returns **1.0** for the batch publish RPC (`/wg.cosmo.platform.v1.PlatformService/PublishFederatedSubgraphs`), matched against the span name / `http.route` / `http.target` / `url.path` / `url.full`;
- falls back to `SENTRY_TRACES_SAMPLE_RATE` for everything else (no change for other transactions);
- honors an upstream sampling decision (`parentSampled`) so distributed traces stay intact.

Child composition spans (e.g. `ComposeGraphsWorker.composeGraphsInWorker`) inherit the parent's sampling decision, so the full trace — including per-federated-graph compose timings — is captured.

The forced-sample list is a single constant (`ALWAYS_SAMPLE_PATHS`), easy to extend to other operations later.

## Notes

- `tracesSampler` takes precedence over `tracesSampleRate`; the sampler returns the configured rate as its default, so overall sampling behaviour is unchanged except for the batch publish path.
- Takes effect after a control plane deploy.

## Summary by CodeRabbit

* **Chores**
  * Improved distributed tracing: selected RPC paths are now always fully sampled, existing upstream parent/child sampling decisions are preserved when present, and all other traces follow the configured sample rate—resulting in more accurate end-to-end trace coverage and clearer diagnostics for publish-related operations.


controlplane/src/core/sentry.config.ts
```diff
@@ -17,6 +17,29 @@ const {
...
+// subgraph publishes are rare, slow, and frequently hit the request timeout — we want a
+// trace for every one so their end-to-end composition cost can be tracked. Child spans
+// (e.g. ComposeGraphsWorker.composeGraphsInWorker) inherit the parent's sampling decision.
+const ALWAYS_SAMPLE_PATHS = ['/wg.cosmo.platform.v1.PlatformService/PublishFederatedSubgraphs'];
+
+const publishAwareTracesSampler: NonNullable<Sentry.NodeOptions['tracesSampler']> = (ctx) => {
+  // Keep distributed traces intact: honor an upstream sampling decision when present.
+  if (typeof ctx.parentSampled === 'boolean') {
+    return ctx.parentSampled ? 1 : SENTRY_TRACES_SAMPLE_RATE;
+  }
```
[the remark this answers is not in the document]
gausie: Good catch — fixed in 2948cf1db. The sampler now honors `parentSampled` exactly (`true → 1`, `false → 0`) instead of falling back to the base rate, so a root that opted out won't leave orphaned child spans.

I reordered rather than just changing `: SENTRY_TRACES_SAMPLE_RATE` to `: 0`: the always-sample path check runs first, *then* the `parentSampled` check. A batch publish is always a root request from the wgc CLI (no incoming `sentry-trace`), so force-sampling it first can't orphan anything, and every other transaction now respects the upstream decision.

controlplane/src/core/sentry.config.ts
```diff
@@ -17,6 +17,32 @@ const {
...
+const ALWAYS_SAMPLE_PATHS = ['/wg.cosmo.platform.v1.PlatformService/PublishFederatedSubgraphs'];
+
+const publishAwareTracesSampler: NonNullable<Sentry.NodeOptions['tracesSampler']> = (ctx) => {
+  const attrs = ctx.attributes ?? {};
+  const target = [ctx.name, attrs['http.route'], attrs['http.target'], attrs['url.path'], attrs['url.full']]
+    .filter(Boolean)
+    .join(' ');
+
+  // Batch publishes are always traced, regardless of the base rate or any upstream decision.
+  if (ALWAYS_SAMPLE_PATHS.some((path) => target.includes(path))) {
```
comatory: I know we'd need to benchmark this to be 100% sure but how about making `ALWAYS_SAMPLE_PATHS` a `Set` and then using `has` with just `target`?

controlplane/src/core/sentry.config.ts
```diff
@@ -17,6 +17,32 @@ const {
...
 
+// RPC paths we always trace at 100%, regardless of SENTRY_TRACES_SAMPLE_RATE. Batch
+// subgraph publishes are rare, slow, and frequently hit the request timeout — we want a
+// trace for every one so their end-to-end composition cost can be tracked. Child spans
+// (e.g. ComposeGraphsWorker.composeGraphsInWorker) inherit the parent's sampling decision.
+const ALWAYS_SAMPLE_PATHS = ['/wg.cosmo.platform.v1.PlatformService/PublishFederatedSubgraphs'];
+
+const publishAwareTracesSampler: NonNullable<Sentry.NodeOptions['tracesSampler']> = (ctx) => {
+  const attrs = ctx.attributes ?? {};
+  const target = [ctx.name, attrs['http.route'], attrs['http.target'], attrs['url.path'], attrs['url.full']]
```
comatory: Why do we need to check all attributes and iterate? I'd rather have something like

```
let target = "";

if (attrs['url.path']) {
  target = attrs['url.path']
```

etc. There's also another iteration down below which make things a bit worse.

comatory (changes requested):
I am worried that this code is in the hot path and I'd try to optimize here. Maybe I'm paranoid? I know we can't be sure if we don't benchmark but still... wDYT?

controlplane/src/core/sentry.config.ts
gausie: You're right, let me sort this

controlplane/src/core/sentry.config.ts
gausie: We're looking for context that contains the paths, not equals it, so this wouldn't be helpful. We will remove this once we've got to the bottom of things anyway.

comatory (approved):

PR #2963, opened 2026-06-11
feat: controlpanel move to a accept first strategy for batch

wilsonrivera:
## Summary by CodeRabbit

* **New Features**
  * Asynchronous subgraph publishing: enqueue a publish, receive a job ID, and track progress.
  * New job status API to query batch publish progress, results, failure reasons, and updated subgraph names.
  * Background cleanup of job details to remove old job records automatically.

* **Enhancements**
  * Improved error and status reporting for batch publishes (composition/deployment counts and sliced diagnostics).

* **Reliability**
  * Per-namespace locking to prevent concurrent batch publish conflicts.

## Checklist

## Open Source AI Manifesto


controlplane/src/redlock.d.ts
```diff
@@ -0,0 +1,154 @@
+/**
+ * Ambient module declaration for `redlock` (v5.0.0-beta.x).
+ *
+ * Only needed if you can't use the package's bundled types
+ * (e.g. moduleResolution quirks, vendored copy, or strict ESM interop).
+ * Place this file somewhere covered by your tsconfig `include`,
+ * e.g. ./types/redlock.d.ts
```
comatory, line 7: `types/` directory is more suitable than `src/`

controlplane/src/redlock.d.ts
```diff
@@ -0,0 +1,154 @@
+/**
+ * Ambient module declaration for `redlock` (v5.0.0-beta.x).
+ *
+ * Only needed if you can't use the package's bundled types
+ * (e.g. moduleResolution quirks, vendored copy, or strict ESM interop).
```
comatory, line 5: I think the reason is that this module does not provide ESM support at all?

controlplane/package.json
```diff
@@ -93,6 +93,7 @@
     "pino": "catalog:",
     "postgres": "^3.4.5",
     "prom-client": "^15.1.3",
+    "redlock": "5.0.0-beta.2",
```
comatory, line 96: It looks a bit that this library is not actively maintained. This is a beta release which is 4+ years old. Are there any specific reason to use it?
I have quickly looked around and found this as a recommendation: https://github.com/alexpota/redlock-universal

It is also compatible with `ioredis`

package.json
```diff
@@ -90,6 +90,7 @@
       "backslash": "<0.2.1",
       "qs": "6.14.1",
       "mdast-util-to-hast": ">=13.2.1",
+      "msgpackr": "1.12.1",
```
comatory: Why is this change here?

comatory (commented):
I did a quick pass on this, mostly not sure about the new dependency.
I think it'd be better to do synchronous review on this one.

controlplane/src/redlock.d.ts
wilsonrivera: It is, however, there is no `@types/` for version `v5.0.0-beta.x` and the one provided is not compatible as it has changed

controlplane/src/redlock.d.ts
wilsonrivera: Yes

controlplane/package.json
wilsonrivera: I played with this at first too, my concern is that is a very young package and not as widely adopted.

package.json
wilsonrivera: The issue is that version `1.10.x` used by the version of `bullmq` we have, is not compatible with node 24

comatory: Ah I missed that it's in overrides

endigma (approved):

controlplane/src/core/workers/DeleteBatchPublishJobDetailsWorker.ts
```diff
@@ -0,0 +1,117 @@
...
+import { IQueue, IWorker } from './Worker.js';
+
+const QueueName = 'batch_publish.delete_job_details';
+const WorkerName = 'DeleteBatchPublishJobDetailsWorker';
+
+export interface DeleteBatchPublishJobDetailsInput {
+  organizationId: string;
+  jobId: string;
+}
+
```
comatory, line 16: Can we rename this file so it  reflects the name of the actual class?

comatory, line 16: ... or perhaps place the queue class into a separate file?

controlplane/test/subgraph/batch-publish-subgraphs.test.ts
```diff
@@ -623,4 +629,94 @@ describe('Batch publish subgraphs tests', () => {
...
+    let statusResp: GetBatchPublishJobStatusResponse;
+    for (;;) {
+      statusResp = await client.getBatchPublishJobStatus({ jobId: resp.jobId });
+      expect(statusResp.response?.code).toBe(EnumStatusCode.OK);
+      expect(statusResp.status).not.toBe(BatchPublishJobStatus.FAILED);
+      if (statusResp.status === BatchPublishJobStatus.COMPLETED) {
+        break;
+      }
+
+      await new Promise((resolve) => setTimeout(resolve, 1000));
```
comatory, line 695: Why is this timeout here? It seems like this could start causing leaky tests. The infinite loop does not look right either.

comatory (changes requested):
Just few observations, otherwise looks fine to me 👍

controlplane/test/subgraph/batch-publish-subgraphs.test.ts
wilsonrivera: The timeout is there to give time composition to complete instead of bombarding the server instance

controlplane/src/core/workers/DeleteBatchPublishJobDetailsWorker.ts
wilsonrivera: I'm following the same standard for Worker/Queue we already have, we can restructure the Worker/Queue pair into separated files in a different PR to not poison the purpose of this one

comatory (approved):

PR #2965, opened 2026-06-12
test(controlplane): fix flaky keycloak "Realm not found" failures

gausie:
Controlplane tests share one Keycloak realm (`test`) and run in parallel with no readiness gate. Each parallel file races to `realms.create("test")`; the loser of the create→durably-readable window fails its first test with `Realm not found` (e.g. [this run](https://github.com/wundergraph/cosmo/actions/runs/27351429197/job/80839129318) — CheckSubgraphSchema, feature-flag-integration).

**Fix:** a vitest `globalSetup` (`controlplane/test/global-setup.ts`) that, once before any worker spawns, waits for Keycloak to be reachable and creates the `test` realm exactly once — removing the concurrent-create race and adding the readiness gate CI lacked. A bounded retry on user-create in `test-util.ts` covers residual Keycloak cache lag.

`globalSetup` is added to the existing `vite.config.ts` (not a new `vitest.config.ts`, which would shadow it and drop the graphql alias/pool settings).

**Verification:** the race only manifests under CI cold-start-under-load timing; it does not reproduce on a warm local Keycloak (confirmed: old code with wiped realm + parallelism shows zero realm errors). Locally the fix typechecks and runs clean across repeated cold-realm cycles. Real confirmation is several consecutive green controlplane CI runs.

## Summary by CodeRabbit

* **Tests**
  * Centralized Keycloak realm setup that runs once before parallel tests to avoid race conditions.
  * Added broader test coverage and stability: new retry/backoff behaviors, abort handling, and scenarios for transient auth/realm readiness and duplicate entries.
* **New Features**
  * Introduced a reusable bounded retry-with-backoff utility used by tests for reliable polling and retries.


controlplane/test/global-setup.ts
```diff
@@ -0,0 +1,86 @@
+import { pino } from 'pino';
+import Keycloak from '../src/core/services/Keycloak.js';
+
+// The realm every test shares. Must match `realm` in test/test-util.ts.
```
comatory: Don't we have some central config file or `.env` that we could place this in and then use that at all call-sites that require it?

controlplane/test/global-setup.ts
```diff
@@ -0,0 +1,86 @@
...
+ * Doing this here removes the race where ~16 parallel test files each call
+ * `realms.create('test')` in their first `SetupKeycloak`, and the loser of the
+ * create-vs-durably-readable window fails its first test with "Realm not found".
+ */
+export default async function setup() {
+  const logger = pino({ level: 'silent' });
+  const keycloakClient = new Keycloak({
+    apiUrl: process.env.KC_API_URL || 'http://localhost:8080',
+    realm: loginRealm,
+    clientId: 'studio',
```
comatory: `process.env.KC_CLIENT_ID`?

controlplane/test/global-setup.ts
```diff
@@ -0,0 +1,86 @@
...
+    adminUser: 'admin',
```
comatory: `process.env.KC_ADMIN_USER`?

controlplane/test/global-setup.ts
```diff
@@ -0,0 +1,86 @@
...
+    adminPassword: 'changeme',
```
comatory: `KC_ADMIN_PASSWORD`?

controlplane/test/global-setup.ts
```diff
@@ -0,0 +1,86 @@
...
+ *   1. Wait for Keycloak to become reachable (CI starts it in the background with
+ *      no readiness gate), and
+ *   2. Create the shared `test` realm exactly once.
+ *
+ * Doing this here removes the race where ~16 parallel test files each call
+ * `realms.create('test')` in their first `SetupKeycloak`, and the loser of the
+ * create-vs-durably-readable window fails its first test with "Realm not found".
+ */
+export default async function setup() {
+  const logger = pino({ level: 'silent' });
```
comatory, line 13: Q: Why silent?

controlplane/test/global-setup.ts
```diff
@@ -0,0 +1,86 @@
...
+/**
+ * Vitest global setup. Runs once in the main process before the worker pool is
+ * spawned, so it is the single point where we:
+ *   1. Wait for Keycloak to become reachable (CI starts it in the background with
+ *      no readiness gate), and
+ *   2. Create the shared `test` realm exactly once.
+ *
+ * Doing this here removes the race where ~16 parallel test files each call
+ * `realms.create('test')` in their first `SetupKeycloak`, and the loser of the
+ * create-vs-durably-readable window fails its first test with "Realm not found".
```
comatory: Q: What does `create-vs-durably-readable`  mean?

controlplane/test/global-setup.ts
```diff
@@ -0,0 +1,86 @@
...
+    try {
+      await task();
+      return;
+    } catch (e) {
+      lastError = e;
+      await new Promise((resolve) => setTimeout(resolve, delayMs));
+    }
+  }
+  throw lastError;
+}
```
comatory: In Node we could use promise-based timeout and tighten the function a bit:

```
import { setTimeout as sleep } from 'node:timers/promises';

  async function retry<T>(
    task: () => Promise<T>,
    { attempts, delayMs }: { attempts: number; delayMs: number },
  ): Promise<T> {
    for (let attempt = 1; ; attempt++) {
      try {
        return await task();
      } catch (error) {
        if (attempt >= attempts) {
          throw error;
        }
        await sleep(delayMs);
      }
    }
  }
```

I think we can just throw the error instead of having intermediary variable to hold it. We don't need to check for attempts, we only care about them in error paths.

controlplane/test/global-setup.ts
```diff
@@ -0,0 +1,86 @@
...
+  await retry(
+    async () => {
+      try {
+        await keycloakClient.client.realms.create({
+          realm,
+          enabled: true,
+          displayName: realm,
+          registrationEmailAsUsername: true,
+        });
+      } catch (e: any) {
```
comatory:
```suggestion
      } catch (e: unknown) {
```

controlplane/test/global-setup.ts
```diff
@@ -0,0 +1,86 @@
...
+        if (e.response?.status === 409) {
```
comatory:
```suggestion
        if (e instanceof Error && e.response?.status === 409) {
```

comatory (changes requested):

comatory:
@gausie Check `poll-with-backoff`, I created that function a while back. Maybe it'd be possible to have a common `retry` function that wraps it. It could be cancellable with a signal which can also be useful. I'll leave it up to you if you want to consider it.

controlplane/test/global-setup.ts
gausie: We are expecting failures and it's noisy. This is not applied to the actual tests

controlplane/test/global-setup.ts
gausie: Rephrased

controlplane/src/core/util/timers.ts
```diff
@@ -55,15 +73,56 @@ export async function pollWithBackoff(
...
+    if (!isLastAttempt) {
+      const delay = computeDelay(baseInterval, maxInterval, attempt, jitter);
+      if ((await sleep(delay, signal)) === 'aborted') {
+        throw signal?.reason;
+      }
+    }
+  }
+
+  throw lastError;
+}
```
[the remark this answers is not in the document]
gausie: Good catch — added a `RangeError` guard for `attempts < 1` (plus a test) in bf8d22511.

comatory (changes requested):
Can we rename the file `poll-with-backoff.ts` to something like `timers.ts`. It seems now that the util file is more general.

Otherwise all good.

comatory (approved):

PR #2974, opened 2026-06-15
fix(router): avoid blocking token validation across JWKS providers

ebachle:
## Summary

When the router is configured with **more than one JWKS provider** under `authentication.jwt.jwks` and `refresh_unknown_kid` is enabled, a token signed by any provider **other than the first-listed one** incurred an avoidable, rate-limiter-bounded **blocking delay on the request path** — up to `refresh_unknown_kid.max_wait` (default **2m**) per request under sustained load.

### Why it happened

`keyFuncWrapper` in `router/pkg/authentication/jwks_token_decoder.go` validated a token by trying each configured provider **sequentially, in config order**. On a kid cache miss with `refresh_unknown_kid` enabled, the underlying `jwkset` client blocks on its rate limiter and performs a live HTTP refresh **before** reporting "key not found" and letting the next provider try.

Because two providers with separate identity sources have **disjoint `kid` sets by design**, a token signed by provider #2 *always* missed provider #1's cache and tripped provider #1's blocking unknown-kid path. The cost scaled with the number of providers listed before the one that owns the key.

### The fix

Add a **non-blocking, refresh-free cache pre-pass** to `keyFuncWrapper`. When a token carries a `kid`, each provider's in-memory storage is consulted directly (`Storage.KeyRead`, an in-memory lookup that never triggers the rate-limited refresh). If a provider already holds the `kid` — the steady state for known / already-rotated keys — the token is validated against that provider and returned **without ever touching another provider's rate limiter or slow endpoint**.

Tokens without a `kid`, and kids that no provider has cached, fall through to the existing sequential loop, so:
- **genuinely-unknown kids keep their current behavior** (still rate-limited + refreshed), and
- **JWKS HTTP refresh volume is unchanged** (the rate limiter still governs real refreshes).

The per-entry audience/algorithm validation is extracted into a `tryEntry` helper so the pre-pass and the fallback loop share identical semantics.

## How to test

New integration tests live in `router-tests/security/authentication_test.go` under `TestHttpJwksAuthorization`.

```bash
cd router-tests

# The two regression cases. On the previous behavior these FAIL on latency
# (~2s for a 2-provider setup, ~4s for 3 providers); with this change they pass in <500ms.
go test ./security/ -run 'TestHttpJwksAuthorization/(known_kid_owned_by_a_later_provider|known_kid_owned_by_the_third_provider)' -v -count=1

# Guards that behavior is otherwise unchanged:
#  - a key owned by the FIRST provider still validates fast (no regression)
#  - a genuinely-unknown kid still blocks (>=700ms), i.e. semantics preserved
go test ./security/ -run 'TestHttpJwksAuthorization/(known_kid_owned_by_the_first_provider|genuinely_unknown_kid_still_blocks)' -v -count=1

# Full auth suite — no regressions:
go test ./security/ -run 'TestAuthentication|TestHttpJwksAuthorization|TestNonHttpAuthorization|TestAudienceValidation|TestAlgorithmMismatch|TestSupportedAlgorithms|TestUseCustomization|TestMultipleKeys|TestOidcDiscovery|TestIntrospectionAuthentication' -count=1
```

The tests drive the latency deterministically: a non-owning provider is made slow via `Server.SetRespondTime`, with `max_wait > interval` so the limiter proceeds into the (slow) refresh rather than short-circuiting. The measured request asserts validation of a later-provider key completes in `< 500ms`.

## Notes

- **No config surface change.** This is an internal latency fix; existing YAML/behavior for users is unchanged apart from the removed blocking.
- Follow-ups discussed but intentionally **out of scope** here: issuer/`iss`-based provider routing (also fixes the genuinely-unknown-kid and garbage-token cases) and an opt-in async "trigger-and-return" refresh. Both would change config surface and/or semantics and are better tracked separately.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved token validation efficiency across multiple authentication providers by checking cached signing keys before performing slower refresh operations.
  * Known keys are resolved without unnecessary delays, while genuinely unknown keys continue through the standard validation process.

* **Tests**
  * Added coverage for provider ordering, cached key lookups, unknown keys, and consistent validation performance across providers.


ebachle:
I promise I'm not stale

ebachle:
I still believe in you little PR!

endigma:
Hi @ebachle , FYI we've ticketed reviewing this internally now, the way we review OSS contributions is changing slightly so sorry for the delay

ebachle:
> Hi @ebachle , FYI we've ticketed reviewing this internally now, the way we review OSS contributions is changing slightly so sorry for the delay

All good!  I just don't want the bot to close it while y'all are still reviewing

PR #2980, opened 2026-06-18
feat(composition): @openfed__entityCache directive (1/5)

SkArchon:
This PR implements the `@openfed__entityCache` directive.

`@openfed__entityCache` — marks a `@key` entity type as cacheable. Introduces the entity-caching **proto base**: the `EntityCaching` container message (field 1) + `EntityCacheConfiguration`, regenerated Go/TS bindings, the `EntityCacheConfig` type, normalization/validation, error messages, and tests.

More documentation on this: https://app.notion.com/p/wundergraph/Directives-37bb19e0a0ec80b9b505d896beed5897

## Release Notes

* **New Features**
  * Added entity-level caching via the `openfed__entityCache` directive, supporting max age, negative caching TTL, header inclusion, partial cache loading, and shadow mode.
  * Plumbed per-entity caching settings through router configuration and protobuf messages for datasource setup.
* **Bug Fixes**
  * Added validation and clearer error messages for missing key directives and invalid `maxAge` / `negativeCacheTTL` values.
* **Tests**
  * Added a directive test suite covering validation and default/explicit configuration extraction.


composition/src/errors/errors.ts
```diff
@@ -2050,3 +2050,15 @@ export function nonEqualComposeDirectiveMajorVersionError(directiveName: Directi
 export function unknownSubgraphNameError(subgraphName: SubgraphName): Error {
   return new Error(`Internal Error: Expected subgraph "${subgraphName}" to be a valid record.`);
 }
+
+export function entityCacheWithoutKeyErrorMessage(typeName: string): string {
```
Aenimus: typeName: TypeName

composition/src/errors/errors.ts
```diff
@@ -2050,3 +2050,15 @@ export function nonEqualComposeDirectiveMajorVersionError(directiveName: Directi
...
+  return `Type "${typeName}" has @openfed__entityCache but no @key directive.`;
```
Aenimus: Can this error occur only on Objects?

Aenimus: Type "${typeName}" declares directive "@openfed__entityCache" but does not a define a "@key" directive.`

composition/src/errors/errors.ts
```diff
@@ -2050,3 +2050,15 @@ export function nonEqualComposeDirectiveMajorVersionError(directiveName: Directi
...
+}
+
+export function maxAgeNotPositiveIntegerErrorMessage(directiveName: string, value: number): string {
```
Aenimus: Please create and provides a params object for passing more than one parameter

Aenimus: type MaxAgeNotPositiveIntegerErrorParams = {
  directiveName: DirectiveName;
  value: number;
}

composition/src/errors/errors.ts
```diff
@@ -2050,3 +2050,15 @@ export function nonEqualComposeDirectiveMajorVersionError(directiveName: Directi
...
+  return `@${directiveName} maxAge must be a positive integer, got "${value}".`;
```
Aenimus: `The directive "@${directiveName}" argument "maxAge" must be provided a positive integer; received "${value}".`

Aenimus: Are the coords supplied through the top level of this error? If not, please add them

composition/src/errors/errors.ts
```diff
@@ -2050,3 +2050,15 @@ export function nonEqualComposeDirectiveMajorVersionError(directiveName: Directi
...
+}
+
+export function negativeCacheTTLNotNonNegativeIntegerErrorMessage(directiveName: string, value: number): string {
```
Aenimus: Params object here too please. With proper type aliases.

composition/src/errors/errors.ts
```diff
@@ -2050,3 +2050,15 @@ export function nonEqualComposeDirectiveMajorVersionError(directiveName: Directi
...
+  return `@${directiveName} negativeCacheTTL must be zero or a positive integer, got "${value}".`;
```
Aenimus: `The directive "@${directiveName}" argument "negativeCacheTTL" must be provided zero or a positive integer; received "${value}".`

You should probably make this error and the one above a single error and pass in the `argumentName: ArgumentName`

composition/src/router-configuration/types.ts
```diff
@@ -98,6 +98,30 @@ export type ConfigurationData = {
   keys?: RequiredFieldConfiguration[];
   requireFetchReasonsFieldNames?: Array<FieldName>;
   requires?: RequiredFieldConfiguration[];
+  entityCaching?: EntityCachingConfiguration;
+};
+
+// Extracted from @openfed__entityCache(maxAge: Int!, negativeCacheTTL: Int, includeHeaders: Boolean,
+// partialCacheLoad: Boolean, shadowMode: Boolean) on OBJECT types. Defines per-entity cache TTL and behavior.
+export type EntityCacheConfig = {
```
Aenimus: Configuration not Config

composition/src/errors/errors.ts
Aenimus: I see the directive location now

Aenimus: What about InterfaceObjects?

composition/src/v1/constants/directive-definitions.ts
```diff
@@ -818,3 +824,42 @@ export const TAG_DEFINITION: DirectiveDefinitionNode = {
...
+      kind: Kind.INPUT_VALUE_DEFINITION,
+      name: stringToNameNode(INCLUDE_HEADERS),
+      type: stringToNamedTypeNode(BOOLEAN_SCALAR),
+      defaultValue: { kind: Kind.BOOLEAN, value: false },
+    },
+    {
+      kind: Kind.INPUT_VALUE_DEFINITION,
+      name: stringToNameNode(PARTIAL_CACHE_LOAD),
+      type: stringToNamedTypeNode(BOOLEAN_SCALAR),
+      defaultValue: { kind: Kind.BOOLEAN, value: false },
```
Aenimus: Might be worth making a constant for a false boolean at this point

composition/src/v1/normalization/types/types.ts
```diff
@@ -137,6 +139,21 @@ export type ComposeDirectiveArgumentNode = {
...
+  readonly name: NameNode;
+  readonly loc?: Location;
+};
+
+export type EntityCacheArgumentNode = {
+  readonly kind: Kind.ARGUMENT;
+  readonly name: NameNode;
+  // maxAge/negativeCacheTTL are Int; includeHeaders/partialCacheLoad/shadowMode are Boolean.
+  // validateDirectives() guarantees each argument's value matches its declared type.
+  readonly value: IntValueNode | BooleanValueNode;
```
Aenimus: Please make a separate type for each argument type.

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -3987,6 +4000,87 @@ export class NormalizationFactory {
     }
   }
 
+  extractEntityCacheDirectives() {
+    for (const [typeName, parentData] of this.parentDefinitionDataByTypeName) {
```
Aenimus: We loop through the parent data in lots of places. Please just add a function that handles this directive to the existing iterator

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -3987,6 +4000,87 @@ export class NormalizationFactory {
...
+          ]),
+        );
+        continue;
+      }
+      // validateDirectives() (run earlier in normalize()) has already guaranteed each argument's type —
+      // Int for maxAge/negativeCacheTTL, Boolean for the flags — so the generic ConstDirectiveNode is
+      // narrowed once to the precise typed node, mirroring RequestScopedDirectiveNode/ComposeDirectiveNode.
+      // Optional arguments may be absent (definition defaults are not materialized onto the usage AST),
+      // so the config starts at the directive's documented defaults and each present argument overrides it.
+      const directive = entityCacheDirectives[0] as EntityCacheDirectiveNode;
```
Aenimus: Please handle this properly. Access index 0, and if it doesn't exist, produce a fatal error.

composition/src/v1/normalization/types/types.ts
Aenimus: In fact it should be specific per argument including name. Then you switch becomes a lot cleaner in the business logic.

composition/tests/v1/directives/entity-cache.test.ts
```diff
@@ -0,0 +1,151 @@
...
+  invalidDirectiveError,
+  maxAgeNotPositiveIntegerErrorMessage,
+  negativeCacheTTLNotNonNegativeIntegerErrorMessage,
+  ROUTER_COMPATIBILITY_VERSION_ONE,
+} from '../../../src';
+import { createSubgraphWithDefault, normalizeSubgraphFailure, normalizeSubgraphSuccess } from '../../utils/utils';
+
+// @openfed__entityCache marks an entity type as cacheable. It requires @key (so the router can
+// construct cache keys) and a positive maxAge (TTL in seconds).
+describe('@openfed__entityCache', () => {
```
Aenimus: Please ensure the test names prescribe to the conventional format in the project; i.e., `test that <predicate>`

shared/src/router-config/builder.ts
```diff
@@ -67,6 +69,41 @@ function costsToCostConfiguration(costs?: Costs): CostConfiguration | undefined
...
+          notFoundCacheTtlSeconds: BigInt(ec.notFoundCacheTtlSeconds),
+          includeHeaders: ec.includeHeaders,
+          partialCacheLoad: ec.partialCacheLoad,
+          shadowMode: ec.shadowMode,
+        }),
+      );
+    }
+  }
+  if (entityCacheConfigurations.length === 0) {
+    return undefined;
```
Aenimus: just return

shared/src/router-config/builder.ts
```diff
@@ -67,6 +69,41 @@ function costsToCostConfiguration(costs?: Costs): CostConfiguration | undefined
...
 
+/**
+ * Convert the entity-caching configuration spread across a subgraph's `ConfigurationData` into the
+ * `EntityCaching` message. Add new entity-caching field types to the collection loop and the
+ * emptiness check below.
+ *
+ * @returns The `EntityCaching` message, or `undefined` when empty so the field is omitted (like
+ * `costsToCostConfiguration`).
+ */
+function toEntityCaching(dataByTypeName?: Map<TypeName, ConfigurationData>): EntityCaching | undefined {
```
Aenimus: Should this be renamed `extractEntityCaching`?

shared/src/router-config/builder.ts
```diff
@@ -67,6 +69,41 @@ function costsToCostConfiguration(costs?: Costs): CostConfiguration | undefined
...
+  if (!dataByTypeName) {
+    return undefined;
```
Aenimus: Just return

shared/src/router-config/builder.ts
Aenimus: EntityCaching | void perhaps?

Aenimus: `EntityCaching` feels like a poor name to me.

composition/src/errors/errors.ts
SkArchon: let me check

composition/src/v1/normalization/normalization-factory.ts
Aenimus: I had a look, and I think you should add this extraction method here:
https://github.com/wundergraph/cosmo/blob/aec1738c071e028893239b6c4f9bf492cf175f36/composition/src/v1/normalization/normalization-factory.ts#L4146

composition/src/errors/errors.ts
Aenimus: Let's ignore InterfaceObjects for now, but this is something that should be tracked.

composition/src/errors/errors.ts
SkArchon: This is only used for entityCache, so removed directiveName and hardcoded it in. So will not introduce a type.

shared/src/router-config/builder.ts
SkArchon: Kept it as it is, the generated protobuf binding needs undefined | entitycaching.

SkArchon: Renaming to `EntityCachingConfiguration` similar to `CostConfiguration`.

composition/src/errors/errors.ts
SkArchon: You mean the directive coords? if so yes. (See "Product")

```
`The 1st instance of the directive "@openfed__entityCache" declared on coordinates "Product" is invalid for the following reason:
The directive "@openfed__entityCache" argument "maxAge" must be provided a positive integer; received "0".`
```

composition/src/directive-definition-data/directive-definition-data.ts
```diff
@@ -982,7 +983,7 @@ export const ENTITY_CACHE_DEFINITION_DATA = newDirectiveDefinitionData({
       INCLUDE_HEADERS,
       newDirectiveArgumentData({
         directive: `@${OPENFED_ENTITY_CACHE}`,
-        defaultValue: { kind: Kind.BOOLEAN, value: false },
+        defaultValue: BOOLEAN_FALSE_VALUE_NODE,
```
Aenimus: I think `FALSE_BOOLEAN_VALUE_NODE` makes more sense

composition/src/errors/errors.ts
```diff
@@ -2050,3 +2051,18 @@ export function nonEqualComposeDirectiveMajorVersionError(directiveName: Directi
...
+
+export function entityCacheWithoutKeyErrorMessage(typeName: TypeName): string {
+  return `Type "${typeName}" declares the directive "@openfed__entityCache" but does not define a "@key" directive.`;
+}
+
+export function maxAgeNotPositiveIntegerErrorMessage({
+  directiveName,
+  value,
+}: MaxAgeNotPositiveIntegerErrorParams): string {
+  return `The directive "@${directiveName}" argument "maxAge" must be provided a positive integer; received "${value}".`;
```
Aenimus: Is `0` an invalid value for `MaxAge`?

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -444,6 +456,10 @@ export class NormalizationFactory {
   directiveDefinitionDataByName = initializeDirectiveDefinitionDatas();
   doesParentRequireFetchReasons = false;
   edfsDirectiveReferences = new Set<string>();
+  // Cached entity configs keyed by type name, populated by extractEntityCacheDirective() from
```
Aenimus: Nitpick: please use multi-line comments `*/ /*` for comments spanning multiple lines

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -3987,6 +4000,106 @@ export class NormalizationFactory {
     }
   }
 
+  extractEntityCacheDirective({ directivesByName, kind, name: typeName }: ParentDefinitionData) {
```
Aenimus: If you change ParentDefinitionData to ObjectDefinitionData you don't need the guard

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -3987,6 +4000,106 @@ export class NormalizationFactory {
...
+    if (kind !== Kind.OBJECT_TYPE_DEFINITION) {
+      return;
+    }
+    const entityCacheDirectives = directivesByName.get(OPENFED_ENTITY_CACHE);
+    if (!entityCacheDirectives || entityCacheDirectives.length == 0) {
```
Aenimus: can be simplified to `!entityCacheDirectives?.length`

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -3987,6 +4000,106 @@ export class NormalizationFactory {
...
+    }
+    if (!this.keyFieldSetDatasByTypeName.has(typeName)) {
+      this.errors.push(
+        invalidDirectiveError(OPENFED_ENTITY_CACHE, typeName, FIRST_ORDINAL, [
+          entityCacheWithoutKeyErrorMessage(typeName),
+        ]),
+      );
+      return;
+    }
+    // validateDirectives() (run earlier in normalize()) has already guaranteed each argument's type —
```
Aenimus: Nitpick: multi-line comment

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -4149,6 +4262,10 @@ export class NormalizationFactory {
           const isEntity = this.entityDataByTypeName.has(parentTypeName);
           const operationTypeNode = this.operationTypeNodeByTypeName.get(parentTypeName);
           const isObject = parentData.kind === Kind.OBJECT_TYPE_DEFINITION;
+
+          // Extract @openfed__entityCache here (Object-only) so we reuse this iterator instead of a separate pass.
```
Aenimus: This comment is pointless please remove it

composition/tests/utils/utils.ts
```diff
@@ -132,3 +132,7 @@ export function createSubgraph(name: SubgraphName, sdlString: string): Subgraph
     url: '',
   };
 }
+
+export function createSubgraphWithDefault(sdlString: string): Subgraph {
```
Aenimus: can you rename to `createSubgraphWithDefaultName` and make the name a constant, please?

composition/tests/v1/directives/entity-cache.test.ts
```diff
@@ -0,0 +1,151 @@
...
```
Aenimus: `'@openfed__entityCache tests'` to follow convention please

composition/tests/v1/directives/entity-cache.test.ts
```diff
@@ -0,0 +1,151 @@
...
+  describe('validation', () => {
```
Aenimus: `Validation tests` to follow convention please

composition/tests/v1/directives/entity-cache.test.ts
```diff
@@ -0,0 +1,151 @@
...
+      expect(errors).toHaveLength(1);
+      expect(errors[0]).toStrictEqual(
+        invalidDirectiveError(OPENFED_ENTITY_CACHE, 'Product', FIRST_ORDINAL, [
+          negativeCacheTTLNotNonNegativeIntegerErrorMessage(-1),
+        ]),
+      );
+    });
+  });
+
+  describe('configuration extraction', () => {
```
Aenimus: `Configuration extraction tests` to follow convention please

composition/tests/v1/directives/entity-cache.test.ts
```diff
@@ -0,0 +1,151 @@
...
+      );
+      expect(errors).toHaveLength(1);
+      expect(errors[0]).toStrictEqual(
+        invalidDirectiveError(OPENFED_ENTITY_CACHE, 'Product', FIRST_ORDINAL, [
+          maxAgeNotPositiveIntegerErrorMessage({ directiveName: OPENFED_ENTITY_CACHE, value: 0 }),
+        ]),
+      );
+    });
+
+    test('that a negative maxAge is rejected', () => {
```
Aenimus, line 57: I think we need a test for a maxAge of 0

Aenimus: We have it; my bad

composition/src/errors/types/params.ts
```diff
@@ -94,3 +94,8 @@ export type InvalidArgumentValueErrorParams = {
   value: string;
   expectedTypeString: string;
 };
+
+export type MaxAgeNotPositiveIntegerErrorParams = {
```
Aenimus: As discussed, I believe we can remove this.

composition/src/errors/errors.ts
SkArchon: yes it is value and its validated

composition/src/v1/normalization/normalization-factory.ts
SkArchon: I know, but I don't like using `.length` as a check. I usually prefer being explicit, especially as not all people reading the codebase would realyze 0 could be falsy.

composition/src/errors/errors.ts
```diff
@@ -2050,3 +2050,15 @@ export function nonEqualComposeDirectiveMajorVersionError(directiveName: Directi
 export function unknownSubgraphNameError(subgraphName: SubgraphName): Error {
   return new Error(`Internal Error: Expected subgraph "${subgraphName}" to be a valid record.`);
 }
+
+export function entityCacheWithoutKeyErrorMessage(typeName: TypeName): string {
+  return `Type "${typeName}" declares the directive "@openfed__entityCache" but does not define a "@key" directive.`;
```
Aenimus: Just realised this is a nested error (my bad), so we should change it to
`Object "${typeName}" does not define a "@key" directive.`

composition/src/errors/errors.ts
```diff
@@ -2050,3 +2050,15 @@ export function nonEqualComposeDirectiveMajorVersionError(directiveName: Directi
...
+}
+
+export function maxAgeNotPositiveIntegerErrorMessage(value: number): string {
+  return `The argument "maxAge" must be provided a positive integer; received "${value}".`;
+}
+
+export function negativeCacheTTLNotNonNegativeIntegerErrorMessage(value: number): string {
+  return `The argument "negativeCacheTTL" must be provided a zero or positive integer; received "${value}".`;
```
Aenimus: nitpick: `provided zero or a`

composition/src/router-configuration/types.ts
```diff
@@ -98,6 +98,30 @@ export type ConfigurationData = {
...
+  // When true, allows partial cache hits — the router fetches only missing entities from the subgraph.
+  partialCacheLoad: boolean;
+  // When true, the cache runs in shadow mode — cache reads/writes happen but responses always come
+  // from the subgraph. Useful for warming caches or validating correctness without affecting traffic.
+  shadowMode: boolean;
+};
+
+export type EntityCachingConfiguration = {
+  // Attached to an entity type's ConfigurationData (e.g. "Product") from @openfed__entityCache.
+  entityCacheConfigurations?: Array<EntityCacheConfiguration>;
```
Aenimus: I mentioned this in the other PR; I think it doesn't make sense for this to be optional. We can handle the empty array. It just makes appending harder, IMO

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -3987,6 +4001,112 @@ export class NormalizationFactory {
...
+      return;
+    }
+
+    this.entityCacheConfigByTypeName.set(typeName, config);
+    const configurationData = getValueOrDefault(this.configurationDataByTypeName, typeName, () =>
+      newConfigurationData(true, typeName),
+    );
+    configurationData.entityCaching = {
+      ...configurationData.entityCaching,
+      entityCacheConfigurations: [...(configurationData.entityCaching?.entityCacheConfigurations ?? []), config],
```
Aenimus: We can avoid this messy code if it's never undefined (see comment on other file)

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -4149,6 +4269,9 @@ export class NormalizationFactory {
           const isEntity = this.entityDataByTypeName.has(parentTypeName);
           const operationTypeNode = this.operationTypeNodeByTypeName.get(parentTypeName);
           const isObject = parentData.kind === Kind.OBJECT_TYPE_DEFINITION;
+
+          this.extractEntityCacheDirective(parentData);
```
Aenimus: you can wrap this with `if (isObject) { }`, then the param can be `ObjectDefinitionData`, and you don't need the nested guard.

composition/src/errors/errors.ts
SkArchon: Changed. to what you requested, but isn't this correct "a zero or positive integer" implies that "zero" is still an integer?

composition/src/v1/normalization/normalization-factory.ts
SkArchon: I thought I posted this comment, but had left it in "pending"

SkArchon:
@Aenimus pushed changes for your review comments

composition/src/v1/normalization/normalization-factory.ts
Aenimus: `===` not `==`

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -3987,6 +4005,108 @@ export class NormalizationFactory {
...
+      this.errors.push(...entityCacheErrors);
+      return;
+    }
+
+    this.entityCacheConfigByTypeName.set(typeName, config);
+    const configurationData = getValueOrDefault(this.configurationDataByTypeName, typeName, () =>
+      newConfigurationData(true, typeName),
+    );
+
+    const entityCaching = getOrInitializeEntityCaching(configurationData);
```
Aenimus: You don't need to allocate; `getOrInitializeEntityCaching(configurationData).push(config)` will do

shared/src/router-config/builder.ts
```diff
@@ -67,6 +69,43 @@ function costsToCostConfiguration(costs?: Costs): CostConfiguration | undefined
...
+ */
+function extractEntityCachingConfiguration(
+  dataByTypeName?: Map<TypeName, ConfigurationData>,
+): EntityCachingConfiguration | undefined {
+  if (!dataByTypeName) {
+    return;
+  }
+  const entityCache: EntityCacheConfiguration[] = [];
+  for (const data of dataByTypeName.values()) {
+    for (const ec of data.entityCaching?.entityCacheConfigurations ?? []) {
```
Aenimus: The array can never be undefined now

composition/src/v1/normalization/normalization-factory.ts
SkArchon: Okay will change to 
```
getOrInitializeEntityCaching(configurationData).entityCacheConfigurations.push(config);
```
Since it was long I thought it might be cleaner this way, but will change.

Aenimus (approved):
LGTM!

Aenimus:
Removed approval because we discussed moving the directive extraction upstream.

Aenimus (changes requested):
As discussed, move the directive extraction upstream.

Aenimus (approved):
LGTM!

PR #2983, opened 2026-06-19
feat(composition): @openfed__cacheInvalidate directive (2/5)

SkArchon:
`@openfed__cacheInvalidate` — evicts the returned entity from the cache after a Mutation/Subscription operation completes. Adds `CacheInvalidateConfiguration` proto message (field 2) + regenerated bindings, the `CacheInvalidateConfig` type, normalization/validation, error messages, and tests.

More documentation on this: https://app.notion.com/p/wundergraph/Directives-37bb19e0a0ec80b9b505d896beed5897

## Summary by CodeRabbit

* **New Features**
  * Added support for `@openfed__cacheInvalidate` to evict cached entities when applied to Mutation and Subscription fields.
  * Router now collects cache-invalidation rules alongside existing entity-caching configuration.
* **Bug Fixes**
  * Enhanced validation and error messages for incorrect directive placement and return types.
* **Tests**
  * Added a dedicated test suite covering both successful normalization output and validation failures for `@openfed__cacheInvalidate`.
* **Chores**
  * Extended protocol buffer and builder plumbing to transport cache-invalidation configuration end-to-end.


SkArchon:
@coderabbit review

composition/src/errors/errors.ts
```diff
@@ -2066,3 +2066,11 @@ export function maxAgeNotPositiveIntegerErrorMessage({
 export function negativeCacheTTLNotNonNegativeIntegerErrorMessage(value: number): string {
   return `The directive "@openfed__entityCache" argument "negativeCacheTTL" must be provided a zero or positive integer; received "${value}".`;
 }
+
+export function cacheInvalidateOnNonMutationSubscriptionFieldErrorMessage(fieldCoords: string): string {
+  return `@openfed__cacheInvalidate is only valid on Mutation or Subscription fields, found on "${fieldCoords}".`;
```
Aenimus: Because this is a nested error, I would change to `Coordinates "${fieldCoords}" are not a Mutation or Subscription root field. `

composition/src/errors/errors.ts
```diff
@@ -2066,3 +2066,11 @@ export function maxAgeNotPositiveIntegerErrorMessage({
...
+}
+
+export function cacheInvalidateOnNonEntityReturnTypeErrorMessage(fieldCoords: string, returnType: string): string {
+  return `Field "${fieldCoords}" has @openfed__cacheInvalidate but returns non-entity type "${returnType}".`;
```
Aenimus: `Coordinates "${fieldCoords}" return non-entity type "${returnType}". `

composition/src/router-configuration/types.ts
```diff
@@ -119,9 +119,19 @@ export type EntityCacheConfiguration = {
   shadowMode: boolean;
 };
 
+// Extracted from @openfed__cacheInvalidate on Mutation/Subscription fields.
+// Tells the router to evict the returned entity from the cache after the operation completes.
+export type CacheInvalidateConfig = {
+  fieldName: FieldName;
+  operationType: string;
```
Aenimus: We have a type alias for this, depending on what it is you want.

composition/src/errors/errors.ts
```diff
@@ -2066,3 +2066,11 @@ export function maxAgeNotPositiveIntegerErrorMessage({
 export function negativeCacheTTLNotNonNegativeIntegerErrorMessage(value: number): string {
   return `The directive "@openfed__entityCache" argument "negativeCacheTTL" must be provided a zero or positive integer; received "${value}".`;
 }
+
+export function cacheInvalidateOnNonMutationSubscriptionFieldErrorMessage(fieldCoords: string): string {
+  return `@openfed__cacheInvalidate is only valid on Mutation or Subscription fields, found on "${fieldCoords}".`;
+}
+
+export function cacheInvalidateOnNonEntityReturnTypeErrorMessage(fieldCoords: string, returnType: string): string {
```
Aenimus: A Params object for multiple params please

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -4100,6 +4104,75 @@ export class NormalizationFactory {
     };
   }
 
+  // Dispatches the per-field caching directives. Must run after extractEntityCacheDirective() (reads
+  // entityCacheConfigByTypeName). All object types are walked, not just root operation types: these
+  // directives are declared `on FIELD_DEFINITION`, so they can be (mis)placed on any field.
+  // getOperationTypeNodeForRootTypeName() returns undefined for non-root types; each extractor then reports
+  // the misplacement via its operation-type check, rather than silently ignoring it.
+  processRootFieldCacheDirectives() {
```
Aenimus: Ignore my comments about this. Let's remove this and handle everything in `extractCacheInvalidateConfig`

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -4100,6 +4104,75 @@ export class NormalizationFactory {
...
+          this.extractCacheInvalidateConfig(parentTypeName, configurationTypeName, fieldName, fieldData, operationType);
+        }
+      }
+    }
+  }
+
+  // Extracts @openfed__cacheInvalidate from Mutation/Subscription fields. The return type must be a cached
+  // entity (@key + @openfed__entityCache). A non-Mutation/Subscription placement (including non-root fields,
+  // where operationType is undefined) is reported, never silently ignored.
+  extractCacheInvalidateConfig(
```
Aenimus: I think you just pass in a `FieldData`, and then call this method in all the places we iterate over fields when adding them to the document.

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -4100,6 +4104,75 @@ export class NormalizationFactory {
...
+    parentTypeName: string,
+    configurationTypeName: string,
+    fieldName: string,
+    fieldData: FieldData,
+    operationType: OperationTypeNode | undefined,
+  ) {
+    const fieldCoords = `${parentTypeName}.${fieldName}`;
```
Aenimus: I would add `if (!fieldData.directivesByName.has(OPENFED_CACHE_INVALIDATE)) { return }` here

composition/src/v1/normalization/normalization-factory.ts
Aenimus: Also just FYI, in the current implementation where this is called, you're already iterating over the field datas. 
You iterate over the fieldDatas, call this, then iterate over the parents, then iterate over its field datas.

composition/src/v1/normalization/normalization-factory.ts
Aenimus: You can call this from `addFieldNamesToConfigurationData` actually

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -4100,6 +4104,75 @@ export class NormalizationFactory {
...
+    if (operationType !== OperationTypeNode.MUTATION && operationType !== OperationTypeNode.SUBSCRIPTION) {
```
Aenimus: You can do `const operationType = this.getOperationTypeNodeForRootTypeName(fieldData.parentTypeName);`

If operationType is undefined or `query`, then you raise an error.

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -4100,6 +4104,75 @@ export class NormalizationFactory {
...
+      this.errors.push(
+        invalidDirectiveError(OPENFED_CACHE_INVALIDATE, fieldCoords, FIRST_ORDINAL, [
+          cacheInvalidateOnNonMutationSubscriptionFieldErrorMessage(fieldCoords),
+        ]),
+      );
+      return;
+    }
+    const returnTypeName = getTypeNodeNamedTypeName(fieldData.node.type);
+    if (!this.keyFieldSetDatasByTypeName.has(returnTypeName) || !this.entityCacheConfigByTypeName.has(returnTypeName)) {
```
Aenimus: Do you need to check them both?

Aenimus: I think you can just check `this.entityDataByTypeName.has(returnTypeName)` ?

composition/tests/v1/directives/cache-invalidate.test.ts
```diff
@@ -0,0 +1,209 @@
...
+  SUBSCRIPTION,
+  type TypeName,
+} from '../../../src';
+import {
+  cacheInvalidateOnNonEntityReturnTypeErrorMessage,
+  cacheInvalidateOnNonMutationSubscriptionFieldErrorMessage,
+} from '../../../src/errors/errors';
+import { createSubgraphWithDefault, normalizeSubgraphFailure, normalizeSubgraphSuccess } from '../../utils/utils';
+
+describe('@openfed__cacheInvalidate', () => {
```
Aenimus: Please apply the test naming conventions (as discussed in PR #1)

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -4371,6 +4444,10 @@ export class NormalizationFactory {
     this.addValidConditionalFieldSetConfigurations();
     // this is where @key configurations are added to the ConfigurationData
     this.addValidKeyFieldSetConfigurations();
+    // per-field caching directives (@openfed__cacheInvalidate etc.) — must run after entityCache
+    // (entityCacheConfigByTypeName is populated earlier by extractEntityCacheDirective() during the
+    // parentDefinitionData iteration in normalize())
+    this.processRootFieldCacheDirectives();
```
Aenimus: You can call the extract function here directly and pass in fieldData

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -4100,6 +4104,75 @@ export class NormalizationFactory {
...
+      this.errors.push(
+        invalidDirectiveError(OPENFED_CACHE_INVALIDATE, fieldCoords, FIRST_ORDINAL, [
+          cacheInvalidateOnNonEntityReturnTypeErrorMessage(fieldCoords, returnTypeName),
+        ]),
+      );
+      return;
+    }
+    const config: CacheInvalidateConfig = {
+      fieldName,
+      operationType: operationType === OperationTypeNode.MUTATION ? MUTATION : SUBSCRIPTION,
```
Aenimus: you can pass in the operationType directly from the code suggested in the other comment

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -4100,6 +4104,75 @@ export class NormalizationFactory {
...
+      entityTypeName: returnTypeName,
+    };
+    const configurationData = getValueOrDefault(this.configurationDataByTypeName, configurationTypeName, () =>
+      newConfigurationData(false, configurationTypeName),
+    );
+
+    const existingCacheInvalidates = configurationData.entityCaching?.cacheInvalidateConfigurations ?? [];
+    configurationData.entityCaching = {
```
Aenimus: This creates a new object. Please can we check for the array and either append to the existing one or create a new one, leaving the original object in tact?

composition/src/v1/normalization/normalization-factory.ts
SkArchon: The caller already validates this, so I think we can skip (the caller will need to do the validating as it needs to decide between cacheInvalidate, cachePopulate, etc)

composition/src/v1/normalization/normalization-factory.ts
SkArchon: I am assuming you mean without passing it from the parent?

composition/src/v1/normalization/normalization-factory.ts
SkArchon: Ignoring this for your other comment

composition/src/v1/normalization/normalization-factory.ts
SkArchon: Didn't get what you mean by `check for the array`, do you mean something like this?

```
if (!configurationData.entityCaching) {
      configurationData.entityCaching = {};
    }
    configurationData.entityCaching.cacheInvalidateConfigurations = [...existingCacheInvalidates, config];
```

composition/src/v1/normalization/normalization-factory.ts
SkArchon: to verify

composition/src/v1/normalization/normalization-factory.ts
SkArchon: could you elaborate a bit on this?

Aenimus: What is the point of `extractFieldCacheInvalidateConfig`?

composition/src/v1/normalization/normalization-factory.ts
Aenimus: Are you saying that `extractFieldCacheInvalidateConfig` will handle multiple of these directives or?

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -4107,6 +4111,69 @@ export class NormalizationFactory {
...
+    const config: CacheInvalidateConfig = {
+      fieldName,
+      operationType: operationType,
+      entityTypeName: returnTypeName,
+    };
+    const configurationData = getValueOrDefault(this.configurationDataByTypeName, configurationTypeName, () =>
+      newConfigurationData(false, configurationTypeName),
+    );
+
+    const existingCacheInvalidates = configurationData.entityCaching?.cacheInvalidateConfigurations ?? [];
```
Aenimus: I do wonder what the point of having an undefined array is. It just seems to make life harder.

If it's important the proto receives undefined, you can simply not pass the array if it's length 0.

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -4280,8 +4347,13 @@ export class NormalizationFactory {
             parentData.fieldDataByName.delete(SERVICE_FIELD);
             parentData.fieldDataByName.delete(ENTITIES_FIELD);
           }
+
+          const newParentTypeName = getParentTypeName(parentData);
+
           const externalInterfaceFieldNames: Array<string> = [];
           for (const [fieldName, fieldData] of parentData.fieldDataByName) {
+            this.extractFieldCacheInvalidateConfig(parentData, newParentTypeName, fieldName, fieldData);
```
Aenimus: In the comment where you asked me to elaborate, I was suggesting to call the inner function here. I don't understand what the wrapper is doing.

I also don't see why you need to pass anything into here besides `fieldData`.

Aenimus: You can also wrap it in `if (isObject) {}`, which is the only reason parentData is provided as an argument

composition/tests/v1/directives/cache-invalidate.test.ts
```diff
@@ -0,0 +1,181 @@
...
+  });
+});
+
+// Returns the ConfigurationData for a type. Entity-caching config is nested under `.entityCaching`.
+function getConfigForType(sg: Subgraph, typeName: string): ConfigurationData | undefined {
+  const result = new BatchNormalizer({ subgraphs: [sg] }).batchNormalize() as BatchNormalizationSuccess;
+  expect(result.success).toBe(true);
+  const internal = result.internalSubgraphByName.get(sg.name);
+  expect(internal).toBeDefined();
+  return internal!.configurationDataByTypeName.get(typeName as TypeName);
```
Aenimus: Why are you casting typeName here? And why isn't typeName just aliased as TypeName in the parameters for this helper?

composition/src/v1/normalization/normalization-factory.ts
SkArchon: nvm ignore my comment above

composition/src/v1/normalization/normalization-factory.ts
SkArchon: Since we hold multiple directives (upto 5) in here `configurationData.entityCaching = {};`, we can have an undefined array for cacheInvalidate if before this the `entityCache` directive was processed. So we need to do
```
const existingCacheInvalidates = configurationData.entityCaching?.cacheInvalidateConfigurations ?? [];
```
the alternative is to basically, any time this is called
```
if (!configurationData.entityCaching) {
      configurationData.entityCaching = {};
}
```
instead of initializing to `{}` we will set defaults, either via a constructor, etc

Lmk wdyt, we can also discuss on Monday

composition/src/v1/normalization/normalization-factory.ts
SkArchon: I removed this, we now call `extractCacheInvalidateConfig` directly.

composition/src/v1/normalization/normalization-factory.ts
Aenimus: I'm saying I think the array being undefined serves no purpose; it just makes life harder. You can initialize the object with a helper that initializes the nested array. This makes the most sense to me.

composition/src/v1/normalization/normalization-factory.ts
Aenimus: Even that call would need to be wrapped in `isObject`

SkArchon: Updated, only additionally passing in parentKind to check for object, can also do the validation outside the function

Aenimus: I think you misunderstood—the parameter you need to pass in is FieldData. Everything else is inside that object.

composition/src/router-configuration/types.ts
```diff
@@ -119,9 +120,19 @@ export type EntityCacheConfiguration = {
...
+  fieldName: FieldName;
+  operationType: OperationTypeNode;
+  entityTypeName: TypeName;
+};
+
 export type EntityCachingConfiguration = {
   // Attached to an entity type's ConfigurationData (e.g. "Product") from @openfed__entityCache.
   entityCacheConfigurations: Array<EntityCacheConfiguration>;
+  // Attached to the Mutation/Subscription type's ConfigurationData from @openfed__cacheInvalidate.
+  cacheInvalidateConfigurations: Array<CacheInvalidateConfig>;
```
Aenimus: The type should be `CacheInvalidationConfiguration`

Aenimus: The property should be `cacheInvalidationConfigurations`

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -4107,6 +4111,44 @@ export class NormalizationFactory {
...
+    const returnTypeName = getTypeNodeNamedTypeName(fieldData.node.type);
+    if (!this.entityCacheConfigByTypeName.has(returnTypeName)) {
+      this.errors.push(
+        invalidDirectiveError(OPENFED_CACHE_INVALIDATE, fieldCoords, FIRST_ORDINAL, [
+          cacheInvalidateOnNonEntityReturnTypeErrorMessage({ fieldCoords, returnType: returnTypeName }),
+        ]),
+      );
+      return;
+    }
+    const configurationData = getValueOrDefault(this.configurationDataByTypeName, fieldData.renamedParentTypeName, () =>
```
Aenimus, line 4148: Make sure you have a test that checks a renamed root node, e.g., `Mutations`.

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -4107,6 +4111,44 @@ export class NormalizationFactory {
...
+      newConfigurationData(false, fieldData.renamedParentTypeName),
+    );
+
+    const config: CacheInvalidateConfig = {
+      fieldName: fieldData.name,
+      operationType: operationType,
+      entityTypeName: returnTypeName,
+    };
+
+    const entityCaching = getOrInitializeEntityCaching(configurationData);
```
Aenimus: Without allocation
`getOrInitializeEntityCaching(configurationData).push(config);`

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -4212,6 +4254,17 @@ export class NormalizationFactory {
     }
     // Check all key field sets for @external fields to assess whether they are conditional
     this.evaluateExternalKeyFields();
+    /*
+     * Register every @openfed__entityCache config before the main loop below extracts @openfed__cacheInvalidate.
+     * cacheInvalidate validation requires the return type's entity-cache config to already exist
+     * (see extractCacheInvalidateConfig), and the return type may be iterated after the field that references it.
+     */
+    for (const parentData of this.parentDefinitionDataByTypeName.values()) {
```
Aenimus: I don't understand this loop.

composition/tests/v1/directives/cache-invalidate.test.ts
```diff
@@ -0,0 +1,224 @@
...
+  type TypeName,
+} from '../../../src';
+import {
+  cacheInvalidateOnNonEntityReturnTypeErrorMessage,
+  cacheInvalidateOnNonMutationSubscriptionFieldErrorMessage,
+} from '../../../src/errors/errors';
+import { createSubgraphWithDefaultName, normalizeSubgraphFailure, normalizeSubgraphSuccess } from '../../utils/utils';
+
+describe('@openfed__cacheInvalidate directive tests', () => {
+  describe('Mutation field tests', () => {
```
Aenimus, line 24: Add a test for a renamed root type, e.g., the `Mutation` node is named `Mutations`

composition/src/v1/normalization/normalization-factory.ts
SkArchon: updated in other pr

composition/tests/v1/directives/cache-invalidate.test.ts
SkArchon: done already

composition/src/v1/normalization/normalization-factory.ts
SkArchon: might be missing some context, but done already I believe

composition/src/v1/normalization/normalization-factory.ts
SkArchon: done already

composition/src/router-configuration/types.ts
SkArchon: I changed it, but was thinking if this is correct? Since the name of the directive is `cacheIndvalidate`?

shared/src/router-config/builder.ts
```diff
@@ -101,12 +103,22 @@ function extractEntityCachingConfiguration(
...
+        new CacheInvalidateConfiguration({
+          fieldName: ci.fieldName,
+          operationType: ci.operationType,
+          entityTypeName: ci.entityTypeName,
+        }),
+      );
+    }
   }
-  if (entityCache.length === 0) {
+  if (entityCache.length === 0 && cacheInvalidateConfigurations.length === 0) {
```
Aenimus: I think I wrote in PR #3 that we should do `entityCache.length > 0 || cacheInvalidateConfigurations.length >0 ...)` and return the entityCachingConfiguration there

composition/src/router-configuration/types.ts
SkArchon: reverted

SkArchon: reverted

composition/src/router-configuration/types.ts
```diff
@@ -119,9 +120,19 @@ export type EntityCacheConfiguration = {
   shadowMode: boolean;
 };
 
+// Extracted from @openfed__cacheInvalidate on Mutation/Subscription fields.
+// Tells the router to evict the returned entity from the cache after the operation completes.
+export type CacheInvalidateConfiguration = {
+  fieldName: FieldName;
+  operationType: OperationTypeNode;
+  entityTypeName: TypeName;
```
Aenimus: nitpick: alphabetical order

composition/src/router-configuration/types.ts
```diff
@@ -119,9 +120,19 @@ export type EntityCacheConfiguration = {
...
+};
+
 export type EntityCachingConfiguration = {
   // Attached to an entity type's ConfigurationData (e.g. "Product") from @openfed__entityCache.
   entityCacheConfigurations: Array<EntityCacheConfiguration>;
+  // Attached to the Mutation/Subscription type's ConfigurationData from @openfed__cacheInvalidate.
+  cacheInvalidateConfigurations: Array<CacheInvalidateConfiguration>;
```
Aenimus, line 133: nitpick: alphabetical order

composition/src/router-configuration/utils.ts
```diff
@@ -20,6 +20,7 @@ export function getOrInitializeEntityCaching(configurationData: ConfigurationDat
   if (!configurationData.entityCaching) {
     configurationData.entityCaching = {
       entityCacheConfigurations: [],
+      cacheInvalidateConfigurations: [],
```
Aenimus, line 22: nitpick: alphabetical order

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -4111,6 +4121,43 @@ export class NormalizationFactory {
...
+      return;
+    }
+    const configurationData = getValueOrDefault(this.configurationDataByTypeName, fieldData.renamedParentTypeName, () =>
+      newConfigurationData(false, fieldData.renamedParentTypeName),
+    );
+
+    const config: CacheInvalidateConfiguration = {
+      fieldName: fieldData.name,
+      operationType: operationType,
+      entityTypeName: returnTypeName,
```
Aenimus: nitpick: alphabetical order

shared/src/router-config/builder.ts
```diff
@@ -101,13 +103,22 @@ function extractEntityCachingConfiguration(
...
+        }),
+      );
+    }
   }
-  if (entityCache.length === 0) {
-    return;
+  if (entityCache.length > 0 || cacheInvalidateConfigurations.length > 0) {
+    return new EntityCachingConfiguration({
+      entityCache,
+      cacheInvalidateConfigurations,
```
Aenimus, line 118: nitpick: alphabetical order

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -4282,8 +4330,13 @@ export class NormalizationFactory {
             parentData.fieldDataByName.delete(SERVICE_FIELD);
             parentData.fieldDataByName.delete(ENTITIES_FIELD);
           }
+
           const externalInterfaceFieldNames: Array<string> = [];
           for (const [fieldName, fieldData] of parentData.fieldDataByName) {
+            if (isObject) {
+              this.extractCacheInvalidateConfig(fieldData);
+            }
```
Aenimus: add an else if for the predicate below so we don't evaluate twice

Aenimus (approved):
LGTM!

PR #2988, opened 2026-06-19
fix(router): accept JSON content type with parameters in MCP requests

asoorm:
## Summary by CodeRabbit

* **Bug Fixes**
  * Router now correctly accepts MCP `tools/call` requests when `Content-Type` includes media-type parameters (for example, `application/json; charset=utf-8`), returning `200 OK` instead of `415 Unsupported Media Type`.
* **Tests**
  * Expanded and refactored MCP `tools/call` and header-forwarding coverage, including explicit assertions for `Content-Type` with parameters and more reliable request body construction.
* **Chores**
  * Updated Go module dependency versions used by the router and its test module.

## Summary

The router's MCP endpoint rejected requests whose `Content-Type` carried a
parameter, such as `application/json; charset=utf-8` (which some JVM-based
HTTP clients add automatically, outside the caller's control), with
`415 Unsupported Media Type`.

The pinned `modelcontextprotocol/go-sdk` v1.4.1 compared the header for an
exact match against `"application/json"`, so any parameter caused it to fail.
v1.6.0+ parses the header with `mime.ParseMediaType` and matches on the base
media type only.

This PR bumps the SDK to v1.6.1 in both the `router` and `router-tests`
modules. No router code changes are needed; the fix is entirely in the
dependency.

## Behavior change

- `application/json; charset=utf-8` was previously rejected with 415, now it
  is accepted.
- The base media type must still be `application/json`; parameters are
  ignored. This also means a request like `application/json; foo=bar` is now
  accepted rather than rejected. The body must still be valid JSON, so
  malformed input fails at decode time with a proper JSON-RPC error.

## Tests

- Added a regression test asserting `application/json; charset=utf-8` is
  accepted.
- Updated the existing test that asserted the old exact-match behavior.
- Verified locally against v1.6.1: `pkg/mcpserver` unit tests, the full
  `TestMCP` integration suite (with NATS running), and the MCP OAuth/auth
  tests all pass.


router-tests/protocol/mcp_test.go
```diff
@@ -1144,14 +1191,11 @@ input UserInput {
...
-			// and silently stripped them. The new SDK's StreamableHTTPHandler rejects them with 415
-			// Unsupported Media Type at the transport level before our code runs.
-			//
-			// This is the correct behavior per the MCP spec which requires "application/json".
-			// No legitimate MCP client sends custom content-type params.
-			t.Run("Non-standard Content-Type params are rejected by the SDK", func(t *testing.T) {
+			// The StreamableHTTPHandler parses the Content-Type media type rather than doing a
+			// strict string comparison, so JSON requests that carry parameters (e.g.
+			// "application/json; foo=bar") are accepted. Only the base media type must be
+			// "application/json"; any parameters are ignored.
```
endigma: This comment seems relevant to something internal in the library? I don't think we need this. If this ever fails in the future, we need to point at the library again as the likely culprit clearly.

router-tests/protocol/mcp_test.go
```diff
@@ -712,6 +712,53 @@ Important Notes:
 		})
 	})
 
+	t.Run("Content-Type Handling", func(t *testing.T) {
+		// Some MCP clients (notably JVM-based HTTP stacks) send
+		// `Content-Type: application/json; charset=utf-8`. The router must accept JSON
+		// media types that carry parameters by parsing the media type rather than doing
+		// a strict string comparison against "application/json".
```
endigma: I don't think "rather than" is useful here, we never did this, it was upstream bug, this whole thing is more like a regression test should the upstream MCP library break again

router-tests/protocol/mcp_test.go
```diff
@@ -712,6 +712,53 @@ Important Notes:
...
+						"arguments": map[string]any{
+							"criteria": map[string]any{},
+						},
+					},
+				}
+
+				requestBody, err := json.Marshal(mcpRequest)
+				require.NoError(t, err)
+
+				req, err := http.NewRequest("POST", mcpAddr, strings.NewReader(string(requestBody)))
```
endigma: `bytes.NewReader(requestBody)`

endigma (changes requested):

endigma (approved):

PR #2989, opened 2026-06-19
fix: pagination in proposals page

JivusAyrus:
## Summary by CodeRabbit

* **New Features**
  * Proposal listings now include a server-provided total count of matching proposals, enabling accurate pagination across the full filtered dataset.

* **Bug Fixes**
  * Pagination has been corrected to rely on the server-computed total instead of deriving page count from only the current page results, preventing incorrect navigation on large result sets.
  * The total count is now consistently returned (including a default of 0 when requests fail), improving reliability of the proposals page.

## Checklist


proto/wg/cosmo/platform/v1/platform.proto
```diff
@@ -3103,6 +3103,7 @@ message GetProposalsByFederatedGraphResponse {
   Response response = 1;
   repeated Proposal proposals = 2;
   bool isProposalsEnabled = 3;
+  int32 totalProposalsCount = 4;
```
comatory: I would just use `totalCount` name for the property. It's what we use most of the time I think.

thisisnithin (approved):

controlplane/src/core/bufservices/proposal/getProposalsByFederatedGraph.ts
```diff
@@ -107,6 +111,12 @@ export function getProposalsByFederatedGraph(
       offset: req.offset,
     });
 
+    const totalProposalsCount = await proposalRepo.countByFederatedGraphId({
```
comatory: Couldn't we instead get the count in `proosalRepo.ByFederatedGraphId`? It seems like we're doing multiple round trips to the database this way and we already pass limit/offset etc to the method.

comatory (changes requested):

controlplane/src/core/bufservices/proposal/getProposalsByFederatedGraph.ts
JivusAyrus: we cant do it in the same query, as the main query will have limits and offsets and the count shouldnt have them

proto/wg/cosmo/platform/v1/platform.proto
JivusAyrus: Done

comatory (approved):

SkArchon (approved):

PR #2994, opened 2026-06-20
feat: router start more lenient on cosmo streams problems

alepane21:
Introduce a way to start the router even if Cosmo Streams provider are not available or not responding.
If the provider is not defined, the router will start, log out an error, and that's it.
If the provider is defined but not reachable, it will start, log the error and try to connect to the provider, without the router even needing to reload.

To test the behaviour I decided to create a small tcp proxy adapter (almost fully AI generated) to simulate those condition. The other way would have been to introduce something like toxiproxy, but I think it would have been an overkill at the moment.

By default I left the current behaviour, only if the new mode is enabled (with config or env), it will be enabled.

## Summary by CodeRabbit

* **New Features**
  * Added a new option to keep the router starting even when some event providers are missing or temporarily unreachable.
  * Documented the new configuration flag and environment variable across router docs.
  * Expanded event-provider support for Kafka, NATS, and Redis with more resilient startup and recovery behavior.

* **Bug Fixes**
  * Improved handling of unavailable event providers so only affected fields are disabled instead of stopping the whole router.
  * Strengthened Redis, NATS, and Kafka connection handling to better recover when brokers become available again.

## Checklist

## Open Source AI Manifesto


router/internal/rediscloser/rediscloser.go
```diff
@@ -21,6 +21,11 @@ type RedisCloserOptions struct {
 	URLs           []string
 	ClusterEnabled bool
 	Password       string
+	// Context bounds the initial connectivity check (Ping). When nil, context.Background()
+	// is used. Callers that must guarantee NewRedisCloser returns within a deadline (e.g. a
+	// bounded provider startup) should pass a context with a timeout so a black-holed broker
+	// cannot block the ping for the full go-redis dial timeout.
+	Context context.Context
```
dkorittki, line 28: Instead of having this here I would put it as the first parameter on `NewRedisCloser`

router/core/graph_server.go
```diff
@@ -2310,16 +2310,39 @@ func providersActionWithTimeout(ctx context.Context, providers []datasource.Prov
 	providersGroup := new(errgroup.Group)
 	for _, provider := range providers {
 		providersGroup.Go(func() error {
+			// Each provider is bounded by its own timer. A single shared timer would only
+			// ever deliver its fire to one goroutine, so multiple slow/unreachable providers
+			// would block the others forever and hang startup.
+			timer := time.NewTimer(timeout)
+			defer timer.Stop()
```
dkorittki: I think we should redesign this and make it more idiomatic. I propose to remove `providersActionWithTimeout()` and modify each providers `Startup()` method to accept contexts as a parameter and handle cancellation themselves based on that. `startPubsubProviders()` derives a timeout based `ctx` and calls `Startup(ctx)` methods. In "strict mode" `startPubsubProviders()` bubbles up an error when providers can't be started in time. In "leniant mode" it's the same up until this point. But instead of bubbling up an error it spawns a goroutine, which continues trying to start the providers forever.

This has the advantage that we can keep any retry logic out of the providers themselves and keep this inside `startPubsubProviders()`. It allows the same startup behaviour for all providers types instead of reyling on library-specific retry mechanisms and trying to cope with it if some can't do it. At the same time we move the timeout logic into providers `Startup()` and move away from timers.

dkorittki (changes requested):
Not completely finished but before digging further I think it makes sense to discuss some broad ideas first

router/pkg/pubsub/pubsub.go
```diff
@@ -66,6 +66,13 @@ func BuildProvidersAndDataSources(
 	if store == nil {
 		store = metric.NewNoopStreamMetricStore()
 	}
+	if logger == nil {
+		logger = zap.NewNop()
+	}
+
+	if config.SkipUnavailableProviders {
+		logger.Warn("EDFS lenient mode is enabled (events.skip_unavailable_providers=true): the router will start even if an event provider referenced by the execution config is undefined or unreachable, disabling only the affected fields")
```
dkorittki: I don't think the warning is necessary. When the router can't connect to the provider we log errors anyway. If we can connect there is no problem so we don't need this warning, too.

router/core/graph_server.go
dkorittki: We will postpone this in favor to get a working solution first

router/core/graph_server.go
```diff
@@ -2310,16 +2310,39 @@ func providersActionWithTimeout(ctx context.Context, providers []datasource.Prov
...
+			// resilient client that reconnects in the background, so those fields recover
+			// without a restart once the broker is reachable. When the cause is a
+			// configuration error (e.g. an invalid URL) the provider cannot recover until the
+			// configuration is fixed; the underlying cause is attached as the error field.
+			s.logger.Error("EDFS provider could not be started at startup; the router will keep running and the fields backed by this provider are temporarily unavailable. An unreachable broker reconnects and recovers automatically without a restart; see the error for the cause",
+				zap.String("provider_id", provider.ID()),
+				zap.String("provider_type", provider.TypeID()),
+				zap.Error(actionErr),
+			)
+			return nil
```
dkorittki: The lates pull from main broke your code. The timer is not needed anymore, got replaced by `timeoutCtx`. Here's a suggestion from me. After that this method needs to be made again a method of `graphServer` and `continueOnError` parameter needs to be readded to method signature.

```suggestion
			actionDone := make(chan error, 1)

			go func() {
				actionDone <- action(timeoutCtx, provider)
			}()

			var err error
			select {
			case err = <-actionDone:
			case <-timeoutCtx.Done():
				err = errors.New(timeoutMessage)
			}

			if err == nil || !continueOnError {
				return err
			}

			s.logger.Error("EDFS provider could not be started at startup; the router will keep running and the fields backed by this provider are temporarily unavailable. An unreachable broker reconnects and recovers automatically without a restart; see the error for the cause",
				zap.String("provider_id", provider.ID()),
				zap.String("provider_type", provider.TypeID()),
				zap.Error(err),
			)
			return nil
```

router/internal/rediscloser/rediscloser.go
alepane21: Haven't changed that because it's not required and not needed in places outside of this. Redisclose is also used in PQ and potentially other places, didn't wanted to change them.

router/core/graph_server.go
```diff
@@ -2333,23 +2338,46 @@ func (s *graphServer) Shutdown(ctx context.Context) error {
...
-				return errors.New(timeoutMessage)
+				err = errors.New(timeoutMessage)
 			}
+
+			if err == nil || !continueOnError {
+				return err
+			}
+
+			if l != nil {
+				l.Error("EDFS provider could not be started at startup; the router will keep running and the fields backed by this provider are temporarily unavailable. An unreachable broker reconnects and recovers automatically without a restart; see the error for the cause",
```
dkorittki: I'd probably consider this a warning level log rather than error level but you can decide this at your own discretion

dkorittki (approved):

router/core/graph_server.go
alepane21: You are right, better to be a warn

PR #2998, opened 2026-06-23
docs(rfc): disallow inline argument values (ENG-9586)

jensneuse:
## Summary

RFC for an optional Cosmo Router policy that detects — and optionally rejects — GraphQL operations that carry hardcoded **inline argument values** instead of variables.

- Customer ask: ticket #2498 / internal ENG-9586.
- Motivation: inline literals explode unique-query cardinality in downstream query stores; forcing variables collapses them.

This PR adds the design document only (`rfc/disallow-inline-args/RFC.md`); no router code yet.

## Design highlights

- **Detection during `Parse()`** in graphql-go-tools `parseArgumentList` — no extra AST walk, sees the operation before any normalization, and rejects before any internal cache is written.
- **Comprehensive flagging**: field arguments, directive arguments (`@skip`/`@include` and any directive), introspection-field arguments, and arguments inside selections that `@skip`/`@include` would normalize out.
- **Three modes**: `off` (default) / `enabled-non-enforcing` (structured log + `extensions` migration hint on a successful response) / `enabled-enforcing` (reject with configurable HTTP status, error code, and message).
- **Persisted operations** are exempt by default (stored server-side, intentional); opt in via `include_persisted_operations`.
- Migration path `off → log-only → enforce`, with structured logs and an OTEL per-client attribute to find affected clients before enforcing.

## Status

Draft — design review only. Open questions are tracked in §9 (subscriptions exemption, nested-path granularity, per-argument allowlist).

## Summary by CodeRabbit

* **Documentation**
  * Added an RFC describing a new router security policy that detects GraphQL operations with hardcoded inline argument literal values (instead of variables). Includes configurable modes for logging/annotation versus strict rejection, details on covered argument contexts, and an end-to-end testing plan.

### DIFF
--- a/rfc/disallow-inline-args/RFC.md
+++ b/rfc/disallow-inline-args/RFC.md
@@ -1,0 +1,647 @@
+# RFC: Disallow Inline Argument Values (Force Variables)
+
+| Field | Value |
+| --- | --- |
+| Title | Disallow Inline Argument Values (Force Variables) |
+| Status | Draft |
+| Date | 2026-06-17 |
+| Internal ID | ENG-9586 |
+| Ticket | #2498 |
+| Author | _TBD (placeholder)_ |
+
+## 1. Summary
+
+This RFC proposes an optional Cosmo Router security policy that detects,
+and optionally rejects,
+GraphQL operations that carry hardcoded inline argument values instead of variables.
+
+The canonical example the policy targets:
+
+```graphql
+query GetUserById {
+  userById(userId: "12345") {
+    loginName
+  }
+}
+```
+
+Under the policy this operation can be flagged or rejected,
+and the client is guided to rewrite it as:
+
+```graphql
+query GetUserById($userId: ID!) {
+  userById(userId: $userId) {
+    loginName
+  }
+}
+```
+
+The feature ships in three modes: `off` (default), `enabled-non-enforcing` (log plus an extensions hint on a successful response), and `enabled-enforcing` (reject with a GraphQL error and a configurable HTTP status code).
+
+## 2. Motivation
+
+The use case driving this request:
+
+> A team validates unique queries from a runtime query store against draft supergraph schemas.
+> Too many unique queries (caused by inline values) overwhelm their tooling.
+> They want the router to optionally force clients to use variables and prevent operations with hardcoded inline argument values,
+> and optionally surface the normalized form / guidance.
+
+The core problem is cardinality.
+Every distinct inline literal produces a distinct raw operation string,
+so `userById(userId: "12345")` and `userById(userId: "12346")` look like two unique queries to a downstream query store,
+even though they are the same operation shape.
+When the operations are parameterized with variables,
+they collapse to a single unique query,
+which is what such validation tooling expects.
+
+### Why detection must run on the ORIGINAL, un-normalized operation
+
+The policy must flag an inline argument **whenever the client sent one**, with no exceptions for
+directive arguments or introspection, and even when normalization would later erase the evidence.
+Normalization mutates and prunes the operation in three ways that destroy that evidence:
+
+- **Variable extraction rewrites literals into variables.**
+  `variablesExtractionVisitor.EnterArgument` sets `arg.Value.Kind = ast.ValueKindVariable`
+  (`variables_extraction.go` lines 79-80 and 119-120).
+  After it runs, the literals look like variables.
+- **`@skip` / `@include` normalization removes whole nodes.**
+  `directiveIncludeSkipKeepNodes` deletes fields whose `@skip(if: true)` / `@include(if: false)`
+  resolves at normalization time, via `RemoveNodeFromSelectionSetNode`
+  (`directive_include_skip.go:128-135`).
+  If a removed field carried an inline argument, that argument is gone before any later pass sees it.
+- **Fragment inlining and dead-fragment removal** reshape where arguments live.
+
+The requirement is explicit: we want to flag the inline argument **even if** a `@skip`/`@include`
+would normalize that part of the query out, **even if** it is a directive argument
+(`@include(if: true)` itself counts), and **even if** the operation is an introspection query
+(`__type(name: "User")` counts).
+The only document where all of these are simultaneously visible and intact is the operation
+exactly as parsed, before any normalization stage runs.
+
+Therefore detection runs during parsing itself (`operationKit.Parse()`, `graphql_prehandler.go:687`),
+which is the first and only pass that sees the operation fully intact,
+before any normalization stage and before the router writes any internal cache.
+An enforce-mode rejection then short-circuits without polluting a single cache — see § 4.1 and § 4.3.
+
+## 3. Goals / Non-Goals
+
+### Goals
+
+- Detect inline argument values on the original operation exactly as the client sent it.
+- Flag inline arguments **comprehensively**: field arguments, directive arguments (`@skip`/`@include` and any other directive), introspection-field arguments, and arguments inside selections that `@skip`/`@include` would normalize out.
+- Provide three modes: off, log-only (non-enforcing), and enforce.
+- Make HTTP status code, GraphQL error code, and error message configurable.
+- In non-enforcing mode, return a successful response with an extensions annotation that lists the offending paths so clients can migrate.
+- Emit structured logs and an OTEL attribute so operators can discover affected clients before enforcing.
+- Allow persisted operations by default (they are stored server-side and therefore intentional), with a config flag to apply the policy to persisted operations as well.
+
+### Non-Goals
+
+- Rewriting the client's operation for them (Cosmo already normalizes internally; this policy does not change the executed operation).
+- Changing the internal normalization behavior or the executed plan.
+- Detecting inline values inside variable-definition default values (`$x: Int = 5` is a definition default, not an argument the client passed to a field or directive — out of scope; see Open Questions).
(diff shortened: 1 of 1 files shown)


rfc/disallow-inline-args/RFC.md
```diff
@@ -0,0 +1,647 @@
...
+   on a different path (variable-definition parsing), not through `parseArgumentList` — so the one
+   intended exclusion requires no special-casing.
+
+```go
+type InlineArgumentInfo struct {
+    ArgumentName string        // e.g. "userId" — available as name.Literal at parser.go:430
+    ValueKind    ast.ValueKind // e.g. ValueKindString — from value.Kind
+    InDirective  string        // directive name if this is a directive argument, else ""
+    Position     position.Position
+}
```
Noroth: I'm not sure if it is a good idea to include this feature in the parser logic. I think we can just include the extra step in one of our validation walkers. Otherwise all parsing will additionally have to check each time if `reportInlineArguments` is set to true.

I'm personally against the recommended approach here

rfc/disallow-inline-args/RFC.md
```diff
@@ -0,0 +1,647 @@
...
+name supplied by the calling parse frame; richer nested paths are a later refinement (§ 9).
+
+This is additive to graphql-go-tools and changes no existing parse or normalization behavior when the flag is off.
+
+#### Comparison of approaches
+
+| # | Approach | Flags directive / introspection / skip-removed args? | Reject before 1st cache write? | Extra walk? | Verdict |
+| --- | --- | --- | --- | --- | --- |
+| A | Detect during parse in `parseArgumentList` (RECOMMENDED) | **Yes** — sees the un-normalized operation | **Yes** — runs during `Parse()`, before `NormalizeOperation` and any cache | **No** — rides the parse already performed | Chosen |
+| B | Dedicated argument visitor over the parsed op, after `Parse()` | **Yes** — also sees the un-normalized operation | Yes | Yes — one extra argument-only walk | Fallback (see § 11) |
```
Noroth: No sure if we need a dedicated visitor here, I think we already have some validation in place that could potentially be re-used? @devsergiy

rfc/disallow-inline-args/RFC.md
```diff
@@ -0,0 +1,647 @@
...
+          {
+            "path": "userById.userId",
+            "argument": "userId",
+            "valueKind": "String"
+          }
+        ]
+      }
+    }
+  ]
+}
```
Noroth, line 408: Something seems off here. We seem to have a path which is dot separated and the last element is the actual argument. not the enclosing field. while we have "argument" as well

so basically we have some kind of duplication here.
"path": "field.field.argument"
"argument": "argument"

If we use the proposed data structure as a reference to create errors we won't have the information about either path or enclosing field

```
type InlineArgumentInfo struct {
    ArgumentName string        // e.g. "userId" — available as name.Literal at parser.go:430
    ValueKind    ast.ValueKind // e.g. ValueKindString — from value.Kind
    InDirective  string        // directive name if this is a directive argument, else ""
    Position     position.Position
}
```

rfc/disallow-inline-args/RFC.md
```diff
@@ -0,0 +1,647 @@
...
+   It naturally **excludes** variable-definition defaults (`$x: Int = 5`), because those are parsed
+   on a different path (variable-definition parsing), not through `parseArgumentList` — so the one
+   intended exclusion requires no special-casing.
+
+```go
+type InlineArgumentInfo struct {
+    ArgumentName string        // e.g. "userId" — available as name.Literal at parser.go:430
+    ValueKind    ast.ValueKind // e.g. ValueKindString — from value.Kind
+    InDirective  string        // directive name if this is a directive argument, else ""
+    Position     position.Position
```
Noroth: Positions might not be always correct, right? We sometimes have multiple stages of parsing where we also append to operation documents or rewrite certain things (e.g. astnormalization). 

We might end up reparsing again for different visitors so the position might be different to the position in the original operation that was sent to the router.

rfc/disallow-inline-args/RFC.md
```diff
@@ -0,0 +1,647 @@
...
+
+Under the policy this operation can be flagged or rejected,
+and the client is guided to rewrite it as:
+
+```graphql
+query GetUserById($userId: ID!) {
+  userById(userId: $userId) {
+    loginName
+  }
+}
```
Noroth, line 36: The idea is to not have too many distinct operations, but technically we could also have different variable names. I suppose this would be up to the user to use the feature properly? Or do we also want to flag use of different variable names for the same operations?

```graphql
query GetUserById($userId1: ID!) {
  userById(userId: $userId2) {
    loginName
  }
}

query GetUserById($userId2: ID!) {
  userById(userId: $userId2) {
    loginName
  }
}
```

rfc/disallow-inline-args/RFC.md
jensneuse: I think we can do this much simpler. After parsing, we simply look into the AST and verify if we have any inline arguments that are not variables. If that's the case we flag this as error and stop the rest of the request pipeline. For that we don't need any visitors but just a single lookup in the request pipeline that we otherwise don't do. So the operation by itself is rather cheap and in case of disabled check a no op.

rfc/disallow-inline-args/RFC.md
jensneuse: we cannot prevent this

rfc/disallow-inline-args/RFC.md
```diff
@@ -101,99 +102,91 @@ An enforce-mode rejection then short-circuits without polluting a single cache 
...
-    InDirective  string        // directive name if this is a directive argument, else ""
-    Position     position.Position
+// in the router prehandler, immediately after operationKit.Parse(), only when the feature is enabled
+for ref := range o.kit.doc.Arguments {
+    if o.kit.doc.Arguments[ref].Value.Kind != ast.ValueKindVariable {
+        // inline argument found
+        // enforce mode: record it and stop the pipeline here
+        // log mode: collect and continue
+    }
 }
```
Noroth, line 142: Imagine a more complex operation. Can we still provide good UX/DX here without knowing where the argument is coming from?

```graphql
query GetDashboard(
  $tenantId: ID!
  $userId: ID!
  $limit: Int!
  $includeArchived: Boolean!
) @cacheControl(maxAge: 60, scope: "PRIVATE")
  @trace(enabled: true, sampleRate: 0.25)
{
  tenant(id: $tenantId)
    @auth(role: "admin")
    @cacheControl(maxAge: 30, scope: "PUBLIC")
  {
    id

    users(
      filter: { status: "ACTIVE", role: "admin" }
      limit: $limit
      orderBy: { field: "createdAt", direction: "DESC" }
    ) @include(if: true)
      @rateLimit(limit: 100, window: "1m")
    {
      id
      name

      posts(
        filter: { published: true, tag: "graphql" }
        limit: 10
      ) @skip(if: false)
      {
        id
        title

        comments(
          filter: { flagged: false }
          limit: 5
        ) @include(if: $includeArchived)
        {
          id
          text
        }
      }
    }

    auditLogs(
      filter: { actorId: $userId, action: "LOGIN" }
      limit: 50
    ) @cacheControl(maxAge: 5, scope: "PRIVATE")
    {
      id
      action
      createdAt
    }
  }
}
```

We would get something like:

```json
{
  "extensions": {
    "inlineArguments": {
      "code": "INLINE_ARGUMENT_VALUES_NOT_ALLOWED",
      "message": "Inline argument values are not allowed. Use variables instead.",
      "arguments": [
        {
          "argument": "maxAge",
          "valueKind": "Int",
          "line": 5,
          "column": 30
        },
        {
          "argument": "scope",
          "valueKind": "String",
          "line": 5,
          "column": 43
        },
        {
          "argument": "role",
          "valueKind": "String",
          "line": 9,
          "column": 17
        },
        {
          "argument": "filter",
          "valueKind": "Object",
          "line": 14,
          "column": 7
        },
        {
          "argument": "limit",
          "valueKind": "Int",
          "line": 15,
          "column": 7
        },
        {
          "argument": "filter",
          "valueKind": "Object",
          "line": 24,
          "column": 9
        },
        {
          "argument": "limit",
          "valueKind": "Int",
          "line": 25,
          "column": 9
        },
        {
          "argument": "filter",
          "valueKind": "Object",
          "line": 34,
          "column": 11
        },
        {
          "argument": "limit",
          "valueKind": "Int",
          "line": 35,
          "column": 11
        }
      ]
    }
  }
}
```

I think something beter could be:

Field:
```json
{
  "path": "tenant.users.posts",
  "kind": "Field",
  "name": "posts",
  "argument": "filter",
  "valueKind": "Object",
  "line": 24,
  "column": 9
}
```
Directive
```json
{
  "path": "tenant.users.posts",
  "kind": "Directive",
  "name": "skip",
  "argument": "if",
  "valueKind": "Boolean",
  "line": 27,
  "column": 15
}
```

gausie:
The RFC specifies different shapes for the two modes, is that intentional?

In enforce mode, `extensions.inlineArguments` is a flat array (inside the GraphQL error object), with `code` and `message` at the `extensions` level. In non-enforce mode, `extensions.inlineArguments` is an object with `code`, `message`, and a nested `arguments` array.

Clients reading the argument list need to branch on mode: `extensions.inlineArguments` in enforce vs `extensions.inlineArguments.arguments` in non-enforce. If the difference is intentional, worth documenting why. If not, which shape should both converge to?

PR #3004, opened 2026-06-24
fix(router): clear slowplancache entries on Close

arutkowski00:
## Problem

When the router hot-reloads execution config (CDN or manifest watcher), memory usage climbs and **does not return to the pre-reload baseline**. After several reloads, pods can hit memory limits.

One contributor: `slowplancache.Close()` stopped the background goroutine but **left entries in the underlying `sync.Map`**. Those entries can hold query plans that reference schema AST nodes, keeping the previous graph generation alive until the `Cache` struct itself is garbage-collected.

## Fix

Clear all entries from the map when `Close()` runs, and reset the size counter.

## Validation

- [x] Unit test `TestCache_CloseReleasesEntries` — map is empty after `Close()`
- [ ] CI

This is the first commit in a larger fix set; see #3005 for the remaining router reload changes.

Support ticket: monday.com #3221

## Summary by CodeRabbit

* **Bug Fixes**
  * Closing the cache now fully clears stored entries and resets its size, helping release memory more reliably.
* **Tests**
  * Added coverage to verify that closing the cache removes all retained entries and leaves it empty.

### DIFF
--- a/router/pkg/slowplancache/slow_plan_cache.go
+++ b/router/pkg/slowplancache/slow_plan_cache.go
@@ -222,5 +222,14 @@ func (c *Cache[V]) Close() {
 		// This downside is also there in ristretto (if set is called concurrently)
 		// it is even documented in the ristretto code as a comment
 		close(c.writeCh)
+
+		// Drop cached entries so Close releases references to stored values
+		// (e.g. query plans holding schema AST pointers) before the Cache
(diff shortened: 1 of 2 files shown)


arutkowski00:
Consolidating into #3005 — slowplancache fix is already the first commit there.

PR #3005, opened 2026-06-24
fix(router): release memory on config hot-reload

arutkowski00:
## Problem

When the router hot-reloads execution config (CDN poller or manifest file watcher), it builds a new `graphServer` and federation schema AST (~200 MB transient spike is expected). The bug is that **settled heap after GC does not return to the pre-reload baseline** — each reload permanently retains memory from the previous generation.

We reproduced this on staging with forced-GC heap profiles (`/debug/pprof/heap?gc=1`) before and after repeated config reloads.

## Root causes addressed

| Area | Issue | Fix in this PR |
|------|-------|----------------|
| slowplancache | Entries survive `Close()`, pinning cached plans | Clear map on `Close()` |
| Cached plans | `planWithMetaData` stored full `schemaDocument` (~200 MB), never read | Remove the field |
| Hot-reload path | `OnRouterConfigReload()` not called before building new graph server | Call it at start of `newServer()` |
| Plan cache shutdown | `OnEvict` during intentional `Close()` migrated entries into slowplancache | Disable OnEvict during mux shutdown |
| Plan cache shutdown | Cache closed before active WebSocket subscriptions drained | Drain WS connections first |
| Executor | Old `RouterSchema` / `ClientSchema` / `PlanConfig` retained after mux shutdown | `Executor.Close()` nils refs at end of shutdown |
| Upstream subscriptions | One subscription client per subgraph → many idle `pingLoop` goroutines | Shared client; noop client when subscriptions unused |
| graphMux caches | Ristretto plan/operation caches closed but not nilled on shut-down mux | `releaseOperationCaches()` — Close + nil all caches |
| Execution config proto | Decoded proto strings retained after reload | Hash `mapper.json`; skip unchanged reloads; `proto.Reset(old)` after swap |
| Profiling | Missing `/debug/pprof/heap` route | Register heap/allocs/goroutine handlers |

## Measured impact

**Production staging** (fresh pod, 3 config reloads, forced-GC heap via pprof):

| Metric | Before these fixes | After partial fixes (Executor.Close, etc.) | After full PR |
|--------|-------------------|---------------------------------------------|---------------|
| Duplicate schema AST (`AddType`) | ~2× retained | +3 MB (flat) | (pending deploy) |
| RSS growth per reload | ~260 Mi | ~96 Mi | (pending deploy) |
| inuse growth per reload | ~200+ MB | ~68 MB | (pending deploy) |

**Local dev** (repeated manifest reloads against a production config snapshot, forced-GC heap, 2 reloads):

| Metric | Before graphMux nil + proto release | After |
|--------|-------------------------------------|-------|
| inuse growth per reload | ~74 MB | ~flat (within noise) |

We also tried passing `Changes`/`Hashes` on the manifest watcher to reuse unchanged graph muxes — it did not improve our reload benchmark and is **not included**.

## Test plan

- [x] `go test ./router/core/...` (executor, noop subscription client)
- [x] Local repeated-reload heap comparison (forced GC before/after)
- [ ] Staging deploy + pprof validation
- [ ] CI

Happy to share heap profiles (`heap?gc=1` before/after reload diffs) on request.

Support ticket: monday.com #3221

## Summary by CodeRabbit

* **New Features**
  * Added broader profiling endpoints for heap, allocations, and goroutine inspection.
  * Improved support for websocket and subscription handling, including safer fallback behavior when upstream subscriptions aren’t needed.

* **Bug Fixes**
  * Fixed shutdown and cleanup so cached data, websocket connections, and planning state are released more reliably.
  * Reduced unnecessary reloads by skipping configuration refreshes when manifest content hasn’t changed.
  * Improved hot-reload behavior to better preserve and update router state.

### DIFF
--- a/router/cmd/main.go
+++ b/router/cmd/main.go
@@ -46,19 +46,6 @@ func Main() {
 	// Parse flags before calling profile.Start(), since it may add flags
 	flag.Parse()
 
-	// Re-read profiling env after flag.Parse() — flag defaults are captured at package
-	// init, before embedders (e.g. platform-api-cosmo-router) can Setenv in main().
-	if *pprofListenAddr == "" {
-		if addr := os.Getenv("PPROF_ADDR"); addr != "" {
-			*pprofListenAddr = addr
-		}
-	}
-	if *pyroscopeAddr == "" {
-		if addr := os.Getenv("PYROSCOPE_ADDR"); addr != "" {
-			*pyroscopeAddr = addr
-		}
-	}
-
 	if *help {
 		flag.PrintDefaults()
 		os.Exit(0)
--- a/router/cmd/main.go
+++ b/router/cmd/main.go
@@ -46,6 +46,19 @@ func Main() {
 	// Parse flags before calling profile.Start(), since it may add flags
 	flag.Parse()
 
+	// Re-read profiling env after flag.Parse() — flag defaults are captured at package
+	// init, before embedders (e.g. platform-api-cosmo-router) can Setenv in main().
+	if *pprofListenAddr == "" {
+		if addr := os.Getenv("PPROF_ADDR"); addr != "" {
+			*pprofListenAddr = addr
+		}
+	}
+	if *pyroscopeAddr == "" {
+		if addr := os.Getenv("PYROSCOPE_ADDR"); addr != "" {
+			*pyroscopeAddr = addr
+		}
+	}
+
 	if *help {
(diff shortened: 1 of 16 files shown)


arutkowski00:
Replaced by 8 focused bugfix PRs (one fix each) per review feedback. Deferred: noop/shared subscription client, mapper hash skip, pprof routes.

PR #3006, opened 2026-06-24
fix(router): clear slowplancache entries on Close

arutkowski00:
## Problem
After `slowplancache.Close()`, entries remained in the underlying `sync.Map`, keeping references to cached query plans (and any schema pointers they hold) until the Cache struct was garbage-collected.

## Fix
Clear all entries and reset the size counter when `Close()` runs.

## Test plan
- [x] `TestCache_CloseReleasesEntries`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cache shutdown so stored entries are fully cleared when the cache is closed.
  * Ensured cached data is released immediately, with internal cache size reset to zero.

* **Tests**
  * Added coverage to verify that closing the cache removes all stored entries and leaves no retained size.


dkorittki:

PR #3007, opened 2026-06-24
fix(router): stop pinning schema AST in cached plan entries

arutkowski00:
## Problem
`planWithMetaData` stored `schemaDocument` (full router schema AST, ~200MB) in every Ristretto plan cache entry but never read it. After config reload, cached plans pinned the previous schema generation.

## Fix
Remove the unused `schemaDocument` field.

## Test plan

## Summary by CodeRabbit

* **Refactor**
  * Streamlined internal request planning by removing an unused piece of stored metadata.
  * Kept the visible operation handling flow the same, with no expected change in user behavior.

### DIFF
--- a/router/core/operation_planner.go
+++ b/router/core/operation_planner.go
@@ -18,8 +18,8 @@ import (
 )
 
 type planWithMetaData struct {
-	preparedPlan                      plan.Plan
-	operationDocument, schemaDocument *ast.Document
+	preparedPlan       plan.Plan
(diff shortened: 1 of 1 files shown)


dkorittki (changes requested):
Hey @arutkowski00, thanks for the contribution. This is safe to merge. Only needs a gofmt on router/core/operation_planner.go. Would do it myself but can't push to your branch. Can you do it?

dkorittki:

PR #3008, opened 2026-06-24
fix(router): call OnRouterConfigReload before config hot-reload

arutkowski00:
## Problem
Supervisor restarts call `OnRouterConfigReload()` to extract slow-plan cache entries before replacing the graph server. CDN and manifest hot-reload paths skipped this hook, leaving stale cache references on the old graphMux.

## Fix
Call `OnRouterConfigReload()` at the start of `newServer()`.

## Test plan

### DIFF
--- a/router/core/router.go
+++ b/router/core/router.go
@@ -627,6 +627,10 @@ func (r *Router) serverTLSConfig() (*tls.Config, error) {
 
 // newGraphServer creates a new server.
 func (r *Router) newServer(ctx context.Context, response *routerconfig.Response) error {
+	// Extract slow-plan cache entries before building the new graph server, which
+	// overwrites ReloadPersistentState cache references and before the old graphMux shuts down.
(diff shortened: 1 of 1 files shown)


dkorittki:
Good catch! Unfortunately the way it is right now it adds a regression on the supervisor Reload path. The supervisor already calls `OnRouterConfigReload()` on the old router before shutting it down (supervisor.go:162). The new router's `newServer()` call will now hit `OnRouterConfigReload()` a second time with the map already in snapshot form. After the first call the values are already `[]*nodev1.Operation`, so the second call's type assertion fails for every entry, nothing gets added to `fallbackMap` and `queriesForFeatureFlag` is overwritten with an empty map.

A quick fix would be to modify `extractQueriesAndOverridePlanCache` to preserve entries of type `[]*nodev1.Operation` but I need to check if there are better ways of doing it. Will come back to this.

PR #3009, opened 2026-06-24
fix(router): skip plan cache OnEvict during graphMux shutdown

arutkowski00:
## Problem
`planCache.Close()` invokes `OnEvict` for every entry, migrating them into `slowplancache` immediately before both caches are torn down during mux shutdown.

## Fix
Disable `OnEvict` during intentional mux shutdown and wait for pending slow-cache writes before closing plan caches.

## Test plan

### DIFF
--- a/router/core/graph_server.go
+++ b/router/core/graph_server.go
@@ -681,6 +681,8 @@ type graphMux struct {
 	mux    *chi.Mux
 	reused atomic.Bool
 
+	planCacheOnEvictEnabled atomic.Bool
+
 	planCache                   *ristretto.Cache[uint64, *planWithMetaData]
 	planFallbackCache           *slowplancache.Cache[*planWithMetaData]
 	persistedOperationCache     *ristretto.Cache[uint64, NormalizationCacheEntry]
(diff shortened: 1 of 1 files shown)


router/core/graph_server.go
```diff
@@ -961,6 +967,11 @@ func (s *graphMux) Shutdown(ctx context.Context) error {
 	// cancel the graph muxes context to close its resources like websocket connections, resolvers, etc.
 	s.cancel()
 
+	// ristretto Close() invokes OnEvict for every entry; skip slow-cache migration during shutdown.
+	s.planCacheOnEvictEnabled.Store(false)
+	if s.planFallbackCache != nil {
+		s.planFallbackCache.Wait()
+	}
```
dkorittki, line 974: Actually I don't think we need to wait for pending writes to the cache at this point because we are about to delete the cache anyways.

dkorittki (changes requested):
Looks good to me generally. It does not make sense to call the OnEvict hook to migrate lots of slow plan caches to the fallback cache when the new graph server is already created (and thus has created its plancache from oldplancache already). It's unnecessary work. I just added one comment which should be handled before I can approve.

dkorittki:

PR #3011, opened 2026-06-24
fix(router): release executor schema refs on graphMux shutdown

arutkowski00:
## Problem
After config hot-reload, the shut-down `Executor` retained federation/client schema AST and plan config references, keeping the previous graph generation alive.

## Fix
Add `Executor.Close()` to nil schema, plan config, and resolver references. Call it from `graphMux.Shutdown()` after plan caches are closed.

## Test plan
- [x] `TestExecutorCloseReleasesSchemaReferences`

### DIFF
--- a/router/core/executor.go
+++ b/router/core/executor.go
@@ -51,6 +51,19 @@ type Executor struct {
 	TrackUsageInfo  bool
 }
 
+// Close releases schema and planner references held by the executor so a replaced
+// graph mux can be garbage-collected after config reload.
+func (e *Executor) Close() {
+	if e == nil {
+		return
+	}
+	e.ClientSchema = nil
+	e.RouterSchema = nil
+	e.PlanConfig = plan.Configuration{}
+	e.RenameTypeNames = nil
+	e.Resolver = nil
+}
+
 type ExecutorBuildOptions struct {
 	EngineConfig                   *nodev1.EngineConfiguration
 	Subgraphs                      []*nodev1.Subgraph
--- a/router/core/executor_test.go
+++ b/router/core/executor_test.go
@@ -1,0 +1,37 @@
+package core
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"github.com/wundergraph/graphql-go-tools/v2/pkg/ast"
+	"github.com/wundergraph/graphql-go-tools/v2/pkg/engine/plan"
+)
+
+func TestExecutorCloseReleasesSchemaReferences(t *testing.T) {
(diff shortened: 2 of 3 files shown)


router/core/executor.go
```diff
@@ -51,6 +51,19 @@ type Executor struct {
...
+// graph mux can be garbage-collected after config reload.
+func (e *Executor) Close() {
+	if e == nil {
+		return
+	}
+	e.ClientSchema = nil
+	e.RouterSchema = nil
+	e.PlanConfig = plan.Configuration{}
+	e.RenameTypeNames = nil
+	e.Resolver = nil
```
dkorittki, line 64: The problem is websockets use this Executor and they are closed async. When this function is called websocket connections are still alive (got signaled to close but might take a moment) and might use the Executor, which has just nilled its components. It can cause nil panics or unwanted errors at least. I need to check this in more depth and how to solve it.

dkorittki (changes requested):

router/core/executor.go
dkorittki, line 64:
```suggestion
```

I think we can exclude this one. Compared to the other fields it's not heavy and reduces the risk surface. Also this executor instance the pointer points to is referenced by other pointers in other places as well (the WS handler/graphql handler), which is still active when we are here. So GC can't clean it up immediately anyways.

EDIT: Also fixed by #3010 but review there still pending

dkorittki:
I did a deeper analysis and found other places this could lead to nil panics. A good risk reduction is described in https://github.com/wundergraph/cosmo/pull/3011#discussion_r3489882093 but this alone is not sufficient. There are still races:

#1: In-flight request handling can now panic. Any request on the old graph mux unlucky enough to be in parse/normalize/validate/plan when we are here can hit it. Can happen on a busy server. These panics are recovered but technically this should not be hard to fix.

#2: When an active WS connection on the old graph mux initiates a new subscription it would panic, no recovery. Best thing would be the WS handler does not accept that anymore since we are about to close it anyways. At least we should have a recovery in place until we come up with a long term solution here.

#3: Persisted query cache rewarming can nil panic and crash the router. Rewarming happens on PQL manifest changes. If we are warming the cache (can take seconds) and then we are here the warming callback parses and plans its stored queries and that accesses these fields. Not so likely to happen but still it can crash the router. At the very least we should have a recovery until we have a long term solution.

So overall it seems to possible if we omit niling `e.Resolver`, which doesn't make sense anyway. I'll probably cherry-pick you commit into a new pull request and work on solving these issues myself

PR #3044, opened 2026-07-02
feat(router): disallow inline argument values (ENG-9586)

gausie:
Implements the `security.disallow_inline_arguments` router policy: `warn` mode logs and annotates responses with `extensions.inlineArguments`, `enforce` mode rejects operations using inline argument values. Applies to both HTTP and WebSocket transports; persisted operations are exempt by default.

RFC: #2998

## Summary by CodeRabbit

* **New Features**
  * Added `security.disallow_inline_arguments` to detect hardcoded inline GraphQL argument values on both HTTP and WebSocket.
  * Supports `off`, `warn`, and `enforce` modes, including configurable enforcement HTTP status and GraphQL error `code`/`message`.
* **Bug Fixes**
  * Enforce/warn behavior is now consistent across HTTP and WebSocket, with persisted operations exempt by default (configurable to include).
* **Documentation**
  * Added dedicated documentation and updated the security configuration reference/navigation.
* **Tests**
  * Added coverage for detection, `extensions.inlineArguments`, telemetry (`wg.operation.inline_arguments.count`), and WebSocket response/error behavior.


router/core/websocket.go
```diff
@@ -353,6 +356,7 @@ func (h *WebsocketHandler) handleUpgradeRequest(w http.ResponseWriter, r *http.R
 		ForwardInitialPayload:        h.config.ForwardInitialPayload,
 		OperationProcessor:           h.operationProcessor,
 		OperationBlocker:             h.operationBlocker,
+		InlineArgumentsChecker:       h.inlineArgumentsChecker,
```
gausie: I added this because it feels technically correct but... do we actually want to do this check for subscriptions?

PR #3045, opened 2026-07-03
fix: router @include variable fails with "not provided" on repeated requests (ENG-9772)

gausie:
## Motivation and Context

Fixes ENG-9772: an operation whose Boolean variable is used **both** as a field argument **and** in a `@skip`/`@include` directive succeeds on the first request, then fails on every subsequent identical request with:

```
Variable "$flag" of required type "Boolean!" was not provided.
```

### Root cause

- On a normalization cache **miss**, the normalizer only strips a skip/include variable from the operation and the request variables when it is not used anywhere else. A dual-use variable (also a field argument) stays declared in the cached normalized representation.
- On a cache **hit**, `normalizeNonPersistedOperation` unconditionally deleted **every** skip/include variable from the request variables. The cached operation still declares the dual-use variable as `Boolean!`, so validation fails.

The first request misses the cache and succeeds; ristretto's `Set` is buffered, so the entry lands shortly after and every subsequent identical request hits the cache and fails until the router restarts.

### Fix

On cache miss, record which skip/include variables normalization actually removed (i.e. are no longer declared in the normalized operation) and store that list in the `NormalizationCacheEntry`. On cache hit, delete only those from the request variables, keeping dual-use variables intact.

The persisted-operation cache path is not affected: its cache key includes the skip/include variable *values*, so a dual-use variable stays consistent between key and entry.

## Checklist

- [x] I have written a test that reproduces the bug (`TestNormalizationCacheHitWithIncludeVariableAlsoUsedAsArgument` fails without the fix)
- [x] Integration test waits deterministically for the first cache HIT via the debug normalization cache response header

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of conditional GraphQL variables so requests using the same value both as an argument and in an `@include` directive return consistent results from cache.
  * Fixed cache behavior across repeated requests, persisted operations, and cache warmup scenarios so responses stay correct after a cache hit.


router-tests/operations/normalization_cache_skip_include_test.go
```diff
@@ -0,0 +1,86 @@
...
+			})
+		}
+
+		// First request populates the cache (miss path) and must succeed.
+		res, err := makeRequest()
+		require.NoError(t, err)
+		require.Equal(t, "MISS", res.Response.Header.Get(core.NormalizationCacheHeader))
+		require.Equal(t, expected, res.Body)
+
+		// Repeat the identical request until the normalization cache entry lands
```
jensneuse: is it possible to not rely on eventual retries but instead somehow observe the ristretto cache OR change its behaviour to synchronous writes such that we don't have to use eventually which might lead to flaky tests

router-tests/operations/normalization_cache_skip_include_test.go
```diff
@@ -0,0 +1,86 @@
...
+//   - On a cache HIT, normalizeNonPersistedOperation unconditionally deletes every
+//     skip/include variable from Request.Variables. The cached normalized operation
+//     still declares the variable as Boolean!, so validation fails.
+//
+// See also TestPersistedOperationSkipIncludeConcurrency for the related (already
+// fixed) persisted-operation cache-key aliasing bug.
+func TestNormalizationCacheHitWithIncludeVariableAlsoUsedAsArgument(t *testing.T) {
+	t.Parallel()
+
+	// $flag is used as the isAvailable argument AND as the @include condition.
```
jensneuse: please also add a test for
1. persisted operations
2. cache warmer
that ensures that the problem as tested here is also fixed when using PO/PQ and/or cache warmer

router/core/operation_processor.go
jensneuse, line 819: is this still right? you're using it for the cache key, but below you're using entry.removedSkipIncludeVariableNames

jensneuse (dismissed):

router/core/operation_processor.go
gausie: Yes, we need the cache to consider all the skip/include variables regardless of whether normalization removed them

router-tests/operations/normalization_cache_skip_include_test.go
gausie: Added both, both passed

router-tests/operations/normalization_cache_skip_include_test.go
gausie: Yes, rsitretto offers Cache.Wait(), which it now uses

alepane21 (approved):

PR #3055, opened 2026-07-07
feat(router): router force use of variables and not inline values

SkArchon:
This PR adds the capability to block inline arguments. This is useful to increase normalization cache hit rates. The feature is broken down into three modes.

* Enforcing: Error when inline args are present
* Non-Enforcing: Don't error but warn log when inline args are present
* Off: Do Nothing

In addition to this we have two additional options which allow users to either exempt (default) or also validate persisted operations, as well as return extensions (two formats for "Enforcing" and "Non-Enforcing").

This PR depends on https://github.com/wundergraph/graphql-go-tools/pull/1577

## Summary by CodeRabbit

* **New Features**
  * Added `engine.disallow_inline_arguments` policy (config/env) to detect inline GraphQL argument values with modes: `off`, `enabled-non-enforcing`, `enabled-enforcing`, including options for persisted operations and optional `extensions.inlineArguments`.
  * Consistent behavior across HTTP, WebSocket subscriptions, and persisted operations.
* **Bug Fixes**
  * Enforcing mode now rejects requests with HTTP 400 using generic GraphQL errors; non-enforcing mode emits warnings (on both cache miss and hit) and preserves behavior across normalization and normalization-cache flows.
* **Tests**
  * Added integration coverage for inline-argument detection, warnings/extensions, and persisted-operation behavior.
* **Chores**
  * Updated `github.com/wundergraph/graphql-go-tools/v2` version.

## Checklist

## Open Source AI Manifesto


router/pkg/config/config.go
```diff
@@ -593,6 +595,33 @@ type CostControl struct {
...
+
+// Enabled reports whether the policy is active in any mode.
+func (d DisallowInlineArguments) Enabled() bool {
+	return d.Mode == DisallowInlineArgumentsModeNonEnforcing || d.Mode == DisallowInlineArgumentsModeEnforcing
+}
+
+// Enforcing reports whether the policy rejects offending operations.
+func (d DisallowInlineArguments) Enforcing() bool {
+	return d.Mode == DisallowInlineArgumentsModeEnforcing
+}
```
Noroth: I'm not 100% confident that we should do the config like this. I'm not against it but maybe we could make it a bit smarter.

Maybe instead of it being `Disallow`-configuration, we could have some kind of `ArgumentValidation`-configuration and inside of it we could then have `DisallowInlineArguments`

Also

```
const (
	DisallowInlineArgumentsModeOff          DisallowInlineArgumentsMode = "off"
	DisallowInlineArgumentsModeNonEnforcing DisallowInlineArgumentsMode = "enabled-non-enforcing"
	DisallowInlineArgumentsModeEnforcing    DisallowInlineArgumentsMode = "enabled-enforcing"
)
```

Would it make sense to define general modes that could be re-used?

And we should probably improve the values here, I think `enabled-non-enforcing` is too long:

Something like:

```
const (
    DisallowInlineArgumentsModeOff         = "off"
    DisallowInlineArgumentsModePermissive  = "permissive"
    DisallowInlineArgumentsModeStrict      = "strict"
)
```

router/core/operation_processor.go
```diff
@@ -1524,16 +1564,19 @@ type parseKitOptions struct {
 	disableExposingVariablesContentOnValidationError       bool
 	relaxSubgraphOperationFieldSelectionMergingNullability bool
 	enableDefer                                            bool
+	disallowInlineArguments                                config.DisallowInlineArguments
 }
 
 func createParseKit(i int, options *parseKitOptions) *parseKit {
+	normalizationOptions, inlineArgumentsValidator := buildNormalizationOptions(options.enableDefer, options.disallowInlineArguments)
```
Noroth: Do we have any simple way of implementing a normalization result in the engine instead?

router/pkg/config/config.go
SkArchon: What about something like this
```
const (
	StrictModeOff        StrictMode = "off"
	StrictModePermissive StrictMode = "permissive"
	StrictModeEnforced   StrictMode = "strict"
)

type ValidateInlineArguments struct {
	Mode                       StrictMode `yaml:"strict_mode,omitempty" envDefault:"off" env:"MODE"`
	EnforceHTTPStatusCode      int        `yaml:"enforce_http_status_code,omitempty" envDefault:"400" env:"ENFORCE_HTTP_STATUS_CODE"`
	ErrorCode                  string     `yaml:"error_code,omitempty" envDefault:"INLINE_ARGUMENT_VALUES_NOT_ALLOWED" env:"ERROR_CODE"`
	ErrorMessage               string     `yaml:"error_message,omitempty" envDefault:"Inline argument values are not allowed. Use variables instead." env:"ERROR_MESSAGE"`
	IncludePersistedOperations bool       `yaml:"include_persisted_operations,omitempty" envDefault:"false" env:"INCLUDE_PERSISTED_OPERATIONS"`
	ReturnInResponseExtensions bool       `yaml:"return_in_response_extensions,omitempty" envDefault:"false" env:"RETURN_IN_RESPONSE_EXTENSIONS"`
}

// Enabled reports whether the policy is active in any mode.
func (d ValidateInlineArguments) Enabled() bool {
	return d.Mode == StrictModePermissive || d.Mode == StrictModeEnforced
}

// Enforcing reports whether the policy rejects offending operations.
func (d ValidateInlineArguments) Enforcing() bool {
	return d.Mode == StrictModeEnforced
}
```

and in EngineConfig

```
ValidateInlineArguments ValidateInlineArguments `yaml:"validate_inline_arguments" envPrefix:"..."
```

SkArchon: Updated to use `EnforcementMode`

router/core/operation_processor.go
SkArchon: I updated this so the normalizationResult is now returned, let me know if the new version is good enough

Noroth (approved):

Noroth (approved):

PR #3056, opened 2026-07-07
feat(router): add pre-fetch field authorization

Noroth:
This PR provides support for pre-fetch field authorization. When enabled, fields protected by @requiresScopes/@authenticated are authorized in a single batch before any subgraph fetch runs, instead of fetching and then filtering them out of the response. That way we don't fetch data the client isn't allowed to see.

It introduces a new config flag (authorization.enable_pre_fetch_field_authorization) which is off by default, so nothing changes unless you opt in.

The existing authorizer was extended to fulfill the engine's new BatchAuthorizer interface. It reuses the same scope-checking logic, so the authorization outcomes and the missingScopes response extension stay the same as before.

Also adds two Makefile helpers to set the engine version, one to pin it to a branch and one to update to the latest release.

## Summary by CodeRabbit

* **New Features**
  * Added `enable_pre_fetch_field_authorization` / `ENABLE_PRE_FETCH_FIELD_AUTHORIZATION` to authorize protected fields in a single pre-fetch batch.
  * Authorization denials now expose `missingScopes`/`actualScopes` via response extensions.
  * Added Makefile targets to pin/update the shared GraphQL tooling version.
* **Bug Fixes**
  * Enabled pre-fetch field authorization consistently for HTTP GraphQL requests and websocket subscriptions.
* **Documentation**
  * Documented the new pre-fetch field authorization option and its behavior.
* **Tests**
  * Added unit and integration coverage for HTTP and websocket pre-fetch authorization, including reject/partial-response and subgraph skipping scenarios.


devsergiy (approved):

router-tests/security/prefetch_field_authorization_test.go
```diff
@@ -0,0 +1,193 @@
...
...uery.employees.startDate', Reason: missing required scopes.","path":["employees",5,"startDate"],"extensions":{"code":"UNAUTHORIZED_FIELD_OR_TYPE"}},{"message":"Unauthorized to load field 'Query.employees.startDate', Reason: missing required scopes.","path":["employees",6,"startDate"],"extensions":{"code":"UNAUTHORIZED_FIELD_OR_TYPE"}},{"message":"Unauthorized to load field 'Query.employees.startDate', Reason: missing required scopes.","path":["employees",7,"startDate"],"extensions":{"code":"UNAUTHORIZED_FIELD_OR_TYPE"}},{"message":"Unauthorized to load field 'Query.employees.startDate', Reason: missing required scopes.","path":["employees",8,"startDate"],"extensions":{"code":"UNAUTHORIZED_FIELD_OR_TYPE"}},{"message":"Unauthorized to load field 'Query.employees.startDate', Reason: missing required scopes.","path":["employees",9,"startDate"],"extensions":{"code":"UNAUTHORIZED_FIELD_OR_TYPE"}}],"data":{"employees":[null,null,null,null,null,null,null,null,null,null]},"extensions":{"authorization":{"missingScopes":[{"coordinate":{"typeName":"Employee","fieldName":"startDate"},"required":[["read:employee","read:private"],["read:all"]]}],"actualScopes":["read:employee"]}}}`, string(data))
```
jensneuse, line 141: Do you think we should return just a single error in this case with the path `Query.employees.startDate` instead of having an actual path pointing to every item in the list?

router/core/websocket.go
```diff
@@ -1136,6 +1136,9 @@ func (h *WebSocketConnectionHandler) executeSubscription(registration *Subscript
 	if h.graphqlHandler.authorizer != nil {
 		resolveCtx = WithAuthorizationExtension(resolveCtx)
 		resolveCtx.SetAuthorizer(h.graphqlHandler.authorizer)
+		if h.graphqlHandler.authorizer.IsPreFetchFieldAuthorizationEnabled() {
```
jensneuse, line 1152: Did you add a test to cover this?

jensneuse (changes requested):

router-tests/security/prefetch_field_authorization_test.go
Noroth: Personally I think we should put this in a different scope. This behavior already exists in our engine and changing it might be breaking for some users.

I talked to Sergiy and he also thinks we shouldn't do it right now.

router/core/websocket.go
Noroth: Done

jensneuse (approved):

Makefile
```diff
@@ -216,3 +216,20 @@ new-gm-data-migration:
 		exit 1; \
 	fi
 	mkdir -p data_migrations/graphqlmetrics/$(shell date +%s)_$(name)
+
```
ysmolski, line 219: might be irrelevant but this is the 1 script i use for those tasks:
```
#!/usr/bin/env bash

# set -ex

COMMIT=$1
(
        cd ~/d/cosmo/router && \
                go get github.com/wundergraph/graphql-go-tools/v2@$1 && \
                go mod tidy && \
                cd ../router-tests && \
                go get github.com/wundergraph/graphql-go-tools/v2@$1 && \
                go mod tidy &&\
                cd .. &&\
                git add router/go.mod router/go.sum router-tests/go.mod router-tests/go.sum
                # git commit --amend \
)
```
`mod tidy` is helpful sometimes

ysmolski:
@Noroth It looks good to me, but could you please switch the engine to the `yury/eng-9855-do-not-charge-actual-cost-for-denied-fields` branch ([PR](https://github.com/wundergraph/graphql-go-tools/pull/1582)) that includes fixes for cost? 

UPDATE: I am working on the engine still. Testing edge-cases.

Makefile
Noroth: The target is doing a tidy after go get

ysmolski:
updated to point to https://github.com/wundergraph/graphql-go-tools/pull/1582

ysmolski (approved):

PR #3058, opened 2026-07-08
Omit empty router config signature metadata

### LINKED ISSUE #3059: Unsigned router config upload sends empty signature metadata
Unsigned router config uploads currently include `signature-sha256` with an empty value when no admission signature is present.

Some S3-compatible storage adapters reject empty metadata values, so `routerconfigs/latest.json` can fail to upload and routers can stay on an older config.

Expected behavior: only send `signature-sha256` when a signature exists.

PR: #3058

umerfarok:
Fixes #3059.

Omits `signature-sha256` metadata for unsigned router configs. Signed configs still include it.

Tested:
- controlplane typecheck
- composer metadata test
- cdn metadata tests

## Summary by CodeRabbit

- **Refactor**
  - Consolidated router configuration metadata handling into a shared utility.
  - Preserved existing upload and cache-warming behavior while standardizing metadata creation.
  - Router metadata now consistently includes the configuration version and includes signature information when available.

- **Tests**
  - Added coverage verifying metadata output with and without signature information.


umerfarok:
Hi @wundergraph/Cosmo, could someone take a look when you have a chance?

This is a small controlplane fix for router config uploads: unsigned configs no longer send an empty `signature-sha256` metadata value. I also checked the CDN side; missing and empty signature metadata both result in no `X-Signature-SHA256` header, so the read path stays the same.

umerfarok:
@wilsonrivera Did you get  a chance to look at this ?

controlplane/src/core/composition/composer.ts
```diff
@@ -64,7 +64,16 @@ export type CompositionResult = {
 
 export interface S3RouterConfigMetadata extends Record<string, string> {
   version: string;
-  'signature-sha256': string;
+}
+
+export function createRouterConfigMetadata(version: string, signatureSha256?: string): S3RouterConfigMetadata {
```
wilsonrivera: Could this be moved to `util/composition.ts` and the parameters be converted to a type with required `version` and optional `signature`, that way it can be reused in the future without having to introduce new parameters

controlplane/test/composer.test.ts
```diff
@@ -0,0 +1,57 @@
+import { create } from '@bufbuild/protobuf';
```
wilsonrivera: Update the test to use the extracted `createRouterConfigMetadata`, under the same folder structure

controlplane/test/composer.test.ts
```diff
@@ -0,0 +1,57 @@
...
+import { RouterConfigSchema } from '@wundergraph/cosmo-connect/dist/node/v1/node_pb';
+import { describe, expect, test, vi } from 'vitest';
+import type { BlobStorage } from '../src/core/blobstorage/index.js';
+import { Composer, createRouterConfigMetadata } from '../src/core/composition/composer.js';
+
+describe('Composer', () => {
+  test('omits router config signature metadata when no signature is available', async () => {
+    const putObject = vi.fn().mockResolvedValue(undefined);
```
wilsonrivera: The test can be simplified to ensure that the returned object contains only the expected elements, the same way the other test is performed

wilsonrivera (changes requested):

controlplane/src/core/composition/composer.ts
umerfarok: Fixed. Moved `S3RouterConfigMetadata` and `createRouterConfigMetadata` into `src/core/util/composition.ts`, and it now takes a `{ version, signature? }` param so future fields won't need new args.

controlplane/test/composer.test.ts
umerfarok: Fixed. Moved the test to `test/util/composition.test.ts` (mirroring the source path) and it imports the extracted `createRouterConfigMetadata`.

controlplane/test/composer.test.ts
umerfarok: Fixed. Simplified to assert the returned object directly with `toStrictEqual`, matching the other test.

umerfarok:
@wilsonrivera  Resolved comments

PR #3061, opened 2026-07-08
Failing regression test for APQ TTL renewal corrupting stored query

benevolent-donut:
See title and https://github.com/wundergraph/cosmo/issues/3062

## Summary by CodeRabbit

* **Tests**
  * Added a regression test for automatic persisted queries to ensure TTL renewal keeps the stored query intact.
  * Covered both initial registration and hash-only replay flows to verify persisted query data remains unchanged after renewal.

### DIFF
--- a/router/core/operation_processor_apq_ttl_test.go
+++ b/router/core/operation_processor_apq_ttl_test.go
@@ -1,0 +1,154 @@
+package core
+
+import (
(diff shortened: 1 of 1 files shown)


endigma:
Hi, thanks for the issue and reproduction, we'll take a look.

PR #3070, opened 2026-07-10
fix: encode url parameters

wilsonrivera:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved URL safety across Studio and CLI by URL-encoding organization, namespace, graph, and other dynamic route/query segments.
  * Fixed “Open in studio” and “Run in playground” destinations, affected-resource warning/JSON output, Slack OAuth redirects, billing/upgrade navigation, invitations, and GitHub commit links.
  * Prevented malformed routes in various navigation paths (including edge cases with special characters).

* **New Features**
  * Added a shared URL-building utility and added test coverage to ensure templates render and encode correctly.

## Checklist

## Open Source AI Manifesto


alepane21 (approved):

comatory (changes requested):
I think it's great we are fixing this. I have a suggestion for you. Since we are now touching the logic extensively for dealing with URLs, how about we introduce a new convention?

I kind of dislike reading stuff like

```
/${encodeURIComponent(organizationSlug ?? '')}/feature-flags?namespace=${encodeURIComponent(namespace)}`
```

It's low abstraction so fair enough, you can scan it, but there is also additional logic for falling back to default values `?? ''`, plus the manually added `?`.

What I'm thinking is abstracting the construction of URL into functions. Such as `createOrganizationFeatureFlagPath` that would accept parameters, such as `organizationSlug` so the call site is something like:

```
createOrganizationFeatureFlagPath({
  organizationSlug,
  namespace
})
```

...and rest of the normalization logic is abstracted away.

I am not advocating for extracting the logic into functions for some kind of re-usability. Although, if we do and have LLM do a pass over these functions, it might recognize we could re-use some of them. But generally, I'd put them in the same file and hide the implementation details. 

Secondly - we do stuff like adding `?` and `encodeURIComponent` calls. I'm wondering, isn't this so messy because we are avoiding absolute URLs? Because if we had absolute URLs, we could use `new URL` constructor`, with setters like `url.searchParams.set('namespace', namespace)` which automatically encodes everything.

I'm not sure how the router would behave with absolute paths though.

wilsonrivera:
> I think it's great we are fixing this. I have a suggestion for you. Since we are now touching the logic extensively for dealing with URLs, how about we introduce a new convention?
> 
> I kind of dislike reading stuff like
> 
> ```
> /${encodeURIComponent(organizationSlug ?? '')}/feature-flags?namespace=${encodeURIComponent(namespace)}`
> ```
> 
> It's low abstraction so fair enough, you can scan it, but there is also additional logic for falling back to default values `?? ''`, plus the manually added `?`.
> 
> What I'm thinking is abstracting the construction of URL into functions. Such as `createOrganizationFeatureFlagPath` that would accept parameters, such as `organizationSlug` so the call site is something like:
> 
> ```
> createOrganizationFeatureFlagPath({
>   organizationSlug,
>   namespace
> })
> ```
> 
> ...and rest of the normalization logic is abstracted away.
> 
> I am not advocating for extracting the logic into functions for some kind of re-usability. Although, if we do and have LLM do a pass over these functions, it might recognize we could re-use some of them. But generally, I'd put them in the same file and hide the implementation details.
> 
> Secondly - we do stuff like adding `?` and `encodeURIComponent` calls. I'm wondering, isn't this so messy because we are avoiding absolute URLs? Because if we had absolute URLs, we could use `new URL` constructor`, with setters like `url.searchParams.set('namespace', namespace)` which automatically encodes everything.
> 
> I'm not sure how the router would behave with absolute paths though.

I think we can abstract to a generic method `joinUrlPath` which could look like:
```
joinUrlPath(organizationSlug, namespace, 'something');
```

And this method could skip and normalize the provided parameters

comatory:
I mean I'd prefer if the function used object parameter as that's a bit more maintainable, but that's implementation detail in my opinion and does not matter that much.

What I'm trying to say is that since we now touch this area, we could re-think how we build paths in general. What do you think about my idea with `URL` constructor? I think that would simplify a lot of things.

We would still have these functions per-component (or shared if needed), but inside of them one would construct URL object. I find this useful especially with query parameters as you could skip the encoding step there. Then the output of that function would be a call to `url.toString()`.

I know `URL` class needs absolute URL, so not sure how the Next router will behave. If the router can't accept that, then let's just use these functions but if it does, we could use it internally.

comatory:
We could also consider adding a dependency for this, I found [page-path](https://www.npmjs.com/package/page-path) but not sure how good that is. There's also `fast-url` but that one is probably aimed at absolute URLs as well.

studio/src/lib/build-url.ts
```diff
@@ -0,0 +1,99 @@
...
+  : PathParams<Template>;
+
+const allowedTypes = ['string', 'number', 'boolean'];
+
+export function buildUrl(template: string): string;
+export function buildUrl<Template extends string>(template: Template, params: BuildUrlParams<Template>): string;
+export function buildUrl<
+  Template extends string,
+  ExtraParams extends Record<string, Primitive> = Record<string, Primitive>,
+>(template: Template, params: Simplify<BuildUrlParams<Template> & ExtraParams>): string;
```
comatory: Are these definitions even needed? We have implementation in the same file which can already infer the types.

studio/src/lib/build-url.ts
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * Receives a template string and builds an URL based on the provided parameters, ignoring any
+ * empty values and adding any remaining to the search parameters of the URL.
```
comatory, line 31: Please add `@example` block.

studio/src/lib/build-url.ts
```diff
@@ -0,0 +1,99 @@
...
+ *
+ * @param template The template string to render
+ * @param params The parameters to substitute when rendering the template
```
comatory:
```suggestion
 * @param params The parameters to substitute when rendering the template, unused params become query params
```

studio/src/lib/build-url.ts
```diff
@@ -0,0 +1,99 @@
...
+ */
+export function buildUrl<
+  Template extends string = string,
+  ExtraParams extends Record<string, Primitive> = Record<string, Primitive>,
+>(template: Template, params?: Simplify<BuildUrlParams<Template> & ExtraParams>): string {
+  const finalPathSegments: string[] = [];
+  const parametersMap = new Map<string, Primitive>(Object.entries(params ?? {}));
+
+  let url = new URL(
```
comatory, line 58: FYI I'd double check the behaviour with absolute links but if it works this is a good approach I think 👍

studio/src/lib/build-url.ts
```diff
@@ -0,0 +1,99 @@
...
+    template,
+    typeof window === 'undefined' ? process.env.NEXT_PUBLIC_COSMO_STUDIO_URL : window.location.origin,
+  );
+
+  const templateSegments = url.pathname.split('/');
+  for (const segment of templateSegments) {
+    if (!segment) {
```
comatory: How does segment become falsy? What if the path contains `0`? I think that's a valid part of the path. I think `if (segment.length === 0)` is better WDYT?

comatory: Shouldn't it be possible to use `isValueAllowed` here? We could then omit the usage of this function on L87

studio/src/lib/build-url.ts
```diff
@@ -0,0 +1,99 @@
...
+  ? { [K in Param | keyof PathParams<Rest>]: Primitive }
+  : Path extends `${string}:${infer Param}`
+    ? { [K in Param]: Primitive }
+    : {};
+
+type BuildUrlParams<Template extends string> = Template extends `${infer Path}?${infer Query}`
+  ? Simplify<PathParams<Path> & Omit<QueryParams<Query>, keyof PathParams<Path>>>
+  : PathParams<Template>;
+
+const allowedTypes = ['string', 'number', 'boolean'];
```
comatory:
```suggestion
const allowedTypes = new Set(['string', 'number', 'boolean'] as const);
```

I think using set is faster for lookup instead of array.

studio/src/lib/build-url.ts
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    url.searchParams.set(key, encodeURIComponent(value));
+  }
+
+  return url.toString();
+}
+
+function isValueAllowed(value: unknown): value is string | number | boolean {
+  return value !== undefined && value !== null && value !== '' && allowedTypes.includes(typeof value);
```
comatory:
```suggestion
return value !== undefined && value !== null && value !== '' && allowedTypes.has(typeof value);
```

studio/src/lib/build-url.ts
```diff
@@ -0,0 +1,99 @@
...
+      finalPathSegments.push(segment);
+    }
+  }
+
+  // Update the path and search parameters of the URL to the with the corresponding values
+  url.pathname = finalPathSegments.length === 0 ? '/' : finalPathSegments.join('/');
+  for (const [key, value] of parametersMap.entries().toArray()) {
+    if (!isValueAllowed(value)) {
+      continue;
+    }
```
comatory, line 105:
```suggestion
```

This could be omitted, see comment: https://github.com/wundergraph/cosmo/pull/3070/changes/7cd2ccc24fcc63c7c501308219d5bc96e90820c8..c7c2a236cc58cf0af63a4e5bfe8cad16a7a1cf87#r3600779140

studio/src/lib/build-url.ts
```diff
@@ -0,0 +1,99 @@
...
+      finalPathSegments.push(encodeURIComponent(value));
+    } else {
+      // The segment doesn't represent a parameter, push it as is
+      finalPathSegments.push(segment);
+    }
+  }
+
+  // Update the path and search parameters of the URL to the with the corresponding values
+  url.pathname = finalPathSegments.length === 0 ? '/' : finalPathSegments.join('/');
+  for (const [key, value] of parametersMap.entries().toArray()) {
```
comatory, line 102: This can be completely omitted by doing this:

```
url.search = new URLSearchParams(parametersMap)
```

This assigns all of it at once and it should also take care of encoding each value and concatenating everything when `.toString` is called.

studio/src/lib/build-url.ts
```diff
@@ -0,0 +1,99 @@
...
+      }
+
+      finalPathSegments.push(encodeURIComponent(value));
+    } else {
+      // The segment doesn't represent a parameter, push it as is
+      finalPathSegments.push(segment);
+    }
+  }
+
+  // Update the path and search parameters of the URL to the with the corresponding values
```
comatory: Misleading comment, the next line does not do this. I think it does some kind of normalization, like adding a `/` at the end or otherwise concatenating unused parameters.

Can we make this more clear? Also can we ensure trailing `/` is used consistently?

comatory (changes requested):
I really like this approach because you can still see the representation of the path (looks like `urlcat` library btw :) )

I have only few suggestion on the implementation, I think it could be simplified a little bit.

studio/src/__tests__/build-url.test.ts
```diff
@@ -0,0 +1,37 @@
...
+    slug: 'org',
+    namespace: 'default',
+    name: 'feat/graph',
+  });
+
+  expect(url).toBe('http://localhost:3000/org/default/graph%20with%20space/feat%2Fgraph');
+});
+
+test('that segments with missing parameters are skipped', () => {
+  // @ts-ignore: TypeScript errors as is missing a required parameter
```
comatory: Seems like it's not needed, which parameter is missing?

studio/src/__tests__/build-url.test.ts
```diff
@@ -0,0 +1,37 @@
+import { buildUrl } from '../lib/build-url';
```
comatory, line 1: I would test following cases as well:

1. The order of query parameters. I don't mean that we should sort them in any way, but we should test that query parameters appear in the order of the input. I know the order of properties on object literals is not guaranteed, but on `Map` it mostly is, so I'd just test the implementation so we get failing test if it ever changes in the future.
2. Explicit test for parameters that have falsy values (`0`, `undefined`, `null`, `""`)
3. Explicit test for trailing `/`, we should have one way of handling it and it should be tested

cli/src/commands/clients/commands/delete.ts
```diff
@@ -215,7 +215,9 @@ export default (opts: BaseCommandOptions) => {
 
     if ((previewResp.persistedOperationsCount > 0 || previewResp.hasTraffic) && !options.force) {
       const studioUrlObj = new URL(
-        `${previewResp.organizationSlug}/${options.namespace}/graph/${options.graphName}/operations`,
+        `${encodeURIComponent(previewResp.organizationSlug)}/${encodeURIComponent(options.namespace)}/graph/${encodeURIComponent(
```
comatory, line 218: What if we used `buildUrl` in CLI as well? I know it introduces a sharing mechanism so it would need to be a separate package.

It's probably out of scope for this PR, but you can create Linear ticket and we can follow up on it. I think it's important to have a consistent way of building URLs.

comatory (changes requested):

studio/src/__tests__/build-url.test.ts
wilsonrivera: In this case the template contains `:slug` which is not provided in the parameters, although the parameter value accept `null` and `undefined`, you still need to provide it. The point is to let you know that something is missing

studio/src/lib/build-url.ts
wilsonrivera: Yes, however IDEs (specially Idea based ones) will try to autocomplete to `buildUrl<any, any>`

studio/src/lib/build-url.ts
wilsonrivera: It does but it will replace existing query parameters, imagine a template that contains `/….?schemaType=client`

studio/src/lib/build-url.ts
wilsonrivera: And empty string is `falsy`, can update to `segment.length === 0` for more clarity

studio/src/lib/build-url.ts
comatory: Oh ok that's right. But why don't we do this for `parametersMap`:
`new Map(url.searchParams)` which will add existing ones to it. We can then use that.

comatory (approved):

PR #3072, opened 2026-07-12
fix(router): renew APQ TTL without corrupting stored query

### LINKED ISSUE #3062: APQ corrupts stored queries with @skip and @includeIf directives on TTL renewal with normalized versions
### Component(s)

router

### Component version

v0.277.0

### wgc version

N/A

### controlplane version

N/A

### router version

v0.277.0

## Description

APQ corrupts the stored query in Redis on TTL renewal. When a persisted operation is re-executed and its cache key has a TTL, the router re-saves the operation to renew expiration but writes back `NormalizedRepresentation` instead of the original raw query. Normalization evaluates and strips `@skip`/`@include` directives against the current request's variable values. So the stored query gets overwritten with a variable-specific, directive-stripped version. Subsequent requests resolving the same hash with different variable values then receive a query missing the conditionally-included fields, returning a 200 with the fields silently absent. 

The initial registration is correct (it saves the raw `Request.Query`); only TTL-renewal saves on later cache hits corrupt the entry.

Location: https://github.com/wundergraph/cosmo/blob/main/router/core/operation_processor.go#L448-L455
```go
	if fromCache {
		if fromCacheHasTTL, _ := o.persistedOperationCacheKeyHasTtl(clientInfo.Name, includeOperationName); fromCacheHasTTL {
			// if it is an APQ request, we need to save it again to renew the TTL expiration
			if err = o.operationProcessor.persistedOperationClient.SaveOperation(ctx, clientInfo.Name, o.parsedOperation.GraphQLRequestExtensions.PersistedQuery.Sha256Hash, o.parsedOperation.NormalizedRepresentation); err != nil {
				return false, false, err
			}
		}
		return true, false, nil
	}
```

## Steps to Reproduce

1. Enable APQ with a Redis store that has a TTL configured on entries.
2. Register a persisted query containing a conditional field, e.g. `query Q($skipIt: Boolean!) { a b @skip(if: $skipIt) }`.
3. Send request with `{ "skipIt": true }`. The entry has a TTL, so the router re-saves the normalized query, which for these variables is `query Q { a }` (field `b` stripped).
4. Send request with the same hash and `{ "skipIt": false }`.

## Expected Result

Field `b` is returned for the second request, since `skipIt` is false.

## Actual Result

Field `b` is absent from the second response (200, no errors). Redis now holds the stripped query `query Q { a }`, so `b` can never be requested via that hash regardless of variables until the entry expires and the client re-registers.

### Environment information

OS: macOS (Darwin 25.5.0, arm64)
Package Manager: Go modules
Compiler: go 1.26.3
Test framework: go test

### Router configuration

```yaml
automatic_persisted_queries:
    enabled: true
    cache:
      ttl: 300          # any value > 0; the bug requires a TTL to be set
    storage:
      provider_id: my_redis
      object_prefix: cosmo_apq

  storage_providers:
    redis:
      - id: my_redis
        cluster_enabled: false
        urls:
          - "redis://localhost:6379"
```

### Router execution config

```json
N/A
```

### Log output

```shell
N/A
```

### Additional context

PR with failing regression test https://github.com/wundergraph/cosmo/pull/3061

Lawal-create:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved automatic persisted query TTL renewal without altering the original query content.
  * Fixed issues affecting queries that use conditional directives such as `@skip`.
  * Ensured cached queries continue to resolve conditional fields correctly after TTL renewal.

* **Tests**
  * Added regression coverage for TTL renewal across distributed and in-memory persistence scenarios.

## Description

Fixes #3062.

Automatic Persisted Queries (APQ) corrupted the stored query on TTL renewal.
When a persisted operation whose store entry has a TTL is re-executed and
resolved from the in-memory normalization cache, the router re-saved it to renew
the expiration but wrote back the **normalized representation** instead of the
original raw query. Normalization evaluates and strips `@skip`/`@include` using
the current request's variables, so the stored query was overwritten with a
variable-specific, directive-stripped version. Subsequent requests resolving the
same hash with different variables then silently lost the conditionally-included
fields (HTTP 200, fields absent). Only TTL-renewal on later cache hits was
affected; initial registration correctly stored the raw query.

## Root cause

In `FetchPersistedOperation` (`router/core/operation_processor.go`), the APQ
cache-hit-with-TTL branch called `SaveOperation(..., NormalizedRepresentation)`.
On the hash-only replay path the raw query is not in memory, so the normalized
(stripped) form was written to the APQ store (Redis or in-memory).

## Fix

Renew the TTL at the storage layer without rewriting the body:

- `apq.Client.RenewTTL` — re-writes the raw body already present in the store via
  the existing `Get`+`Set` (APQ is content-addressed by `sha256(rawQuery)`, so
  the stored body is invariant). No-op when the entry is absent. Covers both the
  Redis and in-memory stores.
- `persistedoperation.Client.RenewOperationTTL` — wraps it and preserves the
  existing PQL-manifest ownership guard used by `SaveOperation`.
- `FetchPersistedOperation` now calls `RenewOperationTTL` instead of
  `SaveOperation(NormalizedRepresentation)`.

No change to the `apq.KVClient` interface, APQ registration, or
persisted-operation lookup semantics. No generated code changes.

## Tests

- Unit (`router/core/operation_processor_apq_ttl_test.go`): regression tests for
  the KV-backed (Redis-style mock) and in-memory APQ stores — asserting the
  stored body stays the raw query and that a `skip:false` replay still resolves
  the conditional field after a `skip:true` renewal. Based on the regression test
  from #3061.
- Integration (`router-tests/operations/automatic_persisted_queries_test.go`): a
  Redis-backed end-to-end regression in the existing `redis cache` group,
  asserting the Redis entry remains the raw query after renewal and that a
  `skip:false` request returns the conditional field.

Both were verified to fail on the unpatched code and pass with the fix.

## How to test

1. Configure APQ with a Redis store and `automatic_persisted_queries.cache.ttl > 0`.
2. Register `query Q($skip: Boolean!) { a: __typename b: __typename @skip(if: $skip) }`
   with `{"skip": true}` (query + sha in the request).
3. Replay hash-only with `{"skip": true}` (triggers TTL renewal).
4. Send hash-only with `{"skip": false}` → the response must include `b`.
   On the unpatched router, `b` is missing and Redis holds the stripped
   `query Q { a: __typename }`.

Automated:

```
cd router && make test
cd router-tests && go test -race -run '^TestAutomaticPersistedQueries$' ./operations/
```

## Notes for upgraders

Entries already corrupted in Redis by an affected router cannot be reconstructed
from hash-only requests. After upgrading, flush the APQ key prefix or wait for
the TTL to expire; affected clients re-register on the next miss.

## Checklist

- [x] I have discussed my proposed changes in an issue (#3062) and provided a reproduction.
- [x] I have provided instructions for reviewers for manually validating my changes.
- [ ] Documentation has been updated on https://github.com/wundergraph/docs-website (N/A — bug fix, no user-facing config change).
- [x] I have read the Contributors Guide.

## Open Source AI Manifesto

This contribution follows the principles of the [Open Source AI Manifesto](https://human-oss.dev).

### DIFF
--- a/router-tests/operations/automatic_persisted_queries_test.go
+++ b/router-tests/operations/automatic_persisted_queries_test.go
@@ -2,6 +2,8 @@ package integration
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
 	"fmt"
 	"net/http"
 	"os"
@@ -196,6 +198,13 @@ func TestAutomaticPersistedQueries(t *testing.T) {
 				require.Equal(t, `{"data":{"__typename":"Query"}}`, res3.Body)
 			})
 		})
+
+		// Note: the in-memory (local cache) TTL-renewal path for issue #3062 is
+		// covered deterministically by the unit tests in
+		// router/core/operation_processor_apq_ttl_test.go. The in-memory store
+		// (ristretto) is eventually consistent, so an HTTP-level cross-variable read
+		// here would be racy; the Redis (distributed) end-to-end regression below
+		// uses a synchronous store and is the authoritative integration guard.
 	})
 
 	t.Run("redis cache", func(t *testing.T) {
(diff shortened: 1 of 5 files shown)


endigma:
See #3191

PR #3087, opened 2026-07-15
feat(router): make MCP server mount path configurable

gausie:
We want to introduce multi-collection MCP. I want to split into tiny PRs.

- Multi-collection means routing by URL path, but `/mcp` was hardcoded in six places — you literally can't mount a second server on the same listener without the path becoming a parameter first.
- The OAuth discovery chain (the well-known metadata URL, the `WWW-Authenticate` header, the `resource` field) is all derived from that path. If we didn't fix it now, every mounted server would claim to be the same OAuth resource and per-server auth would silently break.

## Changes

- New `WithMountPath` option on `mcpserver.NewGraphQLSchemaServer` (default `/mcp`, unchanged). Not exposed in user config or docs yet — Go API only.
- The RFC 9728 well-known path, WWW-Authenticate `resource_metadata` URL, and metadata `resource` field are all derived from the mount path instead of hardcoded strings.
- Mount paths are normalized (leading slash, no trailing slash) and validated at construction: empty/root paths, empty interior segments, ServeMux pattern metacharacters (`{`, `}`), and whitespace are rejected with a config error instead of panicking or silently misrouting at startup.
- `Serve()`'s mux construction extracted into `buildHandler()` so routing is testable without binding a port.

## Testing

- New unit tests: custom mount path routing, well-known path derivation, path normalization/validation table, constructor rejection cases.
- All existing `router/pkg/mcpserver` and `router-tests/protocol` tests pass unedited (default behavior is byte-for-byte unchanged).

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring a custom MCP endpoint mount path.
  * Updated OAuth protected-resource discovery and metadata URLs to use the configured mount path.
* **Bug Fixes**
  * MCP and related well-known endpoints are now served only at the custom endpoint; the default path returns 404.
  * Improved mount-path normalization and validation (including trailing slashes and safer pattern rejection).
* **Tests**
  * Added integration coverage for custom mount path behavior, including verification of default endpoint non-exposure.

Closes ENG-9868


gausie:
Closing without merging: design review concluded this is the wrong shape for multi-collection MCP. Instead of configurable static mount paths per server, we'll route by path parameter on a single handler (/mcp for current behavior, /mcp/{collection-slug} resolved per request), so the /mcp prefix doesn't need to be configurable at all. A new PR will follow with that shape.

PR #2549, opened 2026-02-23
feat(router): per-message-deflate with context takeover support for websockets

dbinnersley:
## Summary by CodeRabbit

* **New Features**
  * Opt-in WebSocket per-message compression (permessage-deflate) with configurable level (1–9, default 6).
* **Configuration**
  * New compression settings in configuration and JSON schema to enable/disable compression, set level, and client window bits; sensible defaults included.
* **Reliability**
  * Enhanced WebSocket connection and handshake flows with optional compression-capable connection establishment and retry handling.
* **Tests**
  * Extensive tests for compression negotiation, context-takeover, fragmentation, multi-client scenarios, and end-to-end behavior.

This is a follow up of a PR I opened up a couple months ago. I would still like to get this functionality in. In addition to the previous functionality which just added support for per-message-deflate, I've now added support for context takeover. This will allow similar messages to be compressed by ~90% instead of significantly less using stateless compression.

This is a quite a big change to websockets because its handling several different options that are available with the per-message-deflate extension

## Checklist

- [ ] I have discussed my proposed changes in an issue and have received approval to proceed.
- [X] I have followed the coding standards of the project.
- [X] Tests or benchmarks have been added or updated.
- [ ] Documentation has been updated on [https://github.com/wundergraph/cosmo-docs](https://github.com/wundergraph/cosmo-docs).
- [X] I have read the [Contributors Guide](https://github.com/wundergraph/cosmo/blob/main/CONTRIBUTING.md).

PR #3096, opened 2026-07-17
feat: add support for per-message-deflate to websockets

dbinnersley:
## Summary by CodeRabbit

- **New Features**
  - Added optional WebSocket per-message compression (permessage-deflate) with configurable enablement, compression level, and client window size.
  - Supports context takeover and no-context-takeover negotiation, including fragmented frames and repeated messages.
- **Documentation**
  - Updated WebSocket configuration docs with compression settings, environment variables, defaults, and an example YAML configuration.
- **Tests**
  - Expanded coverage for compression negotiation, frame handling, fragmented payloads, compressed subscriptions, and context-takeover behavior.

## Summary

Adds opt-in [RFC 7692](https://datatracker.ietf.org/doc/html/rfc7692) `permessage-deflate` compression for WebSocket connections between clients and the router.

- New config block `websocket.compression` with `enabled` (default `false`), `level` (1–9, default `6`), and `client_max_window_bits` (8–15, default `15`), plus matching env vars (`WEBSOCKETS_COMPRESSION_ENABLED`, `WEBSOCKETS_COMPRESSION_LEVEL`, `WEBSOCKETS_COMPRESSION_CLIENT_MAX_WINDOW_BITS`).
- Compression is negotiated per connection during the upgrade. Both context-takeover and `no_context_takeover` modes are supported in both directions; clients that do not offer the extension are served uncompressed as before. Per RFC 7692 §7.1.2.2, `client_max_window_bits` is only echoed in the response when the client offers it, using the more restrictive of the client offer and server config.
- The read path handles fragmented messages, enforces client-frame masking, and maintains the decompression dictionary (sized by the negotiated window bits) across messages when context takeover is active.
- Performance-focused implementation:
  - `no_context_takeover` paths use `sync.Pool`-ed flate writers (pooled per level) and readers instead of per-message construction. This is the path gorilla/websocket clients always negotiate.
  - Two DEFLATE implementations are used deliberately: `klauspost/compress/flate` for pooled writers and all decompression (6–12× faster writes, ~2× faster reads), and stdlib `compress/flate` for the persistent context-takeover compressor only, because klauspost's fast encoders (levels ≤ 6) skip history matching for small inputs, which would forfeit the cross-message compression benefit that justifies per-connection compressor state. The rationale is documented in the code.

## Benefits

- Significant bandwidth reduction for subscription-heavy workloads: GraphQL payloads are repetitive JSON, and with context takeover, consecutive similar ~600 B messages compress to ~17 bytes after the first (measured at level 6).
- Works out of the box with browsers (which offer permessage-deflate by default), gorilla/websocket, and coder/websocket clients.
- Measured cost of the hot no-takeover write path (compress + frame + write, 600 B payload, level 6): **2,351 ns / 924 B / 5 allocs per message**, versus ~38,000 ns / ~814 KB / 19 allocs for a naive per-message `flate.NewWriter` approach. A benchmark (`BenchmarkWriteCompressedNoContextTakeover`) is included to track regressions.
- Zero behavior change when disabled (the default) or when the client does not offer the extension.

## Risks

- **Memory per connection (context takeover):** when context takeover is negotiated, each connection holds a persistent stdlib flate writer (~800 KB) plus up to a 32 KB decompression dictionary. At high connection counts this is significant. Mitigations available today: leave compression disabled (default), or rely on clients that request `no_context_takeover`. A follow-up config option to force `no_context_takeover` in the negotiation response (permitted by RFC 7692) would make memory flat.
- **CPU:** every outbound message on a compression-negotiated connection is compressed. The level is configurable; level 1 gives most of the ratio on JSON at a fraction of the CPU.
- **Small-message output size in no-takeover mode:** klauspost's encoder emits slightly larger output than stdlib for tiny messages (~84 vs ~73 bytes measured on an 80 B payload) — traded deliberately for the 6–12× throughput gain.
- **Inherent to the extension:** compressing attacker-influenced data on the same stream as secrets carries CRIME/BREACH-class considerations, as with any permessage-deflate implementation. The feature is opt-in and off by default.
- Pooled writers/readers are shared across connections; a missing reset would leak sliding-window state between tenants. This is guarded by dedicated isolation tests (see below), and pooling is confined to the stateless mode by construction.

## Testing Done

**Unit tests (`router/core/websocket_test.go`)**

- Compress/decompress round-trips with and without context takeover, fragmented frames, dictionary maintenance and window-bits sizing, negotiation logic (`no_context_takeover` mirroring, window-bits clamping, upgrade-error fallback).
- Pool isolation: identical messages must compress to byte-identical, independently decompressible streams across repeated writes and across connections sharing the pool — either assertion fails if window state leaks.
- Context-takeover ratio test verifying later messages compress smaller via the shared dictionary (this test caught the klauspost small-input behavior and motivated the hybrid design).

**Integration tests (`router-tests/subscriptions/`)**

- Negotiation and round-trip matrix with gorilla (`no_context_takeover`) and coder (context takeover) clients: enabled/disabled on each side, custom levels, window-bits negotiation, multi-query dictionary reuse on one connection.
- Frame-level test (`websocket_compression_test.go`) with a purpose-built RFC 7692 client: compresses every client message (including sub-128 B ones that no off-the-shelf client compresses) against a persistent dictionary, requires RSV1 on every server frame, and proves real cross-message context takeover — a repeated response frame is not independently decompressible but decodes correctly with the shared dictionary. Also serves as a stdlib-client ↔ klauspost-server interop check.
- Full `TestWebSockets` suite, full `router/core` suite, and config schema/defaults/full-fixture tests all pass.

**Manual validation for reviewers**

Enable compression in the router config:

```yaml
websocket:
  enabled: true
  compression:
    enabled: true
    level: 6
```

Start the router against the demo graph, connect any graphql-ws client with compression enabled (a browser does this by default), and confirm the `Sec-WebSocket-Extensions: permessage-deflate` response header on the upgrade; subscriptions behave identically with payloads compressed on the wire. Or run the focused suites:

```bash
# from router/
go test ./core/ -run 'Wrapper|Negotiat|PooledFlate'

# from router-tests/
go test ./subscriptions/ -run 'TestWebSockets$/compression|TestWebSocketsCoderClient|TestWebSocketCompressionContextTakeoverFrameLevel'
```

## Checklist

- [x] If applicable, I have provided instructions for reviewers for [manually validating my changes](https://github.com/wundergraph/cosmo/blob/main/CONTRIBUTING.md#pull-requests-conventions).
- [x] Documentation has been updated on [https://github.com/wundergraph/docs-website](https://github.com/wundergraph/docs-website).

## Open Source AI Manifesto


dbinnersley:
@endigma  this based on the earlier PR https://github.com/wundergraph/cosmo/pull/2549 before your subscription changes were added in. I've rebased and made a couple changes here.

PR #3100, opened 2026-07-18
feat(router): pubsub event hooks — OnBroadcastEvents + adapter middleware (draft/RFC)

mwisner:
**Draft / RFC — feedback welcome.** Two small, additive, *independent* seams for pubsub event handling, plus a sketch of a third for discussion. No behaviour change when unused.

### Context
Registering an `OnReceiveEvents` handler flips every trigger from the concurrent **broadcast** path onto the **serial per-subscriber** path (`pkg/pubsub/datasource/subscription_event_updater.go`, `case 1` vs `case 2/3`). That serializes resolution and defeats single-flight dedup. The seams below let a module transform/observe events while staying on the broadcast path.

### 1. `OnBroadcastEvents` — per-batch hook on the broadcast path
A hook called **once per received batch**, before per-subscriber fan-out, at the top of `subscriptionEventUpdater.Update`. Unlike `OnReceiveEvents` it does **not** force the serial path. Module interface `StreamBroadcastEventHandler`; delivered via `Hooks.OnBroadcastEvents`.

*Use when:* the transform is the **same for every subscriber**. `OnReceiveEvents` runs once **per active subscriber** (in `computeSubscriberEvents`), with that subscriber's request/auth context, so it can produce per-viewer output (redaction, per-user shaping). `OnBroadcastEvents` runs **once per batch** with **no per-subscriber context**, and its single output is broadcast to all subscribers of the trigger — so it's cheaper (1 transform vs N) but by construction cannot vary per subscriber. Reach for it for viewer-independent work: normalizing an envelope, injecting `__typename`, reshaping/observing a batch. (This is about the payload *transform* only — per-subscriber `@openfed__subscriptionFilter` routing is unaffected and still applies.)

### 2. `PubSubAdapterProvider` / `Hooks.AdapterMiddleware` — wrap the adapter
Lets a module wrap a provider's `Adapter` before `Subscribe` (applied in `PubSubSubscriptionDataSource.Start`). The module can decorate the updater to transform/stage events **inside the adapter**, so the engine only ever receives already-normalized data. Module interface `PubSubAdapterProvider.WrapAdapter(providerID, inner) Adapter`.
*Use when:* you want transformation/side-effects at the adapter boundary, before events enter the engine.

**These are two separate operations** — either can be used alone; #2 does not depend on #1.

### 3. (Not implemented — for discussion) Fully pluggable event adapters
A registry/option to register custom `ProviderBuilder`s — or fully custom `Adapter`s — for a provider type. Today the kafka/nats/redis builders are hardcoded in `pubsub.BuildProvidersAndDataSources`. This would give modules full control of the consume loop (offsets, batching, custom dedup, or emitting a key-only event to the engine while staging the full payload elsewhere). Larger surface than #1/#2 and intentionally **not** included here — flagged as the natural next step if there's interest.

---
Happy to adjust naming/shape. Branch is based on a recent `main`; opening as draft to gather direction before polishing (tests/docs).

### DIFF
--- a/router/core/factoryresolver.go
+++ b/router/core/factoryresolver.go
@@ -528,6 +528,11 @@ func (l *Loader) Load(engineConfig *nodev1.EngineConfiguration, subgraphs []*nod
 		onReceiveEventsFns[i] = NewPubSubOnReceiveEventsHook(fn)
 	}
 
+	onBroadcastEventsFns := make([]pubsub_datasource.OnBroadcastEventsFn, len(l.subscriptionHooks.onBroadcastEvents.handlers))
+	for i, fn := range l.subscriptionHooks.onBroadcastEvents.handlers {
+		onBroadcastEventsFns[i] = NewPubSubOnBroadcastEventsHook(fn, l.logger)
+	}
+
 	subscriptionOnCreateFns := make([]pubsub_datasource.SubscriptionOnCreateFn, len(l.subscriptionHooks.onCreate.handlers))
 	for i, fn := range l.subscriptionHooks.onCreate.handlers {
 		subscriptionOnCreateFns[i] = NewPubSubSubscriptionOnCreateHook(fn)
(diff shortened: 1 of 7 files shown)


dkorittki:
Closing this one because it's commits are part of #3102 . The hook is a genuine good idea, just needed to extend some error path logic and tests. The provider adapter middleware is not part of that PR to keep it scoped. Please let us know if it is important to you!

PR #3102, opened 2026-07-21
feat(router): add `BeforeEventsDispatch` Custom Module

dkorittki:
Adds a new Cosmo Streams Custom Module `OnBroadcastEvents`.  
It runs after receiving events from a message provider. It is related to the `OnReceiveEvents` Custom Module but with an important distinction: It runs only once for all subscribers of a trigger. This is for users who don't need to do per-subscriber decisions but want to change the event for subscribers alike.

It's based on PR #3100 with the following changes:
- Removes the provider adapter middleware --> worth a seperate PR
- Adds timeout handling to handler calls
- Adds panic recovery to handler calls
- Adds unit and integration tests

Documentation will be updated in a seperate PR.

## Summary by CodeRabbit

* **New Features**
  * Added a `BeforeEventsDispatch` hook for transforming or filtering event batches before delivery to subscribers.
  * Hooks run once per batch, before per-subscriber processing, with support for sequential handlers and configurable timeouts.
  * Added support for dropping or replacing batches and accessing subscription metadata.

* **Documentation**
  * Added configuration guidance and examples for the new hook.
  * Clarified the differences between batch-level and per-subscriber event hooks.

* **Bug Fixes**
  * Improved timeout handling and isolation between batch-level and subscriber-level processing.

## Checklist

## Open Source AI Manifesto


router/pkg/pubsub/datasource/subscription_event_updater.go
```diff
@@ -74,9 +83,68 @@ func (s *subscriptionEventUpdater) Update(events []StreamEvent) {
...
+						zap.String("handler_name", "OnBroadcastEvents"),
+						zap.Any("error", r),
+					)
+				res = hookResult{nil, false}
+			}
+			done <- res
+		}()
+
+		evts := events
+		var err error
```
endigma: `err` doesn't seem to be used outside this loop, so I think it should just be inside?

router-tests/modules/streams_before_events_dispatch_test.go
```diff
@@ -0,0 +1,437 @@
...
+
+			for i := 0; i < numSubscribers; i++ {
+				require.NoError(t, clients[i].Close())
+				testenv.AwaitChannelWithT(t, Timeout, clientRunChs[i], func(t *testing.T, err error) {
+					require.NoError(t, err)
+				}, "unable to close client before timeout")
+			}
+
+			// The hook runs once per batch, not once per subscriber.
+			assert.GreaterOrEqual(t, customModule.HookCallCount.Load(), int32(1))
```
endigma, line 132: Shouldn't we be able to assert this call count exactly? This assertion does not match its comment since anything >1 passes.

router-tests/modules/streams_on_broadcast_events_test.go
```diff
@@ -0,0 +1,437 @@
...
+	t.Parallel()
+
+	const Timeout = time.Second * 10
+
+	type kafkaSubscriptionArgs struct {
+		dataValue []byte
+		errValue  error
+	}
+
+	t.Run("Test Broadcast hook could change events for all subscribers", func(t *testing.T) {
```
endigma: I would suggest forming all test names in the form Action + Condition + Expectation<br><br>All of these seem to both have the word test in them which will print as `TestBroadcastHook/Test_broadcast_hook_could_change_events_for_all_subscribers` which is not very descriptive.

router/pkg/pubsub/datasource/subscription_event_updater_onbroadcastevents_test.go
```diff
@@ -0,0 +1,747 @@
+package datasource
+
+// These tests are a reflection of tests in subscription_event_updater_onreceiveevents_test.go,
+// adjusted for the OnBroadcastEvents hook.
```
endigma: Is this comment necessary

endigma (changes requested):

router-tests/modules/streams_before_events_dispatch_test.go
dkorittki: `KafkaPublishUntilReceived` can publish more than one message, so hypthetically there can be more. The comment is stale though. WIll fix that.

router/pkg/pubsub/datasource/subscription_event_updater_onbroadcastevents_test.go
dkorittki: Nope, removed it

router-tests/modules/streams_on_broadcast_events_test.go
dkorittki: Yeah I removed that and used the hooks name

router-tests/modules/streams_before_events_dispatch_test.go
dkorittki: Changed the comment

endigma (approved):

alepane21 (commented):
What about beforePublishEvents instead of onBroadcastEvents? I think we should also update the documentation.

alepane21 (approved):

PR #3111, opened 2026-07-24
feat(mcp): add server/discover support with configurable instructions

asoorm:
## Motivation and Context

The MCP spec revision `2026-07-28` (draft) introduces [`server/discover`](https://modelcontextprotocol.io/specification/draft/server/discover) (SEP-2575) as the stateless successor to the `initialize` handshake. This PR makes the router MCP server answer it, and gives operators control over the server identity it reports.

Implements ROUTER-90.

## What's changed

- **go-sdk bump `v1.6.1` to `v1.7.0`** (router + router-tests): the SDK auto-handles `server/discover`, returning `resultType: complete`, `supportedVersions`, `capabilities`, `serverInfo`, `ttlMs`, `cacheScope`, and optional `instructions`.
- **New `mcp.server.discover.instructions`** (env: `MCP_SERVER_DISCOVER_INSTRUCTIONS`): natural-language guidance served to AI agents in the `server/discover` response. While the legacy `initialize` handshake remains in its deprecation window, the same instructions are also served there, so older clients benefit too.
- **New `mcp.server.version`** (env: `MCP_SERVER_VERSION`): the version reported in `serverInfo` alongside the graph-derived server name, so operators can expose their own API version to agents. Defaults to the router release version (previously a hardcoded `"0.0.1"`).
- **New `mcp.server.title` / `mcp.server.description`** (env: `MCP_SERVER_TITLE` / `MCP_SERVER_DESCRIPTION`): human-readable display name and description reported in `serverInfo` — MCP clients show `title` in UIs instead of the machine name `wundergraph-cosmo-<graph>`.
- Fixed: an empty graph name no longer produces the malformed serverInfo name `wundergraph-cosmo-`.
- Docs: new *Server Discovery* section in `docs-website/router/mcp/configuration.mdx`.

## Notes

- `ttlMs`/`cacheScope` are emitted with the SDK's spec-compliant defaults (`0`/`"public"`) and are not yet configurable — blocked on [modelcontextprotocol/go-sdk#1092](https://github.com/modelcontextprotocol/go-sdk/issues/1092) (no `DiscoverHandler` hook). Config for these should be added under `mcp.server.discover` when the hook lands, ideally with the final v1.7.0 bump.
- Protocol `2026-07-28` is only advertised over streamable HTTP in stateless mode (SDK transport gating, since the new protocol replaces sessions with per-request `_meta`). `session.stateless` defaults to `true`, so this is the out-of-the-box behavior; explicit session-based deployments (`session.stateless: false`) negotiate `2025-11-25` or older for now.
- The SDK bump changed `ToolAnnotations` serialization (`ReadOnlyHint`/`IdempotentHint` no longer `omitempty`); router-tests expectations updated accordingly.

## Checklist

- [x] Tests added: `router/pkg/mcpserver/discover_test.go` (in-memory client over the SEP-2575 discover path), `router-tests/protocol/mcp_discover_test.go` (raw `server/discover` JSON-RPC round-trips: instructions, mandatory fields, stateless version advertising, serverInfo version override + fallback), config parsing/env/schema tests.
- [x] Documentation updated (`docs-website/router/mcp/configuration.mdx`).
- [x] Bumped go-sdk to the final `v1.7.0` release (spec `2026-07-28` shipped Jul 28).

## Summary by CodeRabbit

* **New Features**
  * Added configurable MCP server identity metadata (version, title, description) exposed via `serverInfo`.
  * Added `server/discover` support with configurable discover instructions and protocol-version advertisement/negotiation behavior.
  * Added environment variables for MCP identity/discovery configuration.
* **Documentation**
  * Updated MCP configuration reference with the new discovery fields and protocol-version rules.
* **Configuration**
  * Added event startup resiliency (`events.skip_unavailable_providers`) and pre-dispatch handling controls (`before_events_dispatch.handler_timeout`).
* **Tests**
  * Expanded discovery and `serverInfo` tests; updated tool-discovery expectations to match updated SDK serialization behavior.


router/pkg/mcpserver/server.go
```diff
@@ -281,11 +299,14 @@ func NewGraphQLSchemaServer(ctx context.Context, routerGraphQLEndpoint string, o
 	// Create the MCP server with all options
 	mcpServer := mcp.NewServer(
 		&mcp.Implementation{
-			Name:    "wundergraph-cosmo-" + strcase.ToKebab(options.GraphName),
-			Version: "0.0.1",
+			Name:        "wundergraph-cosmo-" + strcase.ToKebab(cmp.Or(options.GraphName, "graph")),
```
endigma: `Options.GraphName` already defaults to `"graph"` at line 223, so this repeats the literal. It also only covers the `serverInfo` name: `s.graphName` on line 325 still receives `""`, so the `graph_name` log field on line 532 stays empty.

Guarding the setter instead fixes both consumers and keeps the default in one place.

```go
// line 302
Name:        "wundergraph-cosmo-" + strcase.ToKebab(options.GraphName),

// lines 381-386
// WithGraphName sets the graph name. An empty name keeps the default.
func WithGraphName(graphName string) func(*Options) {
	return func(o *Options) {
		o.GraphName = cmp.Or(graphName, o.GraphName)
	}
}
```

router/pkg/mcpserver/server.go
```diff
@@ -329,6 +350,34 @@ func (s *GraphQLSchemaServer) SetHTTPClient(client *http.Client) {
...
+		o.Instructions = instructions
+	}
+}
+
+// WithServerVersion sets the version reported in the MCP serverInfo
+func WithServerVersion(version string) func(*Options) {
+	return func(o *Options) {
+		o.ServerVersion = version
+	}
+}
```
endigma: Same clobber hazard as the graph name, but not covered. `router.go:1205` always calls this setter, so the `ServerVersion: "dev"` default on line 231 is always overwritten.

`core.Version` is injected through ldflags (`Makefile:43`), so an empty `VERSION` yields `""`. `mcp.Implementation.Version` carries `json:"version"` with no `omitempty`, so the server would advertise an empty required field.

```suggestion
// WithServerVersion sets the version reported in the MCP serverInfo.
// An empty version keeps the default.
func WithServerVersion(version string) func(*Options) {
	return func(o *Options) {
		o.ServerVersion = cmp.Or(version, o.ServerVersion)
	}
}
```

router-tests/protocol/mcp_discover_test.go
```diff
@@ -0,0 +1,211 @@
...
+	// stream depending on negotiation; support both.
+	payload := string(body)
+	if strings.Contains(resp.Header.Get("Content-Type"), "text/event-stream") {
+		for _, line := range strings.Split(payload, "\n") {
+			if strings.HasPrefix(line, "data:") {
+				payload = strings.TrimSpace(strings.TrimPrefix(line, "data:"))
+				break
+			}
+		}
+	}
```
endigma: The plain-JSON branch is unreachable. `mcpserver/server.go:499` never sets `StreamableHTTPOptions.JSONResponse`, and the SDK computes `useSSE := !c.jsonResponse || isSubscriptionsListen` (`streamable.go:1650`), so every POST reply is SSE. `server/discover` takes the same path, it is only exempt from the session check (`streamable.go:1516`), not from the response encoding.

Assert the content type instead, so a transport change fails the test rather than silently switching branches.

```suggestion
	// The streamable HTTP transport answers POSTs with an SSE stream: the router
	// does not set StreamableHTTPOptions.JSONResponse.
	require.Contains(t, resp.Header.Get("Content-Type"), "text/event-stream")

	var payload string
	for line := range strings.SplitSeq(string(body), "\n") {
		if after, ok := strings.CutPrefix(line, "data:"); ok {
			payload = strings.TrimSpace(after)
			break
		}
	}
	require.NotEmpty(t, payload, "no SSE data frame in response: %s", string(body))
```

router-tests/protocol/mcp_discover_test.go
```diff
@@ -0,0 +1,211 @@
...
+				},
+				sdkmcp.MetaKeyClientCapabilities: map[string]any{},
+			},
+		},
+	}
+
+	requestBody, err := json.Marshal(discoverRequest)
+	require.NoError(t, err)
+
+	req, err := http.NewRequest(http.MethodPost, xEnv.GetMCPServerAddr(), bytes.NewReader(requestBody))
```
endigma: No context, so the request does not cancel with the test.

```suggestion
	req, err := http.NewRequestWithContext(t.Context(), http.MethodPost, xEnv.GetMCPServerAddr(), bytes.NewReader(requestBody))
```

router-tests/protocol/mcp_discover_test.go
```diff
@@ -0,0 +1,211 @@
...
+				Enabled: true,
+			},
+		}, func(t *testing.T, xEnv *testenv.Environment) {
+			result := postServerDiscover(t, xEnv)
+
+			assert.Equal(t, "complete", result["resultType"])
+			assert.NotEmpty(t, result["supportedVersions"])
+			assert.Contains(t, result, "ttlMs")
+			assert.Equal(t, "public", result["cacheScope"])
+			assert.NotEmpty(t, result["capabilities"])
```
endigma, line 128: `assert.Contains` checks presence only. `Cacheable.TTLMs` is an `int` tagged `json:"ttlMs"` with no `omitempty`, and `setDefaultCacheableValues` sets only `CacheScope`, so the value is a known `0`. Assert it.

```suggestion
			assert.Equal(t, "complete", result["resultType"])
			assert.NotEmpty(t, result["supportedVersions"])
			assert.Equal(t, float64(0), result["ttlMs"])
			assert.Equal(t, "public", result["cacheScope"])
			assert.NotEmpty(t, result["capabilities"])
```

router-tests/protocol/mcp_discover_test.go
```diff
@@ -0,0 +1,211 @@
...
+			},
+		}, func(t *testing.T, xEnv *testenv.Environment) {
+			result := postServerDiscover(t, xEnv)
+
+			assert.Contains(t, result["supportedVersions"], protocolVersion20260728)
+		})
+	})
+}
+
+func TestMCPServerInfoVersion(t *testing.T) {
```
endigma: The name says `Version`, but one subtest covers title and description. The `_meta` to `serverInfo` unwrap also repeats verbatim in all three subtests.

```suggestion
// discoverServerInfo returns the serverInfo object carried in the discover
// result _meta.
func discoverServerInfo(t *testing.T, xEnv *testenv.Environment) map[string]any {
	t.Helper()

	meta, ok := postServerDiscover(t, xEnv)["_meta"].(map[string]any)
	require.True(t, ok, "discover result should carry _meta")
	serverInfo, ok := meta[sdkmcp.MetaKeyServerInfo].(map[string]any)
	require.True(t, ok, "_meta should carry serverInfo")

	return serverInfo
}

func TestMCPServerInfo(t *testing.T) {
```

Each subtest body then collapses to `serverInfo := discoverServerInfo(t, xEnv)`.

router/pkg/mcpserver/discover_test.go
```diff
@@ -0,0 +1,108 @@
...
+		append([]func(*Options){
+			WithLogger(zap.NewNop()),
+			WithOperationsDir(tempDir),
+		}, opts...)...,
+	)
+	require.NoError(t, err)
+	require.NoError(t, srv.Reload(&schemaDoc, nil))
+
+	return srv
+}
```
endigma: `newTestServer` and `connectTestClient` are called as a pair in all five tests, and `srv` is never used afterwards. One helper could cover it.

router/pkg/mcpserver/discover_test.go
```diff
@@ -0,0 +1,108 @@
...
+func TestDiscover_EmptyGraphNameFallsBackToDefault(t *testing.T) {
+	// WithGraphName("") must not clobber the default: an empty graph name would
+	// produce the malformed serverInfo name "wundergraph-cosmo-".
+	srv := newTestServer(t, WithGraphName(""))
+	cs := connectTestClient(t, srv)
+
+	serverInfo := cs.InitializeResult().ServerInfo
+	require.NotNil(t, serverInfo)
+	assert.Equal(t, "wundergraph-cosmo-graph", serverInfo.Name)
+}
```
endigma: Five top-level tests for one subject, with underscores in the names. One top-level test plus lowercase sentence subtests would be cleaner?

router/pkg/config/config_test.go
```diff
@@ -2261,3 +2261,102 @@ persisted_operations:
...
+  server:
+    title: "My Commerce API"
+    description: "Query products, orders and customers."
+`)
+	cfg, err := LoadConfig([]string{f})
+	require.NoError(t, err)
+
+	require.Equal(t, "My Commerce API", cfg.Config.MCP.Server.Title)
+	require.Equal(t, "Query products, orders and customers.", cfg.Config.MCP.Server.Description)
+}
```
endigma: Same point: five top-level tests over one config block. Neighbors in this file (`TestCostControlConfig`, `TestPQLManifestConfig`) already use one parent with subtests.

The parent stays serial on purpose — `t.Setenv` rejects a test with parallel ancestors.

```suggestion
func TestMCPServerConfig(t *testing.T) {
	t.Run("reads discover instructions from yaml", func(t *testing.T) {
		t.Parallel()

		f := createTempFileFromFixture(t, `
version: "1"

graph:
  token: "token"

mcp:
  enabled: true
  server:
    discover:
      instructions: "Use the search tools before executing operations."
`)
		cfg, err := LoadConfig([]string{f})
		require.NoError(t, err)

		require.Equal(t, "Use the search tools before executing operations.", cfg.Config.MCP.Server.Discover.Instructions)
	})

	t.Run("leaves discover instructions empty when unset", func(t *testing.T) {
		t.Parallel()

		f := createTempFileFromFixture(t, `
version: "1"

graph:
  token: "token"

mcp:
  enabled: true
`)
		cfg, err := LoadConfig([]string{f})
		require.NoError(t, err)

		require.Empty(t, cfg.Config.MCP.Server.Discover.Instructions)
	})

	t.Run("reads discover instructions from env", func(t *testing.T) {
		t.Setenv("MCP_SERVER_DISCOVER_INSTRUCTIONS", "Env-provided guidance.")

		f := createTempFileFromFixture(t, `
version: "1"

graph:
  token: "token"

mcp:
  enabled: true
`)
		cfg, err := LoadConfig([]string{f})
		require.NoError(t, err)

		require.Equal(t, "Env-provided guidance.", cfg.Config.MCP.Server.Discover.Instructions)
	})

	t.Run("reads the server version from yaml", func(t *testing.T) {
		t.Parallel()

		f := createTempFileFromFixture(t, `
version: "1"

graph:
  token: "token"

mcp:
  enabled: true
  server:
    version: "1.4.0"
`)
		cfg, err := LoadConfig([]string{f})
		require.NoError(t, err)

		require.Equal(t, "1.4.0", cfg.Config.MCP.Server.Version)
	})

	t.Run("reads the server title and description from yaml", func(t *testing.T) {
		t.Parallel()

		f := createTempFileFromFixture(t, `
version: "1"

graph:
  token: "token"

mcp:
  enabled: true
  server:
    title: "My Commerce API"
    description: "Query products, orders and customers."
`)
		cfg, err := LoadConfig([]string{f})
		require.NoError(t, err)

		require.Equal(t, "My Commerce API", cfg.Config.MCP.Server.Title)
		require.Equal(t, "Query products, orders and customers.", cfg.Config.MCP.Server.Description)
	})
}
```

docs-website/router/mcp/configuration.mdx
```diff
@@ -83,6 +91,26 @@ mcp:
 
 A storage provider **must** be specified to load GraphQL operations. See [Storage Providers](/router/storage-providers) for more details on configuring storage providers.
 
+## Server Discovery
+
+The MCP server implements the `server/discover` method (introduced in MCP protocol version `2026-07-28` as the successor to the `initialize` handshake). Clients use it to query the server's supported protocol versions, capabilities, and identity in a single stateless request.
+
+The **instructions** option lets you provide natural-language guidance that tells AI agents how to use your server effectively:
```
endigma: The first sentence runs to 27 words. `**instructions**` also breaks this file's convention of backticks for config keys, and it names a key that does not exist at that path.

```suggestion
The MCP server implements the `server/discover` method. This method arrived in MCP protocol version `2026-07-28` as the successor to the `initialize` handshake. Clients call it to read the supported protocol versions, capabilities, and identity in one stateless request.

The `discover.instructions` option sends natural-language guidance to MCP clients. Use it to tell AI agents how to work with your graph:
```

docs-website/router/mcp/configuration.mdx
```diff
@@ -83,6 +91,26 @@ mcp:
...
+  server:
+    discover:
+      instructions: |
+        Prefer the pre-defined operation tools over execute_graphql.
+        Employee data is refreshed nightly; do not treat it as real-time.
+```
+
+Instructions are served to every MCP client, whether it connects via `server/discover` or the legacy `initialize` handshake.
+
+Protocol version `2026-07-28` is advertised by default. In session-based mode (`session.stateless: false`), clients negotiate `2025-11-25` or older.
```
endigma: Passive voice with a known actor in both paragraphs.

```suggestion
The server sends the instructions to every MCP client. This covers clients that connect with `server/discover` and clients that still use the legacy `initialize` handshake.

The router advertises protocol version `2026-07-28` by default. In session-based mode (`session.stateless: false`), clients negotiate `2025-11-25` or older.
```

router/pkg/mcpserver/server.go
```diff
@@ -281,11 +299,14 @@ func NewGraphQLSchemaServer(ctx context.Context, routerGraphQLEndpoint string, o
...
-			Version: "0.0.1",
+			Name:        "wundergraph-cosmo-" + strcase.ToKebab(cmp.Or(options.GraphName, "graph")),
+			Title:       options.ServerTitle,
+			Description: options.ServerDescription,
+			Version:     options.ServerVersion,
 		},
 		&mcp.ServerOptions{
-			PageSize: 100,
+			Instructions: options.Instructions,
+			PageSize:     100,
```
endigma, line 309: Nit: the `ttlMs` and `cacheScope` limitation lives only in the PR description today. Worth one line here so it survives.

```suggestion
		&mcp.ServerOptions{
			Instructions: options.Instructions,
			PageSize:     100,
			// ttlMs and cacheScope use the SDK defaults (0 / "public"). They
			// become configurable under mcp.server.discover once the SDK adds a
			// DiscoverHandler hook: modelcontextprotocol/go-sdk#1092.
```

endigma (dismissed):

dkorittki (approved):

Issue #2772, opened 2026-04-16
S3 providers should support temporary AWS credentials (STS/IRSA)


MKL2000:
### Component(s)

router, controlplane

### Is your feature request related to a problem? Please describe.

Self-hosted Cosmo cannot use AWS temporary credentials (STS AssumeRole, IRSA, EKS Pod Identity) for S3 storage. This is the standard credential model in most Kubernetes environments, and the lack of support blocks adoption for anyone following AWS security best practices.

There are two root causes:

1. **No default credential chain fallback.** The Node.js components (controlplane + CDN server) throw an error when credentials are empty, preventing the `@aws-sdk` default credential provider chain from being used. This blocks IRSA and other IAM-based auth that requires zero explicit credential configuration.
2. **No session token passthrough.** The router, controlplane, and CDN server all accept `accessKeyId` / `secretAccessKey` but there is no config field for a `sessionToken`. All AWS temporary credentials consist of a three-tuple (access key + secret key + session token); the token is required at the AWS signing layer, even though in most deployments the SDK can obtain it transparently (see Fix 1). This matters when credentials are provided explicitly via Cosmo's own config.

Related: #2159 (covers CDN server IRSA only — this issue expands to all components + adds session token passthrough).

### Describe the solution you'd like

Two complementary fixes. Priority depends on the component:

- **Node.js components (controlplane + CDN server):** Fix 1 alone covers the most common deployments (IRSA, env vars, instance metadata). Fix 2 is a nice-to-have.
- **Go router:** Fix 1 provides only partial coverage due to `minio-go`'s limited built-in credential chain (see note below). Fix 2 is the more impactful change, letting operators pass STS credentials from an external source.

**Fix 1 — Default credential chain fallback (Node.js)**

When `accessKeyId` and `secretAccessKey` are both empty, omit the `credentials` field instead of throwing. The `@aws-sdk` SDK will then use its [default credential provider chain](https://docs.aws.amazon.com/sdkref/latest/guide/standardized-credentials.html#credentialProviderChain):

1. Environment variables (`AWS_ACCESS_KEY_ID` + `AWS_SECRET_ACCESS_KEY` + `AWS_SESSION_TOKEN`)
2. Shared credentials file (`~/.aws/credentials`)
3. IRSA / EKS Pod Identity — SDK internally calls STS AssumeRoleWithWebIdentity and manages session token rotation transparently
4. EC2 / ECS instance metadata (IMDSv2)

[...]

Issue #2769, opened 2026-04-16
feat: Add AWS Session Token (STS) and default credential chain support for all S3 storage providers


markli-zd:
### Component(s)

router, controlplane, cdn-server

### Is your feature request related to a problem? Please describe.

When deploying Cosmo in Kubernetes environments that use temporary AWS credentials (STS AssumeRole, IRSA, or credential injection sidecars like Zendesk's temp-auth), S3 storage providers cannot authenticate because:

1. There is no way to pass a **session token** alongside access key / secret key.
2. The Node.js components (controlplane + CDN server) **throw an error** when credentials are empty, preventing fallback to the `@aws-sdk` default credential provider chain.

This blocks self-hosted Cosmo deployments in any Kubernetes environment where temporary/rotated credentials are standard.

**Affected code paths (verified against `0b0d42d6e`):**

**Router (Go / minio-go):** The `Static` → `IAM` credential chain does not set `SessionToken` on the Static provider.
- `router/pkg/routerconfig/s3/client.go` — `credentials.Static` (lines 47–52)
- `router/internal/persistedoperation/operationstorage/s3/client.go` — `credentials.Static` (lines 57–62)
- `router/pkg/config/config.go` — `S3StorageProvider` struct has `AccessKey`/`SecretKey` but no `SessionToken`

**Controlplane (Node.js / @aws-sdk):** `createS3ClientConfig()` only passes `accessKeyId` and `secretAccessKey`, and throws an error when either is empty.
- `controlplane/src/core/util.ts` — `createS3ClientConfig()` (lines 395–421)

**CDN Server (Node.js / @aws-sdk):** Identical `createS3ClientConfig()` with the same limitations.
- `cdn-server/src/utils.ts` — `createS3ClientConfig()` (lines 18–44)

Related: #2159 (covers CDN server IRSA only — this issue expands to all components + adds sessionToken passthrough)

### Describe the solution you'd like

Two complementary fixes:

#### Fix 1: Add `sessionToken` passthrough

**Router (Go):** Add `SessionToken` field to `S3StorageProvider` config and both `ClientOptions` structs, pass through to `credentials.Static`:

```go
&credentials.Static{
    Value: credentials.Value{
        AccessKeyID:     options.AccessKeyID,
        SecretAccessKey: options.SecretAccessKey,
        SessionToken:    options.SessionToken,  // new
        SignerType:      credentials.SignatureV4,
    },
}
```

[...]

PR #3116, opened 2026-07-27
fix: allow AWS default credential chain for S3 storage

### LINKED ISSUE #2159: Add IAM IRSA Support for S3 Authentication in CDN Server
### Component(s)

_No response_

When running the CDN server in Amazon EKS, users currently cannot leverage IAM roles for service accounts (IRSA) for S3 authentication. The server requires explicit S3 access key ID and secret access key environment variables, which forces users to either hardcode credentials or manage them separately from their Kubernetes IAM roles. This creates security and operational overhead for EKS deployments.

Modify the S3 client configuration in the CDN server to support the AWS SDK's default credential provider chain when static credentials are not provided. This would allow the server to automatically use IRSA credentials when running in EKS, while maintaining backward compatibility for users who prefer to use static credentials.
The changes would:
1. Make S3 credentials optional in the configuration
2. Only set config.credentials when both access key and secret are explicitly provided
3. Fall back to the AWS SDK's default credential provider chain (which includes IRSA support) when credentials are not set
4. Maintain support for region configuration through environment variables or URL parameters

_No response_

### Additional context

_No response_

davideugui1:
### Description

Removes the hard requirement for an explicit S3 access key ID and secret access key in `createS3ClientConfig()` (`cdn-server/src/utils.ts` and `controlplane/src/core/util.ts`), so that deployments without static credentials configured can fall back to the AWS SDK's default credential provider chain — most notably **IRSA (IAM Roles for Service Accounts)** on EKS, but also plain environment variables and EC2/ECS instance metadata.

Closes #2159 
Related to #2772 - Fix 1 scope 
Related to #2769, which was closed

### Why

Several users run the router/controlplane/cdn-server on EKS with IRSA and don't want to manage long-lived static S3 credentials. Because `createS3ClientConfig()` always set an explicit `credentials` object (and threw if access key/secret were blank) the AWS SDK never got the chance to use its default provider chain, even when the pod's service account already had a valid IAM role attached.

### What changed

- `cdn-server/src/utils.ts` / `controlplane/src/core/util.ts`: `createS3ClientConfig()` now builds the config object and only                                                                                                             attaches `credentials` when both `accessKeyId` and `secretAccessKey` are non-empty. When neither is provided, the                                                                                                             field is omitted entirely, letting the AWS SDK's `fromNodeProviderChain` resolve credentials on its own (IRSA, env vars, instance metadata, etc). The existing check that `region` is required is unchanged. 
- `controlplane/test/utils.s3storage.test.ts`: added coverage for the new behavior — omitting credentials when username/password are empty strings, omitting credentials when they're absent entirely from both the URL and opts, and confirming the region-required check still throws even without explicit credentials. 

## Summary by CodeRabbit

* **Bug Fixes**
  * S3 storage configuration now works without explicitly provided credentials by using available runtime credentials.
  * Missing or incomplete access credentials no longer prevent S3 configuration from being created.
  * Missing S3 regions continue to produce a clear validation error.
  * S3 configuration behavior is now consistent across supported services.
  * Added validation for subgraph routing URLs across event-driven, subscription, gRPC, and feature subgraphs.


fiam:
@gmas @JivusAyrus tagging your for review PTAL

gmas (approved):

Aenimus (approved):
LGTM!

PR #3133, opened 2026-07-31
fix(router): close Kafka consumer client when subscription ends

mwisner:
## Summary by CodeRabbit

* **Bug Fixes**
  * Kafka subscriptions now shut down promptly when canceled or when polling ends.
  * Improved resource cleanup prevents consumer connections from remaining open.
  * Idle subscriptions are reliably unblocked during cancellation, helping avoid lingering background activity.
  * Subscription shutdown is handled safely to prevent duplicate cleanup and improve connection stability.
  * Polling and subscription metrics now stop consistently when a subscription is canceled.

## Description

`kafka.ProviderAdapter.Subscribe` creates a dedicated `kgo.Client` per subscription and never closes it. When a subscription ends, the poller goroutine returns and the client is simply dropped — `Close()` is never called. Because a franz-go client owns background goroutines that keep it reachable, the client is never garbage collected, so **every ended subscription permanently leaks a `kgo.Client` together with its broker connections and buffered fetches**. Heap usage and goroutine count grow with the *cumulative* (not concurrent) number of subscriptions until the process is OOM-killed. The only consumer teardown in the adapter today is for the shared producer (`writeClient`), closed in `Shutdown`.

A second, compounding issue: `topicPoller` blocks in `PollRecords` on the adapter (application) context, not the per-subscription context. A subscription cancelled while its topic is idle therefore never unblocks the poller, so neither the goroutine nor its client is reclaimed even at the loop level.

### How to reproduce

- Point a Kafka EDFS subscription at a topic and churn subscriptions (repeated connect/disconnect, or many short-lived subscribers).
- `inuse_space` heap and `goroutine` count climb monotonically with the total number of subscriptions ever created and never fall — independent of how many subscribers are *currently* active.
- goroutine/heap profiles show accumulating franz-go client stacks (broker read/write loops and `PollRecords`).

## Fix

In the `Subscribe` poller goroutine:

- Close the consumer client on every exit path via a `sync.Once`-guarded `defer` (`Client.Close` is not safe to call twice).
- Register `context.AfterFunc(ctx, closeClient)` on the subscription context, so that when the subscription is cancelled an in-flight `PollRecords` returns `IsClientClosed` and the poller exits promptly — even on an otherwise idle topic.

As a side benefit, `Shutdown` now also reclaims consumer clients through the existing `closeWg.Wait()`. The change is minimal (no new imports, no public API change) and preserves the stateless-consumer model.

## Checklist

## Open Source AI Manifesto

---

### Notes for reviewers

Opening as a **draft** for discussion. This is a bugfix rather than a feature, so no prior issue was opened — happy to file one if you'd prefer. I'd also be glad to add a regression test asserting the consumer client is closed on subscription-context cancellation (e.g. that `PollRecords` returns `IsClientClosed` and the poller goroutine exits) before this is marked ready.

Prepared with assistance from Claude (per the Open Source AI Manifesto's transparency principle).


dkorittki:
Hey @mwisner, thanks for the finding! I had a look at this and was able to reproduce the issue. Are you open to a few adjustments in this PR? Your fix works but ideally I would like to solve it differently: Make the topic poller return immidiately by using their contexts correctly and close the kafka client alongside it.

Currently the topic poller mixes up using `p.ctx` and `ctx` and its blocking for longer than it should be. You could create a new context, which cancels when either `ctx` or `p.ctx` is cancelled. You could do this right after the go routine is started:

```go
p.closeWg.Go(func() {
	defer client.Close()

	pollerCtx, cancel := context.WithCancel(ctx) // cancels pollerCtx when ctx is cancelled
	defer cancel()
	stop := context.AfterFunc(p.ctx, cancel) // cancels pollerCtx when p.ctx is cancelled, too
	defer stop()
	
	err := p.topicPoller(pollerCtx, ...)
	// [...]
}
```

Inside `p.topicPoller` get rid of all direct `p.ctx` references like `case <-p.ctx.Done()` and `client.PollRecords(p.ctx, 10_000)`, etc. The only context being used in there is via its `ctx` argument.

This will lead to the poller goroutine exiting immidiately when either a trigger closes or when the router shuts down or hot reloads. When that happens the `defer client.Close()` ensures the kafka client is closed as well.

If you want I can also close this PR, cherry-pick your commits here into a new PR with these changes. Just let me know.

mwisner:
Thanks for reproducing it and for the guidance — that approach is cleaner, applied in the latest push:

- `topicPoller` no longer references `p.ctx` at all: `PollRecords` and the `streamMetricStore.Consume` call both use the passed `ctx`.
- In `Subscribe`, the goroutine derives `pollerCtx, cancel := context.WithCancel(ctx)` and adds `context.AfterFunc(p.ctx, cancel)`, so the poller context is cancelled when either the subscription context or the application context is cancelled.
- A single `defer client.Close()` reclaims the client once the poller returns, so the `sync.Once` guard is gone.

Result: the poller returns immediately on a trigger close, router shutdown, or hot reload, and the client is closed alongside it. `go build`/`vet`/`test ./pkg/pubsub/kafka/` are green.

Happy to keep it in this PR (no need to re-open a new one), and I can add a regression test asserting the poller exits and the client is closed on subscription-context cancellation if you'd like it here.

dkorittki (approved):
Looks good, thanks! I'll merge it now and provide tests afterwards

dkorittki:
The failing test prevents this PR from being merged but it doesn't have anything to do with your PR. I'll clarify whats going on.

dkorittki:
Failing CI fixed. Merged your PR :-)

PR #3135, opened 2026-08-03
fix(router): ignore header rules for pubsub trigger sources

dkorittki:
9 LOC actual bug fix, the remainder is test code.

Fixes a bug on the routers header rule builder. When a user configures the router to propagate headers to all subgraphs

```yaml
headers:
  all:
    request:
      - op: propagate
        named: X-Test
```

it adds this rule to every subgraph, which includes the Cosmo Streams trigger source. Using the demo environment and performing a query 

```graphql
subscription { employeeUpdatedMyKafka(employeeID:1) { details { forename } hobbies { employees { id } } } }
```

The header builder builds available headers per subgraph in its map 

```
map["<uuid>-kafka"] = { propagate X-Test }
map["employees"] = { propagate X-Test }
map["hobbies"] = { propagate X-Test }
```

The first one is the EDFS trigger source and it does not make sense. In reality this is a message broker. When the router connects to them it does not forward any headers. So having header rules for it does not make sense.

The fix is to avoid the first map entry but keep the others. This way subgraph fetches during resolving still send headers.

This caused a bug, which is how I became aware of it. Headers for trigger sources are used as part of the trigger id hash. Two clients on the same subscription with different headers landed on different triggers. It meant they listened for the same events on the same message queue with two different triggers. Having seperate triggers is necesarry for normal subscriptions but not for Cosmo Streams. Removing the above mentioned first line avoids the hash generation in the engine from considering header values.

EDIT: Had to fix a single flight test assumption that edfs triggers are not deduplicated on different headers.
In contrast to normal subscriptions they now will. The test now expects one trigger will still ensuring deduplication
to subgraphs during resolving still works.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved event-driven subscription handling so propagated client headers no longer affect trigger identity.
  * Preserved correct header behavior for regular responses and other subscription types.
  * Improved deduplication when subscriptions use identical or different authorization headers.

* **Tests**
  * Added coverage for WebSocket subscriptions and authorization headers supplied through connection or payload data.
  * Added validation for consistent event-driven subscription behavior and header-independent trigger matching.

## Checklist

## Open Source AI Manifesto


dkorittki:
By the way there is a very similar pull request, which fixed a very similar issue https://github.com/wundergraph/cosmo/pull/2950 .

In contrast to this one it fixed any headers from the websocket `initial_data` to become part of the triggers `input` byte object. This one doesn't do anything to `input`. The trigger hash generator in the engine calls the header builder explicitely, after examining `input`. It's a seperate issue.

alepane21 (approved):

PR #3137, opened 2026-08-04
feat: in memory adapter for entity caching cache

SkArchon:
This PR introduces the in memory adapter for entity caching cache based on the interface defined here in this PR https://github.com/wundergraph/graphql-go-tools/pull/1618

## Summary by CodeRabbit

* **New Features**
  * Added an in-memory entity cache with batch reads and writes, ordered multi-key results, and per-entry expiration.
  * Rejects batches containing entries without a positive expiration time.
  * Protects cached values from unintended modification and automatically removes expired entries.
* **Bug Fixes**
  * Ensures cache operations honor context cancellation and deadlines.
* **Tests**
  * Added comprehensive coverage for expiration, eviction, ordering, overwrites, empty values, and cache compatibility.
* **Chores**
  * Updated the GraphQL tools dependency.

## Checklist

## Open Source AI Manifesto


router/pkg/entitycaching/cache/in_memory.go
```diff
@@ -0,0 +1,119 @@
...
+	"fmt"
+	"sync"
+	"time"
+
+	enginecache "github.com/wundergraph/graphql-go-tools/v2/pkg/entitycaching"
+)
+
+// ErrMissingTTL reports items that were rejected because they carried no
+// positive TTL. Every cached entry expires, so a TTL is mandatory.
+var ErrMissingTTL = errors.New("cache item requires a positive TTL")
```
dkorittki: This error isn't specific to the in memory cache, right? I would define it in the corresponding graphql-go-tools package then and mention in the methods godoc when its returned

router/pkg/entitycaching/cache/in_memory.go
```diff
@@ -0,0 +1,119 @@
...
+	// expiresAt is always set, an entry without an expiry is never stored.
+	expiresAt time.Time
+}
+
+// InMemoryCache sweeps every expired entry on each call, so both methods take
+// the mutex exclusively, hence a plain Mutex rather than an RWMutex.
+// This cache should only be used for tests OR development and never production
+type InMemoryCache struct {
+	mu      sync.Mutex
+	entries map[string]inMemoryEntry
```
dkorittki: Why not use `map[string]enginecache.Item` ?

router/pkg/entitycaching/cache/in_memory.go
```diff
@@ -0,0 +1,119 @@
...
+
+type inMemoryEntry struct {
+	value []byte
+	// expiresAt is always set, an entry without an expiry is never stored.
+	expiresAt time.Time
+}
+
+// InMemoryCache sweeps every expired entry on each call, so both methods take
+// the mutex exclusively, hence a plain Mutex rather than an RWMutex.
+// This cache should only be used for tests OR development and never production
```
dkorittki: This rather belongs to the `mu` field as a godoc instead of `InMemoryCache`.

Also this could become a major bottleneck. Using a normal mutex instead of an RWMutex forces syncronous reads to the cache for all concurrent requests on the router. You can't use an RWMutex because of `sweepExipred` on `GetMany`. Maybe worth checking if you can use Ristretto like mentioned above so you don't need to care about this kind of problem.

I mean yeah it's only meant to be used in tests but if Ristretto can solve it for free and have it production ready for when the situation calls for it.

router/pkg/entitycaching/cache/in_memory.go
```diff
@@ -0,0 +1,119 @@
...
+	// needs no expiry check of its own.
+	c.sweepExpired(c.now())
+
+	results := make([]enginecache.Result, len(keys))
+	for i, key := range keys {
+		entry, ok := c.entries[key]
+		if !ok {
+			continue
+		}
+		results[i] = enginecache.Result{Value: bytes.Clone(entry.value), Found: true}
```
dkorittki: You're not really making use of `Result.Found` because its always true for returned items. Do we really need the `Found` field?

router/pkg/entitycaching/cache/in_memory.go
```diff
@@ -0,0 +1,119 @@
...
+// GetMany returns one result per key, in the same order as keys. It also drops
+// every expired entry, including those no key in this batch asked about.
+func (c *InMemoryCache) GetMany(ctx context.Context, keys []string) ([]enginecache.Result, error) {
+	if len(keys) == 0 {
+		return nil, nil
+	}
+
+	if err := ctx.Err(); err != nil {
+		return nil, err
+	}
```
dkorittki: I'd put this on the top like in `SetMany`

router/pkg/entitycaching/cache/in_memory.go
```diff
@@ -0,0 +1,119 @@
...
+	expiresAt time.Time
+}
+
+// InMemoryCache sweeps every expired entry on each call, so both methods take
+// the mutex exclusively, hence a plain Mutex rather than an RWMutex.
+// This cache should only be used for tests OR development and never production
+type InMemoryCache struct {
+	mu      sync.Mutex
+	entries map[string]inMemoryEntry
+	now     func() time.Time
```
dkorittki: This is only for test purposes right? Have you considered using `testing/synctest`? It lets you mock time inside the runtime, so you don't need to introduce code, which is only there for testing purposes.

https://go.dev/blog/testing-time

dkorittki (commented):
I'd change package structure a bit. We know we will have at least two implementations of `entitycache.Cache`. I'd prefer to have a seperate packages for each

- router/pkg/entitycaching/in_memory/*.go
- router/pkg/entitycaching/redis/*.go

Gives us better namespacing and allows for better type names like `redisEntity.Cache` and `inMemoryEntity.Cache` by aliasing import names on call sites.

Another question: Have you considered using a ristretto cache as the backing cache technology for this implementation? It comes with all the evict logic we need to deal with TTLs.

router/pkg/entitycaching/cache/in_memory.go
SkArchon: cleaned up

router/pkg/entitycaching/cache/in_memory.go
SkArchon: done

router/pkg/entitycaching/cache/in_memory.go
SkArchon: cleaned up

router/pkg/entitycaching/cache/in_memory.go
SkArchon: switched to ristretto

router/pkg/entitycaching/cache/in_memory.go
SkArchon: switched to ristretto

router/pkg/entitycaching/cache/in_memory.go
SkArchon: moved to graphqlgotools

router/pkg/entitycaching/cache/in_memory/in_memory.go
```diff
@@ -0,0 +1,116 @@
...
+	"fmt"
+	"math"
+	"sync"
+
+	"github.com/dgraph-io/ristretto/v2"
+	enginecache "github.com/wundergraph/graphql-go-tools/v2/pkg/entitycaching"
+)
+
+const entryCost = 1
+const maxSize = math.MaxInt32
```
dkorittki: I'd probably go for something much smaller, like a million or so. This blows OOM way before we reach maxInt32

router/pkg/entitycaching/cache/in_memory/in_memory.go
```diff
@@ -0,0 +1,116 @@
...
+
+		results[key] = enginecache.Item{Key: key, Value: bytes.Clone(value), TTL: ttl}
+	}
+
+	return results, nil
+}
+
+// SetMany stores every item, all of which must carry a positive TTL. A single
+// item without one fails the whole batch with an ErrMissingTTL and nothing is
+// stored. It otherwise only fails on a cancelled context.
```
dkorittki: Since it's an implementation of the engines interface method I'd simply write
```suggestion
// SetMany implements enginecache.SetMany.
```

Any behaviour, guard logic and rules should be documented on the interface method godic as a contract for any implementor (imo)

router/pkg/entitycaching/cache/in_memory/in_memory.go
```diff
@@ -0,0 +1,116 @@
...
+		BufferItems:        64,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to create in memory entity cache: %w", err)
+	}
+
+	return &InMemoryCache{cache: cache}, nil
+}
+
+// GetMany returns the entries it found, keyed by the key they were asked for.
```
dkorittki: Since it's an implementation of the engines interface method I'd simply write

```suggestion
// GetMany implements enginecache.GetMany.
```

Any behaviour, guard logic and rules should be documented on the interface method godic as a contract for any implementor (imo)

router/pkg/entitycaching/cache/in_memory/in_memory.go
```diff
@@ -0,0 +1,116 @@
...
+	if len(items) == 0 {
+		return nil
+	}
+
+	// Map used for deduplications
+	last := make(map[string]enginecache.Item, len(items))
+
+	for _, item := range items {
+		if item.TTL <= 0 {
+			return fmt.Errorf("%w: key %q", enginecache.ErrMissingTTL, item.Key)
```
dkorittki, line 95: On the lib version currently referred to by the routers go.mod there is no `ErrMissingTTL`. I think you need to bumb the engine version

router/pkg/entitycaching/cache/in_memory/in_memory.go
```diff
@@ -0,0 +1,116 @@
...
+	for _, item := range last {
+		c.cache.SetWithTTL(item.Key, bytes.Clone(item.Value), entryCost, item.TTL)
+	}
+
+	c.cache.Wait()
+	return nil
+}
+
+// Close stops the goroutines ristretto runs behind the cache. It is safe to
+// call more than once, but not while a GetMany or a SetMany is in flight.
```
dkorittki: > but not while a GetMany or a SetMany is in flight

We can't control this from here right? I checked and Ristretto will actually panic if you close the cache during a concurrent write. Can you secure it with an RWMutex?

dkorittki (changes requested):

router/pkg/entitycaching/cache/in_memory/in_memory.go
SkArchon: It seems this is the case with other cashes, we however close it on the router after waiting for in flight requests to complete. I would probably defer to the same approach.

image

router/pkg/entitycaching/cache/in_memory/in_memory.go
SkArchon: Will be updated to the latest engine after merging.

dkorittki: Just saw you updated it and now there is no more build error -> resolved

dkorittki (approved):

PR #3139, opened 2026-08-05
feat: redis adapter for entity caching cache

SkArchon:
This PR contains a redis adapter to be used for entity caching, currently it does not have support for managing cache tags.

## Summary by CodeRabbit

* **New Features**
  * Added Redis-backed entity caching with namespaced keys.
  * Added efficient batch read and write operations.
  * Added TTL validation, expiration handling, and partial-write error reporting.
  * Added support for Redis-compatible clients and context cancellation.

* **Bug Fixes**
  * Batch operations now reject empty requests and invalid entries consistently.
  * Improved handling of cache misses, expired entries, and Redis errors.

* **Tests**
  * Added comprehensive coverage for caching, expiration, errors, concurrency, and cancellation.

## Checklist

## Open Source AI Manifesto


dkorittki:
@coderabbitai review

router/pkg/entitycaching/cache/redis.go
```diff
@@ -0,0 +1,89 @@
...
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/redis/go-redis/v9"
+	enginecache "github.com/wundergraph/graphql-go-tools/v2/pkg/entitycaching"
+)
+
+// RedisCache stores entries in Redis. Redis owns expiry, so unlike
+// InMemoryCache there is nothing to sweep here.
```
dkorittki: I wouldn't mention the in memory store here. Has nothing to do with this implementation + I always like to mention the interface this intends to implement.

```suggestion
// RedisCache stores entries in Redis. It implements enginecache.Cache
```

router/pkg/entitycaching/cache/redis.go
```diff
@@ -0,0 +1,89 @@
...
+type RedisCache struct {
+	client redis.UniversalClient
+	prefix string
```
dkorittki: Can you add godoc to fields as well? I.e. whats `prefix` good for?

router/pkg/entitycaching/cache/redis.go
```diff
@@ -0,0 +1,89 @@
...
+	// span slots. It also gives each duplicate key its own command, so they are
+	// looked up independently.
+	pipe := c.client.Pipeline()
+	cmds := make([]*redis.StringCmd, len(keys))
+	for i, key := range keys {
+		cmds[i] = pipe.Get(ctx, c.prefix+key)
+	}
+
+	// A miss surfaces as redis.Nil, which is not a failure of the batch.
+	if _, err := pipe.Exec(ctx); err != nil && !errors.Is(err, redis.Nil) {
```
dkorittki: Nit: too much happening in one line. Can you split it up?

```suggestion
	_, err := pipe.Exec(ctx)
	if err != nil && !errors.Is(err, redis.Nil) {
```

router/pkg/entitycaching/cache/redis.go
```diff
@@ -0,0 +1,89 @@
...
+	for i, cmd := range cmds {
+		value, err := cmd.Bytes()
+		if errors.Is(err, redis.Nil) {
+			continue
+		}
+		if err != nil {
+			// There is no partial read to salvage, the whole batch fails.
+			return nil, fmt.Errorf("get %q: %w", keys[i], err)
+		}
+		results[i] = enginecache.Result{Value: value, Found: true}
```
dkorittki: `cmd.Bytes()` does not copy the value but instead returns the value as an `unsafe.Pointer`. As you return this value its possible someone outside this package will modify the value under the hood. Better copy it.


```suggestion
		results[i] = enginecache.Result{Value: bytes.Clone(value), Found: true}
```

router/pkg/entitycaching/cache/redis_test.go
```diff
@@ -0,0 +1,538 @@
...
+	c, err := NewRedisCache(client, testPrefix)
+	require.NoError(t, err)
+
+	return c, mr
+}
+
+func TestRedisCache(t *testing.T) {
+	t.Parallel()
+
+	ctx := context.Background()
```
dkorittki: Use `t.Context()` instead

router/pkg/entitycaching/cache/redis_test.go
```diff
@@ -0,0 +1,538 @@
...
+			c, mr := newTestRedisCache(t)
+
+			err := c.SetMany(ctx, []enginecache.Item{
+				{Key: "a", Value: []byte("value"), TTL: -time.Minute},
+			})
+			require.ErrorIs(t, err, ErrMissingTTL)
+			require.Empty(t, mr.Keys())
+		})
+
+		t.Run("one bad item rejects the whole batch", func(t *testing.T) {
```
dkorittki: This test contradicts the contract defined in the godoc of `SetMany` where it states "An error means an unspecified subset of the items may already have been stored"

router/pkg/entitycaching/cache/redis_test.go
```diff
@@ -0,0 +1,538 @@
...
+				{Key: "no-ttl", Value: []byte("2")},
+				{Key: "after", Value: []byte("3"), TTL: time.Minute},
+			})
+			require.ErrorIs(t, err, ErrMissingTTL)
+			require.ErrorContains(t, err, "no-ttl")
+
+			require.Empty(t, mr.Keys())
+		})
+
+		t.Run("cache tags are accepted and ignored", func(t *testing.T) {
```
dkorittki: Where exactly are cache tags used in this test? I don't know much about these tags but it does not look obvious from looking at the test

router/pkg/entitycaching/cache/redis_test.go
```diff
@@ -0,0 +1,538 @@
...
+		t.Run("reports a failure to reach redis", func(t *testing.T) {
+			t.Parallel()
+
+			c, mr := newTestRedisCache(t)
+			mr.Close()
+
+			err := c.SetMany(ctx, []enginecache.Item{
+				{Key: "a", Value: []byte("value"), TTL: time.Minute},
+			})
+			require.Error(t, err)
```
dkorittki: Use assert.Contains to broadly verify the error is about failure to reach redis

router/pkg/entitycaching/cache/redis_test.go
SkArchon: This is for ttl pre-validation, will rename

router/pkg/entitycaching/cache/redis_test.go
SkArchon: This test should be removed, I removed the cache tag field for now.

router/pkg/entitycaching/cache/redis.go
```diff
@@ -0,0 +1,89 @@
...
+	}
```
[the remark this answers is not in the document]
SkArchon: This is addressed, please check

router/pkg/entitycaching/cache/redis/redis.go
```diff
@@ -0,0 +1,147 @@
...
+	// client is owned by the caller, not this cache, so it is never closed
+	// here. It is a UniversalClient so a single, cluster or sentinel client all
+	// fit without this cache having to know which one it got.
+	client redis.UniversalClient
+	// prefix is prepended to every key before it reaches redis, so entity cache
+	// entries stay in their own namespace and cannot collide with anything else
+	// sharing the instance. It is applied on the way in and stripped back off on
+	// the way out, so callers only ever see the keys they asked with. An empty
+	// prefix is valid and means the keys are used as they are.
+	prefix string
```
fiam, line 24: question: could we skip `prefix` for now? redis does support multiple DBs on the same server/cluster - typically you'd have a dedicated DB per purpose.

router/pkg/entitycaching/cache/redis/redis.go
```diff
@@ -0,0 +1,147 @@
...
+	if client == nil {
+		return nil, errors.New("redis client is nil")
+	}
+
+	return &RedisCache{client: client, prefix: prefix}, nil
+}
+
+// GetMany implements enginecache.GetMany.
+func (c *RedisCache) GetMany(ctx context.Context, keys []string) (map[string]enginecache.Item, error) {
+	if len(keys) == 0 {
```
fiam, line 43: question: `len(keys)` should be an error, shouldn't it? The caller should not invoke us if it has to keys to retrieve IMHO.

router/pkg/entitycaching/cache/redis/redis.go
```diff
@@ -0,0 +1,147 @@
...
+
+// NewRedisCache returns a cache backed by client, namespacing every key with
+// prefix. The caller keeps ownership of client and is responsible for closing
+// it. rediscloser.RDCloser satisfies redis.UniversalClient, so a client built
+// by rediscloser.NewRedisCloser can be passed straight in.
+func NewRedisCache(client redis.UniversalClient, prefix string) (*RedisCache, error) {
+	if client == nil {
+		return nil, errors.New("redis client is nil")
+	}
+
```
fiam, line 37: nit: I would not check for `nil` here, that is part of the function invariant I'd say. Instead, call `client.Ping()` to validate that the database and its credentials are valid.

router/pkg/entitycaching/cache/redis/redis.go
```diff
@@ -0,0 +1,147 @@
...
+	if err != nil && !errors.Is(err, redis.Nil) {
+		return nil, err
+	}
+
+	// Sized for every key finding something, which is the case worth being
+	// ready for. A miss adds nothing, so the map is only as big as the hits.
+	results := make(map[string]enginecache.Item, len(keys))
+	for i, key := range keys {
+		value, err := values[i].Bytes()
+		if errors.Is(err, redis.Nil) {
```
fiam: nit: this would e more performant inside the `if err != nil` arm at line 81, that way in the successful case we elide with check for `nil`

router/pkg/entitycaching/cache/redis/redis.go
```diff
@@ -0,0 +1,147 @@
...
+			continue
+		}
+		if err != nil {
+			// There is no partial read to salvage, the whole batch fails.
+			return nil, fmt.Errorf("get %q: %w", key, err)
```
fiam: nit: maybe prefix these errors with `redis`? e.g.  `redis get: ...`

router/pkg/entitycaching/cache/redis/redis.go
```diff
@@ -0,0 +1,147 @@
...
+		// stored under: the namespace is this cache's business, not theirs.
+		results[key] = enginecache.Item{Key: key, Value: bytes.Clone(value), TTL: ttl}
+	}
+
+	return results, nil
+}
+
+// SetMany implements enginecache.SetMany.
+func (c *RedisCache) SetMany(ctx context.Context, items []enginecache.Item) error {
+	if len(items) == 0 {
```
fiam, line 107: nit: this should be an error IMHO, not return nil

router/pkg/entitycaching/cache/redis/redis_test.go
```diff
@@ -0,0 +1,823 @@
...
+			require.NoError(t, err)
+
+			results, err := c.GetMany(liveCtx, []string{"a"})
+			require.NoError(t, err)
+			require.Contains(t, results, "a")
+		})
+	})
+}
+
+var _ enginecache.Cache = (*RedisCache)(nil)
```
fiam: question: should this be on a `_test.go` file or in the `.go` file? I think the former is more idiomatic because when you see the type you can see that it implements this interface.

router/pkg/entitycaching/cache/redis/redis_test.go
```diff
@@ -0,0 +1,823 @@
...
+import (
+	"context"
+	"errors"
+	"testing"
+	"time"
+
+	"github.com/alicebob/miniredis/v2"
+	"github.com/redis/go-redis/v9"
+	"github.com/stretchr/testify/require"
+	enginecache "github.com/wundergraph/graphql-go-tools/v2/pkg/entitycaching"
```
fiam, line 12: question: could we import this without the alias? i.e. referring to it as `entitycaching`

router/pkg/entitycaching/cache/redis/redis.go
SkArchon: Don't have a strong opinion, makes sense to change to error as the caller also returns false if num keys is 0.

router/pkg/entitycaching/cache/redis/redis.go
SkArchon: The other places like APQ and Rate Limiter has prefixes, thats why its here. Though I don't have a strong opinion and can remove it. Lmk.

router/pkg/entitycaching/cache/in_memory/in_memory.go
```diff
@@ -50,7 +55,7 @@ func (c *InMemoryCache) GetMany(ctx context.Context, keys []string) (map[string]
 	}
 
 	if len(keys) == 0 {
-		return nil, nil
+		return nil, errors.New("entity cache lookup requires at least one key")
```
SkArchon, line 56: I will move this into the engine with my base implementation PRs (I think it should be in the engine imo)

router/pkg/entitycaching/cache/redis/redis.go
fiam: Since it's already there, let's keep it for now, but let's try to be minimal moving forward. Does that sound good?

fiam (approved):

PR #3146, opened 2026-08-07
feat: merge entity fetches and then schedule optimally

ysmolski:
This PR adds two feature flags for the Cosmo's Query Planner:
                                     
`engine.enable_multi_fetch`

This option merges entity fetches that target the same subgraph and 
execute at the same point in the query plan 
into a single request with aliased `_entities` fields. 
This reduces the number of requests sent to subgraphs. 
env: ENGINE_ENABLE_MULTI_FETCH

`engine.enable_schedule_fetches`

This feature replaces the wave-based fetch execution 
with a dependency-aware fetch scheduler. 
Independent fetch chains progress as soon as 
their own dependencies complete,
instead of waiting for the slowest fetch in each wave.                                                                                                                    
env: ENGINE_ENABLE_SCHEDULE_FETCHES

By default, these features are disabled:
```
engine:
  enable_multi_fetch: false       
  enable_schedule_fetches: false 
```

Built on top of https://github.com/wundergraph/graphql-go-tools/pull/1612

## Summary by CodeRabbit

* **New Features**
  * Added optional multi-fetch execution to merge compatible entity fetches.
  * Added dependency-aware fetch scheduling, including parallel execution of independent fetch chains.
  * Both options are configurable through YAML and environment variables and remain disabled by default.

* **Documentation**
  * Updated router configuration reference and examples with the new execution settings.

* **Tests**
  * Added coverage for multi-fetch behavior, scheduling combinations, parallel chains, query plans, responses, and request counts.


docs-website/router/configuration.mdx
```diff
@@ -2058,6 +2058,8 @@ Configure the GraphQL Execution Engine of the Router.
...
 | ENGINE_ENABLE_DEFER                                           | enable_defer                                           | <Icon icon="square" /> | Enables support for the `@defer` directive, allowing clients to defer parts of a query so that the initial response is returned faster and deferred fields are streamed incrementally.                                                                                                                                                      | false         |
+| ENGINE_ENABLE_MULTI_FETCH                                     | enable_multi_fetch                                     | <Icon icon="square" /> | Merges entity fetches that target the same subgraph and execute at the same point in the query plan into a single request with aliased `_entities` fields. This reduces the number of requests sent to subgraphs.                                                                                                                           | false         |
```
SkArchon, line 2061: I think we should document this in more detail, what effect these do (IF we plan to keep this as flags and not the only option). But we can do this in a follow up PR.

SkArchon (approved):
Code itself is ok so approving.

However my main concern is that we have the "query-plan-skip" tag for query plan changes for the customer (which is what the query planner tests are for), I would verify if at least the error is the same as the baseline error before merging (You could open a no-op PR with a comment change and add the "query-plan" label, and compare the results from there and this PR).

router-tests/go.mod
```diff
@@ -31,7 +31,7 @@ require (
 	github.com/wundergraph/cosmo/router v0.0.0-20260710155145-803a4bc06d92
 	github.com/wundergraph/cosmo/router-plugin v0.0.0-20250808194725-de123ba1c65e
 	github.com/wundergraph/cosmo/speedtrap v0.0.0-00010101000000-000000000000
-	github.com/wundergraph/graphql-go-tools/v2 v2.15.0
+	github.com/wundergraph/graphql-go-tools/v2 v2.15.2-0.20260812134543-2f577793f2e9
```
SkArchon: Ensure to update this before merging

ysmolski: Yes, I will.

docs-website/router/configuration.mdx
ysmolski: Not released with being enabled by default for now. We will have some more explanations later in the docs.

ysmolski:
I have tested QPlanner here: https://github.com/wundergraph/cosmo-celestial/pull/2094. There is a slight performance regression detected which comes from Audit FIxes released in the engine some time ago.

router-tests/go.mod
ysmolski: updated to 2.16.0

PR #3138, opened 2026-08-05
feat: implement ptq rpc and index schemas

wilsonrivera:
## Summary by CodeRabbit

* **New Features**
  * Added AI-powered natural-language query generation for federated graphs.
  * Added schema indexing and readiness checks before generating queries.
  * Added configurable Prompt-to-Query service endpoint settings.
  * Added the `prompt-to-query` feature flag for graph tokens.

* **Bug Fixes**
  * Corrected graph-token feature reporting so enabled features retain their proper names.

* **Tests**
  * Added coverage for query generation, schema indexing, feature flags, cancellation, and service configuration.

## Checklist

- [ ] I have discussed my proposed changes in an issue and have received approval to proceed.
- [x] I have followed the coding standards of the project.
- [x] Tests or benchmarks have been added or updated.
- [ ] If applicable, I have provided instructions for reviewers for [manually validating my changes](https://github.com/wundergraph/cosmo/blob/main/CONTRIBUTING.md#pull-requests-conventions).
- [ ] Documentation has been updated on [https://github.com/wundergraph/docs-website](https://github.com/wundergraph/docs-website).
- [x] I have read the [Contributors Guide](https://github.com/wundergraph/cosmo/blob/main/CONTRIBUTING.md).

## Open Source AI Manifesto

This project follows the principles of the [Open Source AI Manifesto](https://human-oss.dev). Please ensure your contribution aligns with its principles.

PR #3158, opened 2026-08-12
feat(router): expose prompt-to-query via MCP

fiam:
## Summary

- expose a `generate_query` tool through the router MCP server
- send the active `RouterConfig.version` to the control-plane `GenerateQuery` RPC
- resolve that version to the stored SDL in the control plane, call Yoko `EnsureIndex` on demand, and use its returned opaque index ID
- add OAuth scope support and MCP documentation for the tool

## Stack

- Implements [ROUTER-575](https://linear.app/wundergraph/issue/ROUTER-575/router-mcp-query-generation-support)
- Depends on #3138
- This PR is intentionally based on the head branch of #3138. Once #3138 merges, this PR can be retargeted to `main`.
- Prompt-to-query service behavior remains covered by the base PR; this stack focuses on the router/MCP integration.

## Verification

- `CGO_ENABLED=0 go test ./...` in `router`
- control-plane TypeScript build

### DIFF
--- a/router/core/supervisor_instance.go
+++ b/router/core/supervisor_instance.go
@@ -9,6 +9,9 @@ import (
 
 	"github.com/KimMachineGun/automemlimit/memlimit"
 	"github.com/dustin/go-humanize"
+	"github.com/wundergraph/cosmo/router/internal/controlplane"
+	rjwt "github.com/wundergraph/cosmo/router/internal/jwt"
+	"github.com/wundergraph/cosmo/router/internal/prompttoquery"
 	"github.com/wundergraph/cosmo/router/pkg/authentication"
 	"github.com/wundergraph/cosmo/router/pkg/config"
 	"github.com/wundergraph/cosmo/router/pkg/controlplane/selfregister"
@@ -157,6 +160,26 @@ func newRouter(ctx context.Context, params RouterResources, additionalOptions ..
 			return nil, fmt.Errorf("could not create self register: %w", err)
 		}
 		options = append(options, WithSelfRegistration(selfRegister))
+	}
+
+	if cfg.MCP.Enabled && cfg.Graph.Token != "" {
+		claims, err := rjwt.ExtractFederatedGraphTokenClaims(cfg.Graph.Token)
+		if err != nil {
+			return nil, fmt.Errorf("could not inspect graph token features: %w", err)
+		}
+
+		if claims.HasFeature(rjwt.FeaturePromptToQuery) {
+			controlplaneTransport, err := controlplane.NewTransport(cfg.Graph.Token, logger)
+			if err != nil {
+				return nil, fmt.Errorf("could not create controlplane transport: %w", err)
+			}
+
+			promptToQueryClient, err := prompttoquery.New(cfg.ControlplaneURL, controlplaneTransport)
+			if err != nil {
+				return nil, fmt.Errorf("could not create prompt-to-query client: %w", err)
+			}
+			options = append(options, WithPromptToQueryClient(promptToQueryClient))
+		}
 	}
 
 	if opt := optionFromExecutionConfig(&cfg.ExecutionConfig, cfg.RouterConfigPath); opt != nil {
--- a/router/internal/prompttoquery/client.go
+++ b/router/internal/prompttoquery/client.go
@@ -1,0 +1,56 @@
+package prompttoquery
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"time"
+
+	"connectrpc.com/connect"
+	aiv1 "github.com/wundergraph/cosmo/router/gen/proto/wg/cosmo/ai/v1"
+	"github.com/wundergraph/cosmo/router/gen/proto/wg/cosmo/ai/v1/aiv1connect"
+	brotli "go.withmatt.com/connect-brotli"
+)
+
+const clientTimeout = 15 * time.Second
+
+// Client generates GraphQL operations through the control plane.
+type Client struct {
+	aiServiceClient aiv1connect.AIServiceClient
+}
+
+func New(endpoint string, transport http.RoundTripper) (*Client, error) {
+	if endpoint == "" {
+		return nil, fmt.Errorf("controlplane endpoint is required for prompt to query")
+	}
+
+	if transport == nil {
+		return nil, fmt.Errorf("controlplane transport is required for prompt to query")
+	}
+
+	httpClient := &http.Client{
+		Transport: transport,
+		Timeout:   clientTimeout,
+	}
+
+	aiServiceClient := aiv1connect.NewAIServiceClient(httpClient, endpoint,
+		brotli.WithCompression(),
+		connect.WithSendCompression(brotli.Name),
+	)
+
+	return &Client{aiServiceClient: aiServiceClient}, nil
+}
+
+func (c *Client) GenerateQuery(ctx context.Context, schemaVersionID, prompt string) (*aiv1.GenerateQueryResponse, error) {
+	req := connect.NewRequest(&aiv1.GenerateQueryRequest{
+		Version: schemaVersionID,
+		Prompt:  prompt,
+	})
+
+	resp, err := c.aiServiceClient.GenerateQuery(ctx, req)
+	if err != nil {
+		return nil, err
+	}
+
+	return resp.Msg, nil
+}
--- a/router/pkg/mcpserver/server.go
+++ b/router/pkg/mcpserver/server.go
@@ -11,6 +11,7 @@ import (
 	"net/http"
 	"slices"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/hashicorp/go-retryablehttp"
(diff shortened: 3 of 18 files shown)


controlplane/src/core/services/PromptToQueryService.ts
```diff
@@ -10,15 +10,21 @@ import {
...
 import * as z from 'zod';
+import { validate as validateUUID } from 'uuid';
 import { traced } from '../tracing.js';
 import * as schema from '../../db/schema.js';
+import { FederatedGraphRepository } from '../repositories/FederatedGraphRepository.js';
 import { OrganizationRepository } from '../repositories/OrganizationRepository.js';
 
 const validationSchema = z.object({
-  schemaSha: z.string().regex(/^sha256:[\da-f]{64}$/i),
+  version: z.string().refine(validateUUID),
```
wilsonrivera: Could you use `z.string().uuid()` instead

controlplane/src/core/services/PromptToQueryService.ts
```diff
@@ -98,13 +104,38 @@ export class PromptToQueryService {
...
+        response: {
+          code: EnumStatusCode.ERR_NOT_FOUND,
+          details: 'Schema version not found for this federated graph',
+        },
+      });
+    }
+
     // Invoke the `prompt to query` service
     try {
+      const indexId = await this.ensureIndex(schemaVersion.sdl);
```
wilsonrivera: This method is not intended to be called by the query generator. The reason for this is that the service could take a bit to parse and index the schema, use the `crypto` to generate the `SHA256` for the schema and use the value directly. 

If the index doesn't exist, let it fail

wilsonrivera (changes requested):

controlplane/src/core/services/PromptToQueryService.ts
fiam: Thanks, I was not aware of the zod builtin! It should be good now.

controlplane/src/core/services/PromptToQueryService.ts
fiam: Makes sense! PTAL and let me know if this is what you had in mind. Thanks!

endigma: `ensureIndex` is implemented via a task queue so it should actually respond instantly with most fields, but importantly query generation will error if the index is not completed so you may want to poll here or fail fast.

router/core/supervisor_instance.go
```diff
@@ -159,6 +160,16 @@ func newRouter(ctx context.Context, params RouterResources, additionalOptions ..
...
 
+	if cfg.MCP.Enabled && cfg.Graph.Token != "" {
+		promptToQueryClient, err := prompttoquery.New(cfg.ControlplaneURL, cfg.Graph.Token,
+			prompttoquery.WithLogger(logger),
+		)
+		if err != nil {
+			return nil, fmt.Errorf("could not create prompt-to-query client: %w", err)
+		}
+		options = append(options, WithPromptToQueryClient(promptToQueryClient))
+	}
```
endigma, line 175: Unsure about this, see my PM

router/pkg/controlplane/prompttoquery/client.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/hashicorp/go-retryablehttp"
+	nodev1 "github.com/wundergraph/cosmo/router/gen/proto/wg/cosmo/node/v1"
+	"github.com/wundergraph/cosmo/router/gen/proto/wg/cosmo/node/v1/nodev1connect"
+	"go.uber.org/zap"
+	brotli "go.withmatt.com/connect-brotli"
+)
+
+const defaultTimeout = 15 * time.Second
+
+type Option func(*Client)
```
endigma: Is this worth it for just a logger and timeout vs named parameters to `New(...)`? This is internal-only, the default is either used or it isn't

router/internal/prompttoquery/client.go
```diff
@@ -0,0 +1,100 @@
...
+	retryClient := retryablehttp.NewClient()
+	retryClient.RetryWaitMax = 15 * time.Second
+	retryClient.RetryMax = 3
+	retryClient.Backoff = retryablehttp.DefaultBackoff
+	retryClient.Logger = nil
+	retryClient.RequestLogHook = func(_ retryablehttp.Logger, _ *http.Request, retry int) {
+		if retry > 0 {
+			c.logger.Info("Generate query through controlplane", zap.Int("retry", retry))
+		}
+	}
```
endigma: Need to ensure that this does not retry on errors that should not be retryable

router/internal/prompttoquery/client.go
```diff
@@ -0,0 +1,100 @@
...
+
+	return c, nil
+}
+
+func (c *Client) GenerateQuery(ctx context.Context, schemaVersionID, prompt string) (*nodev1.GenerateQueryResponse, error) {
+	req := connect.NewRequest(&nodev1.GenerateQueryRequest{
+		Version: schemaVersionID,
+		Prompt:  prompt,
+	})
+	req.Header().Set("Authorization", "Bearer "+c.graphAPIToken)
```
endigma: I believe this is better done as part of a roundtripper on the client itself

router/pkg/mcpserver/server.go
```diff
@@ -690,6 +727,36 @@ func (s *GraphQLSchemaServer) registerTools() error {
...
+			Name:        "generate_query",
+			Description: "Generates a GraphQL operation for this API from a natural-language prompt.",
+			InputSchema: map[string]any{
+				"type":        "object",
+				"description": "A natural-language description of the GraphQL operation to generate.",
+				"properties": map[string]any{
+					"prompt": map[string]any{
+						"type":        "string",
+						"minLength":   1,
+						"description": "A natural-language description of the GraphQL operation to generate.",
```
endigma, line 742: I believe this may need a more descriptive instruction for how to format the prompt, tbd

endigma (changes requested):

controlplane/src/core/services/PromptToQueryService.ts
wilsonrivera: In the context of Yoko that is correct, however, this is just a helper for ControlPlane to send the schema for indexation. If we want to check whether the index is already there, we need to make a new helper

router/pkg/controlplane/prompttoquery/client.go
fiam: I agree, probably a bit overkill. Hardcoded the timeout and made the logger a required argument. I've also moved the package to be internal so we don't have to care about a stable interface https://github.com/wundergraph/cosmo/commit/b7f1b0fccb5c0bbc85f2c1b4a9e46a4a2eb84540 - PTAL

router/internal/prompttoquery/client.go
fiam: The default policy retries on 429 and >= 500 (except 501). Looking at the CP side, it seems we only return 5xx for unhandled errors, all application errors return 200.

So the only thing that could generate a retry is either an intermediate proxy/infra or either an unhandled exception (which might or might not be worth retrying). However, some errors in k8s ingress or some service meshes also end up generating a 500 when the service is unavailable (e.g. some rollout strategy not properly configured).

Since prompt-to-query is supposed to be idempotent I think it might be safer to over-retry than under-retry (if that makes sense). WDYT?

router/internal/prompttoquery/client.go
fiam: I think that's a cool idea, but as an internal package, so eventually all communication to the control plane could use the same transport. PTAL https://github.com/wundergraph/cosmo/commit/561f7e2c39c65270f7e45cc893948808642fac60

router/internal/prompttoquery/client.go
fiam: @endigma ping on this

router/core/supervisor_instance.go
fiam: This has been updated to take the token claims into account, PTAL

endigma: I think we can actually allow this to fail against Cosmo if it's not enabled on that end, it should give a warning and propagate an error that it is not enabled through MCP to the user I think

endigma: With it requiring the JWT feature they have to re-issue all their router tokens to use it

router/internal/prompttoquery/client.go
endigma: Sounds good to me, but prompt to query is not idempotent, it is side effect free though. Retrying could still cause over billing if we fail after we record usage. Just make sure to test failure modes

router/core/supervisor_instance.go
fiam: That was an accepted tradeoff. We decided that exposing a potentially useless tool was worse than having them re-issue the token.

fiam:
@wilsonrivera @endigma I think all your comments have been addressed. Could you PTAL? Thanks!

PR #3190, opened 2026-08-25
fix: update packages with high severity vulnerabilities

gausie:
Simple version updates to clear the high severity Dependabot alerts behind COSMO-381. Closes 39 of the 40.

find-my-way (#843) is left for a separate PR. There's no 8.x fix — the first patched version is 9.7.0, which only ships with fastify 5 — so closing it means moving controlplane off fastify 4, which shouldn't ride along with a pile of mechanical bumps.

Also removed the postcss-fontpath dependency from keycloak/theme. It wasn't being loaded — Parcel resolves .postcssrc ahead of postcss.config.js, so that config never ran — and it pulled in a second copy of postcss below the minimum acceptable version, which would have kept the alert open. Rebuilt login.css is byte-identical.


gausie:
@coderabbitai summary

pepol (approved):

thisisnithin (approved):

PR #3193, opened 2026-08-26
fix(controlplane): upgrade fastify to v5

gausie:
Closes [find-my-way #843](https://github.com/wundergraph/cosmo/security/dependabot/843) (CVE-2026-47219), the last alert behind [COSMO-381](https://linear.app/wundergraph/issue/COSMO-381/vanta-remediate-high-vulnerabilities-identified-in-packages-are) and the one #3190 left out. There is no 8.x fix — find-my-way 9.7.0 only ships with fastify 5 — so closing it meant moving controlplane off fastify 4. The alert was dismissed as `not_used` (the DDoS path is HTTP/2-only, we serve HTTP/1.1); this is the upgrade that dismissal promised.

Almost none of the v5 breaking-change list applies. The only forced change is `logger` → `loggerInstance` at three constructor sites. Bumping the catalog pino to 9 surfaced two `logger.error('msg', { error })` calls that were silently dropping the error object.

Full suite: 1776 passed, 2 failed — both confirmed pre-existing by re-running against unmodified `main`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error logging for schema-version failures, providing clearer diagnostic details.
  * Updated server logging configuration for compatibility with newer framework versions.

* **Chores**
  * Updated server framework, logging, and related plugin dependencies.
  * Improved compatibility across metrics monitoring and test server configurations.


comatory (approved):
Looks good to me. Have you tried running it locally and see if logging still works? Just to be sure.

comatory:
@gausie how about pushing this one over the finish line?

gausie:
It's on the list!

gausie:
This is now tested and works fine by all accounts

pnpm-workspace.yaml
```diff
@@ -128,7 +128,7 @@ catalog:
   lodash-es: 4.18.1
   hono: 4.12.25
   '@hono/node-server': 1.19.13
-  pino: ^8.19.0
+  pino: ^9.0.0
```
comatory, line 131: Is there anything preventing us from going to latest v10? Version 9 is from 2024. Is this because of compatibility with Fastify, or possibly other packages?

controlplane/package.json
```diff
@@ -50,7 +50,7 @@
     "@connectrpc/connect": "catalog:",
     "@connectrpc/connect-fastify": "catalog:",
     "@connectrpc/connect-node": "catalog:",
-    "@fastify/cors": "^9.0.1",
+    "@fastify/cors": "^11.3.0",
```
comatory, line 53: I think we should decide whether we'll use caret or pin explicit versions. I'd rather start pinning, but we have a mix of both. WDYT?

comatory (changes requested):
Requesting changes.

pnpm-workspace.yaml
gausie: Just the minimal change to get fastify v5 out. I can follow up with another PR, or you can and I can review

controlplane/package.json
gausie: I copied what we had, I am still pro caret but am happy to go with the team. Simple change to just pin the resolved version

pnpm-workspace.yaml
comatory: Sure, let's prioritize fastify update.

controlplane/package.json
comatory: I have a ticket to start pinning deps: https://linear.app/wundergraph/issue/COSMO-179/cosmomonorepo-pin-tsjs-dependencies-to-explicit-versions

We can handle it then all at once.

comatory (approved):

PR #3196, opened 2026-08-27
feat: show feature subgraphs in playground and schema dropdowns

gausie:
Fixes COSMO-415.

The `Querying Graph x` dropdown in the playground and the schema SDL dropdown listed a graph's feature flags but not the feature subgraphs inside them. They now appear nested under their flag.

Also
* Pulled out a `SchemaSelector` plus `FeatureFlagMenuItem` from the SDL dropdown to avoid duplication with the explorer's `GraphSelector`. Also the explorer's dropdown used to replace the whole query string while its own no-flags fallback merged (both merge now).

The diff looks like it's 5000 lines but like 95% of it is generated protobuff stuff, so don't be scared. Manually checked the SDL page, explorer, playground and compositions page against a local graph with two flags pinning different feature subgraph versions.

playground-selector
sdl-mood-boost
sdl-mood-boost2

## Summary by CodeRabbit

* **New Features**
  * View feature subgraphs associated with feature flags in schema selections.
  * Select client or router schemas, feature flags, subgraphs, and feature subgraphs in the Playground and Schema Registry.
  * Explore feature-subgraph schemas with accurate routing and subscription details.
  * Feature-subgraph results respect access permissions and remain tied to the correct flag composition.
  * Identify failed latest compositions while browsing feature-flag schemas.

* **Bug Fixes**
  * Improved loading states when resolving feature-subgraph schemas.

* **Documentation**
  * Added guidance for Playground targets and feature-flag and feature-subgraph schemas in Schema Registry.


gausie:
@coderabbit review

studio/src/components/schema/schema-selection.ts
```diff
@@ -0,0 +1,7 @@
+export type SchemaType = 'client' | 'router';
+
+export interface SchemaSelection {
+  featureFlag?: string;
+  subgraph?: string;
+  schemaType?: string;
```
iliya-i: shouldn't we have `schemaType?: SchemaType;` instead of `schemaType?: string;`

studio/src/components/schema/feature-flag-menu-item.tsx
```diff
@@ -0,0 +1,80 @@
...
+          {hasFailedLatestComposition && <StaleCompositionIcon />}
+        </span>
+      </DropdownMenuSubTrigger>
+      <DropdownMenuPortal>
+        <DropdownMenuSubContent>
+          <DropdownMenuRadioGroup
+            value={isActive && !selection.subgraph ? (selection.schemaType ?? 'client') : ''}
+            onValueChange={(value) => onSelect({ featureFlag: name, schemaType: value })}
+          >
+            <DropdownMenuRadioItem className="w-[170px] items-center justify-between pl-2" value="client">
```
iliya-i: nit: this Client/Router radio pair is duplicated in `schema-selector.tsx` (graph submenu), with slightly different widths (`w-[170px]` here vs `w-[150px]` there). Since this PR is already extracting shared pieces, a tiny shared `SchemaTypeRadioGroup` (or at least a shared width constant) would finish the job and keep the two submenus visually consistent.

studio/src/components/schema/schema-selector.tsx
```diff
@@ -0,0 +1,157 @@
...
+export const SchemaSelector = ({
+  title,
+  supportsFederation,
+  featureFlags,
+  selection,
+  onSelect,
+  subgraphNames,
+  featureSubgraphsOfFlag,
+}: SchemaSelectorProps) => {
+  const graphData = useContext(GraphContext);
```
iliya-i: The graph name is the only thing that we read from `GraphContext`, I think passing `graphName` as a prop would make it self-contained and easier to test in isolation

studio/src/pages/[organizationSlug]/[namespace]/graph/[slug]/schema/sdl.tsx
```diff
@@ -119,36 +129,59 @@ const SDLPage: NextPageWithLayout = () => {
...
+          routingUrl: graphData?.graph?.routingURL ?? '',
+          sdl:
+            activeSchemaType === 'router'
+              ? (federatedGraphSdl?.sdl ?? '')
+              : federatedGraphSdl?.clientSchema || federatedGraphSdl?.sdl,
+          time: graphData?.graph?.lastUpdatedAt ?? '',
+        };
 
-  const isLoading = loadingGraphSDL || loadingSubgraphSDL;
+  const isLoading =
```
Comment (iliya-i) [inline studio/src/pages/[organizationSlug]/[namespace]/graph/[slug]/schema/sdl.tsx]: shouldn't we use `isPlaygroundSchemaLoading`?

studio/src/pages/[organizationSlug]/[namespace]/graph/[slug]/schema/sdl.tsx
```diff
@@ -119,36 +129,59 @@ const SDLPage: NextPageWithLayout = () => {
...
+    loadingFeatureSubgraphSDL ||
+    // Which SDL to fetch is unknown until the flag list has loaded.
+    (isFeatureSubgraphSelected && loadingCompositionFlags);
 
   let content: React.ReactNode;
 
   if (isLoading) {
     content = <Loader fullscreen />;
-  } else if (activeSubgraph && subgraphSdl?.response && subgraphSdl.response?.code === EnumStatusCode.ERR_NOT_FOUND) {
+  } else if (isFeatureSubgraphSelected && !featureSubgraphSdl?.sdl) {
```
Comment (iliya-i) [inline studio/src/pages/[organizationSlug]/[namespace]/graph/[slug]/schema/sdl.tsx]: Two different situations collapse into this branch, and both get the generic "No schema found — use the CLI to publish" guidance, which is misleading here: (1) an unresolvable selection — stale bookmark, renamed flag, or a feature subgraph dropped from the flag — where nothing is wrong with publishing and the helpful action would be clearing the selection back to the graph; (2) a failed `getSdlBySchemaVersion` request (network/500), where an error state would be more honest than an empty state. Worth distinguishing "selection didn't resolve" from "subgraph has no published schema", or at least letting real errors surface as errors.

studio/src/components/schema/schema-selector.tsx
```diff
@@ -0,0 +1,157 @@
...
+  const activeSchemaType = selection.schemaType ?? 'client';
+
+  return (
+    <DropdownMenu>
+      <DropdownMenuTrigger value={title} className="w-full md:ml-auto md:w-max md:min-w-[200px]" asChild>
+        <div className="flex items-center justify-center">
```
iliya-i: do we need this extra div here?

studio/src/pages/[organizationSlug]/[namespace]/graph/[slug]/schema/sdl.tsx
```diff
@@ -119,36 +129,59 @@ const SDLPage: NextPageWithLayout = () => {
...
+    content = <EmptySchema subgraphName={activeSubgraph} />;
+  } else if (
+    activeSubgraph &&
+    !isFeatureSubgraphSelected &&
+    subgraphSdl?.response &&
+    subgraphSdl.response?.code === EnumStatusCode.ERR_NOT_FOUND
+  ) {
     content = <EmptySchema subgraphName={activeSubgraph} />;
```
Comment (iliya-i) [inline studio/src/pages/[organizationSlug]/[namespace]/graph/[slug]/schema/sdl.tsx]: is `<EmptySchema subgraphName={activeSubgraph} />` the correct error message for this case?

studio/src/pages/[organizationSlug]/[namespace]/graph/[slug]/schema/sdl.tsx
```diff
@@ -119,36 +129,59 @@ const SDLPage: NextPageWithLayout = () => {
...
+    // Which SDL to fetch is unknown until the flag list has loaded.
+    (isFeatureSubgraphSelected && loadingCompositionFlags);
 
   let content: React.ReactNode;
 
   if (isLoading) {
     content = <Loader fullscreen />;
-  } else if (activeSubgraph && subgraphSdl?.response && subgraphSdl.response?.code === EnumStatusCode.ERR_NOT_FOUND) {
+  } else if (isFeatureSubgraphSelected && !featureSubgraphSdl?.sdl) {
+    content = <EmptySchema subgraphName={activeSubgraph} />;
```
Comment (iliya-i) [inline studio/src/pages/[organizationSlug]/[namespace]/graph/[slug]/schema/sdl.tsx:168]: is `<EmptySchema subgraphName={activeSubgraph} />` the correct error message for this case? `!featureSubgraphSdl?.sdl` is also `true` when the selection doesn't resolve and this can happen due to renamed flag or stale link and for those cases "publish the subgraph" is misleading.

studio/src/pages/[organizationSlug]/[namespace]/graph/[slug]/playground.tsx
```diff
@@ -723,15 +777,31 @@ const PlaygroundPage: NextPageWithLayout = () => {
...
 
   const loadSchemaGraphId = (router.query.load as string) || graphContext?.graph?.id || '';
-  const type = (router.query.type as string) || 'graph';
+  const typeParam = (router.query.type as string) || 'graph';
+  const configType: ConfigType = isConfigType(typeParam) ? typeParam : 'graph';
+
+  const { data: compositionFlagsData, isLoading: isLoadingCompositionFlags } = useCompositionFlags();
 
-  const compositionFlagsData = useCompositionFlags();
+  const activeFeatureSubgraph = useMemo(() => {
```
Comment (iliya-i) [inline studio/src/pages/[organizationSlug]/[namespace]/graph/[slug]/playground.tsx]: I'm not 100% sure, but shouldn't this mirror the loading/not-found handling like in [sdl.tsx](https://github.com/wundergraph/cosmo/blob/a2f44ba86da269c21d03e7e8478c1858c053f8f9/studio/src/pages/%5BorganizationSlug%5D/%5Bnamespace%5D/graph/%5Bslug%5D/schema/sdl.tsx#L163-L186)?

studio/src/pages/[organizationSlug]/[namespace]/graph/[slug]/playground.tsx
```diff
@@ -723,15 +777,31 @@ const PlaygroundPage: NextPageWithLayout = () => {
...
+    );
+
+    return (compositionFlagsData?.featureSubgraphs ?? []).find(
+      (featureSubgraph) =>
+        featureSubgraph.id === loadSchemaGraphId && featureSubgraph.featureFlagId === featureFlag?.id,
+    );
+  }, [compositionFlagsData, loadSchemaGraphId, router.query.featureFlag, configType]);
 
   const { data, isLoading: isLoadingGraphSchema } = useQuery(getFederatedGraphSDLByName, {
     name: graphContext?.graph?.name,
```
Comment (iliya-i) [inline studio/src/pages/[organizationSlug]/[namespace]/graph/[slug]/playground.tsx]: while the flags are loading featureFlagName resolves to undefined, so this briefly fetches/shows the base graph schema so maybe it's good to add `enabled: configType !== 'featureFlag' || !isLoadingCompositionFlags`

Comment (gausie) [reply studio/src/pages/[organizationSlug]/[namespace]/graph/[slug]/playground.tsx]: Ah great catch

studio/src/pages/[organizationSlug]/[namespace]/graph/[slug]/schema/index.tsx
```diff
@@ -894,149 +871,25 @@ export const GraphSelector = () => {
...
+
+  // The explorer renders supergraph schemas, so subgraphs are not selectable here.
+  return (
+    <SchemaSelector
+      title={activeFeatureFlag || graphName}
+      graphName={graphName}
+      supportsFederation
+      featureFlags={featureFlags}
+      selection={{ featureFlag: activeFeatureFlag, schemaType }}
+      onSelect={(next) => applyParams({ featureFlag: next.featureFlag ?? null, schemaType: next.schemaType ?? null })}
```
Comment (iliya-i) [inline studio/src/pages/[organizationSlug]/[namespace]/graph/[slug]/schema/index.tsx]: Switching schemas keeps `typename` and `category` from the previous schema because `applyParams` merges over the current query. The end URL is:

`http://localhost:3000/wundergraph/default/graph/mygraph/schema?organizationSlug=wundergraph&namespace=default&slug=mygraph&category=objects&typename=AvailabilityWindowA&schemaType=router&fieldName=`

https://github.com/user-attachments/assets/91cc4725-645d-4e62-92cb-196c00c0545e

Passing `typename: null` and `category: null` here restores the reset. The same merge also appends `organizationSlug`, `namespace` and `slug` on every switch; building the URL from the pathname with explicit params, as before, avoids both.

studio/src/pages/[organizationSlug]/[namespace]/graph/[slug]/schema/sdl.tsx
```diff
@@ -52,103 +38,146 @@ const SDLPage: NextPageWithLayout = () => {
...
+    isLoading: loadingFeatureSubgraphSDL,
+  } = useFeatureSubgraphSchema(
+    compositionFlagsData,
+    isFeatureSubgraphSelected ? { featureFlagName: activeFeatureFlag, subgraphName: activeSubgraph } : undefined,
+  );
+
+  const activeSchemaType = toSchemaType(router.query.schemaType as string);
+
+  /** Only one schema can be selected, so every selection clears the other two params. */
+  const selectSchema = (next: SchemaSelection) =>
```
Comment (iliya-i) [inline studio/src/pages/[organizationSlug]/[namespace]/graph/[slug]/schema/sdl.tsx:74]: Same `applyParams` merge as in the `GraphSelector`, every selection here appends `organizationSlug`, `namespace` and `slug` to the URL but here is cosmetic, but worth fixing together I think.

studio/src/pages/[organizationSlug]/[namespace]/graph/[slug]/playground.tsx
```diff
@@ -602,28 +605,56 @@ const ToggleClientValidation = () => {
...
+  );
 
   const selected = (router.query.load as string) || graphContext?.graph?.id || '';
-  const type = (router.query.type as string) || 'graph';
+  const configType = toConfigType(router.query.type as string);
+  const activeFeatureFlag = (router.query.featureFlag as string) || '';
 
   const applyParams = useApplyParams();
 
+  const currentValue = configSelectionValue({
```
Comment (iliya-i) [inline studio/src/pages/[organizationSlug]/[namespace]/graph/[slug]/playground.tsx:640]: The trigger next to "Querying Feature subgraph :" renders blank when the flag in the link is no longer in the latest composition, for example after a breaking publish to one of its feature subgraphs. `currentValue` embeds the flag name, so it matches none of the `SelectItem` values and Radix shows nothing.

Repro with a link the playground generated before ff-b's composition broke:

`http://localhost:3000/wundergraph/default/graph/mygraph/playground?load=1823963c-7d66-4fec-8744-80d7c019be37&type=featureSubgraph&featureFlag=ff-b`

Image

Image

studio/src/pages/[organizationSlug]/[namespace]/graph/[slug]/playground.tsx
```diff
@@ -1210,6 +1299,12 @@ const PlaygroundPage: NextPageWithLayout = () => {
         <div className="hidden h-full flex-1 pl-2.5 md:flex">
           {isLoadingDefaultHeaders ? (
             <Loader fullscreen />
+          ) : isUnresolvedFeatureSubgraph ? (
+            <EmptyState
+              icon={<ExclamationTriangleIcon />}
+              title="Schema not found"
+              description={`The selected feature subgraph is not part of the latest composition of feature flag ${router.query.featureFlag}. The flag may have been renamed, or the feature subgraph removed from it.`}
```
Comment (iliya-i) [inline studio/src/pages/[organizationSlug]/[namespace]/graph/[slug]/playground.tsx]: Can flags even be renamed? I don't see a rename in `wgc feature-flag update`. Same wording is in [sdl.tsx:164](https://github.com/wundergraph/cosmo/blob/25bebd04760dda09471053d6640195a614c619ff/studio/src/pages/%5BorganizationSlug%5D/%5Bnamespace%5D/graph/%5Bslug%5D/schema/sdl.tsx#L164), [use-feature-subgraph-schema.ts:21](https://github.com/wundergraph/cosmo/blob/25bebd04760dda09471053d6640195a614c619ff/studio/src/hooks/use-feature-subgraph-schema.ts#L21), [schema-loading.test.ts:23](https://github.com/wundergraph/cosmo/blob/25bebd04760dda09471053d6640195a614c619ff/studio/src/__tests__/schema-loading.test.ts#L23), [playground.tsx:847](https://github.com/wundergraph/cosmo/blob/25bebd04760dda09471053d6640195a614c619ff/studio/src/pages/%5BorganizationSlug%5D/%5Bnamespace%5D/graph/%5Bslug%5D/playground.tsx#L847) and [sdl.tsx:158](https://github.com/wundergraph/cosmo/blob/25bebd04760dda09471053d6640195a614c619ff/studio/src/pages/%5BorganizationSlug%5D/%5Bnamespace%5D/graph/%5Bslug%5D/schema/sdl.tsx#L158).

studio/src/pages/[organizationSlug]/[namespace]/graph/[slug]/schema/index.tsx
Comment (gausie) [reply studio/src/pages/[organizationSlug]/[namespace]/graph/[slug]/schema/index.tsx]: Good spot. I didn't notice this. Fixed now

studio/src/pages/[organizationSlug]/[namespace]/graph/[slug]/playground.tsx
Comment (gausie) [reply studio/src/pages/[organizationSlug]/[namespace]/graph/[slug]/playground.tsx:640]: Fixed and adjusted the wording here

studio/src/pages/[organizationSlug]/[namespace]/graph/[slug]/playground.tsx
Comment (gausie) [reply studio/src/pages/[organizationSlug]/[namespace]/graph/[slug]/playground.tsx]: Pulled this out to a separate component. Good spot.

studio/src/pages/[organizationSlug]/[namespace]/graph/[slug]/schema/index.tsx
```diff
@@ -872,17 +853,24 @@ export const GraphSelector = () => {
...
     namespace: { name: namespace },
   } = useWorkspace();
 
-  const fullPath = router.asPath;
-  const pathWithHash = fullPath.split('?')[0];
-  const pathname = pathWithHash.split('#')[0];
-
   const applyParams = useApplyParams();
 
+  const selectSchema = (next: SchemaSelection) =>
```
Comment (iliya-i) [inline studio/src/pages/[organizationSlug]/[namespace]/graph/[slug]/schema/index.tsx:863]: Nit: switching Client to Router on the same graph resets to the Query root

`?category=objects&typename=Employee&schemaType=client` -> `?schemaType=router`

Main did the same on graphs with flags but kept the type on graphs without flags. I think that keeping the type when it exists in the new schema, and falling back to the category list when it does not, would be the better direction and would also remove the "No data found" case.

https://github.com/user-attachments/assets/b6bff176-08ae-40cd-9bfa-2a775a4d13e8

iliya-i (commented):
Only one small nit, other than this I think the front-end part looks good 🚀 (I have not looked at the backend 😇)

PR #3191, opened 2026-08-26
fix(router): preserve APQ query when renewing TTL

fiam:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved persisted-query cache renewal to reliably extend cache lifetime when valid query content is available.
  - Prevented empty persisted-query data from being saved during renewal.
  - Improved error handling when cached query data cannot be retrieved.
  - Ensured persisted queries remain available across renewal intervals.
  - Verified that changing request variables continues to produce the expected directive behavior.

## Context

Fixes #3062.

Addresses ROUTER-20 and ENG-9964.

This is a smaller, RGR-structured alternative to #3072. The implementation commit credits the original PR author, @Lawal-create, as a co-author.

## What changed

- Keep the original APQ document in Redis when a cache hit renews its TTL, instead of replacing it with the request-specific normalized document.
- Reproduce the corruption using an APQ document whose `@skip` result changes between requests.
- Keep the TTL-renewal details in a private `OperationKit` helper.

## RGR commits

1. Red: reproduce APQ corruption on TTL renewal.
2. Green: preserve the stored APQ query during renewal.
3. Refactor: extract the TTL-renewal helper from `FetchPersistedOperation`.

## Testing

With Redis available on `localhost:6379`, matching the integration-test environment:

- `go test ./operations -run 'TestAutomaticPersistedQueries/redis_cache/query_renewal_preserves_the_stored_query' -count=1`

## Checklist

- [ ] I have discussed my proposed changes in an issue and have received approval to proceed.
- [x] I have followed the coding standards of the project.
- [x] Tests or benchmarks have been added or updated.
- [x] If applicable, I have provided instructions for reviewers for [manually validating my changes](https://github.com/wundergraph/cosmo/blob/main/CONTRIBUTING.md#pull-requests-conventions).
- [ ] Documentation has been updated on [https://github.com/wundergraph/docs-website](https://github.com/wundergraph/docs-website).
- [ ] I have read the [Contributors Guide](https://github.com/wundergraph/cosmo/blob/main/CONTRIBUTING.md).

## Open Source AI Manifesto

This project follows the principles of the [Open Source AI Manifesto](https://human-oss.dev). Please ensure your contribution aligns with its principles.

PR #3197, opened 2026-08-28
refactor(router): separate APQ storage backends

endigma:
Stacked on #3191.

## Summary

- model Redis and in-memory APQ storage as sibling store implementations
- let each backend renew TTLs without routing operation bodies through the caller
- remove the redundant APQ client wrapper and client-name parameters

## Testing

- `cd router && gotestsum --format testdox -- ./internal/persistedoperation/... ./core`

## Summary by CodeRabbit

* **New Features**
  * Improved automatic persisted query (APQ) caching with configurable in-memory or Redis-backed storage.
  * Added configurable cache sizing, key prefixes, and expiration settings.
  * Persisted operations can renew expiration without reloading the full request body.

* **Bug Fixes**
  * Fixed cache-hit expiration renewal.
  * Improved handling across distributed and in-memory caches.

* **Tests**
  * Added coverage for cache initialization, expiration renewal, and unknown persisted queries.


dkorittki (approved):

router/internal/persistedoperation/apq/memory.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+	return nil
+}
+
+func (m *memoryStore) IsDistributed() bool {
+	return false
+}
+
+func (m *memoryStore) Close() {
+	if m.cache.Cache != nil {
```
fiam: question: isn't this check redundant?

router/internal/persistedoperation/apq/store.go
```diff
@@ -0,0 +1,13 @@
...
+
+import (
+	"context"
+)
+
+type Store interface {
+	Get(ctx context.Context, operationHash string) ([]byte, error)
+	Set(ctx context.Context, operationHash string, operationBody []byte) error
+	Renew(ctx context.Context, operationHash string) error
+	IsDistributed() bool
```
fiam, line 11: nit: I think this is leaking the abstraction a bit - not that we have to improve it now (it was already there), but we should improve that at some point

router/internal/persistedoperation/apq/store.go
```diff
@@ -0,0 +1,13 @@
...
+	Close()
```
fiam: question: would it be more useful to use `io.Closer` here? It seems to be some of the underlying drivers to return an error on close that we're swallowing.

router/internal/persistedoperation/apq/redis.go
```diff
@@ -14,17 +14,17 @@ import (
...
-type redisClient struct {
-	logger *zap.Logger
+type redisStore struct {
 	client rd.RDCloser
 	prefix string
+	ttl    time.Duration
 }
 
-func NewRedisClient(opts *RedisOptions) (KVClient, error) {
+func NewRedisStore(opts *RedisOptions) (*redisStore, error) {
```
fiam, line 27: question: what's the rationale for returning a non-exported type here? Wouldn't it be more idiomatic to return the exported interface instead?

router/internal/persistedoperation/apq/memory_test.go
```diff
@@ -0,0 +1,25 @@
+package apq
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestNewMemoryStore(t *testing.T) {
```
fiam, line 10: nit: t.Parallel()

router/internal/persistedoperation/apq/redis.go
endigma: It's idiomatic to return concrete types where possible in Go, returning the interface means that it can only possibly be used as that interface, not as any interface the concrete type satisfies

router/internal/persistedoperation/apq/store.go
endigma: What specifically? `IsDistributed`?

fiam: Yes, I should have been more specific

router/internal/persistedoperation/apq/store.go
endigma: Improved it a bit

router/internal/persistedoperation/apq/store.go
endigma: It's a bit messy but the thing that uses it cannot be easily made part of the inner function

router/internal/persistedoperation/apq/memory.go
endigma: No, it would panic if the cache was nil via a nil pointer dereference.

fiam (approved):
LGTM

PR #3206, opened 2026-09-01
ci: replace custom node setup action with pnpm/setup

pepol:
---
_Generated by [Claude Code](https://claude.ai/code/session_01FUZDuL5JzWEGofTJEDUoXz)_

## Summary by CodeRabbit

* **Chores**
  * Updated project tooling to require pnpm 11 and Node.js 22.22.2.
  * Standardized Node.js and pnpm versions across containerized builds, CI, and release workflows.
  * Improved dependency installation consistency with caching, pinned setup actions, and explicit frozen-lockfile installs.
  * Added workspace configuration for approved native-package builds and dependency compatibility adjustments.
  * Improved deployment compatibility for the updated package manager.
  * Added linting tools to shared development dependencies.


.github/workflows/release-preview.yaml
```diff
@@ -41,7 +41,9 @@ jobs:
           fetch-depth: 0
           token: ${{ steps.app-token.outputs.token }}
 
-      - uses: ./.github/actions/node
+      - uses: pnpm/setup@v2
+        with:
+          cache: true
```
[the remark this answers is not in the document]
pepol: Fixed in 30ffde71d — added `install: false` to all 16 `pnpm/setup@v2` steps across the repo (also pinned the action to an immutable SHA while at it).

---
_Generated by [Claude Code](https://claude.ai/code)_

admission-server/Dockerfile
```diff
@@ -20,7 +20,7 @@ RUN pnpm run --filter=./admission-server build
 # Deploy
 RUN pnpm --filter=./admission-server --prod deploy pruned
 
-FROM --platform=${TARGETPLATFORM} node:lts-slim
+FROM --platform=${TARGETPLATFORM} node:22.22.2-slim
```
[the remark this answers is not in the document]
pepol: Pre-existing: neither Dockerfile set a non-root `USER` before this PR either (both previously used `node:lts-slim`, which has the same root-by-default behavior). This PR only changes the pinned version, not the user model. Leaving this to a separate follow-up rather than widening PLAT-26's scope.

---
_Generated by [Claude Code](https://claude.ai/code)_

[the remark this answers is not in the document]
pepol: Same as the earlier permissions finding on this PR: none of these workflows declared a `permissions:` block before this change either — PLAT-26 only touches the node/pnpm setup step, not the token permission model. Leaving this to a separate follow-up rather than widening scope here.

---
_Generated by [Claude Code](https://claude.ai/code)_

.github/workflows/cdn-ci.yaml
```diff
@@ -21,7 +21,10 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - uses: ./.github/actions/node
+      - uses: pnpm/setup@84cb39b217b10273981911c288cd62326dc7c6d2 # v2
```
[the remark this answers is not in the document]
pepol: Same as the earlier permissions finding on this PR: none of these workflows declared a `permissions:` block before this change either — PLAT-26 is scoped to replacing the custom node/pnpm setup action, not the token permission model. Leaving this to a separate follow-up rather than widening scope here. (Apologies for the delay — my earlier reply to this finding accidentally landed on a different thread.)

---
_Generated by [Claude Code](https://claude.ai/code)_

package.json
```diff
@@ -35,7 +35,14 @@
   ],
   "engines": {
     "node": "22.22.2",
-    "pnpm": "9"
+    "pnpm": "11"
```
comatory, line 38: This could be a breaking change. I have separate ticket for bumping pnpm to v11. Shouldn't we do this gradually and upgrade `pnpm` in separate PR?

comatory (approved):

.github/workflows/cli-binary-release.yaml
```diff
@@ -24,7 +24,10 @@ jobs:
 
       - uses: oven-sh/setup-bun@v2
 
-      - uses: ./.github/actions/node
+      - uses: pnpm/setup@84cb39b217b10273981911c288cd62326dc7c6d2 # v2
```
comatory, line 27: Do they use some sensible version numbers? Not sure about hash but also not so against it :)

admission-server/Dockerfile
```diff
@@ -18,9 +18,11 @@ COPY . /app/
 RUN pnpm run --filter=./admission-server build
 
 # Deploy
-RUN pnpm --filter=./admission-server --prod deploy pruned
+# pnpm v10+ defaults to only deploying workspaces with inject-workspace-packages=true;
+# use --legacy to keep the previous (pnpm 9) deploy behavior we rely on here.
+RUN pnpm --filter=./admission-server --prod deploy --legacy pruned
```
comatory, line 23: Why do we have mismatch between package manager for dev vs production build? I think we should keep it consistent.

package.json
pepol: Verified this against the upstream source directly (not just my earlier assumption):

- `pnpm/setup`'s README states: *"`pnpm/setup@v2` installs pnpm v11 and newer only — it relies on pnpm's self-contained release binaries and the `pnpm runtime` command, both available from v11. If you need pnpm 10 or older, use [`pnpm/action-setup`](https://github.com/pnpm/action-setup) instead."*
- The action's own `version` input description in `action.yml` says: *"Must resolve to pnpm v11 or newer — the action downloads pnpm's native per-platform executable directly, without needing Node.js or npm."*

So the pnpm 9→11 bump isn't a separable enhancement bundled into this migration — it's a hard technical prerequisite for adopting `pnpm/setup@v2` at all (the whole point of PLAT-26). There's no version of this PR that uses the new action without also bumping pnpm to 11+.

Given you have a separate ticket already tracking the pnpm 11 bump, happy to coordinate: either this PR proceeds with the bump bundled in (and your ticket gets closed as superseded/duplicate), or we hold this migration until your ticket lands first and rebase on top of it. Your call — which do you prefer?

---
_Generated by [Claude Code](https://claude.ai/code)_

.github/workflows/cli-binary-release.yaml
pepol: Yes — `pnpm/setup` publishes normal semver-tagged releases (v1, v2, v2.0.0, v2.1.0, etc.), see https://github.com/pnpm/setup/releases. The SHA `84cb39b217b10273981911c288cd62326dc7c6d2` is the exact commit the `v2` tag currently points to (verified independently via the release page and the GitHub API). The trailing `# v2` comment is just for human readability — pinning to the full commit SHA (with that comment) is GitHub's own recommended pattern for treating a third-party action as immutable, since tags like `v2` are mutable and can be moved. So it's not a bare/meaningless hash — it's a specific, real, current release, just locked down against future tampering or accidental tag moves.

---
_Generated by [Claude Code](https://claude.ai/code)_

admission-server/Dockerfile
pepol: There's actually no mismatched package manager here — the whole build and deploy pipeline uses pnpm 11.25.0 throughout (via `corepack enable && corepack prepare pnpm@11.25.0 --activate` on line 6), matching `packageManager` in package.json exactly.

`--legacy` isn't selecting a different pnpm version — it's a flag to the `pnpm deploy` subcommand itself. Starting in pnpm v10, `pnpm deploy` changed its default packaging behavior: by default it now refuses to deploy a workspace unless `inject-workspace-packages: true` is set, because `workspace:*` dependencies are handled differently (symlinked/injected vs copied) than under pnpm 9. `--legacy` restores the pre-v10 packaging behavior we already relied on, so the deployed `pruned/` directory's contents stay identical to what pnpm 9 produced — same pnpm 11 binary, just told to package the output the old way rather than risk changing the shape of the production artifact as part of this migration.

If you'd rather actually adopt native `inject-workspace-packages` mode instead of `--legacy`, I can look into that as a follow-up, but it needs verifying the pruned output still resolves `workspace:*` deps correctly at container runtime before switching — didn't want to bundle that riskier change into this PR.

---
_Generated by [Claude Code](https://claude.ai/code)_

comatory: Ok let's try fixing it in a followup.

package.json
comatory: Lets try the upgrade then.

pepol: Sounds good — the bump is already in this PR (pnpm 9.12.3 → 11.25.0, lockfile regenerated, `allowBuilds` added for the new build-script approval gate), and CI has been green on it since. I'll leave a note for you to close/update your separate ticket as covered by this PR.

---
_Generated by [Claude Code](https://claude.ai/code)_

comatory (approved):

gmas (approved):

Noroth (approved):

PR #3208, opened 2026-09-02
feat: add an option to set default headers

JivusAyrus:
## Summary by CodeRabbit

- **New Features**
  - Added configurable Playground default headers at graph and personal scopes.
  - Added a dialog to view, add, edit, remove, and preview effective headers.
  - Personal headers can override graph-level values, with built-in defaults used when none are configured.
  - Playground requests now initialize with configured effective headers.
  - Added permission-aware editing, loading states, and feedback for validation or save errors.

- **Bug Fixes**
  - Header names are validated case-insensitively, including duplicate detection and blank-key filtering.
  - Header values safely support spaces and tabs.
  - Clearing or updating headers preserves scope and user-specific behavior.

## Checklist


controlplane/src/core/bufservices/playground/updatePlaygroundDefaultHeaders.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    if (req.personalHeaders && !authContext.rbac.hasFederatedGraphReadAccess(federatedGraph)) {
+      throw new UnauthorizedError();
+    }
+
+    const playgroundRepo = new PlaygroundDefaultHeadersRepository(opts.db, authContext.organizationId);
+    await playgroundRepo.update({
+      federatedGraphId: federatedGraph.id,
+      userId: authContext.userId,
+      graphHeaders: req.graphHeaders?.headers.map((h) => ({ key: h.key, value: h.value })),
```
comatory: Why do we need to call `.map` here? It looks like we're just mapping the same properties. If this needs some kind of type coercion, I'd do it in the validation loop to keep the loops to a minimum. WDYT?

controlplane/src/core/util.ts
```diff
@@ -54,6 +59,32 @@ const namespaceRegex = /^[\da-z]+(?:[_-][\da-z]+)*$/;
 const schemaTagRegex = /^(?![/-])[\d/A-Za-z-]+(?<![/-])$/;
 const graphNameRegex = /^[\dA-Za-z]+(?:[./@_-][\dA-Za-z]+)*$/;
 const pluginVersionRegex = /^v\d+$/;
+// Matches studio/src/lib/playground-headers.ts:isValidHeaderName. Keep the two in sync.
```
comatory, line 63: Can you please create sub-issue in https://linear.app/wundergraph/issue/COSMO-425/shared-typescript-monorepo-package-for-utilities-and-types? I'd like to map which code we want to re-use between packages. We already have few cases where we duplicate and I'd love if we could start sharing some of the code.

comatory:
This reads a bit awkwardly. So we treat string as error and `undefined` as OK. I think we should model the outcomes better

1. Create custom error sub-class which is specific, e.g. `HeaderValidationError` and then use `try/catch` at the call-site to check for `instanceof HeaderValidationError`
2. Make the function `validatePlaygroundHeaders` return a result object with `{ success: true } | { success: false; errors: Array<Error> }`

_Could not attach this as a review thread; GitHub did not expose the created thread. Line link: https://github.com/wundergraph/cosmo/blob/4dcaf086a69774dda68cb294464f145da0fe2031/controlplane/src/core/util.ts#L376_

comatory:
Could we model what's valid return value vs errors more explicitly?

1. Custom error `HeaderValidationError` so it can be intercepted with try/catch + `instanceof HeaderValidationError` check at the call-site
OR
2. Result object: `{ success: true } | { success: false; errors: Array<Error> }`

It's not very ergonomic to check for undefined vs string value to decide what's invalid state.

_Could not attach this as a review thread; GitHub did not expose the created thread. Line link: https://github.com/wundergraph/cosmo/blob/26f8984d40c1c2903ac88c30531394db7507c19a/controlplane/src/core/util.ts#L371_

controlplane/src/core/util.ts
```diff
@@ -351,6 +360,32 @@ export const isValidPluginVersion = (version: string): boolean => {
...
 
+export const isValidHeaderName = (name: string): boolean => {
+  return headerNameRegex.test(name);
+};
+
+/**
+ * Returns the first problem found in a playground header list, or undefined when it is valid.
+ * `scopeLabel` names the list in the returned message, e.g. 'graph' or 'personal'.
+ */
+export const validatePlaygroundHeaders = (headers: PlaygroundHeader[], scopeLabel: string): string | undefined => {
```
comatory: It's hard to figure out what's valid value here vs what's an error. I think result object

```
{ success: true } | { success: false; errors: Array<Error> }
```

would work better here.

controlplane/test/playground/playground-default-headers.test.ts
```diff
@@ -0,0 +1,680 @@
...
+      namespace: 'default',
+      graphHeaders: { headers: [{ key: 'x-tenant-id', value: 'acme' }] },
+    });
+    expect(updateRes.response?.code).toBe(EnumStatusCode.OK);
+
+    const res = await client.getPlaygroundDefaultHeaders({
+      federatedGraphName: graphName,
+      namespace: 'default',
+    });
+    expect(res.graphHeaders.map(({ key, value }) => ({ key, value }))).toEqual([{ key: 'x-tenant-id', value: 'acme' }]);
```
comatory: I think the map is unnecessary: `expect(res.graphHeaders).toEqual(expect.arrayContaining([ expect.objectContaining({ key: ..., value ...`

can you change it to this syntax in all the places?

controlplane/test/playground/playground-default-headers.test.ts
```diff
@@ -0,0 +1,680 @@
...
+    const after = await client.getPlaygroundDefaultHeaders({
+      federatedGraphName: graphName,
+      namespace: 'default',
+    });
+    expect(after.response?.code).toBe(EnumStatusCode.OK);
+    expect(after.graphHeaders).toEqual([]);
+    expect(after.personalHeaders).toEqual([]);
+  });
+
+  test('A case-only duplicate in the personal list must not persist a valid graph list', async () => {
```
comatory: I don't understand this test, maybe it's the wording. But why `X-A` header is not persisted?

studio/src/components/playground/default-headers-dialog.tsx
```diff
@@ -0,0 +1,258 @@
+import { GraphContext } from '@/components/layout/graph-layout';
```
comatory: Can we create a folder in components `default-headers/` and break the components in this file into individual files please? 🙏

studio/src/lib/playground-headers.ts
```diff
@@ -50,3 +58,67 @@ export const substituteHeadersFromEnv = (headers: Record<string, string>, graphI
...
+ * existing key case-insensitively replaces that value in place and contributes
+ * its own spelling of the key. Personal-only keys are appended.
+ */
+export const mergeDefaultHeaders = (
+  graph: DefaultHeaderEntry[],
+  personal: DefaultHeaderEntry[],
+): DefaultHeaderEntry[] => {
+  const withNonEmptyKeys = (entries: DefaultHeaderEntry[]) => entries.filter((entry) => entry.key.trim() !== '');
+
+  const merged = withNonEmptyKeys(graph).map((entry) => ({ ...entry }));
```
comatory: Why are we spreading the object with a `map`? This does not seem to do anything. Can we avoid it?

studio/src/lib/playground-headers.ts
```diff
@@ -50,3 +58,67 @@ export const substituteHeadersFromEnv = (headers: Record<string, string>, graphI
...
+ * Combines the organization-shared graph defaults with the caller's personal
+ * defaults. Graph order is preserved; a personal entry whose key matches an
+ * existing key case-insensitively replaces that value in place and contributes
+ * its own spelling of the key. Personal-only keys are appended.
+ */
+export const mergeDefaultHeaders = (
+  graph: DefaultHeaderEntry[],
+  personal: DefaultHeaderEntry[],
+): DefaultHeaderEntry[] => {
+  const withNonEmptyKeys = (entries: DefaultHeaderEntry[]) => entries.filter((entry) => entry.key.trim() !== '');
```
comatory: I find creating this function inside a function harder to read and unnecessary. I'd rather if we do

```
const merged = graph.filter((entry) => entry.key.trim().length > 0)
...
const personal = graph.filter((entry) => entry.key.trim().length > 0)
```

studio/src/lib/playground-headers.ts
```diff
@@ -50,3 +58,67 @@ export const substituteHeadersFromEnv = (headers: Record<string, string>, graphI
...
+};
+
+/**
+ * Serializes header entries into the JSON string shape GraphiQL's header editor
+ * expects.
+ */
+export const defaultHeadersToJsonString = (entries: DefaultHeaderEntry[]): string => {
+  const asObject: Record<string, string> = {};
+
+  for (const entry of entries) {
```
comatory: What if we refactored this into an anonymous function that gets passed as 2nd argument to `JSON.stringify`? That seems more concise. WDYT?

comatory (changes requested):

controlplane/test/playground/playground-default-headers.test.ts
```diff
@@ -0,0 +1,680 @@
...
+    const afterPersonalCleared = await client.getPlaygroundDefaultHeaders({
+      federatedGraphName: graphName,
+      namespace: 'default',
+    });
+    expect(afterPersonalCleared.graphHeaders.map(({ key, value }) => ({ key, value }))).toEqual([
+      { key: 'x-tenant-id', value: 'acme' },
+    ]);
+    expect(afterPersonalCleared.personalHeaders).toEqual([]);
+
+    // And the symmetric case: restore the personal scope, then clear the graph one.
```
comatory: This should be separate unit test. Too many expects for this unit test IMO.

controlplane/test/playground/playground-default-headers.test.ts
```diff
@@ -0,0 +1,680 @@
...
+      const res = await client.updatePlaygroundDefaultHeaders({
+        federatedGraphName: graphName,
+        namespace: 'default',
+        ...(scope === 'graph' ? { graphHeaders: headers } : { personalHeaders: headers }),
+      });
+
+      expect(res.response?.code).toBe(EnumStatusCode.ERR);
+    },
+  );
+  test('Updating personal headers twice replaces rather than duplicates', async () => {
```
comatory, line 653: I feel like this is identical test to "Updating the same scope twice replaces rather than duplicates"

studio/src/components/playground/default-headers-dialog.tsx
```diff
@@ -0,0 +1,258 @@
...
+
+// A blank key is a row the user has not filled in yet, not an error - it is
+// filtered out before saving.
+// Strips the protobuf message wrapper down to the plain entries the editor uses.
+const fromServer = (headers?: { key: string; value: string }[]): DefaultHeaderEntry[] =>
+  (headers ?? []).map((h) => ({ key: h.key, value: h.value }));
+
+const isInvalidKey = (entry: DefaultHeaderEntry) => entry.key.trim() !== '' && !isValidHeaderName(entry.key);
+
+const HeaderRows = ({ entries, disabled = false, onChange }: HeaderRowsProps) => {
```
comatory: Should go to a separate file.

studio/src/components/playground/default-headers-dialog.tsx
```diff
@@ -0,0 +1,258 @@
...
+              {isInvalid && <p className="mt-1 text-xs text-destructive">Not a valid HTTP header name</p>}
+            </div>
+            <Input
+              aria-label="Header value"
+              placeholder="Value"
+              value={entry.value}
+              disabled={disabled}
+              onChange={(e) => update(index, { value: e.target.value })}
+              className="flex-1"
+            />
```
comatory: This should be a self contained component <HeaderValueInput /> which could be abstracted also for the header values (L65-72). We also discussed on a call that the input should use code editor for inserting headers and its values as a JSON object. I would encapsulate this into this new proposed component.

comatory (changes requested):

controlplane/test/playground/playground-default-headers.test.ts
JivusAyrus: That wont check it completely right? like the array can have more elements in that case right?

controlplane/test/playground/playground-default-headers.test.ts
JivusAyrus: X-A and x-a are duplicates so we error out. Changing the title

studio/src/lib/playground-headers.ts
JivusAyrus: we have to filter the empty keys before sending it to JSON.stringify() right?

controlplane/test/playground/playground-default-headers.test.ts
JivusAyrus: one test is testing with graph headers the other with personal headers

controlplane/test/playground/playground-default-headers.test.ts
comatory: Ah you're right, omit `expect.arrayContaining` then and it should check for

1. order of the items in array
2. number of the items in array

I just think it's better to explicitly list the object containing key + value instead of running `.map` on it. It is easier to read and avoids unnecessary mapping.

studio/src/lib/playground-headers.ts
```diff
@@ -50,3 +58,117 @@ export const substituteHeadersFromEnv = (headers: Record<string, string>, graphI
...
+
+/**
+ * Serializes header entries into the JSON string shape GraphiQL's header editor
+ * expects.
+ */
+export const defaultHeadersToJsonString = (entries: DefaultHeaderEntry[]): string => {
+  const asObject = Object.fromEntries(
+    entries.filter((entry) => entry.key.trim().length > 0).map((entry) => [entry.key, entry.value]),
+  );
+
```
comatory, line 103: I think we should, otherwise {[undefined]:""} becomes `"undefined":""`. I'm just saying to refactor this a bit so the function is not outside of `JSON.stringify`, it seems unnecessarily verbose.

comatory (changes requested):

studio/src/lib/playground-headers.ts
JivusAyrus: but key cant be undefined right?

comatory: Technically it can, but now I see it gets omitted by stringification. IDK why I thought it would become "undefined", I must've confused it with some other behaviour:

image

So I think the logic is fine because I guess you want to omit empty `""` keys. I'm just saying to not have `asObject` as separate function.

comatory (approved):

controlplane/src/core/util.ts
JivusAyrus: Done

PR #3215, opened 2026-09-04
fix: apply rbac correctly when groups are limited to specific subgraphs

wilsonrivera:
## Summary by CodeRabbit

- **New Features**
  - Feature-flag listings now respect access permissions based on assigned namespaces and resources.
  - Users with access to related feature subgraphs can view the corresponding feature flags.
  - Unrestricted roles continue to provide access when no namespace or resource restrictions are configured.

- **Bug Fixes**
  - Prevented unauthorized feature flags from appearing in listings.

- **Tests**
  - Added coverage for scoped, unrestricted, unrelated, and unauthorized access scenarios.

## Checklist

- [x] If applicable, I have provided instructions for reviewers for [manually validating my changes](https://github.com/wundergraph/cosmo/blob/main/CONTRIBUTING.md#pull-requests-conventions).

## Open Source AI Manifesto

### Steps to reproduce before fix
- Create a feature flag with a feature subgraph
- Create a group with the `subgraph-*` role and assign the specific feature subgraph to it
- Assign the new group to a user
- Try to retrieve the list of feature flags with the user assigned to the group


controlplane/src/core/repositories/FeatureFlagRepository.ts
```diff
@@ -324,19 +354,44 @@ export class FeatureFlagRepository {
...
+                .from(featureFlagToFeatureSubgraphs)
+                .innerJoin(subgraphs, eq(subgraphs.id, featureFlagToFeatureSubgraphs.featureSubgraphId))
+                .where(
+                  and(
+                    eq(featureFlagToFeatureSubgraphs.featureFlagId, featureFlags.id),
+                    inArray(subgraphs.targetId, uniqueResources),
+                  ),
+                ),
+            )
+          : inArray(schema.targets.id, uniqueResources),
```
JivusAyrus, line 390: can we do it differently? because in the future if we have to use this function and we do the same mistake as now, it will still error our right?

JivusAyrus (changes requested):

controlplane/src/core/repositories/FeatureFlagRepository.ts
wilsonrivera: It shouldn't. there is a `scope` parameter that will use the expected branch, this method is private, the only reason this would fail in the future is if we pass the wrong scope to this function

JivusAyrus (approved):

PR #3223, opened 2026-09-09
feat: add minimal support for newer federation directives

Aenimus:
## Summary by CodeRabbit

- **New Features**
  - Added directive definitions and validation support for cache tags, context values, federation policies, and related directives.
  - Added optional `label` support to `@override` definitions.

- **Bug Fixes**
  - Improved schema normalization and validation for related directive arguments and scalar values.

- **Warnings**
  - Router compatibility version one now reports unsupported directive usage.
  - Added warnings when `@override(label:)` values are ignored, with guidance for compatible behavior.

- **Tests**
  - Added coverage for unsupported directives, generated scalar definitions, and labeled overrides.

## Checklist

- [ ] I have discussed my proposed changes in an issue and have received approval to proceed.
- [ ] I have followed the coding standards of the project.
- [ ] Tests or benchmarks have been added or updated.
- [ ] If applicable, I have provided instructions for reviewers for [manually validating my changes](https://github.com/wundergraph/cosmo/blob/main/CONTRIBUTING.md#pull-requests-conventions).
- [ ] Documentation has been updated on [https://github.com/wundergraph/docs-website](https://github.com/wundergraph/docs-website).
- [ ] I have read the [Contributors Guide](https://github.com/wundergraph/cosmo/blob/main/CONTRIBUTING.md).

## Open Source AI Manifesto

This project follows the principles of the [Open Source AI Manifesto](https://human-oss.dev). Please ensure your contribution aligns with its principles.

PR #3224, opened 2026-09-09
fix(ci): regenerate stale router test config

gausie:
Router CI's `git-dirty-check` is failing on every PR that touches a router path (#3196, #3208), on a file none of them changed.

#3223 added the `label` argument to `@override`, which changes the composed demo config:

```
- directive @override(from: String!)
+ directive @override(from: String!, label: String)
```

`router-tests/testenv/testdata/config.json` was not regenerated, so `./update-config-no-edg.sh` in CI produces those two lines plus the two corresponding schema version hashes. Regenerated here with the same steps CI runs.

Router CI's path filter didn't include `composition/**`, so #3223 never ran the dirty check and the staleness landed on main unnoticed. Added it, so a composition change that alters the composed config is caught by the PR making it. Happy to drop that half if you'd rather not have Router CI run on composition PRs.

`configWithEdfs.json` is stale the same way, but it comes from `update-config.sh`, which CI doesn't run, so it isn't failing anything and I've left it alone.

## Summary by CodeRabbit

- **Chores**
  - Continuous integration now runs for changes within the composition configuration area.

- **Tests**
  - Updated federation test data to reflect current product subgraph schemas, including support for labeled override directives.
  - Refreshed stored schema and content metadata used during validation.


gausie:
Merged into #3225 to avoid catch 22

PR #3227, opened 2026-09-09
feat: implement script for enabling `split-config-loading`

wilsonrivera:
## Summary by CodeRabbit

* **New Features**
  * Added support for enabling split configuration loading for selected organizations and deploying updated federated graphs and feature configurations.
  * Added primary and failover blob storage support during deployments.
  * Added an optional deployment job that runs during installation or upgrade and waits for required services.

* **Configuration**
  * Configuration output can include optional webhook proxy, CDN base URL, and admission JWT secret values.
  * Added settings for configuring the split configuration loading job, including the target organization.

* **Documentation**
  * Updated deployment configuration documentation with the new job settings.

## Checklist

## Open Source AI Manifesto


controlplane/src/bin/enable-split-config-loading.ts
```diff
@@ -0,0 +1,182 @@
...
+  // Ensure that the feature has not been
+  const feature = await orgRepo.getFeature({
+    organizationId: org.id,
+    featureId: 'split-config-loading',
+  });
+
+  if (feature?.enabled) {
+    console.log(`The feature has already been enable for "${org.name}" (${org.id})`);
+    // eslint-disable-next-line unicorn/no-process-exit
+    process.exit(1);
```
pepol, line 69: Why do we signal an error here?

I think it might be cleaner for the script to be idempotent, i.e. if feature is enabled, just say it's enabled already and signal success (as the intention - to have the feature flag enabled - is fulfilled).

helm/cosmo/charts/controlplane/templates/enable-split-confiog-loading.yaml
```diff
@@ -0,0 +1,176 @@
+{{ if .Values.jobs.enableSplitConfigLoading.enabled }}
```
pepol, line 1: This is not a good idea.

I think a better approach would be for the script to support (idempotently) a list of organizations to enable and then enable the feature for all of them (while skipping the ones already enabled).

This as-is would also cause each deployment to fail, if not disabled after the initial "enable" run, because:
- the post-install,post-upgrade helm hook (this file, L12) would trigger a new run for the same org ID at a new deployment
- the run would return exit code 1 because the feature is already enabled (controlplane/src/bin/enable-split-config-loading.ts, L69)
- this would cause the job to be retried 10 times (this file, L19) before being declared failed
- that would cause the kapp deployment to be marked as failed and return the failure in CI run

pepol (changes requested):

pepol (approved):

JivusAyrus (approved):

thisisnithin (approved):

PR #3229, opened 2026-09-09
fix: strip context arguments and internal scalars from federated schemas

gausie:
Adds federation-level test coverage for `@context`/`@fromContext`, which had none, and fixes what it turned up. The directives themselves remain unsupported. Further PRs will add more validation to these directives, this is mostly just to establish a baseline with what we already introduced in #3223.

* An argument carrying `@fromContext` is populated by the router, so it must not appear in the client schema.
* A subgraph that declared `ContextFieldValue`, `federation__Policy` or `openfed__Scope` itself no logner leaks that scalar into the client schema.

## Summary by CodeRabbit

* **New Features**
  * Improved support for `@context` and `@fromContext` directives during schema composition.
  * Contextual arguments and values are omitted from generated client schemas where appropriate.
  * Added validation and clearer errors for inconsistent contextual arguments and interface implementations.

* **Bug Fixes**
  * Improved handling across federated schemas, extensions, and ordering variations.
  * Corrected treatment of special federation scalars in generated schemas.

* **Tests**
  * Added comprehensive coverage for contextual directives, schema generation, validation, federation behavior, and scalar handling.

### DIFF
--- a/composition/src/errors/errors.ts
+++ b/composition/src/errors/errors.ts
@@ -30,6 +30,7 @@ import {
   ARGUMENT,
   FIELD,
   FIELD_PATH,
+  FROM_CONTEXT,
   IN_UPPER,
   INPUT_FIELD,
   INTERFACE,
@@ -623,6 +624,28 @@ export function invalidRequiredInputValueError(
       ` as optional on all other definitions of that ${typeString} in all other subgraphs.\n`;
   }
   return new Error(message);
+}
+
+export function contextArgumentRequiredError(
+  coords: string,
+  contextSubgraphNames: Array<string>,
+  requiredSubgraphNames: Array<string>,
+): Error {
+  return new Error(
+    `The ${ARGUMENT} "${coords}" is invalid because:\n` +
+      ` It is declared "@${FROM_CONTEXT}" in the following subgraph` +
+      (contextSubgraphNames.length > 1 ? 's' : '') +
+      ': "' +
+      contextSubgraphNames.join(QUOTATION_JOIN) +
+      `"\n` +
+      ` However, it is required in the following subgraph` +
+      (requiredSubgraphNames.length > 1 ? 's' : '') +
+      ': "' +
+      requiredSubgraphNames.join(QUOTATION_JOIN) +
+      `"\n` +
+      ` An ${ARGUMENT} declared "@${FROM_CONTEXT}" is populated by the router rather than the client, so it must` +
+      ` not be required in any subgraph in which it is not.\n`,
+  );
 }
 
 export function duplicateArgumentsError(fieldPath: string, duplicatedArguments: string[]): Error {
--- a/composition/src/v1/constants/strings.ts
+++ b/composition/src/v1/constants/strings.ts
@@ -13,14 +13,12 @@ import {
   EDFS_KAFKA_SUBSCRIBE,
   EDFS_NATS_PUBLISH,
   EDFS_NATS_REQUEST,
-  EDFS_NATS_STREAM_CONFIGURATION,
   EDFS_NATS_SUBSCRIBE,
   EDFS_REDIS_PUBLISH,
   EDFS_REDIS_SUBSCRIBE,
   ENUM_UPPER,
   ENUM_VALUE_UPPER,
   FIELD_DEFINITION_UPPER,
-  FIELD_SET_SCALAR,
   FLOAT_SCALAR,
   FROM_CONTEXT,
   ID_SCALAR,
@@ -32,8 +30,6 @@ import {
   INTERFACE_UPPER,
   KEY,
   LINK,
-  LINK_IMPORT,
-  LINK_PURPOSE,
   NOT_UPPER,
   OBJECT_UPPER,
   ONE_OF,
@@ -47,14 +43,11 @@ import {
   SEMANTIC_NON_NULL,
   STREAM_NAME,
   STRING_SCALAR,
-  SUBSCRIPTION_FIELD_CONDITION,
   SUBSCRIPTION_FILTER,
-  SUBSCRIPTION_FILTER_CONDITION,
-  SUBSCRIPTION_FILTER_VALUE,
   TAG,
   UNION_UPPER,
 } from '../../utils/string-constants';
-import { type DefinitionNode, Kind } from 'graphql';
+import { Kind, type TypeDefinitionNode } from 'graphql';
 import { type DirectiveName, type FieldName, type TypeName } from '../../types/types';
 import {
   CONTEXT_FIELD_VALUE_DEFINITION,
@@ -114,9 +107,9 @@ export const COMPOSITE_OUTPUT_NODE_KINDS: ReadonlySet<Kind> = new Set<Kind>([
   Kind.OBJECT_TYPE_EXTENSION,
 ]);
 
-export const DEPENDENCIES_BY_DIRECTIVE_NAME: ReadonlyMap<DirectiveName, Array<DefinitionNode>> = new Map<
+export const DEPENDENCIES_BY_DIRECTIVE_NAME: ReadonlyMap<DirectiveName, Array<TypeDefinitionNode>> = new Map<
   DirectiveName,
-  Array<DefinitionNode>
+  Array<TypeDefinitionNode>
 >([
   [CONNECT_FIELD_RESOLVER, [FIELD_SET_SCALAR_DEFINITION]],
   [EDFS_NATS_SUBSCRIBE, [EDFS_NATS_STREAM_CONFIGURATION_DEFINITION]],
@@ -139,17 +132,11 @@ export const DEPENDENCIES_BY_DIRECTIVE_NAME: ReadonlyMap<DirectiveName, Array<De
 
 export const IGNORED_FEDERATED_TYPE_NAMES: ReadonlySet<TypeName> = new Set<TypeName>([
   BOOLEAN_SCALAR,
-  EDFS_NATS_STREAM_CONFIGURATION,
-  FIELD_SET_SCALAR,
+  FLOAT_SCALAR,
   ID_SCALAR,
   INT_SCALAR,
-  FLOAT_SCALAR,
-  LINK_IMPORT,
-  LINK_PURPOSE,
   STRING_SCALAR,
-  SUBSCRIPTION_FIELD_CONDITION,
-  SUBSCRIPTION_FILTER_CONDITION,
-  SUBSCRIPTION_FILTER_VALUE,
+  ...[...DEPENDENCIES_BY_DIRECTIVE_NAME.values()].flat().map(({ name }) => name.value),
 ]);
 
 export const ROUTER_FEDERATED_DIRECTIVE_NAMES: ReadonlySet<DirectiveName> = new Set<DirectiveName>([
--- a/composition/src/v1/federation/federation-factory.ts
+++ b/composition/src/v1/federation/federation-factory.ts
@@ -27,6 +27,7 @@ import {
   allChildDefinitionsAreInaccessibleError,
   allExternalFieldInstancesError,
   configureDescriptionPropagationError,
+  contextArgumentRequiredError,
   inaccessibleQueryRootTypeError,
   inaccessibleRequiredInputValueError,
   incompatibleFederatedFieldNamedTypeError,
@@ -145,6 +146,7 @@ import {
   getDefinitionDataCoords,
   getInitialFederatedDescription,
   getSubscriptionFilterValue,
+  isInputValueDataFromContext,
   isLeafKind,
   isNodeDataInaccessible,
   isParentDataCompositeOutputType,
@@ -769,6 +771,10 @@ export class FederationFactory {
     );
     setLongestDescription(targetData, incomingData);
     addIterableToSet({
+      source: incomingData.contextSubgraphNames,
+      target: targetData.contextSubgraphNames,
+    });
+    addIterableToSet({
       source: incomingData.requiredSubgraphNames,
       target: targetData.requiredSubgraphNames,
     });
@@ -1119,6 +1125,7 @@ export class FederationFactory {
   copyInputValueData(sourceData: InputValueData): InputValueData {
     return {
       configureDescriptionDataBySubgraphName: copyObjectValueMap(sourceData.configureDescriptionDataBySubgraphName),
+      contextSubgraphNames: new Set(sourceData.contextSubgraphNames),
       directivesByName: copyArrayValueMap(sourceData.directivesByName),
       federatedCoords: sourceData.federatedCoords,
       fieldName: sourceData.fieldName,
@@ -1934,6 +1941,22 @@ export class FederationFactory {
     const invalidRequiredArguments: InvalidRequiredInputValueData[] = [];
     const fieldCoords = `${fieldData.renamedParentTypeName}.${fieldData.name}`;
     for (const [argumentName, inputValueData] of fieldData.argumentDataByName) {
+      if (isInputValueDataFromContext(inputValueData)) {
+        const requiredSubgraphNames = getEntriesNotInHashSet(
+          inputValueData.requiredSubgraphNames,
+          inputValueData.contextSubgraphNames,
+        );
+        if (requiredSubgraphNames.length > 0) {
+          this.errors.push(
+            contextArgumentRequiredError(
+              inputValueData.federatedCoords,
+              [...inputValueData.contextSubgraphNames],
+              requiredSubgraphNames,
+            ),
+          );
+        }
+        continue;
+      }
       if (fieldData.subgraphNames.size === inputValueData.subgraphNames.size) {
         argumentNames.push(argumentName);
         const argumentNodeResult = routerSchemaInputValueNodeFromData({
--- a/composition/src/v1/normalization/normalization-factory.ts
+++ b/composition/src/v1/normalization/normalization-factory.ts
@@ -303,6 +303,7 @@ import {
   EXTERNAL,
   FIELDS,
   FIRST_ORDINAL,
+  FROM_CONTEXT,
   HYPHEN_JOIN,
   INACCESSIBLE,
   INCLUDE_HEADERS,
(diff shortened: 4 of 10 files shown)


composition/tests/v1/utils/utils.ts
```diff
@@ -21,6 +21,8 @@ export const CONTEXT_DIRECTIVE = `
   directive @context(name: String!) repeatable on INTERFACE | OBJECT | UNION
 `;
 
+export const CONTEXT_FIELD_VALUE_SCALAR = `  scalar ContextFieldValue`;
```
Aenimus, line 24: Why the whitespace? Good shout otherwise.

composition/src/v1/constants/strings.ts
```diff
@@ -139,17 +132,11 @@ export const DEPENDENCIES_BY_DIRECTIVE_NAME: ReadonlyMap<DirectiveName, Array<De
...
   ID_SCALAR,
   INT_SCALAR,
-  FLOAT_SCALAR,
-  LINK_IMPORT,
-  LINK_PURPOSE,
   STRING_SCALAR,
-  SUBSCRIPTION_FIELD_CONDITION,
-  SUBSCRIPTION_FILTER_CONDITION,
-  SUBSCRIPTION_FILTER_VALUE,
+  ...[...DEPENDENCIES_BY_DIRECTIVE_NAME.values()].flat().map(({ name }) => name.value),
```
Aenimus, line 139: I don't like coupling these. For example. some have the same dependencies (which is handled by the Set, but it's still ugly IMO)

composition/tests/v1/utils/utils.ts
gausie: Only because OPENFED_FIELD_SET and OPENFED_SCOPE have them already. I can remove?

Aenimus: Oh, weird. can you remove the whitespace in those as well please?

composition/tests/v1/federation-factory.test.ts
```diff
@@ -622,6 +622,52 @@ describe('FederationFactory tests', () => {
     );
   });
 
+  test('that a manually defined federation__Policy scalar is not included in the federated graph', () => {
```
Aenimus, line 625: Please can you inline the subgraphs with `createSubgraph`? That's the new approach we're taking to writing tests.

Aenimus: And please use interpolation where possible (rather than concatenation—also the new approach)

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -1365,9 +1366,11 @@ export class NormalizationFactory {
       ? `${federatedParentTypeName}${fieldName ? `.${fieldName}` : ''}(${name}: ...)`
       : `${federatedParentTypeName}.${name}`;
     const namedTypeName = getTypeNodeNamedTypeName(node.type);
+    const directivesByName = this.extractDirectives(node, new Map<string, ConstDirectiveNode[]>());
```
Aenimus: the key should be typed as DirectiveName

composition/src/v1/federation/federation-factory.ts
```diff
@@ -1934,6 +1941,22 @@ export class FederationFactory {
     const invalidRequiredArguments: InvalidRequiredInputValueData[] = [];
     const fieldCoords = `${fieldData.renamedParentTypeName}.${fieldData.name}`;
     for (const [argumentName, inputValueData] of fieldData.argumentDataByName) {
+      if (isInputValueDataFromContext(inputValueData)) {
+        const requiredSubgraphNames = getEntriesNotInHashSet(
+          inputValueData.requiredSubgraphNames,
+          inputValueData.contextSubgraphNames,
+        );
+        if (requiredSubgraphNames.length > 0) {
```
Aenimus: fromContext args always need to be non-nullable?

gausie: No they have to be nullable, but I wanted validation to be out of scope for this PR.

composition/tests/v1/utils/utils.ts
gausie: Hm it needs a leading space otherwise you get "}scalar" in the results, now I need to fix that.

composition/tests/v1/federation-factory.test.ts
gausie: I think I've now done this?

composition/src/v1/federation/federation-factory.ts
Aenimus: You don’t need to validate everything in this PR but this is misleading and confusing.

gausie: Ok for now I can just reject if you require the context argument in any subgraph

composition/tests/v1/utils/utils.ts
Aenimus: Never mind then. Keep the whitespace. It will be an issue with `normalizeString`

composition/tests/v1/directives/fieldset-directives.test.ts
```diff
@@ -754,7 +754,8 @@ describe('openfed_FieldSet tests', () => {
           type Entity @key(fields: "id") {
             id: ID!
             name: String! @requires(fields: "__typename")
-          }` +
+          }
+        ` +
```
Aenimus: Is this the test that broke with changing the constant? I think it works fine with a new line and interpolation

composition/tests/v1/directives/context.test.ts
```diff
@@ -0,0 +1,840 @@
...
+import {
+  CONTEXT_DIRECTIVE,
+  CONTEXT_FIELD_VALUE_SCALAR,
+  FROM_CONTEXT_DIRECTIVE,
+  KEY_DIRECTIVE,
+  OPENFED_FIELD_SET,
+  SCHEMA_QUERY_DEFINITION,
+} from '../utils/utils';
+
+const subgraphWithContextOnInterface = createSubgraph(
```
Aenimus: Please in-line these subgraphs in each test. We're moving away from re-used fixtures because it's easier to read if they're just next to the test.

composition/tests/v1/directives/context.test.ts
```diff
@@ -0,0 +1,840 @@
...
+
+          type User implements Node {
+            id: ID!
+            name: String!
+          }
+        `),
+      );
+    });
+
+    it('preserves @context on a union type', () => {
```
Aenimus: `it` is not convention, `test` is preferred

composition/tests/v1/directives/context.test.ts
```diff
@@ -0,0 +1,840 @@
...
+  }
+
+  type Wallet {
+    apply: Int!
+    id: ID!
+  }
+`);
+
+describe('@context and @fromContext directives', () => {
+  describe('normalisation', () => {
```
Aenimus: nit: Normalization tests

composition/tests/v1/directives/context.test.ts
```diff
@@ -0,0 +1,840 @@
...
+            id: ID!
+            locale: String!
+            profile: Profile!
+          }
+        `),
+      );
+    });
+
+    it('strips an argument that only one subgraph declares as a context argument', () => {
+      const { federatedGraphClientSchema, federatedGraphSchema } = federateSubgraphsSuccess(
```
Aenimus, line 586: for example, this test is really hard to read because we don't know what input or output is.

composition/tests/v1/directives/context.test.ts
gausie: I like to make the test structure readable, "describe normalisation" makes sense but "describe normalisation tests" doesn't and is redundant.

composition/tests/v1/directives/context.test.ts
gausie: This builds on what I said about describe - I like to make them readable "It preserves context on a union type". I don't think it's particularly important that the function used is a convention one way or the other though, and would ideally keep this with an eye to moving towards this style in future.

composition/tests/v1/directives/context.test.ts
gausie: I'm happy to do this, I only did it the other way because I thought it was a meaningful existing convention but I didn't love it.

composition/tests/v1/directives/context.test.ts
gausie: Agreed, fixed

composition/tests/v1/directives/context.test.ts
Aenimus: i was referring to z rather than s, which is the standard across the code base

composition/src/errors/errors.ts
```diff
@@ -625,6 +636,27 @@ export function invalidRequiredInputValueError(
...
 }
 
+export function contextArgumentRequiredError(
+  coords: string,
+  contextSubgraphNames: Array<string>,
+  requiredSubgraphNames: Array<string>,
+): Error {
+  return new Error(
+    `The ${ARGUMENT} "${coords}" is invalid because:\n` +
+      ` It is declared "@${FROM_CONTEXT}" in the following subgraph` +
```
Aenimus: It is declared -> It defines

composition/src/errors/errors.ts
```diff
@@ -625,6 +636,27 @@ export function invalidRequiredInputValueError(
...
+      (contextSubgraphNames.length > 1 ? 's' : '') +
+      ': "' +
+      contextSubgraphNames.join(QUOTATION_JOIN) +
+      `"\n` +
+      ` However, it is required in the following subgraph` +
+      (requiredSubgraphNames.length > 1 ? 's' : '') +
+      ': "' +
+      requiredSubgraphNames.join(QUOTATION_JOIN) +
+      `"\n` +
+      ` An ${ARGUMENT} that the router populates rather than the client must not be required in any subgraph.\n`,
```
Aenimus: in any subgraph -> in all subgraphs that reference it

?

composition/src/errors/errors.ts
```diff
@@ -585,6 +586,16 @@ export function invalidInterfaceImplementationError(
           invalidFieldImplementation.originalResponseType +
           `" for "${interfaceName}.${fieldName}".\n`;
       }
+      if (invalidFieldImplementation.invalidContextArguments.size > 0) {
+        message +=
+          `   An ${ARGUMENT} that the router populates rather than the client must be declared "@${FROM_CONTEXT}" on` +
+          ` both the Interface field and its implementation.\n` +
```
Aenimus: I wonder whether this wording is confusing.

Should it be more about the actual definitions than the under the hood functionality? E.g., Arguments that define `fromContext` must have a corresponding context

composition/src/errors/errors.ts
Aenimus: Same question about the wording of router and client population as above

composition/src/errors/errors.ts
```diff
@@ -625,6 +636,27 @@ export function invalidRequiredInputValueError(
   return new Error(message);
 }
 
+export function contextArgumentRequiredError(
```
Aenimus, line 639: I think I would rename this as "requiredContextArgumentError" or possibly "nonOptionalContextArgumentError"

composition/src/errors/errors.ts
```diff
@@ -625,6 +636,27 @@ export function invalidRequiredInputValueError(
...
+  requiredSubgraphNames: Array<string>,
+): Error {
+  return new Error(
+    `The ${ARGUMENT} "${coords}" is invalid because:\n` +
+      ` It is declared "@${FROM_CONTEXT}" in the following subgraph` +
+      (contextSubgraphNames.length > 1 ? 's' : '') +
+      ': "' +
+      contextSubgraphNames.join(QUOTATION_JOIN) +
+      `"\n` +
+      ` However, it is required in the following subgraph` +
```
Aenimus, line 651: it is -> this argument is

composition/src/errors/errors.ts
Aenimus: ` An ${ARGUMENT} that defines ${FROM_CONTEXT} must be optional in any subgraph that references it.`

(note also the removed trailing new line)

composition/src/errors/errors.ts
```diff
@@ -585,6 +586,16 @@ export function invalidInterfaceImplementationError(
           invalidFieldImplementation.originalResponseType +
           `" for "${interfaceName}.${fieldName}".\n`;
       }
+      if (invalidFieldImplementation.invalidContextArguments.size > 0) {
+        message +=
+          `   An ${ARGUMENT} must define "@${FROM_CONTEXT}" on both the Interface field and its implementation,` +
```
Aenimus: I might slightly rephrase this to:
If "@${FROM_CONTEXT}" is defined on the implemented field of an Interface, it must also be defined on the Interface field itself.

composition/src/v1/normalization/normalization-factory.ts
```diff
@@ -1365,9 +1366,11 @@ export class NormalizationFactory {
       ? `${federatedParentTypeName}${fieldName ? `.${fieldName}` : ''}(${name}: ...)`
       : `${federatedParentTypeName}.${name}`;
     const namedTypeName = getTypeNodeNamedTypeName(node.type);
+    const directivesByName = this.extractDirectives(node, new Map<DirectiveName, ConstDirectiveNode[]>());
     const inputValueData: InputValueData = {
       configureDescriptionDataBySubgraphName: new Map<string, ConfigureDescriptionData>(),
-      directivesByName: this.extractDirectives(node, new Map<string, ConstDirectiveNode[]>()),
+      contextSubgraphNames: new Set<SubgraphName>(directivesByName.has(FROM_CONTEXT) ? [this.subgraphName] : []),
```
Aenimus: should we rename this to `fromContextSubgraphNames` just because `@context` is its own thing?

composition/src/v1/federation/federation-factory.ts
```diff
@@ -1934,6 +1948,18 @@ export class FederationFactory {
     const invalidRequiredArguments: InvalidRequiredInputValueData[] = [];
     const fieldCoords = `${fieldData.renamedParentTypeName}.${fieldData.name}`;
     for (const [argumentName, inputValueData] of fieldData.argumentDataByName) {
+      if (isInputValueDataFromContext(inputValueData)) {
```
Aenimus: What do you think about renaming this to "doesArgumentDefineFromContext"?

Input value can mean argument or it can mean Input Object fields, and I believe the location only allows arguments

composition/src/utils/types.ts
```diff
@@ -11,6 +11,8 @@ export type InvalidArgumentImplementation = {
 
 export type InvalidFieldImplementation = {
   implementedResponseType?: string;
+  interfaceContextFields: Set<ArgumentName>;
+  interfaceImplementationContextFields: Set<ArgumentName>;
```
Aenimus: nit: we might be able to remove interface from the name just because there can only be interface implementations

composition/src/errors/errors.ts
```diff
@@ -585,6 +586,24 @@ export function invalidInterfaceImplementationError(
           invalidFieldImplementation.originalResponseType +
           `" for "${interfaceName}.${fieldName}".\n`;
       }
+      if (invalidFieldImplementation.interfaceContextFields.size > 0) {
+        message +=
+          `   "@${FROM_CONTEXT}" cannot be defined on an Interface field.\n` +
+          `    The following argument` +
+          (invalidFieldImplementation.interfaceContextFields.size > 1 ? `s define` : ` defines`) +
+          ` "@${FROM_CONTEXT}": "` +
+          [...invalidFieldImplementation.interfaceContextFields].join(`", "`) +
```
Aenimus: nit: we have QUOTATION_JOIN constant

composition/src/errors/errors.ts
```diff
@@ -585,6 +586,24 @@ export function invalidInterfaceImplementationError(
...
+          `"\n`;
+      }
+      if (invalidFieldImplementation.interfaceImplementationContextFields.size > 0) {
+        message +=
+          `   "@${FROM_CONTEXT}" cannot be defined on the implementation of an Interface field.\n` +
+          `    The following argument` +
+          (invalidFieldImplementation.interfaceImplementationContextFields.size > 1 ? `s define` : ` defines`) +
+          ` "@${FROM_CONTEXT}": "` +
+          [...invalidFieldImplementation.interfaceImplementationContextFields].join(`", "`) +
```
Aenimus: same QUOTATION_JOIN nit

composition/src/errors/errors.ts
```diff
@@ -585,6 +586,24 @@ export function invalidInterfaceImplementationError(
...
+          `   "@${FROM_CONTEXT}" cannot be defined on an Interface field.\n` +
+          `    The following argument` +
+          (invalidFieldImplementation.interfaceContextFields.size > 1 ? `s define` : ` defines`) +
+          ` "@${FROM_CONTEXT}": "` +
+          [...invalidFieldImplementation.interfaceContextFields].join(`", "`) +
+          `"\n`;
+      }
+      if (invalidFieldImplementation.interfaceImplementationContextFields.size > 0) {
+        message +=
+          `   "@${FROM_CONTEXT}" cannot be defined on the implementation of an Interface field.\n` +
```
Aenimus: Should we gather and present the interface coordinates?

composition/src/v1/federation/federation-factory.ts
```diff
@@ -424,16 +428,28 @@ export class FederationFactory {
...
             hasErrors = true;
             hasNestedErrors = true;
             invalidFieldImplementation.unimplementedArguments.add(argumentName);
             continue;
           }
+          // @fromContext cannot be defined on an interface field
+          if (doesArgumentDefineFromContext(inputValueData)) {
+            hasErrors = true;
+            hasNestedErrors = true;
+            invalidFieldImplementation.interfaceContextFields.add(argumentName);
```
Aenimus: I think my comment in the error message is solved if you use the argument coords rather than the name. This is accessible at `inputValueData.federatedCoords`

composition/src/v1/federation/federation-factory.ts
```diff
@@ -424,16 +428,28 @@ export class FederationFactory {
...
+          }
+          // @fromContext cannot be defined on the implementation of an interface field
+          if (doesArgumentDefineFromContext(implementationArgumentData)) {
+            hasErrors = true;
+            hasNestedErrors = true;
+            invalidFieldImplementation.interfaceImplementationContextFields.add(argumentName);
```
Aenimus: Same here: we can provide the Interface field argument coords (`inputValueData.federatedCoords`) to explain where this is invalid

PR #3232, opened 2026-09-10
feat(router): support regex-based CORS origin matching

fiam:
Customers migrating from Apollo Router currently have to expand regex origin rules into literals or wildcard patterns. Add `cors.match_origins` and `CORS_MATCH_ORIGINS` to allow Go regular expressions alongside the existing `allow_origins` list.

Regexes use unanchored matching, as in Apollo Router. Add `^` and `$` to require a match of the entire origin. Patterns are compiled exactly as configured, using Go's default case-sensitive matching; no flags or anchors are injected. Invalid patterns fail startup with an error naming the pattern. `Validate` rebuilds the compiled patterns on `Config` using a slice allocated to the exact pattern count.

The origin is lowercased before literal, wildcard, and regex matching to normalize scheme and host casing. Ports are preserved. Write scheme and host literals in lowercase, or opt into case-insensitive regex matching with `(?i)`. Regex escapes retain their original meaning. Requests, custom origin callbacks, and `Access-Control-Allow-Origin` responses retain the original header value.

The lists are additive. Regex-only configurations must set `allow_origins: []`, since its existing default `["*"]` still permits every origin. `CORS_MATCH_ORIGINS` uses semicolons so commas in repetition counts such as `{1,3}` are preserved.

Implements [ROUTER-656](https://linear.app/wundergraph/issue/ROUTER-656/support-regex-based-cors-origin-matching-in-the-router-apollo).

## How to test

From `router/`, run the focused tests:

```sh
go test -race ./pkg/cors -run 'Test(MatchOrigin|InvalidMatchOrigins)' -count=1 -v
```

The focused tests cover origin normalization, anchoring, additive literal/wildcard/regex rules, and startup validation. Existing middleware tests cover normal and preflight responses.

## Validation

`go test -race ./pkg/cors ./pkg/config ./core ./pkg/mcpserver ./pkg/connectrpc`, `mintlify validate`, formatting, and `git diff --check` pass.

## Checklist

- [x] Changes discussed and reviewed under ROUTER-656.
- [x] Project coding standards followed.
- [x] Tests updated.
- [x] Unit-test validation instructions provided.
- [x] Documentation updated in `docs-website/router/configuration.mdx`.
- [x] Contributors Guide read.

## Open Source AI Manifesto

AI-assisted implementation, with author review.


router/pkg/cors/cors.go
```diff
@@ -102,24 +107,45 @@ func (c *Config) validateAllowedSchemas(origin string) bool {
...
+			return nil, fmt.Errorf("bad origin regex in match_origins %q: %w", pattern, err)
+		}
+		// Group alternatives and use text anchors so inline multiline flags cannot
+		// turn a full-origin match into a match of just one line. Serializing the
+		// parsed expression also prevents a trailing \Q from quoting the anchors.
+		re, err := regexp.Compile(`\A(?:` + parsed.String() + `)\z`)
+		if err != nil {
+			return nil, fmt.Errorf("bad origin regex in match_origins %q: %w", pattern, err)
 		}
+		patterns = append(patterns, re)
```
endigma: Can you include a better explanation for this pre-processing step? Or maybe factor this out and include tests showing its behaviour?

docs-website/router/configuration.mdx
```diff
@@ -950,6 +951,28 @@ pyroscope:
 
 Wildcard origins retain their existing validation and matching behavior. Patterns containing `*`, including `*.example.com`, bypass scheme validation and do not require `custom_schemas`.
 
+### Regular expression origins
+
+Use `match_origins` to allow origins using [Go regular expressions](https://pkg.go.dev/regexp/syntax). The router compiles patterns at startup and rejects invalid patterns with an error naming the pattern. Matching uses Go's linear-time regex engine. Lookarounds and backreferences are not supported.
+
+Each pattern must match the entire origin, including its scheme and optional port. Anchors such as `^` and `$` are optional. For example, `https://([a-z0-9-]+\.)*example\.com` allows `https://example.com` and `https://app.example.com`, but rejects `https://app.example.com.evil.com`. Matching is case-sensitive unless the pattern uses `(?i)`.
```
endigma: Could we normalize the origin to make this not a minor footgun?

router/core/cors_test.go
```diff
@@ -0,0 +1,101 @@
...
+	require.NoError(t, os.WriteFile(configPath, []byte(`
+version: "1"
+cors:
+  allow_origins: []
+  match_origins: ['https://([a-z0-9-]+\.)*example\.com']
+`), 0o600))
+	loaded, err := config.LoadConfig([]string{configPath})
+	require.NoError(t, err)
+	router, err := NewRouter(t.Context(), optionsFromResources(zap.NewNop(), &loaded.Config, nil)...)
+	require.NoError(t, err)
```
endigma: Do we need to load this all the way from YAML here? I think this would be clearer if this was directly configured, we can test loading this in pkg/config if needed.

endigma: This also seems mostly duplicative scope of router-tests/protocol/cors_test, maybe this should be a true E2E?

endigma (changes requested):

docs-website/router/configuration.mdx
fiam: Do you mean making it case insensitive by default?

router/pkg/cors/cors.go
fiam: It was to allow `^` and `$` to be implicit, but you made me realize that was a bad idea. It should be gone now.

router/pkg/config/config.go
```diff
@@ -203,6 +203,7 @@ type Telemetry struct {
 type CORS struct {
 	Enabled          bool          `yaml:"enabled" envDefault:"true" env:"CORS_ENABLED"`
 	AllowOrigins     []string      `yaml:"allow_origins" envDefault:"*" env:"CORS_ALLOW_ORIGINS"`
+	MatchOrigins     []string      `yaml:"match_origins" env:"CORS_MATCH_ORIGINS" envSeparator:";"`
 	CustomSchemes    []string      `yaml:"custom_schemes" env:"CORS_CUSTOM_SCHEMES"`
```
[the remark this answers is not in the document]
fiam: This is just fixing a typo in #3230, which hasn't merged yet. The old keys have never shipped, so there's nothing to migrate.

docs-website/router/configuration.mdx
endigma: Just tolowercase the origin maybe?

router/core/cors_test.go
fiam: I've slimmed these so they look closer to unit tests. Thanks for the suggestion!

endigma (approved):

PR #3238, opened 2026-09-11
fix(router): add SSE server write timeout

### LINKED ISSUE #3175: Router: add a configurable SSE server write timeout
## Component

Router

## Problem

A downstream client can stop reading an SSE subscription while leaving the connection open. The router's response write or flush can then block indefinitely.

For subscriptions backed by a shared trigger, that stalled write can prevent later events from reaching otherwise healthy subscribers. The router already bounds WebSocket writes with `engine.websocket_server_write_timeout`, but SSE delivery has no equivalent safeguard.

## Proposed behavior

Add a separately configurable timeout for downstream SSE writes:

- expose `engine.sse_server_write_timeout` and `ENGINE_SSE_SERVER_WRITE_TIMEOUT`
- keep the timeout disabled by default for backward compatibility
- apply a fresh deadline to the initial SSE response, data events, heartbeats, and completion events
- when configured, fail the affected SSE connection if the deadline cannot be applied or a write/flush exceeds it
- isolate and remove the stalled subscriber so healthy subscribers can continue receiving shared-trigger events
- preserve existing multipart behavior and SSE behavior when the timeout is disabled

This should conceptually mirror the existing WebSocket server write timeout while remaining independently configurable for SSE transports.

## Acceptance criteria

- Negative timeout values are rejected during configuration loading.
- Zero preserves the existing no-deadline behavior.
- A positive timeout bounds each downstream SSE write/flush.
- Header, data, heartbeat, and completion paths are covered by tests.
- A blocked SSE subscriber times out without preventing a healthy subscriber from receiving a subsequent event from the same shared trigger.
- Existing multipart and disabled-timeout behavior remains covered.

## Implementation

Proposed implementation: #3172

endigma:
Closes #3175.

## Summary

Add a write timeout for SSE subscription responses. The router sets a write deadline before each SSE write and flush and clears it after. If a client does not read within the timeout, the write fails and the router closes that subscription. Other subscriptions on the same trigger continue.

This matches the existing `engine.websocket_server_write_timeout`. The new option is `engine.sse_server_write_timeout` or `ENGINE_SSE_SERVER_WRITE_TIMEOUT`. The default is `10s`. Set `0s` to disable it.

A custom module that wraps the response writer without `SetWriteDeadline` or `Unwrap` keeps working. The timeout has no effect for that writer. The custom modules docs describe both methods.

## Changes

- Add `engine.sse_server_write_timeout` with a `10s` default.
- Apply the deadline to the initial SSE header flush and to every data, heartbeat, and completion write.
- Clear the deadline after each flush so an idle HTTP/2 stream does not expire.
- Multipart and plain responses do not change.
- End SSE error responses with a blank line so clients dispatch the event.

## How to test

1. In `router/`, run `go test -race ./core ./pkg/config`.
2. In `router-tests/`, run `go test -run 'TestSSESubscriptions|TestHTTPMultipartSubscriptions' ./subscriptions`.

The SSE tests cover two cases. One subscription stays idle longer than the timeout over HTTP/2 and then receives data. One subscription blocks its writes. The router removes it after the timeout, and a healthy subscription on the same trigger keeps receiving events.

## Summary by CodeRabbit

- **New Features**
  - Added configurable Server-Sent Events (SSE) write timeouts, defaulting to 10 seconds; set to `0s` to disable.
  - Blocked SSE subscribers are disconnected while healthy streams continue operating.
  - Added streaming support for custom response writers that support flushing and write deadlines.
  - Enabled multi-fetch and scheduled fetches by default.

- **Bug Fixes**
  - Improved SSE error handling and event termination when response writes fail.

- **Documentation**
  - Documented SSE timeout settings, custom writer requirements, and complexity behavior.


router-tests/subscriptions/http_subscriptions_test.go
```diff
@@ -303,14 +310,266 @@ func TestHeartbeats(t *testing.T) {
...
+			require.NotEqual(t, beforeTimeout, afterTimeout)
+		})
+	})
 }
 
-func TestNonFlusherWriterSubscriptionError(t *testing.T) {
-	t.Parallel()
+func openCountEmpSSESubscription(
+	t *testing.T,
+	ctx context.Context,
```
fiam: nit: do we need a context.Context here? could we use `t.Context()`?

router/core/subscription_response_writer_test.go
```diff
@@ -137,10 +159,62 @@ func TestGetSubscriptionResponseWriter(t *testing.T) {
 		req := httptest.NewRequest(http.MethodPost, "/graphql", nil)
 		req.Header.Set("Accept", sseMimeType)
 
-		_, _, ok := GetSubscriptionResponseWriter(resolve.NewContext(context.Background()), req, recorder, false)
+		_, _, ok := GetSubscriptionResponseWriter(resolve.NewContext(context.Background()), req, recorder, false, 0)
```
fiam: nit: maybe a good opportunity to change to t.Context()?

router/core/subscription_response_writer.go
```diff
@@ -151,44 +147,58 @@ func (f *HttpFlushWriter) Flush() (err error) {
...
 
-func GetSubscriptionResponseWriter(ctx *resolve.Context, r *http.Request, w http.ResponseWriter, apolloSubscriptionMultipartPrintBoundary bool) (*resolve.Context, resolve.SubscriptionResponseWriter, bool) {
+// writeAndFlush writes b and flushes it to the client. With a writeTimeout, the write
+// and the flush must both finish before the deadline or the connection write fails.
+func (f *HttpFlushWriter) writeAndFlush(b []byte) error {
+	if f.writeTimeout > 0 {
+		// A writer without deadline support returns http.ErrNotSupported and keeps
+		// working without the timeout. The deadline is cleared after the flush so an
+		// idle stream does not expire before its next write.
+		_ = f.rc.SetWriteDeadline(time.Now().Add(f.writeTimeout))
```
fiam, line 168: question: should we handle the error here? what if we get an error that's not `http.ErrNotSupported`?

router-tests/subscriptions/http_subscriptions_test.go
```diff
@@ -303,14 +310,266 @@ func TestHeartbeats(t *testing.T) {
...
+
+func (w *deadlineBlockingResponseWriter) Write(data []byte) (int, error) {
+	if !w.state.armed.CompareAndSwap(true, false) {
+		return w.ResponseWriter.Write(data)
+	}
+
+	close(w.state.writeStarted)
+	defer close(w.state.writeDone)
+
+	wait := time.Until(time.Unix(0, w.deadlineNanos.Load()))
```
fiam: question: is this correct? with `deadlineNanos = 0` we set the deadline to the unix zero time, not to the zero `time.Time{}`.

router/core/subscription_response_writer.go
endigma: SetWriteDeadline is infallible on HTTP2 and HTTP1 only returns err when connection is closed in which case the `Write` will have the same error and return it

router-tests/subscriptions/http_subscriptions_test.go
endigma: It is fine, added a comment

router/core/subscription_response_writer_test.go
endigma: Sure

router-tests/subscriptions/http_subscriptions_test.go
endigma: Done

router/core/subscription_response_writer.go
```diff
@@ -197,7 +207,7 @@ func GetSubscriptionResponseWriter(ctx *resolve.Context, r *http.Request, w http
 		ctx.ExecutionOptions.SendHeartbeat = true
 		// Flush the response head immediately so the client establishes the connection
 		// before the first message, instead of blocking until one is streamed.
-		flusher.Flush()
+		_ = flushWriter.writeAndFlush(nil)
```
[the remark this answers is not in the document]
endigma: Behaviour matches pre-PR state, if the first flush hits deadline the connection is dead, heartbeat or write then kills it via normal path. Checking here would write an error into a dead connection.

router/pkg/config/config.schema.json
```diff
@@ -4281,6 +4281,12 @@
           "default": "30s",
           "description": "The maximum time a subscription fetch can take before it is considered timed out. The period is specified as a string with a number and a unit, e.g. 10ms, 1s, 1m, 1h. The supported units are 'ms', 's', 'm', 'h'."
         },
+        "sse_server_write_timeout": {
+          "type": "string",
+          "format": "go-duration",
+          "default": "10s",
+          "description": "The maximum time for one SSE write and flush to a client. If a write takes longer, the router closes that subscription so it does not block other subscriptions on the same trigger. Set 0s to disable the timeout."
```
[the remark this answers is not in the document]
endigma: Negative value is treated as disabled

router/core/subscription_response_writer.go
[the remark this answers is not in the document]
endigma: What are you suggesting?

dkorittki (approved):

PR #3246, opened 2026-09-14
ci: tolerate missing pull-request-info artifact in query planner CI

ysmolski:
A clean query-planner run currently turns the check red.

`get_pr_url` downloads the `pull-request-info` artifact from the celestial run, but celestial only uploads that artifact when the query plans actually changed and it opened a PR — both the save and upload steps are gated on `steps.cpr.outputs.pull-request-number != `. When the plans are unchanged there is no PR, no artifact, and the `github-script` step throws:

```
Error: Could not find pull-request-info artifact
```

`build_test` had already succeeded at that point, so the query plans were fine — only the URL lookup failed.

Seen on [#3243](https://github.com/wundergraph/cosmo/pull/3243) ([run 34864223756](https://github.com/wundergraph/cosmo/actions/runs/34864223756), celestial [run 34864562463](https://github.com/wundergraph/cosmo-celestial/actions/runs/34864562463) passed).

## Change

- Treat a missing artifact as a skipped lookup rather than a failure, leaving `pullrequest_url` empty. `comment-on-success` and `comment-on-failure` already branch on the empty case, so a clean run now passes silently and a failing run still comments.
- Skip the lookup entirely when `build_test` produced no `workflow_id`, instead of calling the API with an empty `run_id`.

## Verification

Workflow YAML parses and the `github-script` body passes `node --check`. The behaviour change is confined to the artifact-missing path; when the artifact exists the steps run exactly as before.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved query-planner CI reliability when workflow runs or pull-request information are unavailable.
  * Prevented failures during artifact processing when no matching artifact exists.
  * Reports an empty pull request URL when plans remain unchanged.


pepol (approved):

ysmolski:
I will follow-up this with celestial always reporting the status of run explicitely.

PR #3247, opened 2026-09-15
fix(router/mcpserver): mirror omitToolNamePrefix in Annotations.Title

### LINKED ISSUE #3226: MCP tool Annotations.Title ignores omitToolNamePrefix, unlike Name
### Component(s)

router

### Component version

`v0.346.0` (also reproduces on current `main`, `45e6fd9d`)

### wgc version

N/A — this report doesn't involve `wgc`.

### controlplane version

N/A — reproduced in isolation via the router's Go test suite, no controlplane involved (see Runnable Test below).

### router version

`v0.346.0` (and current `main`, `45e6fd9d`)

### Description

`WithOmitToolNamePrefix` (added in #2441) strips the `execute_operation_` prefix from an MCP tool's `Name`, but the corresponding `Annotations.Title` is built unconditionally as `fmt.Sprintf("Execute operation %s", op.Name)` regardless of the option. With the prefix omitted, a client ends up with a wire `Name` of `list_employees` next to a `Title` of `"Execute operation ListEmployees"` — the two fields disagree about whether the prefix is present, with nothing in the config explaining why.

### Steps to Reproduce

Register an operation `ListEmployees`, start the router's MCP server with `WithOmitToolNamePrefix(true)`, and call `tools/list`.

### Expected Result

`Title` tracks the same flag `Name` does — e.g. `ListEmployees` (no `"Execute operation "` prefix), matching `Name`'s `list_employees`.

### Actual Result

| `omitToolNamePrefix` | `Name` | `Annotations.Title` |
| --- | --- | --- |
| `false` (default) | `execute_operation_list_employees` | `Execute operation ListEmployees` |
| `true` | `list_employees` | `Execute operation ListEmployees` ← unchanged |

`Title` stays fixed regardless of the option.

## Root Cause

`router/pkg/mcpserver/server.go`, in the operation-registration loop:

https://github.com/wundergraph/cosmo/blob/45e6fd9dcaa8f887d499984caac7dc654c653224/router/pkg/mcpserver/server.go#L813-L860

```go
toolName := operationToolName
if !s.omitToolNamePrefix {
    toolName = fmt.Sprintf("execute_operation_%s", operationToolName)
} else if slices.Contains(s.registeredTools, operationToolName) || slices.Contains(reservedToolNames, operationToolName) {
    ...
}
...
tool := &mcp.Tool{
    Name:        toolName,
    ...
    Annotations: &mcp.ToolAnnotations{
        ...
        Title: fmt.Sprintf("Execute operation %s", op.Name), // always "Execute operation ...", regardless of s.omitToolNamePrefix
        ...
    },
}
```

`toolName` branches on `s.omitToolNamePrefix`; the `Title` expression a few lines below does not consult it at all. The option's own doc comment says it "removes the `execute_operation_` prefix from MCP tool names" with no carve-out for the display title:

https://github.com/wundergraph/cosmo/blob/45e6fd9dcaa8f887d499984caac7dc654c653224/router/pkg/mcpserver/server.go#L85-L86

This is not new in this router version — it dates back to when `omitToolNamePrefix`/`OmitToolNamePrefix` was introduced in #2441, which added the flag and branched `Name` on it but never touched `Title`.

## Suggested Fix

Mirror the same branch used for `toolName`:

```diff
+		toolTitle := fmt.Sprintf("Execute operation %s", op.Name)
+		if s.omitToolNamePrefix {
+			toolTitle = op.Name
+		}
+
 		tool := &mcp.Tool{
 			Name:        toolName,
 			...
 			Annotations: &mcp.ToolAnnotations{
 				...
-				Title: fmt.Sprintf("Execute operation %s", op.Name),
+				Title: toolTitle,
 				...
 			},
 		}
```

Deliberately the bare, non-humanized form (`op.Name`, e.g. `ListEmployees`) rather than a spaced/Title-Case rewrite — keeps the change minimal and non-breaking for the default (`omitToolNamePrefix: false`) path, where `Title` is untouched.

## Runnable Test

Exercises both flag states through a real in-memory MCP client/server round trip (reusing the existing `newTestSession` harness in `router/pkg/mcpserver/discover_test.go`), asserting `Annotations.Title` directly rather than just `Name`:

<details>
<summary><code>tool_title_test.go</code></summary>

```go
package mcpserver

import (
	"testing"

	"github.com/modelcontextprotocol/go-sdk/mcp"
	"github.com/stretchr/testify/assert"
	"github.com/stretchr/testify/require"
)

// Prior to the fix, the Title annotation always read "Execute operation %s"
// regardless of WithOmitToolNamePrefix -- the one field that setting never
// touched, even though its own doc comment says it "removes the
// execute_operation_ prefix from MCP tool names" with no carve-out for the
// display title. A client showing Title next to Name would read
// "Execute operation ListEmployees" over a wire name of "list_employees",
// with nothing in the config explaining why one lost the prefix and the
// other didn't.
func TestToolTitle_MirrorsNamePrefixState(t *testing.T) {
	t.Run("prefixed (default): title keeps the descriptive form", func(t *testing.T) {
		cs := newTestSession(t, WithOmitToolNamePrefix(false))

		result, err := cs.ListTools(t.Context(), nil)
		require.NoError(t, err)

		tool := findTool(t, result.Tools, "execute_operation_list_employees")
		require.NotNil(t, tool.Annotations)
		assert.Equal(t, "Execute operation ListEmployees", tool.Annotations.Title)
	})

	t.Run("omitted: title drops the same words Name drops", func(t *testing.T) {
		cs := newTestSession(t, WithOmitToolNamePrefix(true))

		result, err := cs.ListTools(t.Context(), nil)
		require.NoError(t, err)

		tool := findTool(t, result.Tools, "list_employees")
		require.NotNil(t, tool.Annotations)
		assert.Equal(t, "ListEmployees", tool.Annotations.Title)
	})
}

func findTool(t *testing.T, tools []*mcp.Tool, name string) *mcp.Tool {
	t.Helper()
	for _, tool := range tools {
		if tool.Name == name {
			return tool
		}
	}
	t.Fatalf("tool %q not found in %d returned tools", name, len(tools))
	return nil
}
```

</details>

Fails on current `main` (both subtests observe `"Execute operation ListEmployees"`); passes with the one-line fix above.

## Additional context

We hit this downstream while auditing MCP tool naming for a per-tool RBAC rollout, and already have this exact patch running in our fork (junipersquare/cosmo#64). Happy to open it here as a PR against `main` if that's welcome — it's the minimal one-line-branch version above, not a broader humanization of the title.

Tyagiquamar:
### What does this PR do?

When `WithOmitToolNamePrefix(true)` is set, the `execute_operation_` prefix is stripped from the MCP tool `Name`, but the tool's `Annotations.Title` was still unconditionally formatted as `fmt.Sprintf("Execute operation %s", op.Name)`.

This change mirrors the `omitToolNamePrefix` branch for `Annotations.Title`, setting `Title` to `op.Name` when prefix omission is enabled while preserving the default `"Execute operation %s"` format when prefix omission is disabled.

Fixes #3226.

### Testing

Added `TestToolTitle_MirrorsNamePrefixState` in `router/pkg/mcpserver/discover_test.go` verifying that:
1. Prefixed state (default): `Annotations.Title` retains `"Execute operation ListEmployees"`.
2. Omitted state: `Annotations.Title` drops the prefix and is `"ListEmployees"`.

## Summary by CodeRabbit

- **Improvements**
  - MCP tool titles now accurately reflect the configured naming style.
  - When name prefixes are enabled, titles retain the descriptive “Execute operation” wording.
  - When prefixes are omitted, titles display the operation name directly for a cleaner, more concise presentation in MCP clients.


router/pkg/mcpserver/server.go
```diff
@@ -849,6 +849,11 @@ func (s *GraphQLSchemaServer) registerTools() error {
 			}
 		}
 
+		toolTitle := fmt.Sprintf("Execute operation %s", op.Name)
```
fiam, line 852: issue: this should not run unconditionally, only when we'll use it. e.g. something like

```go
var tootTitle string
if s.omitToolNamePrefix {
    toolTitle = op.Name
} else {
    toolTitle = fmt.Sprintf("Execute operation %s", op.Name)
}
```

Could you please update that? LGTM otherwise. Thanks!

endigma:
Maybe this should be split into `omitToolTitlePrefix` or something so these two are not tied together?

PR #3248, opened 2026-09-15
refactor: read url state with nuqs and useParams

gausie:
`router.query.x as string` appeared 169 times across 59 files. `router.query.x` is `string | string[] | undefined`, so the cast quietly asserts away both a repeated param and an absent one. The Pages Router also merges dynamic route segments *into* `router.query`, so the cast can't tell `?slug=x` from a `[slug]` segment either.

Both halves have a platform answer, so this uses them rather than inventing hooks:

```ts
const [search, setSearch] = useQueryState('search', parseAsString.withDefault(''));  // nuqs
const { slug, checkId } = useParams<{ slug: string; checkId: string }>();            // next/navigation
```

Fixed some issues that became apparent along the way. And there are still some casts in really complicated areas like useSyncTableWithQuery.

## Summary by CodeRabbit

* **Improvements**
  * Improved URL-based navigation across analytics, caching, checks, schemas, feature flags, subgraphs, compositions, and workspace pages.
  * Standardized pagination controls and URL synchronization across lists, with configurable page-size defaults.
  * Search, filtering, tabs, selections, and detail panels now stay synchronized with URL state.
  * Improved handling of optional route and query values for more reliable requests and links.
  * Members lists now display 20 items per page by default.


studio/src/hooks/use-operations-filters.tsx
```diff
@@ -1,85 +1,59 @@
...
-import { useRouter } from 'next/router';
+import { parseAsArrayOf, parseAsBoolean, parseAsString, parseAsStringLiteral, useQueryStates } from 'nuqs';
 import { OperationsFetchBasedOn } from '@wundergraph/cosmo-connect/dist/platform/v1/platform_pb';
-import { useApplyParams } from '@/components/analytics/use-apply-params';
+
+const fetchBasedOnValues = ['requests', 'latency', 'errors'] as const;
+
+type FetchBasedOn = (typeof fetchBasedOnValues)[number];
+
+const fetchBasedOnEnum: Record<FetchBasedOn, OperationsFetchBasedOn> = {
```
comatory: I'd use this instead

```
  const fetchBasedOnEnum = {
    requests: OperationsFetchBasedOn.REQUESTS,
    latency: OperationsFetchBasedOn.LATENCY,
    errors: OperationsFetchBasedOn.ERRORS,
  } as const satisfies Record<FetchBasedOn, OperationsFetchBasedOn>;
```

It catches missing or additional (or invalid) entries.

studio/src/hooks/use-operations-filters.tsx
```diff
@@ -1,85 +1,59 @@
...
+  requests: OperationsFetchBasedOn.REQUESTS,
+  latency: OperationsFetchBasedOn.LATENCY,
+  errors: OperationsFetchBasedOn.ERRORS,
+};
+
+const enumToString = (enumValue: OperationsFetchBasedOn): FetchBasedOn =>
```
comatory: The previous implementation was more efficient and I'd argue efficiency is important here as re-renders can cause frequent executions.

I'd rather if we could do a mapping or switch statement instead of `.find`

studio/src/hooks/use-operations-filters.tsx
```diff
@@ -1,85 +1,59 @@
...
+  includeOperationsWithDeprecatedFieldsOnly: parseAsBoolean.withDefault(false),
+  operationHash: parseAsString,
+  operationName: parseAsString,
+  clientNames: parseAsArrayOf(parseAsString).withDefault([]),
+  searchQuery: parseAsString.withDefault(''),
+  fetchBasedOn: parseAsStringLiteral(fetchBasedOnValues).withDefault('requests'),
+  sortDirection: parseAsString.withDefault('desc'),
+};
 
 export const useOperationsFilters = () => {
```
comatory, line 35: Could we improve this with JS doc comment? Just specifying what is the purpose of the hook (3 lines at most)

studio/src/components/changelog/changelog.tsx
```diff
@@ -1,17 +1,16 @@
...
 import { PiCubeFocus } from 'react-icons/pi';
 import { Changes, getDiffCount } from './changes';
 import { useWorkspace } from '@/hooks/use-workspace';
 import { useCurrentOrganization } from '@/hooks/use-current-organization';
 import { buildUrl } from '@/lib/build-url';
 
 export const Changelog = ({ entries }: { entries: FederatedGraphChangelogOutput[] }) => {
-  const router = useRouter();
-  const slug = router.query.slug as string;
+  const { slug } = useParams<{ slug: string }>();
```
comatory, line 13: Have you tested whether destructuring is safe in case the param is completely missing?

studio/src/components/checks/changes-table.tsx
```diff
@@ -93,29 +99,28 @@ const Row = ({
   operationHash?: string;
   operationName?: string;
   subgraphName?: string;
+  limit: number;
+  offset: number;
   openUsage: (changeType: string, path?: string) => void;
 }) => {
-  const router = useRouter();
+  const { checkId, slug } = useParams<{ checkId: string; slug: string }>();
```
comatory: There doesn't seem to be a context where `Row` would be used independently of its parent component. I'd extract the `checkId` + `slug` params in `ChangesTable` and pass it as props here.

studio/src/components/checks/changes-table.tsx
```diff
@@ -237,7 +242,7 @@ const Row = ({
                           query: {
                             organizationSlug,
                             namespace,
-                            slug: router.query.slug,
+                            slug: slug,
```
comatory:
```suggestion
slug,
```

studio/src/components/checks/override.tsx
```diff
@@ -168,26 +168,28 @@ const Override = ({
...
   const { toast } = useToast();
-  const applyParams = useApplyParams();
 
   const { data, error, isLoading, refetch } = useQuery(
     getOperationOverrides,
     {
       graphName: graphContext?.graph?.name,
       namespace: graphContext?.graph?.namespace,
-      operationHash,
+      operationHash: operationHash,
```
comatory:
```suggestion
operationHash,
```

studio/src/components/checks/override.tsx
```diff
@@ -281,11 +283,11 @@ export const ConfigureOverride = () => {
                     ? removeIgnoreAll({
                         graphName: graphContext?.graph?.name,
                         namespace: graphContext?.graph?.namespace,
-                        operationHash,
+                        operationHash: operationHash,
```
comatory:
```suggestion
operationHash,
```

studio/src/components/checks/override.tsx
```diff
@@ -281,11 +283,11 @@ export const ConfigureOverride = () => {
...
                       })
                     : createIgnoreAll({
-                        operationHash,
-                        operationName,
+                        operationHash: operationHash,
```
comatory:
```suggestion
operationHash,
```

studio/src/components/checks/override.tsx
```diff
@@ -281,11 +283,11 @@ export const ConfigureOverride = () => {
...
+                        operationName: operationName,
```
comatory:
```suggestion
operationName,
```

studio/src/components/ui/pagination.tsx
```diff
@@ -9,23 +9,18 @@ export const Pagination = ({
   noOfPages,
   pageNumber,
   options,
```
comatory, line 11: Can we move `10` to a constant and re-use it everywhere, including the pagination params hook?

There's also another "default" pagination which has value of `20`, it's in members page and analytics. I'd move it to a constant as well if we're refactoring anyway.

studio/src/components/ui/pagination.tsx
```diff
@@ -9,23 +9,18 @@ export const Pagination = ({
...
+  defaultPageSize = 10,
 }: {
   limit: number;
   noOfPages: number;
   pageNumber: number;
   options?: number[];
```
comatory, line 17: Confusing comment? I don't get it.

studio/src/components/ui/pagination.tsx
```diff
@@ -38,7 +33,7 @@ export const Pagination = ({
           value={`${limit}`}
           onValueChange={(value) => {
             // Reset page when size changes because the number of pages may not be the same
```
comatory, line 36: Is this conversion needed? I figured the hook would do it for us.

studio/src/pages/[organizationSlug]/[namespace]/graph/[slug]/checks/[checkId]/index.tsx
```diff
@@ -346,9 +346,8 @@ const CheckDetails = ({ data, refetch }: { data: GetCheckSummaryResponse; refetc
   const {
     namespace: { name: namespace },
   } = useWorkspace();
-  const slug = router.query.slug as string;
-  const id = router.query.checkId as string;
-  const tab = router.query.tab as string;
+  const { slug, checkId: id } = useParams<{ slug: string; checkId: string }>();
+  const [tab] = useQueryState('tab');
```
Comment (comatory) [inline studio/src/pages/[organizationSlug]/[namespace]/graph/[slug]/checks/[checkId]/index.tsx:350]: L349-350 seems to contain repeated code. I wonder if it'd make sense to extract it into mini hooks to share the functionality.

studio/src/pages/[organizationSlug]/members.tsx
```diff
@@ -54,6 +54,8 @@ import { formatDateTime } from '@/lib/format-date';
 import { MultiGroupSelect } from '@/components/multi-group-select';
 import { buildUrl } from '@/lib/build-url';
 
+const DEFAULT_PAGE_SIZE = 20;
```
Comment (comatory) [inline studio/src/pages/[organizationSlug]/members.tsx:57]: If we can re-use this constant, I'd place it in another file (constants or something similar).

studio/src/pages/_app.tsx
```diff
@@ -108,22 +109,24 @@ function MyApp({ Component, pageProps }: AppPropsWithLayout) {
...
-                    <Toaster />
-                    {getLayout(<Component {...pageProps} />)}
-                  </TooltipProvider>
-                </AppProvider>
-              </OnboardingProvider>
-            </PostHogFeatureFlagProvider>
-          </QueryClientProvider>
-        </ThemeProvider>
-      </PostHogProvider>
+      <NuqsAdapter>
```
comatory: Does it need to be top-level? (I haven't read the documentation so correct me if I'm wrong).
I'm just thinking if it fails rendering for some reason, we'd want PostHog to catch it.

comatory (changes requested):
Few suggestions but really good work on this one.

studio/src/components/changelog/changelog.tsx
gausie: The param is always present because we use getInitialProps in _app.tsx

studio/src/pages/[organizationSlug]/members.tsx
Comment (gausie) [reply studio/src/pages/[organizationSlug]/members.tsx:57]: It's only for this file. everything else uses 10 as a default page size actually.

studio/src/pages/_app.tsx
gausie: Oooh good bet.

studio/src/pages/[organizationSlug]/members.tsx
Comment (comatory) [reply studio/src/pages/[organizationSlug]/members.tsx:57]: Oh ok I thought analytics used 20 somewhere as well.

studio/src/components/ui/pagination.tsx
gausie: Yeah it's required by the types. Radix hands onValueChange a string and parseAsInteger's setter takes a number.

studio/src/pages/[organizationSlug]/[namespace]/graph/[slug]/checks/[checkId]/index.tsx
Comment (gausie) [reply studio/src/pages/[organizationSlug]/[namespace]/graph/[slug]/checks/[checkId]/index.tsx:350]: I made a useCheckParams hook, but tab is still plain since we don't really have a repeated pattern for it. I could follow up with some dressing around Tabs though.

comatory (approved):
Thanks for addressing the feedback. Looks good :+1

PR #3258, opened 2026-09-17
feat(router): cache multi-fetches and cap Cache-Control at cached TTL

ysmolski:
graphql-go-tools v2.21.0 caches merged multi-entity fetches per alias.
A merged fetch can now be served partly from the cache: some aliases come
back with a TTL left, the others are fetched and arrive with a fresh
Cache-Control.

The router builds the client's Cache-Control from every subgraph response,
taking the most restrictive one. Before this change a fetch was either a
hit, with no subgraph header, so the hook wrote the TTL into the empty
header slot, or a miss with the subgraph's header. A partly cached merged
fetch has both, and the TTL was dropped: with 60 seconds left on the
cached half and a subgraph max-age of 600, the client saw 600.

The hook now combines the subgraph's Cache-Control with the TTL into the
most restrictive policy before the response header rules run. It does so
only when a most restrictive cache control rule applies to the fetch, so a
named propagate rule alone no longer forwards a Cache-Control the subgraph
did not send.

The integration test that pinned merged fetches as uncacheable is replaced
by tests for the per-alias cache, the cold-alias request, and the capped
Cache-Control. The docs describe the new limitations of merged fetches.

For https://github.com/wundergraph/graphql-go-tools/pull/1656

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved response caching for merged entity requests by caching aliases independently.
  - Cold aliases are refetched without discarding available warm data.
  - Failed cold fetches no longer prevent dependent fetches from using warm results.
  - Propagated cache lifetimes now respect the most restrictive applicable TTL and preserve existing cache-control directives.
  - Invalid cache-control directives are ignored without disrupting cached responses.

- **Documentation**
  - Clarified that merged entity fetches do not declare cache tags.


router/core/engine_loader_hooks.go
```diff
@@ -136,8 +138,35 @@ func ttlToCacheControl(ttl time.Duration) string {
 	return cacheControl.ToHeaderString()
 }
 
-func (f *engineLoaderHooks) OnFinished(ctx context.Context, ds resolve.DataSourceInfo, responseInfo *resolve.ResponseInfo) {
+// applyResponseCacheLifetime puts the TTL left on the cached entries a fetch was
+// served from into its Cache-Control, where the most restrictive algorithm reads it.
+func applyResponseCacheLifetime(headers http.Header, info *resolve.ResponseInfo) {
```
SkArchon, line 143: I was wondering if there is a better way to do this since we already have logic for handling cache control headers, with the help of claude it came up with this https://github.com/wundergraph/cosmo/commit/e4d85e68d926c7acb7ddac818862395011d327ce

You can do this change in a different pr if you think it makes sense, its fine

SkArchon (approved):
approved with 1 comment

router/core/engine_loader_hooks.go
ysmolski: Thanks for noticing it. I agree that this approach is not the best, but I did not want to refactor code all of the place since I do not (own / know) that code well. There is a need for refactoring the code a bit, but let's unblock master for now and refactor later.

PR #3259, opened 2026-09-17
docs(rfc): at-least-once for Cosmo Streams using cursors v1

dkorittki:
## Summary by CodeRabbit

- **Documentation**
  - Clarified that cursors are cumulative and commits advance only through the longest contiguous processed prefix.
  - Documented safe handling of out-of-order acknowledgments and limits on unacknowledged messages, including errors when the limit is reached.
  - Added guidance that handlers must be idempotent because resumed connections may replay processed messages.
  - Clarified cursor verification retention and signing-key behavior.
  - Refined transport, configuration, middleware, and SSE handshake guidance.

## Checklist

## Open Source AI Manifesto

### DIFF
--- a/rfc/at-least-once-v1.md
+++ b/rfc/at-least-once-v1.md
@@ -9,12 +9,9 @@ Date: 2026.09.10
 ### The Cursor
 
 The cursor is a core aspect of the Cursor-Resume strategy and hence it's defined early in this RFC.
-
 They are created alongside GraphQL response messages and represent the subscriptions position
-of that message in the broker.
-
-The Cursor is designed to provide all data necessary for the router to know where to connect and
-where to resume without any outside context.
+of that message in the broker. The Cursor is designed to provide all data necessary for the router
+to know where to connect and where to resume without any outside context.
 
 A cursor is a position, not a permission. Presenting a valid cursor never confers authority. The
 resumed operation is planned and authorized exactly as a fresh subscription would be.
@@ -247,3 +244,53 @@ Within the payload the same applies one level down. `Cursor.Position` is an inte
 router needs a concrete type before it can unmarshal the position bytes, and it takes that from
 `Cursor.ProviderType`. `ProviderType` therefore has to be decoded before `Position`, which means
 it has to be written before `Position` on the wire.
+
+### Transport
+
+This section specifies how a cursor is transmitted and received between client and router.
+
+#### Foundations
+
+- Clients express their wish to receive a cursor. If that wish is not expressed the router falls back to the usual at-most-once delivery
+- Cursors are managed per subscription, not per connection (you can have multiple subscriptions on  one connection)
+- A Cursor is sent in the wire format described here [here](#wire-format)
+
+#### Official Spec Changes
+
+Its a territory where we aim to enhance official specs like
+[graphql-transport-ws](https://github.com/enisdenjo/graphql-ws),
+[graphql-sse](https://github.com/enisdenjo/graphql-sse) and maybe
+[GraphQL GAP](https://graphql.org/blog/2026-06-01-announcing-gaps/).
+We would like to include things like capability negotations ("dear server I support at-least-once
+methods XYZ, what do you support?") and ACK responses ("dear server I got your message #132").  
+However we decided not to include this aspect in At-Least-Once v1. These spec changes need a
+public, community-driven discussion with potentially huge changes to the initial design idea.
+Its also not clear how long it takes until the spec changes are accepted. We want to keep the
+project independent and be able to deliver it in a reasonably short period of time.  
+However, we want to commit to these spec changes as we think they are valuable to the GraphQL
+ecosystem. We will work on spec changes very soon and integrate them in
+At-Least-Once v2. The big endgoal is to have a public spec for at-least-once and we support it.
+
+For the time being we make use of the extensibility of current transport specs.
+
+#### Client <-> Router handshake
+
+This section specifies how a client can signal the need for cursors to the router.
+
+The handshake has to be initiated on subscription request.
+Both relevant subprotocols, graphql-transport-ws and graphql-sse, support metadata on subscription requests.
+
+- graphql-transport-ws: Specify a free-form map on `payload.extensions` on [suscribe requests](https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md#subscribe)
+- graphql-sse: Specify a free-form map on `payload.extensions` (indirectly via [GraphQL-Over-HTTP subscription requests](https://github.com/graphql/graphql-over-http/blob/main/spec/GraphQLOverHTTP.md#graphql-over-http-request))
+
+On both subprotocols the client can add the key `at-least-once-capabilities`. It accepts a
+comma-seperated list of at-least-once capabilities the client supports and wishes to use.
+The only available value at this time is `cursor`. The value has to be a string.
+
+##### Negotation failure
+
+TBD
+
+#### Transmitting the cursor to clients
+
+
--- a/rfc/at-least-once-v1.md
+++ b/rfc/at-least-once-v1.md
@@ -1,0 +1,246 @@
+# At-Least-Once delivery guarantees for Cosmo Streams v1
+
+Status: Draft  
+Author: Dominik Korittki  
+Date: 2026.09.10
+
+## Implementation
+
+### The Cursor
+
+The cursor is a core aspect of the Cursor-Resume strategy and hence it's defined early in this RFC.
+
+They are created alongside GraphQL response messages and represent the subscriptions position
+of that message in the broker.
+
+The Cursor is designed to provide all data necessary for the router to know where to connect and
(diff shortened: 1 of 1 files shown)


rfc/at-least-once-v1.md
```diff
@@ -0,0 +1,623 @@
...
+Order of messages is only preserved inside a partition
+but not inside a topic. So when the router reads messages from a topic it reads randomly from its
+partitions, or from whatever partition receives a message first.
+
+To describe where the router stands on a Kafka topic when it delivers a message it needs
+to remember all positions of already read messages of all partitions of a topic.
+
+Now we need to remember that on a GraphQL schema you can define multiple topics per EDFS subscription.
+So all in all for Kafka we need to record the message offset for all partitions of all topics of
+a subscription.
```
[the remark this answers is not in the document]
dkorittki: Added a commit to expand on this. Basically the client (and thus probably our middleware) has to be intelligent when doing acks. I.e. do not ack message 3 when message 2 is still in progress. Processing message 2 has finished but so has message 3? Ack on message 3.

rfc/at-least-once-v1.md
```diff
@@ -0,0 +1,623 @@
...
+For cursors to work the adapters needs to be seekable, i.e. start from any given position but
+not every broker can do this. Plain NATS and Redis PubSub don't provide seekable queues. NATS
+Jetstream and Redis Streams however do. To reflect this we invent a new interface for seekable
+adapters, which extends normal adapters:
+
+```go
+type SeekableAdapter interface {
+  Adapter
+  SubscribeFrom(ctx context.Context, cfg SubscriptionEventConfiguration, position CursorPosition, updater SubscriptionEventUpdater) error
+}
```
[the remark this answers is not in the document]
dkorittki: It's an implementation detail, which can be left out of the RFC.

rfc/at-least-once-v1.md
```diff
@@ -0,0 +1,623 @@
...
+
+A cursor that fails any of these steps fails the subscription with a distinct error code. We
+deliberately do not fall back to a fresh subscription: that would silently hide both attacks and
+misconfigured keys behind a subscription that merely looks like it works.
+
+##### Keys
+
+`KeyID` selects the key from a small keyring the router holds. The router signs with the newest key
+and accepts any key in the ring that has not expired yet, which is what makes rotation possible
+without breaking cursors that are already in flight.
```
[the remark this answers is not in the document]
dkorittki: Adressed in latest commit

rfc/at-least-once-v1.md
```diff
@@ -0,0 +1,704 @@
...
+replayed as a valid tag somewhere else that happens to use the same secret.
+
+##### Request data
+
+The following goes into the `requestData` part of the HMAC input:
+
+- `SubscriptionEventConfiguration.ProviderType()`
+- `SubscriptionEventConfiguration.ProviderID()`
+- Root field name
+- alphanumerically sorted list of topics of the subscription (or channels / subjects)
```
alepane21, line 195: What if there are parameters used for filtering in the subscription?

rfc/at-least-once-v1.md
```diff
@@ -0,0 +1,704 @@
...
+In case the router won't support any guarantees the client listed it will terminate the subscription
+with an error. For both subprotocols two fields can be defined which indicate the nature of the
+error:
+
+- `code`: An error code indicating the reason why the server refused
+- `supported`: A string array with supported guarantees
+
+The only allowed value for `code` to return in this case is `DELIVERY_GUARANTEE_UNSUPPORTED`.
+
+Depending on the subprotocol they have to be delivered in different ways.
```
alepane21, line 424: I think there is another error that can happen. The router support the guarantee, but the cursor can not be satisfied anymore. As an example: the broker delete messages older than 5 min and the cursor is older than that. I think the right approach here is to send the error with a specific code (DELIVERY_GUARANTEE_ UNSATISFIABLE?) and keep the subscription open and let the client decide if they don't accept this degradation.

rfc/at-least-once-v1.md
```diff
@@ -0,0 +1,704 @@
...
+If client and router both agree to use cursors the router will confirm this to the client
+
+- graphql-sse: via `x-cosmo-at-least-once-capabilities: cursor` response header, available before
+  the first event
+- graphql-transport-ws: `extensions.at-least-once-capabilities` on the first `next` message, since
+  the subprotocol has no per-subscription ack frame in v1
+
+##### Negotiation failure
+
+In case the router won't support any guarantees the client listed it will terminate the subscription
```
alepane21, line 415: Not sure if terminating is the right approach. Why not allowing the client to decide if he want to close the subscription or not based on the needs? Or maybe this "force closing the sub" could be set to true by default.

rfc/at-least-once-v1.md
```diff
@@ -0,0 +1,704 @@
...
+
+#### Resuming clients
+
+When a client wants to receive messages where it left off it has to provide the cursor upon
+reconnect. Specifically during handshake the client adds the field `cursor` to `payload.extensions`,
+similar to how it adds `delivery-guarantee`, see [handshake](#client---router-handshake). The value
+of this field contains the last processed cursor as chosen by the client in
+[wire format](#wire-format).
+
+The `cursor` field is independent of `delivery-guarantee`. A client may choose to resume from a
```
alepane21, line 491: There could be a security problem here: I could be allowed to see events only from today. If I can have access to older messages, we are allowing clients to access data that they are not allowed to see.

rfc/at-least-once-v1.md
```diff
@@ -0,0 +1,704 @@
...
+This means clients will get messages as fast as the broker supports. This comes with some
+challenges.
+
+- Subgraph load: Every incoming event from the broker need to be resolved into a GraphQL response.
+- Client TCP blocks: If a client reads slower than the router wants to write it reaches a 10s
+  default timeout and the WS/SSE connection is dropped
+
+Since resuming clients live off their own trigger a regression for other clients is not expected.
+
+To help users manage this situation a router config parameter will be introduced to allow message
```
alepane21, line 506: This config is only about resumes or a general config?

rfc/at-least-once-v1.md
```diff
@@ -0,0 +1,704 @@
...
+define event ids, and it would introduce a second cursor channel that works on only one transport.
+
+
+# Todos
+- [x] Goals / Non-Goals
+- [x] Cursors
+- [x] Transport
+- [x] Adapters
+- [x] Config
+- [x] Client Middleware
```
alepane21, line 704: There is something missing: the custom modules. I think we should expose the cursor data to custom modules, to allow customers to customize the behaviour also checking on this. Again: they could want to allow cursor to only go back to when the security authorization has been granted, as an example.

rfc/at-least-once-v1.md
dkorittki: The config option `events.delivery-guarantees.cursor-resume.max_resume_window` prevents that. Cursors older than that are not accepted anymore.

rfc/at-least-once-v1.md
dkorittki: Currently its only meant to enforce during replay but maybe it makes sense to enable it globally. Whats your opinion?

rfc/at-least-once-v1.md
dkorittki: Good point, I rewrote the signing section to be better human readable and I also included the query + variables into it. It means that now the query + variables have to be the same when replaying. My first idea was to allow users to replay with a different query to be able to select different fields but I think its better to be strict here. I think most users want it to be strict.

rfc/at-least-once-v1.md
dkorittki: Good point, added a small new section

rfc/at-least-once-v1.md
dkorittki: Two reasons:
- The dangers are high that it goes unnoticed by the client otherwise
- graphql-ws and sse specs are very weak in that regard. You could only send next events which carry that error. But actually thats misuse. They are designed for GraphQL related errors such as "no data for field" but not transport related errors like this one

For now i'd rather let clients deal with the fallback, i.e. reconnect without cursor on their own if they wish to

rfc/at-least-once-v1.md
dkorittki: I added a new section for custom modules

rfc/at-least-once-v1.md
alepane21: I'm OK with this, if we explain this in the documentation.

rfc/at-least-once-v1.md
alepane21: I think that more than that, custom modules are going to solve the issue, that now are part of the RFC. So I would say this is solved!

rfc/at-least-once-v1.md
alepane21: I think is good to keep it at replay intially.

alepane21 (approved):

PR #3261, opened 2026-09-18
docs: add defer directive docs

devsergiy:
fixes router-681

## Summary by CodeRabbit

- **New Features**
  - Added router support for `@defer` incremental delivery using `multipart/mixed` and `incrementalSpec=v0.2`.
  - Added clearer errors for unsupported response formats and incremental specifications.
  - HTTP/2 responses now omit the HTTP/1.1-only `Transfer-Encoding: chunked` header.

- **Documentation**
  - Added comprehensive Router and Federation documentation for `@defer`, including setup, usage, execution behavior, response formats, and guidance.
  - Added `@defer` to the Federation directives reference.
  - Expanded CORS, SSE timeout, fetch scheduling, and query complexity configuration guidance.
  - Documented updated defaults for multi-fetch and scheduled fetches.

## Checklist

## Open Source AI Manifesto


docs-website/router/defer/usage.mdx
```diff
@@ -0,0 +1,450 @@
...
+  incrementalHandler: new GraphQL17Alpha9Handler(),
+});
+```
+
+The handler sends `Accept: multipart/mixed;incrementalSpec=v0.2` for operations that contain `@defer`, which the router accepts.
+Queries with `@defer` then deliver partial data to `useQuery` and `watchQuery` as parts arrive, with `dataState` moving from `streaming` to `complete`.
+
+Apollo Client rewrites the operation before sending it.
+It adds `__typename` to every selection set and a generated `label` such as `ac_0` to every `@defer` fragment.
+Both are handled by the router, and the labels show up on the `pending` entries.
```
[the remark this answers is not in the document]
devsergiy: not applicable

docs-website/router/defer/when-to-use.mdx
```diff
@@ -0,0 +1,88 @@
...
+## @defer compared to split queries
+
+Splitting a screen into a fast query and a slow query also gets the fast data first.
+`@defer` differs in these points:
+
+- **One HTTP request.**
+  The router plans both parts together and executes them in one HTTP exchange.
+- **Shared fetches.**
+  The entity keys and parent objects fetched for the first part are reused for the deferred fragments.
+  Two queries fetch them twice.
```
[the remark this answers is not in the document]
devsergiy: not applicable, it is an example of current behaviour, it is not replaceable

docs-website/router/defer.mdx
```diff
@@ -0,0 +1,200 @@
...
+      "message": "the router received a query with the @defer directive but the client does not accept multipart/mixed HTTP responses. To enable @defer support, add the HTTP header 'Accept: multipart/mixed'",
+      "extensions": {
+        "code": "DEFER_BAD_HEADER"
+      }
+    }
+  ]
+}
+```
+
+A client that asks for a different incremental delivery format, such as `multipart/mixed;deferSpec=20220824` sent by Apollo Router clients, would parse the response incorrectly and lose the deferred data.
```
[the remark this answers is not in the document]
devsergiy: Reworded, though not to the suggested text.

`deferSpec=20220824` is Apollo's wire format generally — Apollo Router implements it server-side and `Defer20220824Handler` is one client-side implementation of it — so pinning the sentence to that handler trades a vague attribution for an over-specific one. The handler is introduced in `usage.mdx` where client setup is covered; on the overview page the reader doesn't need it yet.

Line 105 now reads `..., the older format used by Apollo Router,`, which matches the phrasing already used in `usage.mdx:446`.

docs-website/router/defer.mdx
```diff
@@ -0,0 +1,200 @@
...
+An operation with all `@defer` directives disabled through `@defer(if: false)` will be executed as a regular operation.
+The router answers it with a single JSON response regardless of the `Accept` header.
+
+## Response format
+
+The router responds with these headers:
+
+```http
+Content-Type: multipart/mixed; boundary="graphql"; incrementalSpec=v0.2
+Transfer-Encoding: chunked
```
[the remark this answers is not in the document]
devsergiy: Fixed, in the code as well as the docs.

The mismatch was wider than the doc line: `defer_response_writer.go` set the header unconditionally, while the sibling streaming path at `subscription_response_writer.go:255` already guards it with `if r.ProtoMajor == 1`. `GetDeferResponseWriter` received the request as `_ *http.Request` and discarded it, so the guard cost a parameter rename.

Nothing was broken on the wire — Go's HTTP/2 server drops `transfer-encoding` unless the value is `trailers` (`h2_bundle.go`, `http2encodeHeaders`) — but it was a dead write plus a doc promising a header HTTP/2 clients never see. Deleting the doc line alone would have hidden that rather than resolved it.

The header block keeps the entry, with a note below it scoping `Transfer-Encoding: chunked` to HTTP/1.1.

docs-website/router/defer/usage.mdx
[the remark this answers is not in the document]
devsergiy: Handled by dropping the `__typename` sentence rather than version-qualifying it.

`__typename` insertion isn't a client-integration detail at all: the router adds its own placeholders during normalization (`astnormalization/defer_ensure_typename.go` in graphql-go-tools) for any field whose children are all deferred, regardless of which client sent the query. Apollo Client also adds `__typename` client-side for cache normalization, but the original sentence implied the router was accommodating something the client did — a reader could conclude non-Apollo clients must send `__typename` themselves.

The paragraph now mentions only the generated `label`, which is genuinely client-specific. The 4.3.0 boundary you raised would still apply to what's left, if the version split is real.

docs-website/router/defer/when-to-use.mdx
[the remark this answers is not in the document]
devsergiy: Qualified rather than replaced.

The rewrite drops "Two queries fetch them twice", which is the only sentence doing comparison work in a section titled "@defer compared to split queries" — softened to "can seed deferred entity fetches", the bullet stops arguing anything.

The underlying point stands though, and for a reason not stated in the comment: the two pages contradicted each other. `how-it-works.mdx:80` says the router "walks up to `employee`, a root field on the same subgraph, and runs an additional query", and the trace at :203-205 labels fragments "second root query to employees".

Kept the bullet and added one line noting that deferred root fields and fields on non-entity types can still need a second root query, linking to that section.

router/core/defer_response_writer.go
```diff
@@ -114,28 +114,70 @@ func GetDeferResponseWriter(ctx *resolve.Context, _ *http.Request, w http.Respon
...
 
-	return false
+	switch {
+	case unsupported != "":
+		return deferAcceptUnsupportedSpec, unsupported
+	case wildcard:
+		return deferAcceptCompatible, ""
+	}
+	return deferAcceptNoMultipart, ""
 }
```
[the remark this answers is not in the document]
devsergiy: Not taking this one.

The mechanics are right — `Header.Get` returns only the first field-line, and RFC 9110 §5.3 makes multiple field-lines equivalent to one comma-joined line. But every defer-capable client in the docs sends a single `Accept`: Apollo Client, curl, browsers, `fetch`. Two `Accept` field-lines come from a proxy appending rather than replacing, and I couldn't construct a case a real client hits.

`subscription_response_writer.go:271` also uses `Header.Get` for its own negotiation, so `Get` is the established pattern here; changing only `deferAccept` would make the two streaming paths inconsistent in the other direction for an unreachable case.

router/core/defer_response_writer.go
```diff
@@ -114,28 +114,70 @@ func GetDeferResponseWriter(ctx *resolve.Context, _ *http.Request, w http.Respon
...
+			// mime.ParseMediaType lowercases parameter names.
+			if v, ok := params["deferspec"]; ok {
+				unsupported = "deferSpec=" + v
+				continue
+			}
+			if v, ok := params["incrementalspec"]; ok && v != deferIncrementalSpec {
+				unsupported = "incrementalSpec=" + v
+				continue
+			}
+			return deferAcceptCompatible, ""
```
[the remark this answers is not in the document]
devsergiy: Not taking this one either, and the stated harm is inverted.

`deferAccept` exists to stop clients from silently losing deferred data. A client that sends a `@defer` operation while declaring `multipart/mixed;q=0` is self-contradictory: it asked for incremental delivery over a transport it says it won't accept. Of the two outcomes, streaming multipart is the more useful one — rejecting with `DEFER_BAD_HEADER` would turn a working request into a hard failure to honour a header nobody sends. `q=0` is near-extinct in practice; the realistic shape (`*/*;q=0.1`) is already covered at `defer_accept_test.go:31`.

There's also a scope trap: in `Accept`, `q` is a delimiter, not an ordinary parameter — media-type parameters precede it, extension parameters follow — and `mime.ParseMediaType` flattens both into one map. Handling `q` correctly means tracking that boundary, which turns a three-way guard into a content-negotiation engine.

fiam (commented):
The docs part looks great. Could you please split the fix into a separate PR so we can review that separately? Thank you!

devsergiy:
> The docs part looks great. Could you please split the fix into a separate PR so we can review that separately? Thank you!

@fiam done

docs-website/router/defer.mdx
```diff
@@ -0,0 +1,218 @@
...
+
+When resolving a regular GraphQL response, a GraphQL server waits for every field to resolve.
+The response latency equals the resolve time of the slowest field.
+This can be a problem for large queries with many fields.
+
+One solution splits the query into smaller queries.
+The client then maintains two queries and makes two HTTP calls.
+
+With the `@defer` directive, a client marks the fragments that may arrive later.
+The router sends every available field in the initial response.
```
ysmolski: This implies that Router makes some decision about availability of fields, but it is not. It uses @defer as a marker for fields to be sent later.

docs-website/router/defer.mdx
```diff
@@ -0,0 +1,218 @@
...
+  }
+}
+```
+
+This query spans three subgraphs:
+- The `employees` subgraph returns `employee` and `teammates`.
+- The `products` subgraph returns `products`.
+- The `hobbies` subgraph returns `hobbies`.
+
+Without `@defer`, the client waits for all three subgraphs.
```
ysmolski: `` Without `@defer`, the client would wait for all three subgraphs. ``

docs-website/router/defer.mdx
```diff
@@ -0,0 +1,218 @@
...
+With `@defer`, the client receives everything except `hobbies` in the initial response.
+The client receives `hobbies` in an incremental response.
+
+## Specification status
+
+`@defer` is not yet part of a released GraphQL specification.
+The router implements the [Incremental Delivery spec draft](https://github.com/graphql/graphql-spec/pull/1110).
+The draft is at RFC stage 2 of the GraphQL specification process.
```
ysmolski: Please omit this stage version since we do not control this narrative and it will become outdated without us knowing it.

docs-website/router/defer.mdx
```diff
@@ -0,0 +1,218 @@
...
+The draft describes the response format as the `pending`, `incremental`, and `completed` format.
+
+graphql-js `17.0.0-alpha.9` and Apollo Client 4.1 implement the same format.
+Apollo Client implements it through its `GraphQL17Alpha9Handler`.
+See [Using @defer](/router/defer/usage) for client setup.
+
+## How to enable it
+
+The router disables `@defer` support by default.
```
ysmolski: `` The router ignores `@defer` by default. ``

docs-website/router/defer.mdx
```diff
@@ -0,0 +1,218 @@
...
+To enable it, add the following to the router configuration:
+
+```yaml config.yaml
+engine:
+  enable_defer: true
+```
+
+The corresponding environment variable is `ENGINE_ENABLE_DEFER`.
+The setting is available since router `0.328.0`.
```
ysmolski: Maybe move version requirements to the top like we have for other features?

docs-website/router/defer.mdx
```diff
@@ -0,0 +1,218 @@
...
+      }
+    }
+  ]
+}
+```
+
+A named format takes priority over a wildcard.
+The router rejects `Accept: multipart/mixed;deferSpec=20220824, */*`, because the client stated which format it can parse.
+
+An operation can disable all its `@defer` directives through `@defer(if: false)`.
```
ysmolski: Not sure about phrasing here: "all its". Do you want to say they particular defer directive can be disabled via parameter? Or did you mean operation level directive?

docs-website/router/defer/when-to-use.mdx
```diff
@@ -0,0 +1,99 @@
...
+  It needs the reviews later.
+  If the client defers the reviews, the page renders before the reviews subgraph answers.
+- **Content that the user cannot see yet.**
+  Data outside the visible area can arrive after the visible data.
+- **Expensive computed fields.**
+  A field with a `@requires` chain or an expensive resolver adds its latency to the whole response.
+  A deferred fragment removes that latency from the initial response.
+- **Secondary panels in a dashboard.**
+  The primary metrics render first.
+  Each secondary panel renders when its fragment arrives.
```
ysmolski: Please avoid such long lists where each item has many sentences. You might wanna make them paragraphs for readability. This applies to all list in the file.

docs-website/router/defer/when-to-use.mdx
```diff
@@ -0,0 +1,99 @@
...
+
+## @defer compared to split queries
+
+A split of the screen into a fast query and a slow query also delivers the fast data first.
```
ysmolski: What is the screen here?

docs-website/router/defer/how-it-works.mdx
```diff
@@ -0,0 +1,287 @@
...
+      }
+    }
+  }
+}
+```
+
+`Details` is a plain object type without `@key`.
+The router cannot fetch it on its own.
+The nearest anchor is `employee`, a root field on the same subgraph.
+The router runs an additional query for that root field.
```
ysmolski: Who is the target audience for this doc file? I am asking because this doc goes into low-level details that might not be needed for the regular users. Also even I have a problem understanding explanation like this one.

docs-website/router/defer/how-it-works.mdx
```diff
@@ -0,0 +1,287 @@
...
+
+The router splits the operation into a primary group of fetches, and one group per deferred fragment.
+
+1. **Primary phase.**
+   The router runs the fetches for every field outside the deferred fragments.
+   The router renders the initial response with `data` and `pending`, and sends it.
+2. **Deferred phase.**
+   The router resolves each deferred fragment through its own fetches.
+   The router merges the result into the same response tree.
+   The router renders the result as an `incremental` entry and sends one part.
```
ysmolski: What is the one part?

ysmolski (approved):
This looks awesome. Thanks!

PR #3267, opened 2026-09-21
feat: support no-cache and no-store directives of Cache-Control

ysmolski:
## Summary by CodeRabbit

- **New Features**
  - Added support for client `Cache-Control` directives when using response caching.
  - `no-cache` bypasses cached responses and refreshes stored data.
  - `no-store` allows cached reads but prevents responses from being stored.
  - Combining both directives bypasses cached data and prevents cache updates.
  - Invalid or unsupported cache directives are ignored.
- **Documentation**
  - Documented response-cache behavior and supported request directives.


router/core/cache_control.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+// readOnlyCache is a store that drops every write. Lookups pass through.
+type readOnlyCache struct {
+	caching.Cache
+}
+
+func (readOnlyCache) SetMany(context.Context, []caching.Item) error {
+	return nil
+}
```
Noroth, line 59: I'm against embedding interfaces like that. You make the calling design based on the implementation of `caching.Cache`. In general this could be error prone if anyone decides to change the interface a bit or add functions and suddenly the purpose of your defined types is not entirely fulfilled anymore.

I think you should rather implement the interface and maybe have a compile time assertion \_ = … to make sure your type satisfies the interface and implement the functions properly calling w.cache.SetMany/GetMany

Noroth (changes requested):

router/core/cache_control.go
ysmolski: Oops. I totally forgot about this. Fixed now

